PR #80, opened 2021-11-09
Fix `Page/ElementHandle.screenshot()`, viewport, and screen emulation

robingustafsson:
Replaces https://github.com/grafana/xk6-browser/pull/63

Implementation based on: https://github.com/microsoft/playwright/blob/master/packages/playwright-core/src/server/screenshotter.ts

### DIFF
--- a/common/browser_context_options.go
+++ b/common/browser_context_options.go
@@ -137,7 +137,7 @@ func (b *BrowserContextOptions) Parse(ctx context.Context, opts goja.Value) erro
 					b.ReducedMotion = ReducedMotionNoPreference
 				}
 			case "screen":
-				screen := NewScreen()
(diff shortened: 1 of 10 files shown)


imiric (changes requested):
I can't comment on the CDP API usage, but the code LGTM, besides the minor suggestions. I like the introduction of `Screenshotter` :+1: The only real blocker is the file cleanup in `TestElementHandleScreenshot`.

@robingustafsson You were following the Playwright TS implementation of this, right? Can you link it in the PR description?

And there's a nil pointer panic in `Response.fetchBody()` :disappointed: I'll create an issue for it. It doesn't seem related to the changes here, right?

imiric (approved):
:+1:

PR #86, opened 2021-11-10
Make `Request.url` into a `url.URL`

imiric:
This is easier to work with for e.g. scheme comparisons, the parsing is done only once, and will now be properly logged if it fails.

### DIFF
--- a/common/network_manager.go
+++ b/common/network_manager.go
@@ -24,7 +24,6 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"net/url"
 	"strconv"
 	"strings"
 	"sync"
@@ -49,6 +48,7 @@ type NetworkManager struct {
 	BaseEventEmitter
 
 	ctx          context.Context
+	logger       *Logger
 	session      *Session
 	parent       *NetworkManager
 	frameManager *FrameManager
@@ -68,10 +68,16 @@ type NetworkManager struct {
 }
 
 // NewNetworkManager creates a new network manager
-func NewNetworkManager(ctx context.Context, session *Session, manager *FrameManager, parent *NetworkManager) (*NetworkManager, error) {
+func NewNetworkManager(
+	ctx context.Context, session *Session, manager *FrameManager, parent *NetworkManager,
+) (*NetworkManager, error) {
+	state := k6lib.GetState(ctx)
 	m := NetworkManager{
-		BaseEventEmitter:               NewBaseEventEmitter(ctx),
-		ctx:                            ctx,
+		BaseEventEmitter: NewBaseEventEmitter(ctx),
+		ctx:              ctx,
+		// TODO: Pass an internal logger instead of basing it on k6's logger?
+		// See https://github.com/grafana/xk6-browser/issues/54
+		logger:                         NewLogger(ctx, state.Logger, false, nil),
 		session:                        session,
 		parent:                         parent,
 		frameManager:                   manager,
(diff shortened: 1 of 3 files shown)


common/network_manager.go
```diff
@@ -338,33 +344,35 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent, interc
...
 	}
 	if frame == nil {
-		debugLog("<networkmanager:onRequest> frame is nil")
+		m.logger.Debugf("NetworkManager", "frame is nil")
 	}
 
-	req := NewRequest(m.ctx, event, frame, redirectChain, interceptionID, m.userReqInterceptionEnabled)
+	req, err := NewRequest(m.ctx, event, frame, redirectChain, interceptionID, m.userReqInterceptionEnabled)
+	if err != nil {
+		m.logger.Errorf("NetworkManager", "error creating Request: %s", err)
```
inancgumus, line 352: It's clear that it's an error. What about:

```golang
m.logger.Errorf("NetworkManager", "cannot create Request: %s", err)
```

common/request.go
```diff
@@ -61,19 +61,28 @@ type Request struct {
...
+	redirectChain []*Request, interceptionID string, allowInterception bool,
+) (*Request, error) {
 	documentID := cdp.LoaderID("")
 	if event.RequestID == network.RequestID(event.LoaderID) && event.Type == "Document" {
 		documentID = event.LoaderID
 	}
+
+	u, err := url.Parse(event.Request.URL)
+	if err != nil {
+		return nil, err
```
inancgumus, line 75: Let's return a richer error message here, like:

```golang
return nil, fmt.Errorf("cannot parse URL: %w", err)
```

It will result in:
`cannot parse URL: parse ":": missing protocol scheme`

common/request_test.go
```diff
@@ -51,7 +52,26 @@ func TestRequest(t *testing.T) {
...
+				URL:      ":",
+				Method:   "POST",
+				Headers:  network.Headers(headers),
+				PostData: "hello",
+			},
+			Timestamp: &ts,
+			WallTime:  &wt,
+		}
+		req, err := NewRequest(ctx, evt, nil, nil, "intercept", false)
+		require.EqualError(t, err, `parse ":": missing protocol scheme`)
```
inancgumus, line 72: Let's check the error in the test whether it contains our error message instead of stdlib's: `cannot parse URL`.

inancgumus (changes requested):
LGTM, minor changes.

common/network_manager.go
imiric: Sure, done in 7dc6ef7.

common/request.go
imiric: Done in 7dc6ef7.

common/request_test.go
imiric: Done in 7dc6ef7.

imiric:
I removed the skipping emission of request metrics for `data:` URLs, and will create a new PR for that.

common/request_test.go
```diff
@@ -51,7 +52,26 @@ func TestRequest(t *testing.T) {
...
+				URL:      ":",
+				Method:   "POST",
+				Headers:  network.Headers(headers),
+				PostData: "hello",
+			},
+			Timestamp: &ts,
+			WallTime:  &wt,
+		}
+		req, err := NewRequest(ctx, evt, nil, nil, "intercept", false)
+		require.EqualError(t, err, `cannot parse URL: parse ":": missing protocol scheme`)
```
inancgumus, line 72: It's fine this way, and I think we don't need to look for an exact match here. It's fine to test if the error just contains `cannot parse URL`.

inancgumus (approved):

PR #91, opened 2021-11-11
Skip request handling of data: URLs

imiric:
This skips handling of `data:` URLs as early as possible, so that we don't emit empty metrics for them.

These are typically found in CSS, e.g.: `background:url("data:image/svg+xml;base64,PHN2ZyB3...")`.

### DIFF
--- a/tests/network_manager_test.go
+++ b/tests/network_manager_test.go
@@ -1,0 +1,56 @@
+/*
+ *
+ * xk6-browser - a browser automation extension for k6
+ * Copyright (C) 2021 Load Impact
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package tests
+
+import (
+	"io/ioutil"
+	"testing"
+
+	"github.com/grafana/xk6-browser/testutils"
+	"github.com/grafana/xk6-browser/testutils/browsertest"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDataURLSkipRequest(t *testing.T) {
+	t.Parallel()
+	bt := browsertest.NewBrowserTest(t)
+	p := bt.Browser.NewPage(nil)
+	t.Cleanup(func() {
+		p.Close(nil)
+		bt.Browser.Close()
+	})
+
+	logHook := &testutils.SimpleLogrusHook{HookedLevels: []logrus.Level{logrus.DebugLevel}}
+	bt.State.Logger.SetLevel(logrus.DebugLevel)
+	bt.State.Logger.AddHook(logHook)
+	bt.State.Logger.SetOutput(ioutil.Discard)
+
+	p.Goto("data:text/html,hello", nil)
+
+	var gotMsg bool
+	for _, evt := range logHook.Drain() {
+		if evt.Message == "skipped request handling of data URL" {
+			gotMsg = true
+		}
+	}
+	assert.True(t, gotMsg)
+}
--- a/tests/network_manager_test.go
+++ b/tests/network_manager_test.go
@@ -21,12 +21,9 @@
 package tests
 
 import (
-	"io/ioutil"
 	"testing"
 
-	"github.com/grafana/xk6-browser/testutils"
 	"github.com/grafana/xk6-browser/testutils/browsertest"
-	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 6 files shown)


testutils/logrus_hook.go
```diff
@@ -0,0 +1,59 @@
+/*
+ *
+ * k6 - a next-generation load testing tool
+ * Copyright (C) 2019 Load Impact
```
imiric, line 4: How do we handle the copyright header for files copied from k6, or well any other project? We have to keep it intact, right? So do we add the xk6-browser one below?

@robingustafsson

In this case we could import `lib.testutils.SimpleLogrusHook` directly from k6, but since this is a test helper and might change I'd rather we copy it.

robingustafsson: @imiric If we copy them from projects under the /grafana gh org and the projects have CLAs we are free to re-license the code but I'd suggest we follow these guidelines:

1. If the repo we're copying code from is a repo under /grafana gh org and it has the same license as xk6-browser we can switch to the xk6-browser copyright header.
2. If the repo we're copying code from is a repo under /grafana gh org but has a different license from xk6-browser (will always be more liberal than AGPLv3) we retain the license from the source repo and add both the license header from the source as well as the xk6-browser license header to cover any modifications.

robingustafsson:
Code LGTM, I can't see why `common.Logger.categoryFilter` would cause a race condition as the `Logf` method where it's used is guarded by a mutex 🤔

imiric:
`Logger.Logf` is guarded, but the race is on a lower level. I've seen it happen with `categoryFilter`, but in https://github.com/grafana/xk6-browser/runs/4176861845 it failed on `Logger.logger.Level`. So b76f78e should avoid that.

tests/network_manager_test.go
```diff
@@ -0,0 +1,56 @@
...
+	p := bt.Browser.NewPage(nil)
+	t.Cleanup(func() {
+		p.Close(nil)
+		bt.Browser.Close()
+	})
+
+	logHook := &testutils.SimpleLogrusHook{HookedLevels: []logrus.Level{logrus.DebugLevel}}
+	bt.State.Logger.SetLevel(logrus.DebugLevel)
+	bt.State.Logger.AddHook(logHook)
+	bt.State.Logger.SetOutput(ioutil.Discard)
```
inancgumus, line 45: We can move the logger into the `browsertest` package where it is used (without thinking about we can use it afterward (which we can but we can also think about that later when the time comes)).

And, we can do it like this instead of adding verbosity to a test (here's the [logger.patch.zip](https://github.com/grafana/xk6-browser/files/7520460/logger.patch.zip)):


```golang
func (bt *BrowserTest) Log() *DebugLog {
	logHook := &DebugLog{HookedLevels: []logrus.Level{logrus.DebugLevel}}
	bt.State.Logger.SetLevel(logrus.DebugLevel)
	bt.State.Logger.AddHook(logHook)
	bt.State.Logger.SetOutput(ioutil.Discard)
	return logHook
}
```

Within the test:

```golang
func TestDataURLSkipRequest(t *testing.T) {
	t.Parallel()
	bt := browsertest.NewBrowserTest(t)
	p := bt.Browser.NewPage(nil)
	t.Cleanup(func() {
		p.Close(nil)
		bt.Browser.Close()
	})

	log := bt.Log() // <----
	p.Goto("data:text/html,hello", nil)

	var gotMsg bool
	for _, evt := range log.Drain() {
		if evt.Message == "skipped request handling of data URL" {
			gotMsg = true
		}
	}
	assert.True(t, gotMsg)
}
```

inancgumus (changes requested):
I have some suggestions.

Testing something through log messages is not the best way, and I understand you, it's the most effective solution for now.

tests/network_manager_test.go
imiric: :-/ Renaming `SimpleLogrusHook` to `DebugLog` wouldn't make sense since the hook can be used for other log levels, and I'd like to keep the same name from k6. It's also not specific to the `browsertest` package, so it doesn't belong there.

I slightly agree that it adds a bit of setup to the test, but hooking on `DebugLevel` is unique for this test, and others might want to do it differently. The verbosity doesn't bother me, so I don't see what we gain from adding a method to `BrowserTest` :man_shrugging:

inancgumus: > :-/ Renaming SimpleLogrusHook to DebugLog wouldn't make sense since the hook can be used for other log levels, 
bt.State.Logger.SetLevel(logrus.DebugLevel)
> hooking on DebugLevel is unique for this test,

Think about `DebugLevel` as a default state. Changing it in other tests won't be a problem because we can set our log levels after calling the method—We can still access `bt.State.Logger.SetLevel`.

We can call it `Log` instead of `DebugLog`. It will mean naturally like a "Browser Test's log".

> and I'd like to keep the same name from k6.

Calling it `Simple` might be OK for k6 seniors, but it doesn't tell me anything about what it represents until I read its code 🤷 

> It's also not specific to the browsertest package, so it doesn't belong there.

`browsertest` is the premiere testing environment we have right now, and all testing stuff goes right through it.

> verbosity doesn't bother me

Less code = fewer bugs 😓 I think verbosity isn't necessary here.

> I don't see what we gain from adding a method to BrowserTest

Easy to read code without thinking about what it does. Just like using functions instead of copy-pasting their inner code everywhere 😅

imiric: OK, see the latest changes and let's compromise :stuck_out_tongue:

I'm against coupling this to the `browsertest` package, but I added a `BrowsertTest.LogHook` method, and renamed `SimpleLogrusHook` to `CacheLogrusHook`. But really man, it's a single page of code with decent documentation...

I'm not sure why this is a blocker for you. You're asking to clean up something of which there's only one use of. If the verbosity bothered someone in the future they could've done something similar then. :man_shrugging:

inancgumus: Man, I'm just reviewing it. I can simply approve it. It doesn't bother me at all 😄  If you don't want me to review this deeply, just tell me, I'll act accordingly 🙇

inancgumus: `CacheLogrusHook` much better by the way 😄

inancgumus (approved):

Issue #93, opened 2021-11-12
GC browser processes


robingustafsson:
Throughout the development of xk6-browser there's been many times when the browser process or processes have not been properly terminated when k6 exits. Most of the time it's related to k6 `panic()`'ing but there's also been situations where the orderly closing of the browser hasn't worked properly.

We need to find a way to avoid ever having orphaned browser processes when k6 exits (find a way to get the equivalent of cross-platform `cmd.SysProcAttr.Pdeathsig = syscall.SIGKILL`). Some ideas:

- We now have [`k6Throw()`](https://github.com/grafana/xk6-browser/blob/main/common/helpers.go#L312) in helpers.go, perhaps we should extend it to cleanup any running browser processes before calling `panic()` via some sort of context-attached hook.
- Implement hooks in k6 for extensions to be notified when an iteration, VU, scenario or test ends, so that the extension can do necessary cleanup actions.

inancgumus:
We can temporarily do something in `k6Throw`, but I think we need the second solution from the k6 side the best: notifying the extension just before an iteration, VU, scenario, or test ends.

PR #110, opened 2021-11-12
Cross-platform browser process killer

inancgumus:
This PR adds a cross-platform browser process killer.
* `k6Throw` kills any lingering browser processes in a cross-platform way.
* Saves the process ID to the context when launching a new browser.

We need to update `k6common.Throw` to our `k6Throw` for this to work stable. And I manually tested this with 100 browsers on my Mac. It cleaned up all the lingering processes. Automated tests are missing for now. I'm planning to do these in a separate PR.

As I said in the PR this is a temporary fix:
https://github.com/grafana/xk6-browser/issues/93#issuecomment-966911692

Closes: #93

### DIFF
--- a/common/context.go
+++ b/common/context.go
@@ -26,6 +26,7 @@ type ctxKey int
 
 const (
 	ctxKeyLaunchOptions ctxKey = iota
+	ctxKeyPid
 	ctxKeyHooks
 )
 
@@ -51,4 +52,15 @@ func GetLaunchOptions(ctx context.Context) *LaunchOptions {
 		return nil
 	}
 	return v.(*LaunchOptions)
+}
+
+// TODO: Test is missing
+func WithProcessID(ctx context.Context, pid int) context.Context {
+	return context.WithValue(ctx, ctxKeyPid, pid)
+}
+
+// TODO: Test is missing
+func GetProcessID(ctx context.Context) int {
+	v, _ := ctx.Value(ctxKeyPid).(int)
+	return v // it will return zero on error
 }
--- a/common/context.go
+++ b/common/context.go
@@ -54,12 +54,12 @@ func GetLaunchOptions(ctx context.Context) *LaunchOptions {
 	return v.(*LaunchOptions)
 }
 
-// TODO: Test is missing
+// WithProcessID saves the browser process ID to the context.
 func WithProcessID(ctx context.Context, pid int) context.Context {
 	return context.WithValue(ctx, ctxKeyPid, pid)
 }
 
-// TODO: Test is missing
+// GetProcessID returns the browser process ID from the context.
 func GetProcessID(ctx context.Context) int {
 	v, _ := ctx.Value(ctxKeyPid).(int)
 	return v // it will return zero on error
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -401,7 +401,7 @@ func (fs *FrameSession) navigateFrame(frame *Frame, url, referrer string) (strin
 }
 
 func (fs *FrameSession) onConsoleAPICalled(event *runtime.EventConsoleAPICalled) {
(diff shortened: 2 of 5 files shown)


robingustafsson (approved):
Great, yes this is basically what I thought of as a temporary fix. I'll let Ivan do a more thorough code review 🙂 ...but LGTM.

common/context.go
```diff
@@ -52,3 +53,14 @@ func GetLaunchOptions(ctx context.Context) *LaunchOptions {
 	}
 	return v.(*LaunchOptions)
 }
+
+// TODO: Test is missing
+func WithProcessID(ctx context.Context, pid int) context.Context {
```
imiric, line 58: Could you document this and `GetProcessID` properly, so that we have less things to do in #58? :)

The `TODO` seems unnecessary as well, we can get to it along with the other missing tests.

common/helpers.go
```diff
@@ -308,12 +309,37 @@ func waitForEvent(ctx context.Context, emitter EventEmitter, events []string, pr
...
+		// this should never happen unless a programmer error
+		panic("cannot find process id")
+	}
+	p, err := os.FindProcess(pid)
+	if err != nil {
+		// optimistically return and don't kill the process
+		return
+	}
+	ps, _ := p.Wait()
+	if err != nil {
```
imiric, line 334: Did you mean to assign to `err` from the `Wait()` call? Otherwise you're just checking the above `err`.

And I'm not sure about the flow here. Wouldn't this wait for the browser to finish exiting normally, and what we want is to kill it prematurely?

imiric (commented):
Just some nitpicks and a question.

I was in the middle of reviewing it when you removed the `psutil` dependency :), but it's better if we can do this with the stdlib.

Some debug logging would be nice to know what happening, and tests of course.

Also, why aren't we replacing the other direct uses of `k6common.Throw()`?

inancgumus:
> I was in the middle of reviewing it when you removed the `psutil` dependency :), but it's better if we can do this with the stdlib.
> 

;)

> Also, why aren't we replacing the other direct uses of `k6common.Throw()`?

:) PR description: "We need to update k6common.Throw to our k6Throw for this to work stable. ... Automated tests are missing for now. I'm planning to do these in a separate PR."

common/frame_session.go
```diff
@@ -402,8 +401,7 @@ func (fs *FrameSession) navigateFrame(frame *Frame, url, referrer string) (strin
 }
 
 func (fs *FrameSession) onConsoleAPICalled(event *runtime.EventConsoleAPICalled) {
-	// TODO: switch to using browser logger instead of directly outputting to k6 logging system
-	rt := k6common.GetRuntime(fs.ctx)
+	// TODO: switch to using browser logger instead of directly outputting to k6 logging system	rt := k6common.GetRuntime(fs.ctx)
```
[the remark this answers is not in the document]
inancgumus: Thanks for catching that ❤️  It's pretty hard to remove these. As there are hundreds of these calls. That's why I want to do it in a separate PR.

common/context.go
inancgumus: Sure ;)

common/helpers.go
inancgumus: Ah, good catch, thanks. 

I thought we can gracefully shut down our extension but yeah, we better kill it prematurely.

imiric (approved):
The code LGTM, but I wasn't able to come up with a manual test for this scenario.

On `main`, stopping the test run with `Ctrl+C` or calling `k6.fail()` in the script properly kills the process and leaves no zombies behind. Panic issues also seem to get rid of the process. :confused: So I'm not sure about the exact scenario that reproduces this, but automated tests (E2E only I suppose) would help.

Since I don't see the code doing any harm, I'll merge this as is.

inancgumus:
Would you please try running Tom’s earliest script within a `for..` in bash? It happens every time for me on OSX. @imiric

Issue #90, opened 2021-11-10
`Request.frame` is sometimes nil


imiric:
Sometimes when we create a new `Request` here:
https://github.com/grafana/xk6-browser/blob/9882ab5e0390cd12cea19165a6786c4b9c463774/common/network_manager.go#L342-L350

... we get a nil `Frame` pointer from `FrameManager.getFrameByID()`.

The logic of how `m.frames` is changed is a bit convoluted, as we delete frames in a couple of places:
- https://github.com/grafana/xk6-browser/blob/9882ab5e0390cd12cea19165a6786c4b9c463774/common/frame_manager.go#L185
- https://github.com/grafana/xk6-browser/blob/9882ab5e0390cd12cea19165a6786c4b9c463774/common/frame_manager.go#L295

We added some debugging information in 0451efb so we can see when it happens, but there's likely a synchronization bug here that we need to address so that anything that needs a `Frame` reference (e.g. when getting the session through it for [`Response.fetchBody()`](https://github.com/grafana/xk6-browser/blob/9882ab5e0390cd12cea19165a6786c4b9c463774/common/response.go#L140) (which we could possibly store directly on `Response` to avoid resolving the reference chain?)) can run before we actually delete the `Frame`.

imiric:
Some more info about this: from testing with the script in #49, when we get a nil `Frame` in [`NetworkManager.onRequest()`](https://github.com/grafana/xk6-browser/blob/9882ab5e0390cd12cea19165a6786c4b9c463774/common/network_manager.go#L342-L350) it's because `event.FrameID == ""`, i.e. there's no `Frame` associated to the `RequestWillBeSent` event to lookup. In all cases these are preflight/`OPTIONS` requests, which typically don't have a body anyway, though [this is not strictly forbidden by the spec](https://stackoverflow.com/questions/11926908/how-to-respond-to-an-http-options-request#comment66715405_11930225). So if the responses have `Content-Length` we would emit metrics for them fine and won't rely on `Response.fetchBody()`/CDP.

That said, the reason `fetchBody()` fails so often is because calling `network.GetResponseBody()` shouldn't be done on the `ResponseReceived` event, but on the `LoadingFinished` one. See https://github.com/ChromeDevTools/devtools-protocol/issues/44#issuecomment-320127404.

[...]

Issue #40, opened 2021-10-27
Implement tracking of `data_sent`/`data_received` metrics


imiric:
Similarly [how this is done for native k6 requests/responses](https://github.com/grafana/k6/blob/b7de6d770d/lib/netext/dialer.go#L103-L116), the extension should emit `data_sent`/`data_received` metrics as well.

PR #117, opened 2021-11-16
More reliable response body size and `data_received` metric emission

imiric:
This is related to #90, but doesn't really fix the issue as the current behavior is correct. See [this comment](https://github.com/grafana/xk6-browser/issues/90#issuecomment-966477394).

Essentially `Frame` being nil is a valid state for responses to preflight/`OPTIONS` requests, so the fix in b2052bb is all we need to do for handling this case.

What this PR does is:
- move the `data_received` metric emission to the `LoadingFinished` event handler, which [according to this comment](https://github.com/ChromeDevTools/devtools-protocol/issues/44#issuecomment-320127404) is more reliable. This results in more accurate `data_received` metric values.
  
  Before, on `main` (54cd293) with the script from [this comment](https://github.com/grafana/xk6-browser/pull/117#issuecomment-1002521620):
  ```
  data_received....................: 17 MB  737 kB/s
  ```
  After, on a9d9868:
  ```
  data_received....................: 39 MB  754 kB/s
  ```

  I confirmed this is more accurate by comparing with the Chrome DevTools Network tab.
  
  This change also produces less `error fetching response body` warning events, though it doesn't eliminate them entirely, except when running the test suite.
- enhance the logging so that we can see the type of request when `Frame` is nil. In all cases I've seen with the script from Tom these are preflight/`OPTIONS` requests which don't have an associated `FrameID`.

### DIFF
--- a/common/network_manager.go
+++ b/common/network_manager.go
@@ -419,7 +419,8 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent, interc
 		frame = m.frameManager.getFrameByID(event.FrameID)
 	}
 	if frame == nil {
-		m.logger.Debugf("NetworkManager", "frame is nil")
+		m.logger.Debugf("NetworkManager:onRequest", "url:%s method:%s type:%s fid:%s frame is nil",
+			event.Request.URL, event.Request.Method, event.Initiator.Type, event.FrameID)
 	}
 
 	req, err := NewRequest(m.ctx, event, frame, redirectChain, interceptionID, m.userReqInterceptionEnabled)
--- a/common/network_manager.go
+++ b/common/network_manager.go
@@ -417,11 +417,16 @@ func (m *NetworkManager) onLoadingFinished(event *network.EventLoadingFinished) 
 		}
 	}
 	req.responseEndTiming = float64(event.Timestamp.Time().Unix()-req.timestamp.Unix()) * 1000
-	if req.url.Scheme != "data" {
+	// Skip data and blob URLs when emitting metrics, since they're internal to the browser.
+	if !isInternalURL(req.url) {
 		m.emitResponseMetrics(req.response, req)
 	}
 	m.deleteRequestByID(event.RequestID)
 	m.frameManager.requestFinished(req)
+}
+
+func isInternalURL(u *url.URL) bool {
+	return u.Scheme == "data" || u.Scheme == "blob"
 }
 
 func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent, interceptionID string) {
@@ -454,8 +459,9 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent, interc
 		m.logger.Errorf("NetworkManager", "cannot create Request: %s", err)
 		return
 	}
-	if req.url.Scheme == "data" {
-		m.logger.Debugf("NetworkManager", "skipped request handling of data URL")
+	// Skip data and blob URLs, since they're internal to the browser.
+	if isInternalURL(req.url) {
+		m.logger.Debugf("NetworkManager", "skipped request handling of %s URL", req.url.Scheme)
 		return
 	}
 	m.reqsMu.Lock()
--- a/common/network_manager.go
+++ b/common/network_manager.go
@@ -391,6 +391,9 @@ func (m *NetworkManager) onLoadingFinished(event *network.EventLoadingFinished) 
 		}
 	}
 	req.responseEndTiming = float64(event.Timestamp.Time().Unix()-req.timestamp.Unix()) * 1000
+	if req.url.Scheme != "data" {
+		m.emitResponseMetrics(req.response)
+	}
 	m.deleteRequestByID(event.RequestID)
 	m.frameManager.requestFinished(req)
 }
@@ -528,9 +531,6 @@ func (m *NetworkManager) onResponseReceived(event *network.EventResponseReceived
 	}
 	resp := NewHTTPResponse(m.ctx, req, event.Response, event.Timestamp)
 	req.response = resp
-	if req.url.Scheme != "data" {
-		m.emitResponseMetrics(resp)
-	}
 	m.frameManager.requestReceivedResponse(resp)
 }
 
--- a/common/network_manager.go
+++ b/common/network_manager.go
@@ -190,32 +190,57 @@ func (m *NetworkManager) emitRequestMetrics(req *Request) {
 	})
 }
 
-func (m *NetworkManager) emitResponseMetrics(resp *Response) {
+func (m *NetworkManager) emitResponseMetrics(resp *Response, req *Request) {
 	state := k6lib.GetState(m.ctx)
+
+	// In some scenarios we might not receive a ResponseReceived CDP event, in
+	// which case the response won't be created. So to emit as much metric data
+	// as possible we set some sensible defaults instead.
+	var (
+		status, bodySize                    int64
+		ipAddress, protocol                 string
+		fromCache, fromPreCache, fromSvcWrk bool
+		url                                 = req.url.String()
+		timestamp                           = time.Now()
+	)
+	if resp != nil {
+		status = resp.status
+		bodySize = resp.Size().Total()
+		ipAddress = resp.remoteAddress.IPAddress
+		protocol = resp.protocol
+		fromCache = resp.fromDiskCache
+		fromPreCache = resp.fromPrefetchCache
+		fromSvcWrk = resp.fromServiceWorker
+		timestamp = resp.timestamp
+		url = resp.url
+	} else {
+		m.logger.Debugf("NetworkManager:emitResponseMetrics",
+			"response is nil url:%s method:%s", req.url, req.method)
+	}
 
 	tags := state.CloneTags()
 	if state.Options.SystemTags.Has(k6stats.TagGroup) {
 		tags["group"] = state.Group.Path
 	}
 	if state.Options.SystemTags.Has(k6stats.TagMethod) {
-		tags["method"] = resp.request.method
+		tags["method"] = req.method
 	}
 	if state.Options.SystemTags.Has(k6stats.TagURL) {
-		tags["url"] = resp.url
+		tags["url"] = url
 	}
 	if state.Options.SystemTags.Has(k6stats.TagIP) {
-		tags["ip"] = resp.remoteAddress.IPAddress
+		tags["ip"] = ipAddress
 	}
 	if state.Options.SystemTags.Has(k6stats.TagStatus) {
-		tags["status"] = strconv.Itoa(int(resp.status))
+		tags["status"] = strconv.Itoa(int(status))
 	}
 	if state.Options.SystemTags.Has(k6stats.TagProto) {
-		tags["proto"] = resp.protocol
+		tags["proto"] = protocol
 	}
 
-	tags["from_cache"] = strconv.FormatBool(resp.fromDiskCache)
-	tags["from_prefetch_cache"] = strconv.FormatBool(resp.fromPrefetchCache)
-	tags["from_service_worker"] = strconv.FormatBool(resp.fromServiceWorker)
+	tags["from_cache"] = strconv.FormatBool(fromCache)
+	tags["from_prefetch_cache"] = strconv.FormatBool(fromPreCache)
+	tags["from_service_worker"] = strconv.FormatBool(fromSvcWrk)
 
 	sampleTags := k6stats.IntoSampleTags(&tags)
 	k6stats.PushIfNotDone(m.ctx, state.Samples, k6stats.ConnectedSamples{
@@ -224,7 +249,7 @@ func (m *NetworkManager) emitResponseMetrics(resp *Response) {
 				Metric: state.BuiltinMetrics.HTTPReqs,
 				Tags:   sampleTags,
 				Value:  1,
-				Time:   resp.timestamp,
+				Time:   timestamp,
 			},
 			{
 				Metric: state.BuiltinMetrics.HTTPReqDuration,
@@ -235,18 +260,19 @@ func (m *NetworkManager) emitResponseMetrics(resp *Response) {
 				// issues with the parsing and conversion to `time.Time`.
 				// Have not spent time looking for the root cause of this in the Chromium source to file a bug report, and neither
 				// Puppeteer nor Playwright seems to care about the `responseTime` value and don't use/expose it.
-				Value: k6stats.D(resp.timestamp.Sub(resp.request.timestamp)),
-				Time:  resp.timestamp,
+				Value: k6stats.D(timestamp.Sub(req.timestamp)),
+				Time:  timestamp,
 			},
 			{
 				Metric: state.BuiltinMetrics.DataReceived,
 				Tags:   sampleTags,
-				Value:  float64(resp.Size().Total()),
-				Time:   resp.timestamp,
+				Value:  float64(bodySize),
+				Time:   timestamp,
 			},
 		},
 	})
-	if resp.timing != nil {
+
+	if resp != nil && resp.timing != nil {
 		k6stats.PushIfNotDone(m.ctx, state.Samples, k6stats.ConnectedSamples{
 			Samples: []k6stats.Sample{
 				{
@@ -283,8 +309,8 @@ func (m *NetworkManager) handleRequestRedirect(req *Request, redirectResponse *n
 	req.response = resp
 	req.redirectChain = append(req.redirectChain, req)
 
+	m.emitResponseMetrics(resp, req)
 	m.deleteRequestByID(req.requestID)
-	m.emitResponseMetrics(resp)
 
 	/*
 		delete(m.attemptedAuth, req.interceptionID);
@@ -392,7 +418,7 @@ func (m *NetworkManager) onLoadingFinished(event *network.EventLoadingFinished) 
 	}
 	req.responseEndTiming = float64(event.Timestamp.Time().Unix()-req.timestamp.Unix()) * 1000
 	if req.url.Scheme != "data" {
-		m.emitResponseMetrics(req.response)
+		m.emitResponseMetrics(req.response, req)
 	}
 	m.deleteRequestByID(event.RequestID)
 	m.frameManager.requestFinished(req)
--- a/common/response.go
+++ b/common/response.go
@@ -24,7 +24,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -191,19 +190,7 @@ func (r *Response) Body() goja.ArrayBuffer {
 }
 
 // bodySize returns the size in bytes of the response body.
-// It first attempts to get the size specified in the content-length
-// header, and if unavailable falls back to the size as returned by CDP in
-// r.fetchBody(). This is because the CDP Network.getResponseBody call
-// is unreliable, see https://github.com/ChromeDevTools/devtools-protocol/issues/12#issuecomment-306947275 .
 func (r *Response) bodySize() int64 {
-	if v, ok := r.headers["content-length"]; ok && len(v) > 0 {
-		cl, err := strconv.ParseInt(v[0], 10, 64)
-		if err == nil {
-			return cl
-		}
-		r.logger.Warnf("cdp", "error parsing content-length header: %s", err)
-	}
-
 	if err := r.fetchBody(); err != nil {
 		r.logger.Warnf("cdp", "error fetching response body for '%s': %s", r.url, err)
 	}
--- a/common/response.go
+++ b/common/response.go
@@ -24,6 +24,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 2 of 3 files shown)


robingustafsson (approved):
LGTM, I'll let @inancgumus comment on the synchronization part :)

common/response.go
```diff
@@ -66,6 +67,7 @@ type Response struct {
 	url               string
 	status            int64
 	statusText        string
+	bodyMx            sync.RWMutex
```
inancgumus, line 70: Let's use the conventional `bodyMu` name instead of `bodyMx`.

common/response.go
```diff
@@ -133,15 +135,20 @@ func NewHTTPResponse(ctx context.Context, req *Request, resp *network.Response,
 }
 
 func (r *Response) fetchBody() error {
+	r.bodyMx.RLock()
```
inancgumus, line 138: I fear the usage of locking here can be error-prone and difficult to follow.

What about something like this?

```golang
cached := func() bool {
    r.bodyMu.Lock()
    defer r.bodyMu.RUnlock()

    return r.body != nil || r.request.frame == nil
}
if cached() {
    return nil
}
```

common/response.go
```diff
@@ -133,15 +135,20 @@ func NewHTTPResponse(ctx context.Context, req *Request, resp *network.Response,
...
 	if r.body != nil || r.request.frame == nil {
+		r.bodyMx.RUnlock()
 		return nil
 	}
+	r.bodyMx.RUnlock()
 	action := network.GetResponseBody(r.request.requestID)
 	body, err := action.Do(cdp.WithExecutor(r.ctx, r.request.frame.manager.session))
 	if err != nil {
 		return err
```
inancgumus, line 147: Let's enhance the error message while in here.

inancgumus (changes requested):
Wow, a lot of work, good job. Minor requests.

common/response.go
imiric: I wasn't aware `Mu` was a convention, I've seen both usages. And kind of prefer `Mx` because it's short for **m**utual e**x**clusion. But sure, `Mu` is more common in this project, so I'll change it.

common/response.go
imiric: I moved this fix to #124 so it can be merged quicker, and I used your suggestion there.

common/response.go
imiric: Done in 8718977.

imiric:
After some more research into this, for a reason I'm unable to explain, we just don't receive a `ResponseReceived` CDP event for some requests. :confused:

We get `RequestWillBeSent` and `LoadingFinished`, but between those we should also get `ResponseReceived`, which happens for most request flows. This is not only the case for preflight/`OPTIONS` requests, but some standard `GET` or `POST` requests as well. I haven't noticed a pattern where it happens for certain requests or URLs only; it can basically happen at any point with the Vistaprint script.

So it's not a matter of synchronizing how events are processed (thankfully! :sweat_smile:), but the fact we just don't get the `ResponseReceived` event.

I did a [hacky fallback](https://github.com/grafana/xk6-browser/commit/405ee49edc33dc144294661f80c98ba41fcd7685) so that we at least don't panic when this happens, but metrics for these requests will be less useful.

I'll pick this back up when I'm back from vacation, but in the meantime @inancgumus, if you have some time, please test the Vistaprint script and any others you have on this branch and compare it to `main`. From my tests `error fetching response body` is much less frequent (it's not printed anymore in tests when running with `go test -v`), and the `data_received` metric seems more accurate now (I compared it with the DevTools Network tab), though it's probably not 100% accurate.

inancgumus:
> From my tests, error fetching response body is much less frequent 

It was displayed 16 times on the main branch:

```
# main
$ go test ./tests -v -count=1 |& grep 'error fetching response body' | wc -l
16
```

I'm no longer observing these error log messages on this branch:

```
# fix/90-nil-request-frame
$ go test ./tests -v -count=1 |& grep 'error fetching response body' | wc -l
0
```

Using Tom's original script, I see the errors three times on the main branch:

```
# main
$ XK6_BROWSER_LOG=debug xk6 run -q ../tests/tom.js |& grep 'error fetching response body' | wc -l
3
```

However, I see them twice on this branch:

```
# fix/90-nil-request-frame
$ XK6_BROWSER_LOG=debug xk6 run -q ../tests/tom.js |& grep 'error fetching response body' | wc -l
2
```

I wrote an improved version of Tom's script. Using it, I see the errors 5 times on the main branch. However, **I see them 23 times on this branch.**

<details>
<summary>Here's the modified script.</summary>

```javascript
import { sleep, group } from 'k6';
import launcher from 'k6/x/browser';

import { randomIntBetween } from 'https://jslib.k6.io/k6-utils/1.1.0/index.js';

export const options = {};

let browser, context, page;

const pauseMin = 0;
const pauseMax = 1;

export default function() {
  initialize();

  navigateToHomepage();
  clickBusinessCards();
  clickBrowseDesigns();
  clickDesign();
  clickCustomizeDesign();
  clickNext();
  continueWithoutBack();
  customerReviewed();
  addToCart();
  continuePastRecommendations();
  continueToCart();
  goToCheckout();
}

function initialize() {
  browser = launcher.launch('chromium', {
    args: [],
    debug: false,
    devtools: false,
    env: {},
    executablePath: null,
    headless: false,
    ignoreDefaultArgs: [],
    proxy: {},
    slowMo: '500ms',
    timeout: '30s',
  });

  context = browser.newContext({
    ignoreHTTPSErrors: true,
    screen: {width: 1920, height: 1080},
    viewport: {width: 1920, height: 1080},
  });

  page = context.newPage();
}

function navigateToHomepage() {
  console.log('Starting navigateToHomepage...');

  group("Navigate to Homepage", () => {
    page.goto('https://vistaprint.ca', { waitUntil: 'networkidle' });
  });

  sleep(randomIntBetween(pauseMin, pauseMax));
}

function clickBusinessCards() {
  console.log('Starting clickBusinessCards...');

  group("Click Business Cards", () => {
    page.click("//span[text()='Business Cards']");
  });

  sleep(randomIntBetween(pauseMin, pauseMax));
}

function clickBrowseDesigns() {
  console.log('Starting clickBrowseDesigns...');

  group("Click Browse Designs", () => {
    page.waitForSelector("//a[text()='Browse designs']");
    page.click("//a[text()='Browse designs']");
  });

  sleep(randomIntBetween(pauseMin, pauseMax));
}

function clickDesign() {
  console.log('Starting clickDesign...');

  const designOption = page.waitForSelector(`li.design-tile >> nth=0`);
  console.log(designOption.textContent());
  
  group("Select Design Option", () => {
    page.click(`li.design-tile >> nth=0`)
  });

  sleep(randomIntBetween(pauseMin, pauseMax));
}

function clickCustomizeDesign() {
  console.log('Starting clickCustomizeDesign...');

  
  group("Click Customize Design", () => {
    page.click("//a[text()='Customize this design']");
  });

  sleep(randomIntBetween(pauseMin, pauseMax));
}

function clickNext() {
  console.log('Starting clickNext...');

  group("Click Next", () => {
    page.click("//button[text()='Next']");
  });

  sleep(randomIntBetween(pauseMin, pauseMax));
}

function continueWithoutBack() {
  console.log('Starting continueWithoutBack...');

  group("Click Continue Without Back", () => {
    page.click("//button[text()='Continue without back']");
  });

  sleep(randomIntBetween(pauseMin, pauseMax));
}

function customerReviewed() {
  console.log('Starting customerReviewed...');

  page.click("//input[@value='customerReviewed']");

  sleep(5);
  
  group("Click Customer Reviewed", () => {
    page.click("//button[text()='Continue']");
  });

  sleep(randomIntBetween(pauseMin, pauseMax));
}

function addToCart() {
  console.log('Starting addToCart...');

  group("Click Add to Cart", () => {
    page.click("//button[text()='Add to cart']");
  });

  sleep(randomIntBetween(pauseMin, pauseMax));
}

function continuePastRecommendations() {
  console.log('Starting continuePastRecommendations...');

  group("Click Continue (Recommendations)", () => {
    page.waitForSelector("//a[text()='Continue']");
    page.click("//a[text()='Continue']");
  });

  sleep(randomIntBetween(pauseMin, pauseMax));
}

function continueToCart() {
  console.log('Starting continueToCart...');

  group("Click Continue to Cart", () => {
    page.click("//a[text()='Continue to cart']");
  });

  sleep(randomIntBetween(pauseMin, pauseMax));
}

function goToCheckout() {
  console.log('Starting goToCheckout...');

  group("Click Checkout", () => {
    page.click("//a[text()='Checkout']");
  });

  sleep(randomIntBetween(pauseMin, pauseMax));
}
```
</details

I rebased #169 onto this branch, used the modified script, and saw the errors 100 times. However, when I used the modified script on #169, I saw the errors 29 times.

imiric:
@inancgumus It's difficult to judge the improvement based on that Vistaprint script since it's a very dynamic site and we use random sleeps in the script you're testing with, so the amount of requests will be different each time. Moreover, I see the `error fetching response body` message for blob URLs (e.g. `blob:https://www.vistaprint.ca/1a44f750-8ddd-4d3b-a454-8f1ee270946e`), which we should probably ignore for determining body size like we do for data URLs, since they're internal URLs used only in the browser.

This PR isn't meant to completely eliminate the `error fetching response body` message, caused by an underlying `No data found for resource with given identifier (-32000)` error, for which we don't have a fix yet. What it does is move response metric emission to the `LoadingFinished` event handler, which results in more accurate `data_received` metrics (you can confirm this within Chrome DevTools, or by packet capture).

E.g. on `main` running that example script:

```
data_received....................: 17 MB  737 kB/s
```

And on this branch rebased on current `main`:

```
data_received....................: 39 MB  754 kB/s
```

That said, it's surprising that you get a ~4x increase of the error on this branch, or that you have that many to begin with. My difference is +/- ~5, and usually not more than 10 in total on either branch.

I'll run some more tests and add ignoring of blob URLs.

inancgumus:
> It's difficult to judge the improvement based on that Vistaprint script since it's a very dynamic site and we use random sleeps in the script you're testing with, so the amount of requests will be different each time.
> That said, it's surprising that you get a ~4x increase of the error on this branch, or that you have that many to begin with. My difference is +/- ~5, and usually not more than 10 in total on either branch.

I agree 👍  I was measuring this in an earlier version of the main branch, things probably have changed since then 🤷 

> we should probably ignore for determining body size like we do for data URLs, since they're internal URLs used only in the browser

👍

imiric:
@inancgumus This is ready for review. Please test again when compared to `main`.

Like I mentioned above, comparing the amount of `error fetching response body` messages is unreliable, since the Vistaprint test script is so dynamic. For example, sometimes running it on `main` would fail after ~30s with the error `cannot call function on expression` and would finish at the `Click Add to Cart` stage, while running it on this branch would run for longer (~1m), but still fail at the `Click Continue (Recommendations)` stage, so it makes sense that more error messages would be produced.

Instead of focusing on the message count, you can confirm that `data_received` reporting is more accurate now, and that the message is gone when running the test suite.

I'm frankly unsure if we should merge this as it's difficult to judge the improvement or whether it's not breaking anything else, but from a metrics standpoint it seems like an improvement.

imiric:
@inancgumus WDYT about merging this?

imiric:
I pushed a rebase on `main`.

A smaller test that shows the improvement is the `fillform.js` example I've been testing for #200.

When running on current `main` (0f2ba32):

```shell
$ XK6_BROWSER_LOG=info XK6_HEADLESS=1 xk6 run examples/fillform.js
[...]
WARN[0001] error fetching response body for 'https://test.k6.io/': No data found for resource with given identifier (-32000)  category=cdp elapsed="0 ms" goroutine=75
ERRO[0003] unknown command or event "Page.frameScheduledNavigation"
ERRO[0003] unknown command or event "Page.frameClearedScheduledNavigation"
WARN[0003] error fetching response body for 'https://test.k6.io/my_messages.php': No data found for resource with given identifier (-32000)  category=cdp elapsed="0 ms" goroutine=75
ERRO[0007] unknown command or event "Page.frameScheduledNavigation"
ERRO[0007] unknown command or event "Page.frameClearedScheduledNavigation"
WARN[0007] error fetching response body for 'https://test.k6.io/login.php': No data found for resource with given identifier (-32000)  category=cdp elapsed="0 ms" goroutine=75
```

And on this branch (3295ede):

```shell
$ XK6_BROWSER_LOG=info XK6_HEADLESS=1 xk6 run examples/fillform.js
[...]
ERRO[0003] unknown command or event "Page.frameScheduledNavigation"
ERRO[0003] unknown command or event "Page.frameClearedScheduledNavigation"
ERRO[0007] unknown command or event "Page.frameScheduledNavigation"
ERRO[0007] unknown command or event "Page.frameClearedScheduledNavigation"
```

inancgumus:
OK let me review this today.

imiric:
And to confirm that the `data_received` metric is now more accurate, here's a simple example script:

```javascript
// simple.js
import launcher from 'k6/x/browser';
import { sleep } from 'k6';

export default function() {
    headless: true,
  });

  page.goto('https://test.k6.io/', { waitUntil: 'networkidle' });
  sleep(1);  // waitFor workaround

  page.close();
  browser.close();
}
```

Running this on `main` (0f2ba32) with `XK6_BROWSER_LOG=info xk6 run -o json=out-main.json examples/simple.js` and then filtering out the `data_received` samples:
```shell
$ jq -c '. | select(.type=="Point" and .metric == "data_received" and .data.tags.url)' out-main.json
{"type":"Point","data":{"time":"2022-02-04T06:44:31.082861+01:00","value":174,"tags":{"from_cache":"false","from_prefetch_cache":"false","from_service_worker":"false","group":"","method":"GET","proto":"http/1.1","scenario":"default","status":"200","url":"https://test.k6.io/"}},"metric":"data_received"}
{"type":"Point","data":{"time":"2022-02-04T06:44:31.287852+01:00","value":5076,"tags":{"from_cache":"false","from_prefetch_cache":"false","from_service_worker":"false","group":"","method":"GET","proto":"http/1.1","scenario":"default","status":"200","url":"https://test.k6.io/static/css/site.css"}},"metric":"data_received"}
{"type":"Point","data":{"time":"2022-02-04T06:44:31.87362+01:00","value":10112,"tags":{"from_cache":"false","from_prefetch_cache":"false","from_service_worker":"false","group":"","method":"GET","proto":"http/1.1","scenario":"default","status":"200","url":"https://test.k6.io/static/js/prisms.js"}},"metric":"data_received"}
```

Only the sizes:
```shell
$ jq -c '. | select(.type=="Point" and .metric == "data_received" and .data.tags.url) | {url: .data.tags.url, received: .data.value}' out-main.json
{"url":"https://test.k6.io/","received":174}
{"url":"https://test.k6.io/static/css/site.css","received":5076}
{"url":"https://test.k6.io/static/js/prisms.js","received":10112}
```

Summing up:
```shell
$ jq -c '. | select(.type=="Point" and .metric == "data_received" and .data.tags.url) | .data.value' out-main.json | paste -sd+ | bc
15362
```

Comparing this to `fix/90-nil-request-frame` (3295ede) with `XK6_BROWSER_LOG=info xk6 run -o json=out-117.json examples/simple.js`:

```shell
$ jq -c '. | select(.type=="Point" and .metric == "data_received" and .data.tags.url) | {url: .data.tags.url, received: .data.value}' out-117.json
{"url":"https://test.k6.io/","received":11277}
{"url":"https://test.k6.io/static/css/site.css","received":5076}
{"url":"https://test.k6.io/static/js/prisms.js","received":10112}
```

Sum:

```shell
$ jq -c '. | select(.type=="Point" and .metric == "data_received" and .data.tags.url) | .data.value' out-117.json | paste -sd+ | bc
26465
```

To confirm which one is correct:

```shell
$ jq -c '. | select(.type=="Point" and .metric == "data_received" and .data.tags.url) | .data.tags.url' out-main.json | xargs -I{} sh -c 'echo {}; curl -is {} | wc --bytes'
https://test.k6.io/
11277
https://test.k6.io/static/css/site.css
5076
https://test.k6.io/static/js/prisms.js
10112
```

:smile:

inancgumus (approved):
Overall, it's good, glad that we can finally get rid of these ancient error messages!

But I'd like to separate k6 concerns from the browser concerns as much as we can.

```golang
sampleTags := k6stats.IntoSampleTags(&tags)
k6stats.PushIfNotDone(m.ctx, state.Samples, k6stats.ConnectedSamples{
	...
})
```

So, would you like to move this part of the code to another function like:

```golang
func (m *NetworkManager) emitResponseMetrics(resp *Response) {
	...
	emitK6Metrics(...)
	...
}

func emitK6Metrics(...) {
	sampleTags := k6stats.IntoSampleTags(&tags)
	k6stats.PushIfNotDone( ... )
	...
	k6stats.PushIfNotDone( ... )
	...
}
```

inancgumus:
Thanks for the short `jq` tutorial as well, nice work! 😄 👍

inancgumus:
Btw, are we going to have a test for this?

imiric:
I'm not sure how to test this TBH without an E2E test like that script.

Hhmm though we might be able to get the metric for some `httpmultibin` endpoint and assert on that?

I'll look into it.

inancgumus:
> Hhmm though we might be able to get the metric for some `httpmultibin` endpoint and assert on that?

It sounds like a plan!

common/network_manager.go
```diff
@@ -391,6 +391,9 @@ func (m *NetworkManager) onLoadingFinished(event *network.EventLoadingFinished)
 		}
 	}
 	req.responseEndTiming = float64(event.Timestamp.Time().Unix()-req.timestamp.Unix()) * 1000
+	if req.url.Scheme != "data" {
```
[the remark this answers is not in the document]
imiric: Sure, resolved by 3369bfc.

common/response.go
```diff
@@ -191,19 +190,7 @@ func (r *Response) Body() goja.ArrayBuffer {
...
-// is unreliable, see https://github.com/ChromeDevTools/devtools-protocol/issues/12#issuecomment-306947275 .
 func (r *Response) bodySize() int64 {
-	if v, ok := r.headers["content-length"]; ok && len(v) > 0 {
-		cl, err := strconv.ParseInt(v[0], 10, 64)
-		if err == nil {
-			return cl
-		}
-		r.logger.Warnf("cdp", "error parsing content-length header: %s", err)
-	}
-
```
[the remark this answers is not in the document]
imiric: Yeah, I dropped the revert commit and this is now gone.

common/network_manager.go
```diff
@@ -391,6 +417,9 @@ func (m *NetworkManager) onLoadingFinished(event *network.EventLoadingFinished)
 		}
 	}
 	req.responseEndTiming = float64(event.Timestamp.Time().Unix()-req.timestamp.Unix()) * 1000
+	if req.url.Scheme != "data" && req.url.Scheme != "blob" {
```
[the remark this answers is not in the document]
imiric: Sure, resolved by 3369bfc.

common/network_manager.go
```diff
@@ -416,7 +445,8 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent, interc
 		frame = m.frameManager.getFrameByID(event.FrameID)
 	}
 	if frame == nil {
-		m.logger.Debugf("NetworkManager", "frame is nil")
+		m.logger.Debugf("NetworkManager:onRequest", "frame is nil url:%s method:%s type:%s fid:%s",
+			event.Request.URL, event.Request.Method, event.Initiator.Type, event.FrameID)
```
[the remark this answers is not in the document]
imiric: OK, resolved by a696100.

Though we really need to switch to structured logging... While informative, this is a mess to write.

common/response.go
```diff
@@ -197,7 +198,8 @@ func (r *Response) bodySize() int64 {
 	}
 
 	if err := r.fetchBody(); err != nil {
-		r.logger.Warnf("cdp", "error fetching response body for '%s': %s", r.url, err)
+		r.logger.Warnf("Response:bodySize:fetchBody",
+			"url:%s method:%s err:%s", err, r.url, r.request.method)
```
inancgumus, line 202: You forgot `err` at the end :)

common/response.go
```diff
@@ -205,7 +211,8 @@ func (r *Response) bodySize() int64 {
 	}
 
 	if err := r.fetchBody(); err != nil {
-		r.logger.Warnf("cdp", "error fetching response body for '%s': %s", r.url, err)
+		r.logger.Warnf("Response:bodySize:fetchBody",
+			"err:%s url:%s method:%s", err, r.url, r.request.method)
```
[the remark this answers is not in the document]
imiric: Sure, resolved by b26247d.

common/response.go
imiric: Fixed in b26247d.

imiric:
> I'd like to separate k6 concerns from the browser concerns as much as we can.

I think that makes sense, but this function is not the place to do it. All of these metrics are "k6 metrics". The custom ones we emit is mostly done from `onPageLifecycle()`.

inancgumus:
I don't mean that but never mind, no big deal.

common/network_manager.go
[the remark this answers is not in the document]
inancgumus: Yeah, I agree...

imiric:
@inancgumus A test for this would be similar to e6db54c9, and we decided to revert that because of `waitFor*` issues (#200). So let's wait until that is fixed to add a test for both #40 and this. It would likely be the same test, actually.

Let me know if there's anything pending here, otherwise let's merge it.

(Sorry for the last force push... It was only for adding a test in f29064e.)

inancgumus:
Fair enough, let's merge it 🚀  Maybe create an issue for testing both, so we don't forget? And maybe add a note next to #200.

PR #118, opened 2021-11-16
Fix nil pointer panic on Frame.networkIdleCtx

### LINKED ISSUE #109: Nil pointer: Network Idle Context Done
This one is coming from `doneCh := f.networkIdleCtx.Done()`.

I used Tom's script in #49 with `headless` mode `true`.

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1c372dc]
default   [--------------------------------------] 1 VUs  00m05.5s/10m0s  0/1 iters, 1 per VU
goroutine 2536 [running]:
github.com/grafana/xk6-browser/common.(*Frame).startNetworkIdleTimer.func1()
	github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame.go:290 +0x5c
created by github.com/grafana/xk6-browser/common.(*Frame).startNetworkIdleTimer
	github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame.go:288 +0x135
2021/11/12 13:52:46 [ERROR] exit status 2
```

imiric:
For very quick requests there's a race condition between when we [set `f.networkIdleCtx` to nil](https://github.com/grafana/xk6-browser/blob/baaf58caef73370f1659c30b3d3a989ddcd8da27/common/frame.go#L273) as part of the `FrameManager.requestStarted()` call, and when we [start the idle timer again in `FrameManager.requestFinished()`](https://github.com/grafana/xk6-browser/blob/baaf58caef73370f1659c30b3d3a989ddcd8da27/common/frame.go#L290).

This fix uses a channel instead of storing a mutex, which is safer.

Fixes #109

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -126,9 +126,7 @@ type Frame struct {
 
 	loadingStartedTime time.Time
 
-	networkIdleMu       sync.Mutex
-	networkIdleCtx      context.Context
-	networkIdleCancelFn context.CancelFunc
+	networkIdleCh chan struct{}
 
 	inflightRequestsMu sync.RWMutex
 	inflightRequests   map[network.RequestID]bool
@@ -153,6 +151,7 @@ func NewFrame(ctx context.Context, m *FrameManager, parentFrame *Frame, frameID 
 		utilityExecutionContextCh: make(chan bool, 1),
 		inflightRequests:          make(map[network.RequestID]bool),
 		currentDocument:           &DocumentInfo{},
+		networkIdleCh:             make(chan struct{}),
 	}
 }
 
@@ -266,12 +265,9 @@ func (f *Frame) recalculateLifecycle() {
 }
 
 func (f *Frame) stopNetworkIdleTimer() {
-	f.networkIdleMu.Lock()
-	defer f.networkIdleMu.Unlock()
-	if f.networkIdleCancelFn != nil {
-		f.networkIdleCancelFn()
-		f.networkIdleCtx = nil
-		f.networkIdleCancelFn = nil
+	select {
+	case f.networkIdleCh <- struct{}{}:
+	default:
 	}
 }
 
@@ -279,16 +275,11 @@ func (f *Frame) startNetworkIdleTimer() {
 	if f.hasLifecycleEventFired(LifecycleEventNetworkIdle) || f.detached {
 		return
 	}
-	if f.networkIdleCancelFn != nil {
-		f.networkIdleCancelFn()
-	}
-	networkIdleCtx, networkIdleCancelFn := context.WithCancel(f.ctx)
-	f.networkIdleMu.Lock()
-	f.networkIdleCtx, f.networkIdleCancelFn = networkIdleCtx, networkIdleCancelFn
-	f.networkIdleMu.Unlock()
+	f.stopNetworkIdleTimer()
 	go func() {
 		select {
-		case <-networkIdleCtx.Done():
+		case <-f.ctx.Done():
+		case <-f.networkIdleCh:
 		case <-time.After(LifeCycleNetworkIdleTimeout):
 			f.manager.frameLifecycleEvent(f.id, LifecycleEventNetworkIdle)
 		}
--- a/common/frame.go
+++ b/common/frame.go
@@ -282,15 +282,13 @@ func (f *Frame) startNetworkIdleTimer() {
 	if f.networkIdleCancelFn != nil {
 		f.networkIdleCancelFn()
 	}
+	networkIdleCtx, networkIdleCancelFn := context.WithCancel(f.ctx)
 	f.networkIdleMu.Lock()
-	f.networkIdleCtx, f.networkIdleCancelFn = context.WithCancel(f.ctx)
+	f.networkIdleCtx, f.networkIdleCancelFn = networkIdleCtx, networkIdleCancelFn
 	f.networkIdleMu.Unlock()
 	go func() {
-		f.networkIdleMu.Lock()
-		doneCh := f.networkIdleCtx.Done()
-		f.networkIdleMu.Unlock()
 		select {
-		case <-doneCh:
+		case <-networkIdleCtx.Done():
 		case <-time.After(LifeCycleNetworkIdleTimeout):
 			f.manager.frameLifecycleEvent(f.id, LifecycleEventNetworkIdle)
 		}


robingustafsson (approved):
LGTM

inancgumus (approved):
LGTM, good catch, however:

I feel like there's still a race condition here.

We might miss to call the cancel funcs because of the `if f.networkIdleCancelFn != nil` checks. These are not protected by locks.

It's because `stopNetworkIdleTimer` can set the cancel func to `nil` while `startNetworkIdleTimer` is still working. This can result in goroutine leaks for uncancelled funcs (as the context canceller fires up goroutines if we don't call the cancel funcs).

imiric:
Yeah, this doesn't fix the several race conditions here, just avoids this specific nil panic.

I'm wondering if we can do this without storing the `Context` on `Frame` (which is an anti-pattern; from [the docs](https://pkg.go.dev/context): "Do not store Contexts inside a struct type"). And this is not a "timer" anyway :)

We essentially only need this context to dispose of the goroutine that waits on `LifeCycleNetworkIdleTimeout`. Could it be done with a channel instead? I'll see if I can refactor it...

inancgumus:
> Yeah, this doesn't fix the several race conditions here, just avoids this specific nil panic.

Yep ;)


> I'm wondering if we can do this without storing the `Context` on `Frame` (which is an anti-pattern; from [the docs](https://pkg.go.dev/context): "Do not store Contexts inside a struct type"). And this is not a "timer" anyway :)

It's actually not an anti-pattern for _pro_s :) It's ok to store a context in a struct as long as it belongs to a single request-flow chain. Also see [this](https://github.com/golang/go/issues/22602) discussion.

> We essentially only need this context to dispose of the goroutine that waits on `LifeCycleNetworkIdleTimeout`. Could it be done with a channel instead? I'll see if I can refactor it...

What about this (_not yet tried it :)_)?

```golang
func (f *Frame) stopNetworkIdleTimer() {
	f.networkIdleMu.Lock()
	defer f.networkIdleMu.Unlock()
	if f.networkIdleCancelFn == nil {
		return
	}
	f.networkIdleCancelFn()
}


func (f *Frame) startNetworkIdleTimer() {
	if f.hasLifecycleEventFired(LifecycleEventNetworkIdle) || f.detached {
		return
	}
	go func() {
		f.networkIdleMu.Lock()
		{
			f.networkIdleCtx, f.networkIdleCancelFn = context.WithTimeout(f.ctx, LifeCycleNetworkIdleTimeout)
			defer f.networkIdleCancelFn()
			f.networkIdleMu.Unlock()
		}
		<-f.networkIdleCtx.Done()
		f.manager.frameLifecycleEvent(f.id, LifecycleEventNetworkIdle)
	}()
}
```

imiric:
Yeah, I've seen that discussion, but notice that the mentioned "good" use case for doing so is if it's still passed along as a parameter. Not what we're doing here with overwriting it, nil checks, etc. So I think our usage is what shouldn't be done. And you can see that it complicates things with locks, etc.

So if we can find a different approach it woud be better. I'll try out your suggestion if nothing else works. :)

inancgumus:
> Yeah, I've seen that discussion, but notice that the mentioned "good" use case for doing so is if it's still passed along as a parameter. Not what we're doing here with overwriting it, nil checks, etc. So I think our usage is what shouldn't be done. And you can see that it complicates things with locks, etc.
> 

Yeah, setting it to `nil` etc. seems like an anti-pattern.

> So if we can find a different approach it would be better. I'll try out your suggestion if nothing else works. :)

It will probably not work but it could with some tweaks, IDK :)

imiric:
@inancgumus Would something like 2f50e4d be a solution?

From my tests it seems to work fine for all the examples, but the updated script from Tom fails with:

```
ERRO[0015] unable to evaluate expression: Cannot find context with specified id (-32000)
default at file:///home/ivan/Projects/grafana/xk6-browser/examples/49-vistaprint-20211115.js:76:26(9)
        at go.k6.io/k6/js/common.Bind.func1 (native)
        at clickBusinessCards (file:///home/ivan/Projects/grafana/xk6-browser/examples/49-vistaprint-20211115.js:73:17(17))
        at file:///home/ivan/Projects/grafana/xk6-browser/examples/49-vistaprint-20211115.js:18:21(8)  executor=per-vu-iterations scenario=default source=stacktrace
panic: cannot create frame session: unable to add script to evaluate for isolated world: context canceled
running (00m13.9s), 1/1 VUs, 0 complete and 0 interrupted iterations
goroutine 132 [running]:-------------------------] 1 VUs  00m13.9s/10m0s  0/1 iters, 1 per VU
go.k6.io/k6/js/common.Throw(0x111b3c0, {0x15802e0, 0xc00009bd40})
        go.k6.io/k6@v0.35.0/js/common/util.go:36 +0x55
github.com/grafana/xk6-browser/common.k6Throw({0x159fe28, 0xc000e0f740}, {0x132303d, 0x1f}, {0xc0039a3f30, 0x1, 0x1})
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/helpers.go:338 +0x1ae
github.com/grafana/xk6-browser/common.(*FrameSession).onAttachedToTarget(0xc0025e7800, 0xc000e12720)
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame_session.go:653 +0x1d4
github.com/grafana/xk6-browser/common.(*FrameSession).initEvents.func1()
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame_session.go:217 +0x2f2
created by github.com/grafana/xk6-browser/common.(*FrameSession).initEvents
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame_session.go:180 +0xe7
```

This also happens on `main` so it's not related to this change, just mentioning it here. It seems like a new issue. :disappointed:

inancgumus:
> @inancgumus Would something like [2f50e4d](https://github.com/grafana/xk6-browser/commit/2f50e4dbf37e644870d003c5ec09d68f65b69fe1) be a solution?
> 
> From my tests, it seems to work fine for all the examples, but the updated script from Tom fails with:
> 

Yep, this is the same error that I've seen so far (_with Tom's script_).

inancgumus (approved):
LGTM!

Issue #126, opened 2021-11-18
Invalid memory in frame.currentDocument.request


inancgumus:
After fixing #109 with #118, a new issue has popped up.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x1c3239f]
default   [--------------------------------------] 1 VUs  00m08.1s/10m0s  0/1 iters, 1 per VU
goroutine 85 [running]:
github.com/grafana/xk6-browser/common.(*Frame).clearLifecycle(0xc002b9c480)
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame.go:201 +0x17f
github.com/grafana/xk6-browser/common.(*FrameManager).frameNavigated(0xc001c9e120, {0xc00190efe0, 0x0}, {0x0, 0x0}, {0xc00190f000, 0x20}, {0x0, 0x0}, {0xc000d35f20, ...}, ...)
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame_manager.go:227 +0x588
github.com/grafana/xk6-browser/common.(*FrameSession).onFrameNavigated(0xc00138a480, 0xc000493f80, 0x20)
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame_session.go:509 +0x10d
github.com/grafana/xk6-browser/common.(*FrameSession).initEvents.func1()
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame_session.go:195 +0x3ec
created by github.com/grafana/xk6-browser/common.(*FrameSession).initEvents
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame_session.go:180 +0xe7
2021/11/18 17:17:40 [ERROR] exit status 2
```

This time, it's coming from here (`frame.currentDocument.request` becomes `nil`):

https://github.com/grafana/xk6-browser/blob/b7f55930199bebb70ee553bc53627f8965aca866/common/frame.go#L201

According to my research, `currentDocument`'s request can be nil. For example:

https://github.com/grafana/xk6-browser/blob/b7f55930199bebb70ee553bc53627f8965aca866/common/frame_manager.go#L197-L225

A fix is coming shortly...

imiric:
This seems like a duplicate of #119, and I think I've seen the stack trace before that.

[...]

Issue #49, opened 2021-11-01
Cannot find context with specified id


inancgumus:
Reported by Tom on Oct, 30th:

* I was getting nil pointer exceptions here and somewhere here as a result of frame being nil. Moving the if blocks inside the preceding if `frame != nil` block got me past these. Not sure what the implications are beyond that though, but I guess having a nil frame is expected if there's already code checking for that condition?! :sweat_smile:
* Currently seeing a `ERRO[0026] unable to evaluate expression: Cannot find context with specified id (-32000)`. Not sure why - I think this was working before I pulled the latest `xk6-browser`.
* Also seeing a `panic: send on closed channel` but I think that might be happening as a result of earlier errors (like the above), i.e. during shutdown.

<details>
<summary>Example stack trace</summary>

```
panic: send on closed channel

goroutine 46 [running]:
go.k6.io/k6/stats.PushIfNotDone(...)
        go.k6.io/k6@v0.34.1/stats/stats.go:433
github.com/grafana/xk6-browser/common.(*NetworkManager).emitResponseReceived(0xc0010c0280, 0xc0006085b0)
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/network_manager.go:131 +0x87a
github.com/grafana/xk6-browser/common.(*NetworkManager).onResponseReceived(0xc0010c0280, 0xc0005de8c0)
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/network_manager.go:356 +0xc5
github.com/grafana/xk6-browser/common.(*NetworkManager).handleEvents(0xc0010c0280, 0xc00057a5a0)
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/network_manager.go:245 +0x1cf
github.com/grafana/xk6-browser/common.(*NetworkManager).initEvents.func1()
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/network_manager.go:220 +0x37
created by github.com/grafana/xk6-browser/common.(*NetworkManager).initEvents
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/network_manager.go:219 +0x125
```

</details>

Here's the JS code I'm running, with the failure happening in `clickBrowseDesigns` (which is odd as it is very similar to the preceding `clickBusinessCards`).

<details>

<summary>Script</summary>

```javascript
import { sleep, group } from 'k6';
[...]

PR #127, opened 2021-11-18
Fix clearLifecycle bug of nil request

inancgumus:
See: #126 for the explanation.

This also fixes Tom's oldest script problems from #49 while in `headless: true` mode. But still no luck with `headless: false` mode.

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -196,11 +196,12 @@ func (f *Frame) clearLifecycle() {
 	// in f.inflightRequests
 	f.inflightRequestsMu.Lock()
 	{
+		// currentDocument may not always have a request
+		// associated with it. see: frame_manager.go
+		cdr := f.currentDocument.request
+
 		inflightRequests := make(map[network.RequestID]bool)
 		for req := range f.inflightRequests {
-			// currentDocument may not always have a request
-			// associated with it. see: frame_manager.go
-			cdr := f.currentDocument.request
 			if cdr != nil && req != cdr.requestID {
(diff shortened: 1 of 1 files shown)


imiric (commented):
While this fixes the panic, do you think we should look into why we're setting it to nil and if we can synchronize `Frame.clearLifecycle()` with `FrameManager.frameNavigated()`?

inancgumus:
> While this fixes the panic, do you think we should look into why we're setting it to nil and if we can synchronize `Frame.clearLifecycle()` with `FrameManager.frameNavigated()`?

I'm not sure why we're setting it to `nil` but Robin explicitly set it to `nil`.
So it should be the expected behavior.

robingustafsson:
@inancgumus The `currentDocument.request` is set to `nil` when we have received a `Page.frameNavigated` event without a prior `Network.requestWillBeSent` (where a `loaderID` is set and `type` is "Document", meaning we're seeing a new document being loaded).

inancgumus:
> @inancgumus The `currentDocument.request` is set to `nil` when we have received a `Page.frameNavigated` event without a prior `Network.requestWillBeSent` (where a `loaderID` is set and `type` is "Document", meaning we're seeing a new document being loaded).

@imiric In `frame.clearLifecycle` we remove the inflight requests by looking for a `requestID` in a new Document, which won't exist. I think this PR fixes the right thing.

imiric (approved):
Sounds good :shipit:

Issue #125, opened 2021-11-18
Go panic: cannot create frame session


imiric:
We've been noticing this error recently from Tom's updated script:

```
ERRO[0014] unable to evaluate expression: Cannot find context with specified id (-32000)
running at reflect.methodValueCall (native)
default at file:///home/ivan/Projects/grafana/xk6-browser/examples/49-vistaprint-20211115.js:76:26(9)
        at go.k6.io/k6/js/common.Bind.func1 (native)
        at clickBusinessCards (file:///home/ivan/Projects/grafana/xk6-browser/examples/49-vistaprint-20211115.js:73:17(17))
        at file:///home/ivan/Projects/grafana/xk6-browser/examples/49-vistaprint-20211115.js:18:21(8)  executor=per-vu-iterations scenario=default source=stacktrace
WARN[0014]                                               category=cdp error="error fetching response body: context canceled" method=GET url="https://www.vistaprint.ca/visage/latest/v5/all/library/styles/components/core.ef3572a618c168760ba19432b84f7447.css"
panic: cannot create frame session: unable to add script to evaluate for isolated world: context canceled
running (00m13.4s), 1/1 VUs, 0 complete and 0 interrupted iterations
goroutine 52 [running]:--------------------------] 1 VUs  00m13.4s/10m0s  0/1 iters, 1 per VU
go.k6.io/k6/js/common.Throw(0x11203c0, {0x1585ca0, 0xc003298aa0})
        go.k6.io/k6@v0.35.0/js/common/util.go:36 +0x55
github.com/grafana/xk6-browser/common.k6Throw({0x15a57a8, 0xc0037d43c0}, {0x1328705, 0x1f}, {0xc002b49f30, 0x1, 0x1})
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/helpers.go:338 +0x1ae
github.com/grafana/xk6-browser/common.(*FrameSession).onAttachedToTarget(0xc003292280, 0xc003288400)
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame_session.go:653 +0x1d4
github.com/grafana/xk6-browser/common.(*FrcaameSession).initEvents.func1()
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame_session.go:217 +0x2f2
created by github.com/grafana/xk6-browser/common.(*FrameSession).initEvents
[...]

PR #129, opened 2021-11-22
Fix/125 cannot create frame session (context canceled)

inancgumus:
### Update (2021-12-23)

I added a few more bug fixes after [receiving another script that fails with this error](https://community.k6.io/t/unable-to-properly-navigate-to-url/2567/4) (_from a user in our community_). Now we no longer try to attach a frame or worker if the browser is already closed. These commits solved the user's issues: 77949eb and a3aea00.

### Update (2021-12-20)

This PR only focuses on fixing and gracefully handling _"cannot create frame session"_ problem in #125 with this afe4550c commit. Later on, within other commits, it fixes related data race issues, makes some refactorings, and cleans things up.

The problem was about the propagation of #49 into this one. We throw an error when we cannot find a context with a given id while a new `FrameSession` is trying to boot up. So the extension is already in a terminating state but `FrameSession` doesn't know about it. This PR prevents `FrameSession` to rethrow if the context was canceled and instead log.

However, it's not focusing on fixing #49 as it's a whole world of a different problem, and we don't have a good fix for it yet (see: 6382e42 and d4e3900).

### First Update

**In this PR:**
* I changed the handling of execution contexts.
* Fixed most of the data races.
* Refactored some parts of the code.
* Fixed _cannot create frame session (context canceled)_ problem.
* ~Skipped creating frame sessions for frames with empty URLs (such as _about:blank_)~

**With these, this PR seemed to fix the problems to some extent:**
* Tom's script was failing on the 2nd group before, `clickBusinessCards`—**16% of the whole script**
* Now it works until the `customerReviewed` group—**%67 of the whole script**

I discovered some reasons why it still fails.

**I noticed we receive [deprecated events](https://github.com/chromium/chromium/commit/4005fcf1c724629b63ac6fc390d40a2f8f40be3d) like**:
* `Page.frameScheduledNavigation`
* `Page.frameClearedScheduledNavigation`

So, I believe the last steps don't work because of these and we can create new issues for handling them. So I let them return as errors:

```
ERRO[0094] unknown command or event "Page.frameScheduledNavigation"
ERRO[0094] unknown command or event "Page.frameClearedScheduledNavigation"
```

### DIFF
--- a/common/element_handle.go
+++ b/common/element_handle.go
@@ -238,8 +238,12 @@ func (h *ElementHandle) checkHitTargetAt(apiCtx context.Context, p Position) (bo
 			return injected.checkHitTargetAt(node, point);
 		}
 	`)
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: true,
+	}
 	result, err := h.execCtx.evaluate(
-		apiCtx, true, true, pageFn, []goja.Value{
+		apiCtx, opts, pageFn, []goja.Value{
 			rt.ToValue(injected),
 			rt.ToValue(h),
 			rt.ToValue(p),
@@ -270,8 +274,12 @@ func (h *ElementHandle) checkElementState(apiCtx context.Context, state string) 
 			return injected.checkElementState(node, state);
 		}
 	`)
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: true,
+	}
 	result, err := h.execCtx.evaluate(
-		apiCtx, true, true, pageFn, []goja.Value{
+		apiCtx, opts, pageFn, []goja.Value{
 			rt.ToValue(injected),
 			rt.ToValue(h),
 			rt.ToValue(state),
@@ -397,8 +405,12 @@ func (h *ElementHandle) dispatchEvent(apiCtx context.Context, typ string, eventI
 			injected.dispatchEvent(node, type, eventInit);
 		}
 	`)
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: true,
+	}
 	_, err = h.execCtx.evaluate(
-		apiCtx, true, true, pageFn, []goja.Value{
+		apiCtx, opts, pageFn, []goja.Value{
 			rt.ToValue(injected),
 			rt.ToValue(h),
 			rt.ToValue(typ),
@@ -421,8 +433,12 @@ func (h *ElementHandle) fill(apiCtx context.Context, value string) (interface{},
 				return injected.fill(node, value);
 			}
 		`)
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: true,
+	}
 	result, err := h.execCtx.evaluate(
-		apiCtx, true, true, pageFn, []goja.Value{
+		apiCtx, opts, pageFn, []goja.Value{
 			rt.ToValue(injected),
 			rt.ToValue(h),
 			rt.ToValue(value),
@@ -450,8 +466,12 @@ func (h *ElementHandle) focus(apiCtx context.Context, resetSelectionIfNotFocused
 			return injected.focusNode(node, resetSelectionIfNotFocused);
 		}
 	`)
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: true,
+	}
 	result, err := h.execCtx.evaluate(
-		apiCtx, true, true, pageFn, []goja.Value{
+		apiCtx, opts, pageFn, []goja.Value{
 			rt.ToValue(injected),
 			rt.ToValue(h),
 			rt.ToValue(resetSelectionIfNotFocused),
@@ -473,7 +493,11 @@ func (h *ElementHandle) getAttribute(apiCtx context.Context, name string) (inter
 		return element.getAttribute('` + name + `');
 	}`
 	rt := k6common.GetRuntime(apiCtx)
-	return h.execCtx.evaluate(apiCtx, true, true, rt.ToValue(js), rt.ToValue(h))
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: true,
+	}
+	return h.execCtx.evaluate(apiCtx, opts, rt.ToValue(js), rt.ToValue(h))
 }
 
 func (h *ElementHandle) hover(apiCtx context.Context, p *Position) error {
@@ -485,7 +509,11 @@ func (h *ElementHandle) innerHTML(apiCtx context.Context) (interface{}, error) {
 	js := `(element) => {
 		return element.innerHTML;
 	}`
-	return h.execCtx.evaluate(apiCtx, true, true, rt.ToValue(js), rt.ToValue(h))
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: true,
+	}
+	return h.execCtx.evaluate(apiCtx, opts, rt.ToValue(js), rt.ToValue(h))
 }
 
 func (h *ElementHandle) innerText(apiCtx context.Context) (interface{}, error) {
@@ -493,7 +521,11 @@ func (h *ElementHandle) innerText(apiCtx context.Context) (interface{}, error) {
 	js := `(element) => {
 		return element.innerText;
 	}`
-	return h.execCtx.evaluate(apiCtx, true, true, rt.ToValue(js), rt.ToValue(h))
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: true,
+	}
+	return h.execCtx.evaluate(apiCtx, opts, rt.ToValue(js), rt.ToValue(h))
 }
 
 func (h *ElementHandle) inputValue(apiCtx context.Context) (interface{}, error) {
@@ -504,7 +536,11 @@ func (h *ElementHandle) inputValue(apiCtx context.Context) (interface{}, error) 
 		}
 		return element.value;
 	}`
-	return h.execCtx.evaluate(apiCtx, true, true, rt.ToValue(js), rt.ToValue(h))
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: true,
+	}
+	return h.execCtx.evaluate(apiCtx, opts, rt.ToValue(js), rt.ToValue(h))
 }
 
 func (h *ElementHandle) isChecked(apiCtx context.Context, timeout time.Duration) (bool, error) {
@@ -543,8 +579,12 @@ func (h *ElementHandle) offsetPosition(apiCtx context.Context, offset *Position)
 			return injected.getElementBorderWidth(node);
 		}
 	`)
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: true,
+	}
 	result, err := h.execCtx.evaluate(
-		apiCtx, true, true, pageFn, []goja.Value{
+		apiCtx, opts, pageFn, []goja.Value{
 			rt.ToValue(injected),
 			rt.ToValue(h),
 		}...)
@@ -709,8 +749,12 @@ func (h *ElementHandle) selectOption(apiCtx context.Context, values goja.Value) 
 				return injected.selectOptions(node, values);
 			}
 		`)
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: false,
+	}
 	result, err := h.execCtx.evaluate(
-		apiCtx, true, false, pageFn, []goja.Value{
+		apiCtx, opts, pageFn, []goja.Value{
 			rt.ToValue(injected),
 			rt.ToValue(h),
 			rt.ToValue(convValues),
@@ -738,8 +782,12 @@ func (h *ElementHandle) selectText(apiCtx context.Context) error {
 				return injected.selectText(node);
 			}
 		`)
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: true,
+	}
 	result, err := h.execCtx.evaluate(
-		apiCtx, true, true, pageFn, []goja.Value{
+		apiCtx, opts, pageFn, []goja.Value{
 			rt.ToValue(injected),
 			rt.ToValue(h),
 		}...)
@@ -789,7 +837,11 @@ func (h *ElementHandle) textContent(apiCtx context.Context) (interface{}, error)
 	js := `(element) => {
 		return element.textContent;
 	}`
-	return h.execCtx.evaluate(apiCtx, true, true, rt.ToValue(js), rt.ToValue(h))
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: true,
+	}
+	return h.execCtx.evaluate(apiCtx, opts, rt.ToValue(js), rt.ToValue(h))
 }
 
 func (h *ElementHandle) typ(apiCtx context.Context, text string, opts *KeyboardOptions) error {
@@ -811,7 +863,11 @@ func (h *ElementHandle) waitAndScrollIntoViewIfNeeded(apiCtx context.Context, fo
 			element.scrollIntoViewIfNeeded(true);
 			return [window.scrollX, window.scrollY];
 		}`)
-		return h.execCtx.evaluate(apiCtx, true, true, pageFn, rt.ToValue(h))
+		opts := evaluateOptions{
+			forceCallable: true,
+			returnByValue: true,
+		}
+		return h.execCtx.evaluate(apiCtx, opts, pageFn, rt.ToValue(h))
 	}
 	actFn := elementHandleActionFn(h, []string{"visible", "stable"}, fn, force, noWaitAfter, timeout)
 	_, err := callApiWithTimeout(h.ctx, actFn, timeout)
@@ -832,8 +888,12 @@ func (h *ElementHandle) waitForElementState(apiCtx context.Context, states []str
 			return injected.waitForElementStates(node, states, timeout);
 		}
 	`)
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: true,
+	}
 	result, err := h.execCtx.evaluate(
-		apiCtx, true, true, pageFn, []goja.Value{
+		apiCtx, opts, pageFn, []goja.Value{
 			rt.ToValue(injected),
 			rt.ToValue(h),
 			rt.ToValue(states),
@@ -873,8 +933,12 @@ func (h *ElementHandle) waitForSelector(apiCtx context.Context, selector string,
 			return injected.waitForSelector(selector, scope, strict, state, 'raf', timeout, ...args);
 		}
 	`)
+	eopts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: false,
+	}
 	result, err := h.execCtx.evaluate(
-		apiCtx, true, false, pageFn, []goja.Value{
+		apiCtx, eopts, pageFn, []goja.Value{
 			rt.ToValue(injected),
 			rt.ToValue(parsedSelector),
 			rt.ToValue(h),
@@ -1166,7 +1230,11 @@ func (h *ElementHandle) OwnerFrame() api.Frame {
 			return injected.getDocumentElement(node);
 		}
 	`)
-	res, err := h.execCtx.evaluate(h.ctx, true, false, pageFn, []goja.Value{rt.ToValue(injected), rt.ToValue(h)}...)
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: false,
+	}
+	res, err := h.execCtx.evaluate(h.ctx, opts, pageFn, []goja.Value{rt.ToValue(injected), rt.ToValue(h)}...)
 	if err != nil {
 		k6common.Throw(rt, fmt.Errorf("failed getting document element: %w", err))
 	}
@@ -1226,8 +1294,12 @@ func (h *ElementHandle) Query(selector string) api.ElementHandle {
 			return injected.querySelector(selector, scope || document, false);
 		}
 	`)
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: false,
+	}
 	result, err := h.execCtx.evaluate(
-		h.ctx, true, false, pageFn, []goja.Value{
+		h.ctx, opts, pageFn, []goja.Value{
 			rt.ToValue(injected),
 			rt.ToValue(parsedSelector),
 			rt.ToValue(h),
@@ -1267,8 +1339,12 @@ func (h *ElementHandle) QueryAll(selector string) []api.ElementHandle {
 			return injected.querySelectorAll(selector, scope || document, false);
 		}
 	`)
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: false,
+	}
 	result, err := h.execCtx.evaluate(
-		h.ctx, true, false, pageFn, []goja.Value{
+		h.ctx, opts, pageFn, []goja.Value{
 			rt.ToValue(injected),
 			rt.ToValue(parsedSelector),
 			rt.ToValue(h),
--- a/common/execution_context.go
+++ b/common/execution_context.go
@@ -36,7 +36,12 @@ import (
 
 const evaluationScriptURL = "__xk6_browser_evaluation_script__"
 
-var /* const */ sourceURLRegex = regexp.MustCompile(`^(?s)[\040\t]*//[@#] sourceURL=\s*(\S*?)\s*$`)
+var sourceURLRegex = regexp.MustCompile(`^(?s)[\040\t]*//[@#] sourceURL=\s*(\S*?)\s*$`)
+
+const (
+	mainExecutionContext    = "main"
+	utilityExecutionContext = "utility"
+)
 
 // ExecutionContext represents a JS execution context
 type ExecutionContext struct {
--- a/common/execution_context.go
+++ b/common/execution_context.go
@@ -45,6 +45,10 @@ const (
 	utilityWorld executionWorld = "utility"
 )
 
+func (ew executionWorld) valid() bool {
+	return ew == mainWorld || ew == utilityWorld
+}
+
 type evaluateOptions struct {
 	forceCallable, returnByValue bool
 }
--- a/common/execution_context.go
+++ b/common/execution_context.go
@@ -39,8 +39,8 @@ const evaluationScriptURL = "__xk6_browser_evaluation_script__"
 var sourceURLRegex = regexp.MustCompile(`^(?s)[\040\t]*//[@#] sourceURL=\s*(\S*?)\s*$`)
 
 const (
-	mainExecutionContext    = "main"
-	utilityExecutionContext = "utility"
+	mainWorld    = "main"
+	utilityWorld = "utility"
 )
 
 // ExecutionContext represents a JS execution context
--- a/common/execution_context.go
+++ b/common/execution_context.go
@@ -164,7 +164,10 @@ func (e *ExecutionContext) adoptElementHandle(eh *ElementHandle) (*ElementHandle
 
 // evaluate will evaluate provided callable within this execution context
 // and return by value or handle
-func (e *ExecutionContext) evaluate(apiCtx context.Context, opts evaluateOptions, pageFunc goja.Value, args ...goja.Value) (res interface{}, err error) {
+func (e *ExecutionContext) evaluate(
+	apiCtx context.Context,
+	opts evaluateOptions, pageFunc goja.Value, args ...goja.Value,
+) (res interface{}, err error) {
 	e.logger.Debugf(
 		"ExecutionContext:evaluate",
 		"sid:%s stid:%s fid:%s ectxid:%d furl:%q %s",
--- a/common/execution_context.go
+++ b/common/execution_context.go
@@ -38,9 +38,11 @@ const evaluationScriptURL = "__xk6_browser_evaluation_script__"
 
 var sourceURLRegex = regexp.MustCompile(`^(?s)[\040\t]*//[@#] sourceURL=\s*(\S*?)\s*$`)
 
+type executionWorld string
+
 const (
-	mainWorld    = "main"
-	utilityWorld = "utility"
+	mainWorld    executionWorld = "main"
+	utilityWorld executionWorld = "utility"
 )
 
 type evaluateOptions struct {
--- a/common/execution_context.go
+++ b/common/execution_context.go
@@ -43,6 +43,14 @@ const (
 	utilityWorld = "utility"
 )
 
+type evaluateOptions struct {
+	forceCallable, returnByValue bool
+}
+
+func (ea evaluateOptions) String() string {
+	return fmt.Sprintf("forceCallable:%t returnByValue:%t", ea.forceCallable, ea.returnByValue)
+}
+
 // ExecutionContext represents a JS execution context
 type ExecutionContext struct {
 	ctx            context.Context
@@ -150,23 +158,19 @@ func (e *ExecutionContext) adoptElementHandle(eh *ElementHandle) (*ElementHandle
 
 // evaluate will evaluate provided callable within this execution context
 // and return by value or handle
-func (e *ExecutionContext) evaluate(
-	apiCtx context.Context,
-	forceCallable bool, returnByValue bool,
-	pageFunc goja.Value, args ...goja.Value,
-) (res interface{}, err error) {
+func (e *ExecutionContext) evaluate(apiCtx context.Context, opts evaluateOptions, pageFunc goja.Value, args ...goja.Value) (res interface{}, err error) {
 	e.logger.Debugf(
 		"ExecutionContext:evaluate",
-		"sid:%s stid:%s fid:%s ectxid:%d furl:%q forceCallable:%t returnByvalue:%t",
-		e.sid, e.stid, e.fid, e.id, e.furl, forceCallable, returnByValue)
+		"sid:%s stid:%s fid:%s ectxid:%d furl:%q %s",
+		e.sid, e.stid, e.fid, e.id, e.furl, opts)
 
 	suffix := `//# sourceURL=` + evaluationScriptURL
 
 	var (
-		isCallable = forceCallable
+		isCallable = opts.forceCallable
 		expression = pageFunc.ToString().String()
 	)
-	if !forceCallable {
+	if !isCallable {
 		_, isCallable = goja.AssertFunction(pageFunc)
 	}
 	if !isCallable {
@@ -181,7 +185,7 @@ func (e *ExecutionContext) evaluate(
 		)
 		action := runtime.Evaluate(expressionWithSourceURL).
 			WithContextID(e.id).
-			WithReturnByValue(returnByValue).
+			WithReturnByValue(opts.returnByValue).
 			WithAwaitPromise(true).
 			WithUserGesture(true)
 		if remoteObject, exceptionDetails, err = action.Do(cdp.WithExecutor(apiCtx, e.session)); err != nil {
@@ -193,7 +197,7 @@ func (e *ExecutionContext) evaluate(
 		if remoteObject == nil {
 			return
 		}
-		if returnByValue {
+		if opts.returnByValue {
 			res, err = valueFromRemoteObject(apiCtx, remoteObject)
 			if err != nil {
 				return nil, fmt.Errorf("cannot extract value from remote object (%s): %w", remoteObject.ObjectID, err)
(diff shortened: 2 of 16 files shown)


inancgumus:
@robingustafsson, do you have any idea why **the session gets closed first**:

Closed session ID: `EBD975C5F3C41A8557C46D7AB3EFD06A`

```
INFO[0005] sid=1241375EAFBAD72F5ED96C4053A617FE method="Network.getResponseBody"  category="\x1b[35mSession:Execute\x1b[0m" elapsed="0 ms" goroutine=44
INFO[0005] sid=1241375EAFBAD72F5ED96C4053A617FE method="Network.getResponseBody"  category="\x1b[35mSession:Execute:s.conn.send\x1b[0m" elapsed="0 ms" goroutine=44
INFO[0005] sid=1241375EAFBAD72F5ED96C4053A617FE cannot unmarshal: unknown command or event ""  category="\x1b[35mSession:readLoop:<-s.readCh\x1b[0m" elapsed="0 ms" goroutine=38

ERRO[0005] sid:EBD975C5F3C41A8557C46D7AB3EFD06A wsURL:"ws://127.0.0.1:60570/devtools/browser/7ba29ff7-6edb-48c7-a03e-95b3ce0e5941", closeSession  category="\x1b[35mConnection:recvLoop:EventDetachedFromTarget\x1b[0m" elapsed="0 ms" goroutine=68
ERRO[0005] sid=EBD975C5F3C41A8557C46D7AB3EFD06A wsURL=ws://127.0.0.1:60570/devtools/browser/7ba29ff7-6edb-48c7-a03e-95b3ce0e5941  category="\x1b[35mConnection:closeSession\x1b[0m" elapsed="0 ms" goroutine=68
INFO[0005] sid=EBD975C5F3C41A8557C46D7AB3EFD06A          category="\x1b[35mSession:close\x1b[0m" elapsed="0 ms" goroutine=68
```

And then, `Page.addScriptToEvaluateOnNewDocument` message tries to inject the script to the closed session:

```
ERRO[0015] wsURL:"ws://127.0.0.1:60570/devtools/browser/7ba29ff7-6edb-48c7-a03e-95b3ce0e5941" sid:1241375EAFBAD72F5ED96C4053A617FE, msg err:Cannot find context with specified id (-32000)  category="\x1b[35mConnection:send\x1b[0m" elapsed="0 ms" goroutine=30
ERRO[0015] unable to evaluate expression: Cannot find context with specified id (-32000)
	at file:///Users/inanc/grafana/tom-12nov.js:78:26(15)
	at go.k6.io/k6/js/common.Bind.func1 (native)
	at clickBusinessCards (file:///Users/inanc/grafana/tom-12nov.js:74:17(17))
	at file:///Users/inanc/grafana/tom-12nov.js:44:21(8)  executor=per-vu-iterations scenario=default source=stacktrace
ERRO[0015] sid=EBD975C5F3C41A8557C46D7AB3EFD06A method="Page.addScriptToEvaluateOnNewDocument", returns  category="\x1b[35mSession:Execute:<-evCancelCtx.Done()\x1b[0m" elapsed="1 ms" goroutine=970
ERRO[0015] returning, ctx.Err: "context canceled"        category="\x1b[35mconnection:sendLoop\x1b[0m" elapsed="0 ms" goroutine=69
ERRO[0015] wsURL:"ws://127.0.0.1:60570/devtools/browser/7ba29ff7-6edb-48c7-a03e-95b3ce0e5941" sid:EBD975C5F3C41A8557C46D7AB3EFD06A err:context canceled  category="\x1b[35mConnection:send:<-c.ctx.Done()\x1b[0m" elapsed="0 ms" goroutine=47
INFO[0015] sid:EBD975C5F3C41A8557C46D7AB3EFD06A tid:AC46CC93BEE8796F78585F9692325947 err:unable to add script to evaluate for isolated world: context canceled  category="\x1b[35mNewFrameSession:initIsolatedWorld\x1b[0m" elapsed="0 ms" goroutine=47
```

Resulting in:
`panic: cannot create frame session: unable to add script to evaluate for isolated world: context canceled`

robingustafsson:
Can't say for sure from just the those log lines you've posted.

The close call says that we received a `EventDetachedFromTarget`, what target did we detach from?

Regarding the resulting error, the only way to arrive at that error message `unable to add script to evaluate for isolated world` is when creating a new FrameSession for an iframe (well, besides the main frame session).

So could this be in relation to the destruction of an iframe at the same time as a new iframe is created perhaps and we handle that poorly right now? 🤔

inancgumus:
> Can't say for sure from just the those log lines you've posted.
> 
> The close call says that we received a `EventDetachedFromTarget`, what target did we detach from?
> 

@robingustafsson I’ll try to find out the target tomorrow and let you know. Thanks for pointing it out!

> Regarding the resulting error, the only way to arrive at that error message `unable to add script to evaluate for isolated world` is when creating a new FrameSession for an iframe (well, besides the main frame session).
> 
> So could this be in relation to the destruction of an iframe at the same time as a new iframe is created perhaps and we handle that poorly right now? 🤔

Thanks! I read about the article you linked in the code about isolated contexts, and yeah, I’m suspicious of an iframe handling issue. I’ll first try to create a test (if not exists) if I can.

inancgumus:
> The close call says that we received a `EventDetachedFromTarget`, what target did we detach from?

@robingustafsson
Can you point me in the code where I should investigate next?

* I've enabled the `targetID` in session logs.
* The same targetID and sessionID are getting closed.
  * TargetID (tid): A3090F4ECB94747A6660763EA5A92D8B
  * SessionID (sid): C2EEAA73BF1B1ECE0FC2A188FA7A8E94
* Then we try to inject a script to a closed target.

```
INFO[0008] <- {"method":"Network.loadingFailed","params":{"requestId":"4352BADD39733710C3721B3C17C0F012","timestamp":107412.566322,"type":"Document","errorText":"net::ERR_ABORTED","canceled":true},"sessionId":"28234C8BA995F6FE6E9CDF4C01592852"}  category="cdp:recv" elapsed="2 ms" goroutine=65
INFO[0008] <- {"method":"Target.detachedFromTarget","params":{"sessionId":"C2EEAA73BF1B1ECE0FC2A188FA7A8E94","targetId":"A3090F4ECB94747A6660763EA5A92D8B"},"sessionId":"28234C8BA995F6FE6E9CDF4C01592852"}  category="cdp:recv" elapsed="0 ms" goroutine=65
INFO[0008] sid:C2EEAA73BF1B1ECE0FC2A188FA7A8E94 tid:A3090F4ECB94747A6660763EA5A92D8B wsURL:"ws://127.0.0.1:50119/devtools/browser/7a9081be-5e4e-4346-bf52-96e1a04c2cbe", closeSession  category="Connection:recvLoop:EventDetachedFromTarget" elapsed="0 ms" goroutine=65
INFO[0008] sid:C2EEAA73BF1B1ECE0FC2A188FA7A8E94 tid:A3090F4ECB94747A6660763EA5A92D8B wsURL:ws://127.0.0.1:50119/devtools/browser/7a9081be-5e4e-4346-bf52-96e1a04c2cbe  category="Connection:closeSession" elapsed="0 ms" goroutine=65
INFO[0008] sid:C2EEAA73BF1B1ECE0FC2A188FA7A8E94 tid:A3090F4ECB94747A6660763EA5A92D8B  category="Session:close" elapsed="0 ms" goroutine=65
INFO[0008] sid:C2EEAA73BF1B1ECE0FC2A188FA7A8E94 tid:A3090F4ECB94747A6660763EA5A92D8B  category="Session:readLoop:<-s.done" elapsed="1 ms" goroutine=1030
INFO[0008] <- {"method":"Page.frameStoppedLoading","params":{"frameId":"A3090F4ECB94747A6660763EA5A92D8B"},"sessionId":"28234C8BA995F6FE6E9CDF4C01592852"}  category="cdp:recv" elapsed="3 ms" goroutine=65
INFO[0008] <- {"method":"Page.frameDetached","params":{"frameId":"A3090F4ECB94747A6660763EA5A92D8B","reason":"remove"},"sessionId":"28234C8BA995F6FE6E9CDF4C01592852"}  category="cdp:recv" elapsed="11 ms" goroutine=65

...

INFO[0042] sid:28234C8BA995F6FE6E9CDF4C01592852 stid:2AC2D468203DA753654C16F69409EE55 ptid:2AC2D468203DA753654C16F69409EE55 selector:"//a[text()='Browse designs']"  category="Page:WaitForSelector" elapsed="563 ms" goroutine=49
INFO[0042] sid:28234C8BA995F6FE6E9CDF4C01592852 tid:2AC2D468203DA753654C16F69409EE55 method:"Runtime.callFunctionOn"  category="Session:Execute" elapsed="0 ms" goroutine=49
INFO[0042] sid:28234C8BA995F6FE6E9CDF4C01592852 tid:2AC2D468203DA753654C16F69409EE55 method:"Runtime.callFunctionOn"  category="Session:Execute:s.conn.send" elapsed="0 ms" goroutine=49
INFO[0042] -> {"id":82,"sessionId":"28234C8BA995F6FE6E9CDF4C01592852","method":"Runtime.callFunctionOn","params":{"functionDeclaration":"\n\t\t(injected, selector, scope, strict, state, timeout, ...args) =\u003e {\n\t\t\treturn injected.waitForSelector(selector, scope, strict, state, 'raf', timeout, ...args);\n\t\t}\n\t\n//# sourceURL=__xk6_browser_evaluation_script__\n","arguments":[{"objectId":"2562850493479125420.3.2"},{"value":{"selector":"//a[text()='Browse designs']","parts":[{"name":"xpath","body":"//a[text()='Browse designs']"}],"capture":null}},{"objectId":"2562850493479125420.3.1"},{"value":false},{"value":"visible"},{"value":30000}],"userGesture":true,"awaitPromise":true,"executionContextId":3}}  category="cdp:send" elapsed="0 ms" goroutine=66
INFO[0042] <- {"id":82,"error":{"code":-32000,"message":"Cannot find context with specified id"},"sessionId":"28234C8BA995F6FE6E9CDF4C01592852"}  category="cdp:recv" elapsed="1 ms" goroutine=65
INFO[0042] sid:28234C8BA995F6FE6E9CDF4C01592852 tid:2AC2D468203DA753654C16F69409EE55 cannot unmarshal: unknown command or event ""  category="Session:readLoop:<-s.readCh" elapsed="0 ms" goroutine=55
INFO[0042] sid:28234C8BA995F6FE6E9CDF4C01592852 tid:2AC2D468203DA753654C16F69409EE55 wsURL:"ws://127.0.0.1:50119/devtools/browser/7a9081be-5e4e-4346-bf52-96e1a04c2cbe", msg err:Cannot find context with specified id (-32000)  category="Connection:send" elapsed="0 ms" goroutine=49
ERRO[0042] unable to evaluate expression: Cannot find context with specified id (-32000)
	at file:///Users/inanc/grafana/tom-12nov.js:78:26(15)
	at go.k6.io/k6/js/common.Bind.func1 (native)
	at clickBusinessCards (file:///Users/inanc/grafana/tom-12nov.js:74:17(17))
	at file:///Users/inanc/grafana/tom-12nov.js:44:21(8)  executor=per-vu-iterations scenario=default source=stacktrace
INFO[0042] returning, ctx.Err: "context canceled"        category="connection:sendLoop" elapsed="0 ms" goroutine=66

...

INFO[0042] sid:C2EEAA73BF1B1ECE0FC2A188FA7A8E94 tid: wsURL:"ws://127.0.0.1:50119/devtools/browser/7a9081be-5e4e-4346-bf52-96e1a04c2cbe" err:context canceled  category="Connection:send:<-c.ctx.Done()" elapsed="0 ms" goroutine=47
INFO[0042] sid:C2EEAA73BF1B1ECE0FC2A188FA7A8E94 tid:A3090F4ECB94747A6660763EA5A92D8B err:unable to add script to evaluate for isolated world: context canceled  category="NewFrameSession:initIsolatedWorld" elapsed="0 ms" goroutine=47
INFO[0042] sid:28234C8BA995F6FE6E9CDF4C01592852 tid:2AC2D468203DA753654C16F69409EE55, returns  category="NewFrameSession:initEvents:go" elapsed="0 ms" goroutine=47
INFO[0042] sid:C2EEAA73BF1B1ECE0FC2A188FA7A8E94 tid:A3090F4ECB94747A6660763EA5A92D8B method:"Page.addScriptToEvaluateOnNewDocument", returns  category="Session:Execute:<-evCancelCtx.Done()" elapsed="0 ms" goroutine=1053
panic: cannot create frame session: unable to add script to evaluate for isolated world: context canceled
```

inancgumus:
I fixed the problem, but I couldn't write a unit test because of the dependencies, and I didn't want to use a mocking framework either. So, I'll use the integration test route.

inancgumus:
@imiric Please read [the description](https://github.com/grafana/xk6-browser/pull/129#issue-785823159) while you're reviewing this :)

common/frame_session.go
```diff
@@ -745,7 +751,7 @@ func (fs *FrameSession) onAttachedToTarget(event *target.EventAttachedToTarget)
 	session := fs.page.browserCtx.conn.getSession(event.SessionID)
 	targetID := event.TargetInfo.TargetID
 
-	if event.TargetInfo.Type == "iframe" {
+	if eti := event.TargetInfo; eti.Type == "iframe" && eti.URL != "" {
```
robingustafsson, line 754: I question whether this is something we want to do (Playwright doesn't for example), discarding iframe target attachments because URL is empty. Why is this problematic? Iframes can be created without a `src` attribute (will load `about:blank`) and then have the `src` set programmatically, and if you remove the `src` attribute it will also automatically load `about:blank` according to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

robingustafsson (commented):
Looks mostly good, great that you fix all my go noob mistakes (`make()` etc.) 🙌😅

You mention two deprecated events, afaict Playwright doesn't handle these events either so should not be necessary to support Tom's script. I'd suggest running the equivalent of Tom's script in Playwright (I did that a lot with other scripts, analyzing the CDP traffic differences with xk6-browser).

common/frame.go
```diff
@@ -431,22 +417,38 @@ func (f *Frame) setID(id cdp.FrameID) {
...
+		default:
+			ok = f.hasContext(world)
 		}
-	} else if world == "utility" && atomic.CompareAndSwapInt32(&f.utilityExecutionContextHasWaited, 0, 1) {
-		select {
-		case <-f.ctx.Done():
-		case <-f.utilityExecutionContextCh:
+		if !ok {
+			// TODO: change sleeping with something else
+			time.Sleep(time.Millisecond * 50)
```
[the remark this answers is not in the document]
inancgumus: Yeah, I tried a lot of stuff while trying to fix it 😅  I'll optimize this part today!

common/session.go
```diff
@@ -98,15 +98,19 @@ func (s *Session) readLoop() {
 		select {
 		case msg := <-s.readCh:
 			ev, err := cdproto.UnmarshalMessage(msg)
+			if errors.Is(err, cdp.ErrUnknownCommandOrEvent("")) && msg.Method == "" {
```
imiric, line 101: I don't think this `errors.Is()` check will work. `Is()` checks against a specific value (e.g. [`ErrNotFound`](https://pkg.go.dev/os/exec#pkg-variables)), whereas `As()` checks against a specific type, which [`ErrUnknownCommandOrEvent`](https://github.com/chromedp/cdproto/blob/master/cdp/types.go#L71) is. So we should use `errors.As()` here.

common/execution_context.go
```diff
@@ -64,21 +80,31 @@ func NewExecutionContext(
...
+		err       error
+	)
 
 	action := dom.ResolveNode().
 		WithBackendNodeID(backendNodeID).
 		WithExecutionContextID(e.id)
+
 	if remoteObj, err = action.Do(cdp.WithExecutor(e.ctx, e.session)); err != nil {
-		return nil, fmt.Errorf("unable to get DOM node: %w", err)
+		return nil, fmt.Errorf("cannot resolve dom node: %w", err)
```
imiric, line 96: We should keep the original capitalization.
```suggestion
		return nil, fmt.Errorf("cannot resolve DOM node: %w", err)
```

common/execution_context.go
```diff
@@ -91,95 +117,108 @@ func (e *ExecutionContext) adoptElementHandle(elementHandle *ElementHandle) (*El
 
 	action := dom.DescribeNode().WithObjectID(elementHandle.remoteObject.ObjectID)
 	if node, err = action.Do(cdp.WithExecutor(e.ctx, e.session)); err != nil {
-		return nil, fmt.Errorf("unable to get DOM node: %w", err)
+		return nil, fmt.Errorf("cannot describe dom node: %w", err)
```
imiric, line 120:
```suggestion
		return nil, fmt.Errorf("cannot describe DOM node: %w", err)
```

imiric (changes requested):
Nice work! This does indeed work much better with the Vistaprint script.

Although I'm not sure it fixes the `Cannot find context with specified id` issue, since it failed for me with that error:

```
DEBU[0106] sid:85A2399926F5DB8BFC14DD8BB958EE64 tid:CA8F61C073936905D509C09E1BDA2141 wsURL:"ws://127.0.0.1:33215/devtools/browser/e750ffc1-e59e-4dd6-8e2c-5808bc23e6bb", msg err:Cannot find context with specified id (-32000)  category="Connection:send" elapsed="0 ms" goroutine=117
ERRO[0106] unable to evaluate expression: Cannot find context with specified id (-32000)
default at file:///home/ivan/Downloads/tom-12nov-original.js:198:26(9)
        at go.k6.io/k6/js/common.Bind.func1 (native)
        at customerReviewed (file:///home/ivan/Downloads/tom-12nov-original.js:195:17(34))
        at file:///home/ivan/Downloads/tom-12nov-original.js:23:19(26)  executor=per-vu-iterations scenario=default source=stacktrace
```

I'll make another pass or two of this soon, just left a few comments and suggestions for now. Also, please try to leave large logging changes for separate PRs, since they clutter the purpose of a "fix" PR. Refactorings are mostly fine if they're relevant to the fix, but otherwise those also belong in separate PRs. I mostly review commit by commit anyway, but it's difficult to see the final picture if the code is changed in the final diff.

common/frame.go
```diff
@@ -220,50 +168,69 @@ func (f *Frame) clearLifecycle() {
...
+	f.lifecycleEventsMu.RUnlock()
 
 	// Only consider a life cycle event as fired if it has triggered for all of subtree.
 	f.childFramesMu.RLock()
-	for child := range f.childFrames {
-		child.(*Frame).recalculateLifecycle()
-		for k := range events {
-			if !child.(*Frame).hasSubtreeLifecycleEventFired(k) {
-				delete(events, k)
+	{
```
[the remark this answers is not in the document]
inancgumus: I think they make the protected area in the code clearer. You like indentation and don't use early returns pretty much, and I like this one. Can we agree on compromising some minor style differences in our code?

common/frame_session.go
inancgumus: > Why is this problematic?

I wonder why it solves most of the issues when we do this? 🤷  I'm open to any help. I feel like I'm stuck, and it seems like I can't see any other solutions for this issue 😓 

I guess this fixes most of the issues because our system is not a 100% match of Playwright, so ours is acting differently.  I suggest merging and keeping this fix for a while and seeing it in action. WDYT?

> I'd suggest running the equivalent of Tom's script in Playwright (I did that a lot with other scripts, analyzing the CDP traffic differences with xk6-browser).

I wrote a script to run in Playwright, and it worked. I'll keep analyzing the CDP traffic differences.

> Looks mostly good, great that you fix all my go noob mistakes (`make()` etc.) 🙌😅

You're welcome! ❤️

common/frame.go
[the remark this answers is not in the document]
inancgumus: > Nice work! This does indeed work much better with the Vistaprint script. It's a shame it's difficult to add a test for this, as we could really use some. Maybe some smaller unit tests for specific method changes could be added?

Thanks! I tried that as we discussed on Slack but it's challenging and I couldn't come up with good solutions while trying to debug thousands of logs—kinda mind stuck here.

> Although I'm not sure it fixes the `Cannot find context with specified id` issue, since it failed for me with that error:

Yes, we don't have a 100% fix and I don't know why. But I think we at least have 50%+ more improvement than before (in Tom's script). Also, this `Cannot find context with specified id` was for #49 but it looked like it solves some of the problems here as well, they're kind of connected. I can leave out that fix from this PR, though? And move it to #49.

> Also, please try to leave large logging changes for separate PRs, since they clutter the purpose of a "fix" PR. Refactorings are mostly fine if they're relevant to the fix, but otherwise those also belong in separate PRs. I mostly review commit by commit anyway, but it's difficult to see the final picture if the code is changed in the final diff.

Debugging is hard and I didn't want to throw away the work I've done. And doing it in another PR complicates things further for me.

I tried hard to make the commits easier for you to review by separating them into responsibilities, and I even added commits that fix the issue with explanations like: 

* [Fix cannot find context with specified id](https://github.com/grafana/xk6-browser/pull/129/commits/4f95b1be990f744c0bb4348ec3f2eb4e1ea4b860)
* [Fix cannot create frame session bug](https://github.com/grafana/xk6-browser/pull/129/commits/fab6b1aac529d14ddea0c4f203dfe97806234d0a)

🤷 😓 

Not like this [one](https://github.com/grafana/xk6-browser/pull/130/commits/4e2e427ac7860c9af42e26887eacf5a5a93eb79d) 😅

inancgumus:
> Nice work! This does indeed work much better with the Vistaprint script. It's a shame it's difficult to add a test for this, as we could really use some. Maybe some smaller unit tests for specific method changes could be added?

Thanks! I tried that as we discussed on Slack but it's challenging and I couldn't come up with good solutions while trying to debug thousands of logs—kinda mind stuck here.

> Although I'm not sure it fixes the `Cannot find context with specified id` issue, since it failed for me with that error:

Yes, we don't have a 100% fix and I don't know why. But I think we at least have 50%+ more improvement than before (in Tom's script). Also, this `Cannot find context with specified id` was for #49 but it looked like it solves some of the problems here as well, they're kind of connected. I can leave out that fix from this PR, though? And move it to #49.

> Also, please try to leave large logging changes for separate PRs, since they clutter the purpose of a "fix" PR. Refactorings are mostly fine if they're relevant to the fix, but otherwise those also belong in separate PRs. I mostly review commit by commit anyway, but it's difficult to see the final picture if the code is changed in the final diff.

Debugging is hard and I didn't want to throw away the work I've done. And doing it in another PR complicates things further for me.

I tried hard to make the commits easier for you to review by separating them into responsibilities, and I even added commits that fix the issue with explanations like: 

* [Fix cannot find context with specified id](https://github.com/grafana/xk6-browser/pull/129/commits/4f95b1be990f744c0bb4348ec3f2eb4e1ea4b860)
* [Fix cannot create frame session bug](https://github.com/grafana/xk6-browser/pull/129/commits/fab6b1aac529d14ddea0c4f203dfe97806234d0a)

🤷 😓 

Not like this [one](https://github.com/grafana/xk6-browser/pull/130/commits/4e2e427ac7860c9af42e26887eacf5a5a93eb79d) 😅

common/execution_context.go
inancgumus: See: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

common/session.go
inancgumus: It works because that's a conversion, not a new error.

common/frame.go
```diff
@@ -431,22 +417,38 @@ func (f *Frame) setID(id cdp.FrameID) {
...
+			ok = true
+		default:
+			ok = f.hasContext(world)
 		}
-	} else if world == "utility" && atomic.CompareAndSwapInt32(&f.utilityExecutionContextHasWaited, 0, 1) {
-		select {
-		case <-f.ctx.Done():
-		case <-f.utilityExecutionContextCh:
+		if !ok {
+			// TODO: change sleeping with something else
```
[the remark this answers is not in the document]
inancgumus: I think is with the sleeping part not the `Timer` vs `sleep`. Maybe I might use a `sync.Cond` here, WDYT?

common/frame_session.go
```diff
@@ -777,6 +797,17 @@ func (fs *FrameSession) onAttachedToTarget(event *target.EventAttachedToTarget)
...
+			fs.logger.Debugf("FrameSession:onAttachedToTarget:NewWorker:<-ctx.Done",
+				"sid:%v tid:%v esid:%v etid:%v ebid:%v type:%q",
+				fs.session.id, fs.targetID, event.SessionID,
+				event.TargetInfo.TargetID, event.TargetInfo.BrowserContextID,
+				event.TargetInfo.Type)
+			return
+		default:
+			k6Throw(fs.ctx, "cannot create frame session (worker): %w", err)
+		}
 		k6Throw(fs.ctx, "cannot create new worker: %w", err)
```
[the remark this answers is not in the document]
inancgumus: You're right, good catch!

common/frame.go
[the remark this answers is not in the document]
imiric: Sure, if you feel strongly about this, of course we can compromise :smile:

To be fair, I can see how it would make the code clearer in cases where the lock is held between large blocks (which is actually a code smell since locks should be released as quickly as possible), as it highlights exactly where the lock is acquired and released. But doing it for <~10 LOCs or [one like in this case](https://github.com/grafana/xk6-browser/blob/235bc474b20634870b8ccf703b24505b31cd0762/common/page.go#L163-L167) just clutters the code and actually makes it _less_ readable. Especially since in most cases it can be refactored into a separate function that does `defer m.Unlock()`, or just done inline without the extra block.

My other issue with it is that it functionally changes behavior, as I need to be aware of an additional scope introduced for style purposes.

> You like indentation and don't use early returns pretty much

I think there was a misunderstanding :smile: I don't _like_ indentation, in fact I'm arguing against it right now... What I commented in your other PR was that refactoring 10 LOCs indented 4 levels deep just to remove 1 level of indentation doesn't make much sense, and just introduces noise. I actually quite like that kind of refactoring you've done in other places, including in this PR. I just suggested being pragmatic and not applying that for small blocks of code.

I don't have strong opinions on early returns, they're probably good in most situations, even though I prefer [single function exit points](https://wiki.c2.com/?SingleFunctionExitPoint) when possible (though this is impractical in Go with `err != nil` checks everywhere).

Anyway, it's fine, I won't bring it up again. I would just suggest being pragmatic about where it's used. I mean, if we want to apply it globally, it can become a linter, no?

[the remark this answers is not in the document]
inancgumus: Good to know your thoughts, thanks for clarifying :)

> suggest being pragmatic about where it's used

+1, alright, I'll see what I can do afterward 😅 

> if we want to apply it globally, it can become a linter, no?

Some linters won't hurt :)

common/frame.go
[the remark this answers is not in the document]
imiric: Ah, I see. Then I would prefer Robin's suggestion of using a channel in `setContext()` for signaling, as it was before. `sync.Cond` doesn't seem necessary.

common/session.go
imiric: Ah, so you're checking that we received an empty `ErrUnknownCommandOrEvent`. OK, it makes sense after reading the comment below.

common/frame.go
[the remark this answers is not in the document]
inancgumus: Channel solution is more complicated compared to the possible `Cond` solution. WDYT `Cond` is unnecessary?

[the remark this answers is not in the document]
imiric: I'm in favor of whatever is the simplest solution, and if we don't need to synchronize more than 2 goroutines, a channel should be simpler than using `sync.Cond`. In this case I think all synchronization is happening internally in `Frame`, so the previous channel approach seems fine to me.

If you can explain what was wrong with the previous approach and why `sync.Cond` would be simpler and necessary, by all means go with that.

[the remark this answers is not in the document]
inancgumus: Me too. The synchronization happens between `FrameSession`s and `Frame`s, and `VU` and `Frames`. If it were only the `Frame` there would be deadlocks 🤷

common/frame_session.go
inancgumus: > Iframes can be created without a src attribute (will load about:blank) and then have the src set programmatically, and if you remove the src attribute it will also automatically load

Maybe this is the reason why skipping frames with empty src works.

inancgumus:
@imiric Please see the updated PR description before reviewing, thanks!

common/frame_session.go
```diff
@@ -827,17 +845,19 @@ func (fs *FrameSession) onAttachedToTarget(event *target.EventAttachedToTarget)
 	// Handle new worker
 	w, err := NewWorker(fs.ctx, session, targetID, event.TargetInfo.URL)
 	if err != nil {
-		if !fs.page.browserCtx.browser.connected && strings.Contains(err.Error(), "websocket: close 1006 (abnormal closure)") {
-			// If we're no longer connected to browser, then ignore WebSocket errors
+		// ignore errors if we're no longer connected to browser
+		// this happens when there is no browser but we still want to
+		// attach a worker to it.
+		if !fs.page.browserCtx.browser.connected {
```
imiric, line 851: Should we use `browser.IsConnected()` here like on line 811, to make sure it's read locked?

common/frame_session.go
```diff
@@ -827,17 +845,19 @@ func (fs *FrameSession) onAttachedToTarget(event *target.EventAttachedToTarget)
...
 			fs.logger.Debugf("FrameSession:onAttachedToTarget:NewWorker:<-ctx.Done",
 				"sid:%v tid:%v esid:%v etid:%v ebctxid:%v type:%q",
 				fs.session.id, fs.targetID, event.SessionID,
-				event.TargetInfo.TargetID, event.TargetInfo.BrowserContextID, event.TargetInfo.Type)
-
+				event.TargetInfo.TargetID, event.TargetInfo.BrowserContextID,
+				event.TargetInfo.Type)
 			return
 		default:
 			k6Throw(fs.ctx, "cannot create frame session (worker): %w", err)
```
imiric, line 863: Maybe unrelated to this PR, but the below `k6Throw()` (line 865) will never be reached, so we can remove it. (GitHub doesn't allow me to comment on that line...)

imiric (changes requested):
Nicely done, this works much better than on `main` from my tests.

Just left some minor things and nitpicks.

common/element_handle.go
```diff
@@ -238,8 +238,12 @@ func (h *ElementHandle) checkHitTargetAt(apiCtx context.Context, p Position) (bo
 			return injected.checkHitTargetAt(node, point);
 		}
 	`)
+	opts := evaluateOptions{
+		forceCallable: true,
+		returnByValue: true,
+	}
```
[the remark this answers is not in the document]
inancgumus: 😄 Not here because config structs that are not expected to grow or don't need to be passed as a `nil` value are good. Having them optional here won't do any good because the purpose was to make things more straightforward and reduce possible bugs in the future. Accepting successive primitive values is not a good pattern.

common/execution_context.go
```diff
@@ -145,23 +164,19 @@ func (e *ExecutionContext) adoptElementHandle(eh *ElementHandle) (*ElementHandle
 
 // evaluate will evaluate provided callable within this execution context
 // and return by value or handle
-func (e *ExecutionContext) evaluate(
-	apiCtx context.Context,
-	forceCallable bool, returnByValue bool,
-	pageFunc goja.Value, args ...goja.Value,
-) (res interface{}, err error) {
+func (e *ExecutionContext) evaluate(apiCtx context.Context, opts evaluateOptions, pageFunc goja.Value, args ...goja.Value) (res interface{}, err error) {
```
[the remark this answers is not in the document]
inancgumus: Lol :) No worries, I wasn't going to discuss it. Actually, I was going to ask you for a comfortable line length before then, I forgot. 80-120 is OK then.

common/frame.go
```diff
@@ -715,29 +700,46 @@ func (f *Frame) DispatchEvent(selector string, typ string, eventInit goja.Value,
 }
 
 // Evaluate will evaluate provided page function within an execution context
-func (f *Frame) Evaluate(pageFunc goja.Value, args ...goja.Value) interface{} {
+func (f *Frame) Evaluate(pageFunc goja.Value, args ...goja.Value) (result interface{}) {
```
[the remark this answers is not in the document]
inancgumus: Yes, it was for documentation purposes but it's not necessary, so I'm going to remove it.

common/frame_session.go
inancgumus: Good catch!

common/frame.go
```diff
@@ -1427,6 +1437,22 @@ func (f *Frame) WaitForTimeout(timeout int64) {
 	}
 }
 
+func (f *Frame) adoptBackendNodeID(world executionWorld, id cdp.BackendNodeID) (*ElementHandle, error) {
+	f.executionContextMu.RLock()
+	defer f.executionContextMu.RUnlock()
+
+	ec := f.executionContexts[world]
+	return ec.adoptBackendNodeID(id)
```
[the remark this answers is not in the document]
inancgumus: Good point, thanks!

common/frame_session.go
inancgumus: Good catch 🙇

common/frame.go
```diff
@@ -719,25 +712,42 @@ func (f *Frame) Evaluate(pageFunc goja.Value, args ...goja.Value) interface{} {
...
-	f.waitForExecutionContext("main")
-	handle, err := f.mainExecutionContext.EvaluateHandle(f.ctx, pageFunc, args...)
+
+	f.waitForExecutionContext(mainWorld)
+
+	var err error
+	f.executionContextMu.RLock()
+	{
+		ec := f.executionContexts[mainWorld]
+		handle, err = ec.EvaluateHandle(f.ctx, pageFunc, args...)
```
imiric, line 744: Here too we could use a nil check.

Since this is commonly used, let's add a `Frame.getContext(world)` method, and `Frame.hasContext(world)` could call it.

common/frame.go
```diff
@@ -573,10 +564,12 @@ func (f *Frame) waitForSelector(selector string, opts *FrameWaitForSelectorOptio
...
 	// We always return ElementHandles in the main execution context (aka "DOM world")
-	if handle.execCtx != f.mainExecutionContext {
+	f.executionContextMu.RLock()
+	defer f.executionContextMu.RUnlock()
+
+	if ec := f.executionContexts[mainWorld]; handle.execCtx != ec {
 		defer handle.Dispose()
-		handle, err = f.mainExecutionContext.adoptElementHandle(handle)
-		if err != nil {
+		if handle, err = ec.adoptElementHandle(handle); err != nil {
```
imiric, line 572: Another place where `ec` could be nil.

common/frame.go
inancgumus: Thanks. Yeah, but we'll still need to check for nil and type the same errors. I'm not sure what benefit `getContext` will provide us.

common/frame.go
inancgumus: Thanks!

common/frame.go
imiric: > I'm not sure what benefit `getContext` will provide us.

Encapsulate the usage of `f.executionContextMu`. But yeah, we'd still need to check for nil.

imiric (approved):
LGTM. It seems we hit #161 in CI.

Were you able to reproduce the issues reported on the forum? It would be great if we could add E2E tests for them in a follow-up PR, and probably do the E2E/example tests split at that point as well, which Robin suggested in a PR a few weeks ago.

inancgumus:
😒 Any ideas on reproducing these errors in an E2E test without using an external website? I’d appreciate your help because I couldn’t reproduce these problems without using external websites.

> Were you able to reproduce the issues reported on the forum? 

Yes, I used the website the user tests on.

imiric:
> Any ideas on reproducing these errors in an E2E test without using an external website?

Ah, it's the same situation as with Vistaprint. :disappointed: I thought it could be easily replicated.

We'd need to spend time reverse engineering each site and building a minimal test case that reproduced it, which would vary in complexity/effort.

Alternatively, we could try doing a WARC dump using something like [ArchiveBox](https://github.com/ArchiveBox/ArchiveBox), and then testing against that. It probably won't work for the more complex cases, but should for a lot of them, and it would avoid spending time on re-implementing each scenario ourselves. Or by using some recording/replay proxy, maybe [warcprox](https://github.com/internetarchive/warcprox), though it doesn't inspire much confidence.

Failing all of the above, we could just test against the sites themselves, and configure CI to run these jobs on, say, PR merge and release only, to minimize any impact it could have on them. This could of course break at any point, and we'd have to keep updating the tests, so probably not worthwhile. :smile:

inancgumus:
Thanks 😄 It seems like a lot of dead-ends. I can try WARC dump and then keep cleaning up the dump until the error doesn't fade away. Maybe we can have a reproducible result and then create a Go/JS test for it 🤷 😄

imiric:
> It seems like a lot of dead-ends.

Not really, I think [pywb](https://github.com/webrecorder/pywb) looks promising. I'll create an issue.

inancgumus:
So pywb -> Clean up the dump -> Go/Js test?

imiric:
I don't think any cleanup would be needed, we could load the archive as is. But let's discuss that in the issue whenever we start looking into this. I can't say for sure what the process is without giving it a try. :)

PR #131, opened 2021-11-23
Fix/54 proper logger

### LINKED ISSUE #54: Proper logger
A proper logger can make us see the underlying problems, find and fix them faster.

**Features:**
* Enable/disable logging on purpose
* Enable/disable components log output like `networkManager`, `frameManager`, ...
* Enable/disable debugs/warnings/errors/timeouts
* `TraceID` to see the call graph for tracking each command of a JS test script
* Nice to have: colored output :)

inancgumus:
This PR is for #54 (Proper logger):
* Passes the same logger extension-wide
* Adds `Debugf` to the entities in the extension
* Updates tests to use a null logger when needed

With this, I'm able to see the problems in #125 and #49.

So I think they are helpful for us to diagnose issues in the future as well.

Closes #54

### DIFF
--- a/common/connection.go
+++ b/common/connection.go
@@ -182,7 +182,7 @@ func (c *Connection) closeConnection(code int) error {
 }
 
 func (c *Connection) closeSession(sid target.SessionID, tid target.ID) {
-	c.logger.Errorf("Connection:closeSession", "sid:%v tid:%v wsURL:%v", sid, tid, c.wsURL)
+	c.logger.Debugf("Connection:closeSession", "sid:%v tid:%v wsURL:%v", sid, tid, c.wsURL)
 	c.sessionsMu.Lock()
 	if session, ok := c.sessions[sid]; ok {
 		session.close()
@@ -192,16 +192,16 @@ func (c *Connection) closeSession(sid target.SessionID, tid target.ID) {
 }
 
 func (c *Connection) createSession(info *target.Info) (*Session, error) {
-	c.logger.Errorf("Connection:createSession", "")
+	c.logger.Debugf("Connection:createSession", "")
 
 	var sessionID target.SessionID
 	var err error
 	action := target.AttachToTarget(info.TargetID).WithFlatten(true)
 	if sessionID, err = action.Do(cdp.WithExecutor(c.ctx, c)); err != nil {
-		c.logger.Errorf("Connection:createSession", "err:%v", err)
+		c.logger.Debugf("Connection:createSession", "err:%v", err)
 		return nil, err
 	}
-	c.logger.Errorf("Connection:createSession", "sid:%v", sessionID)
+	c.logger.Debugf("Connection:createSession", "sid:%v", sessionID)
 	return c.getSession(sessionID), nil
 }
 
@@ -222,9 +222,9 @@ func (c *Connection) handleIOError(err error) {
 	}
 	select {
 	case c.closeCh <- code:
-		c.logger.Errorf("Connection:handleIOError:c.closeCh <-", "code:%d", code)
+		c.logger.Debugf("Connection:handleIOError:c.closeCh <-", "code:%d", code)
 	case <-c.done:
-		c.logger.Errorf("Connection:handleIOError:<-c.done", "")
+		c.logger.Debugf("Connection:handleIOError:<-c.done", "")
 	}
 }
 
@@ -293,7 +293,7 @@ func (c *Connection) recvLoop() {
 			evt := ev.(*target.EventDetachedFromTarget)
 			sid := evt.SessionID
 			tid := c.findSessionTargetID(sid)
-			c.logger.Errorf("Connection:recvLoop:EventDetachedFromTarget", "sid:%v tid:%v wsURL:%q, closeSession", sid, tid, c.wsURL)
+			c.logger.Debugf("Connection:recvLoop:EventDetachedFromTarget", "sid:%v tid:%v wsURL:%q, closeSession", sid, tid, c.wsURL)
 			c.closeSession(sid, tid)
 		}
 
@@ -305,24 +305,23 @@ func (c *Connection) recvLoop() {
 				continue
 			}
 			if msg.Error != nil && msg.Error.Message == "No session with given id" {
-				c.logger.Errorf("Connection:recvLoop", "sid:%v tid:%v wsURL:%q, closeSession #2", session.id, session.targetID, c.wsURL)
+				c.logger.Debugf("Connection:recvLoop", "sid:%v tid:%v wsURL:%q, closeSession #2", session.id, session.targetID, c.wsURL)
 				c.closeSession(session.id, session.targetID)
 				continue
 			}
 
 			select {
 			case session.readCh <- &msg:
-				// c.logger.Errorf("Connection:recvLoop:session.readCh<-", "sid:%v wsURL:%v crashed:%t", session.id, c.wsURL, session.crashed)
 			case code := <-c.closeCh:
-				c.logger.Errorf("Connection:recvLoop:<-c.closeCh", "sid:%v tid:%v wsURL:%v crashed:%t", session.id, session.targetID, c.wsURL, session.crashed)
+				c.logger.Debugf("Connection:recvLoop:<-c.closeCh", "sid:%v tid:%v wsURL:%v crashed:%t", session.id, session.targetID, c.wsURL, session.crashed)
 				_ = c.closeConnection(code)
 			case <-c.done:
-				c.logger.Errorf("Connection:recvLoop:<-c.done", "sid:%v tid:%v wsURL:%v crashed:%t", session.id, session.targetID, c.wsURL, session.crashed)
+				c.logger.Debugf("Connection:recvLoop:<-c.done", "sid:%v tid:%v wsURL:%v crashed:%t", session.id, session.targetID, c.wsURL, session.crashed)
 				return
 			}
 
 		case msg.Method != "":
-			c.logger.Errorf("Connection:recvLoop:msg.Method:emit", "method:%q", msg.Method)
+			c.logger.Debugf("Connection:recvLoop:msg.Method:emit", "method:%q", msg.Method)
 			ev, err := cdproto.UnmarshalMessage(&msg)
 			if err != nil {
 				c.logger.Errorf("cdp", "%s", err)
@@ -331,11 +330,11 @@ func (c *Connection) recvLoop() {
 			c.emit(string(msg.Method), ev)
 
 		case msg.ID != 0:
-			c.logger.Errorf("Connection:recvLoop:msg.ID:emit", "method:%q", msg.ID)
+			c.logger.Debugf("Connection:recvLoop:msg.ID:emit", "method:%q", msg.ID)
 			c.emit("", &msg)
 
 		default:
-			c.logger.Errorf("cdp", "ignoring malformed incoming message (missing id or method): %#v (message: %s)", msg, msg.Error.Message)
+			c.logger.Debugf("cdp", "ignoring malformed incoming message (missing id or method): %#v (message: %s)", msg, msg.Error.Message)
 		}
 	}
 }
@@ -344,14 +343,14 @@ func (c *Connection) send(msg *cdproto.Message, recvCh chan *cdproto.Message, re
 	select {
 	case c.sendCh <- msg:
 	case err := <-c.errorCh:
-		c.logger.Errorf("Connection:send:<-c.errorCh", "wsURL:%q sid:%v, err:%v", c.wsURL, msg.SessionID, err)
+		c.logger.Debugf("Connection:send:<-c.errorCh", "wsURL:%q sid:%v, err:%v", c.wsURL, msg.SessionID, err)
 		return err
 	case code := <-c.closeCh:
-		c.logger.Errorf("Connection:send:<-c.closeCh", "wsURL:%q sid:%v, websocket code:%v", c.wsURL, msg.SessionID, code)
+		c.logger.Debugf("Connection:send:<-c.closeCh", "wsURL:%q sid:%v, websocket code:%v", c.wsURL, msg.SessionID, code)
 		_ = c.closeConnection(code)
 		return &websocket.CloseError{Code: code}
 	case <-c.done:
-		c.logger.Errorf("Connection:send:<-c.done", "wsURL:%q sid:%v", c.wsURL, msg.SessionID)
+		c.logger.Debugf("Connection:send:<-c.done", "wsURL:%q sid:%v", c.wsURL, msg.SessionID)
 		return nil
 	}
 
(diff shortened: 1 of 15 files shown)


imiric (commented):
Nice job with this! :clap: It will surely help troubleshoot thorny issues like #125, though now running with `--verbose` is _super_ noisy. It would be good to place the [`cdp:recv`](https://github.com/grafana/xk6-browser/blob/0e0f966ee440e5d1d278a188b2e212bdbe4a9b4c/common/connection.go#L236) event a level below `Debug` (`Trace`) that could be activated with a second `--verbose`, though that's probably not supported right now and would need changes in k6. Otherwise putting it behind an environment variable (`XK6_BROWSER_TRACE`) might be a good idea until `-vv` is supported. From the logrus side it would be a matter of simply calling `Tracef()` instead of `Debugf()`. Since those make the majority of logged events (70% by my rough estimate), and we usually don't care about them unless we're debugging the CDP layer itself, it would make debugging our code less noisy.

I left a few minor suggestions, and here are some general comments:

- Using structured logging for all this like you suggested over Slack would be much better. E.g. all places where `err:%v` is logged would be better with `WithError(err)` instead. It would also cut down on all the repetition, since you could initialize the logger with all fields upfront (in a constructor or at the start of the method) and then simply call `Debug()` with just the message. But I don't think this is a requirement for this PR and we can improve on that later.
- Another related thing I was thinking last week would be to abandon our `common.Logger` altogether and use `logrus` directly. Some of the things `Logger.Logf` does are of questionable importance (the `elapsed` field, `debugOverride` is always `false`, the fallback to `fmt.Printf()`...), though standardizing the `category` field and logging the goroutine ID seem useful.
  OTOH, if we ever decide to switch logging libraries, as [we've been planning to do in k6 for years now](https://github.com/grafana/k6/issues/216), having a custom logging API would be very convenient. So maybe we should keep it but simplify it considerably.
- If we do decide to keep our wrapper, we might want to abstract away the logging of the calling function names. This adds more work when functions are renamed/moved, and doing it with LSP won't pick it up. We can easily get the caller function name with some `runtime` introspection, see [this SO answer](https://stackoverflow.com/a/58232098). I actually had a patch I never pushed that does this when I was debugging some things last week, and it worked quite well. Obviously it won't pick up all the nuances like adding `:go` for goroutines or custom categories based on `if` branches, but those aren't the biggest concern when refactoring.

common/connection.go
```diff
@@ -297,10 +330,11 @@ func (c *Connection) recvLoop() {
 			c.emit(string(msg.Method), ev)
 
 		case msg.ID != 0:
+			c.logger.Debugf("Connection:recvLoop:msg.ID:emit", "method:%q", msg.ID)
 			c.emit("", &msg)
 
 		default:
-			c.logger.Errorf("cdp", "ignoring malformed incoming message (missing id or method): %#v (message: %s)", msg, msg.Error.Message)
+			c.logger.Debugf("cdp", "ignoring malformed incoming message (missing id or method): %#v (message: %s)", msg, msg.Error.Message)
```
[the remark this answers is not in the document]
inancgumus: I made this `Debugf` because there were a lot of these errors. But, ok, let's make it `Errorf`.

common/logger.go
```diff
@@ -84,23 +84,26 @@ func (l *Logger) Logf(level logrus.Level, category string, msg string, args ...i
 	if now == elapsed {
 		elapsed = 0
 	}
+	defer func() {
+		l.lastLogCall = now
```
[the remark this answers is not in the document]
inancgumus: `elapsed` was helpful for me when I was debugging the context error.

For example, it showed me where the code was stuck for 20 seconds and failed.

common/page.go
```diff
@@ -293,6 +310,7 @@ func (p *Page) AddStyleTag(opts goja.Value) {
 
 // BrintToFront activates the browser tab for this page
 func (p *Page) BrintToFront() {
+	p.logger.Debugf("Page:BrintToFront", "sid:%v", p.session.id)
```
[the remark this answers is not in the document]
inancgumus: According to UrbanDictionary:

> Brint, is a sweet, funny, caring, loving, and gentle guy. but when you make him angry he will snap at revenge and dispise you greatly.

Anyways, +1 :)

inancgumus:
> Nice job with this! 👏 It will surely help troubleshoot thorny issues like #125 
Thanks! ❤️ 

> though now running with `--verbose` is _super_ noisy. It would be good to place the [`CDP:recv`](https://github.com/grafana/xk6-browser/blob/0e0f966ee440e5d1d278a188b2e212bdbe4a9b4c/common/connection.go#L236) event a level below `Debug` (`Trace`) that could be activated with a second `--verbose`, though that's probably not supported right now and would need changes in k6. Otherwise putting it behind an environment variable (`XK6_BROWSER_TRACE`) might be a good idea until `-vv` is supported. From the logrus side it would be a matter of simply calling `Tracef()` instead of `Debugf()`. Since those make the majority of logged events (70% by my rough estimate), and we usually don't care about them unless we're debugging the CDP layer itself, it would make debugging our code less noisy.
> 

We use environment flags only on the testing side right now, so we can add a `XK6_BROWSER_TRACE` there.

However, we use flags for the test runs (like the `debug` flag). So, we might also add a `trace` flag to the launch options and enable CDP messages when it's on? And replace all CDP `Debugf`s with `Tracef`s and update the logger accordingly. WDYT?

> * Using structured logging for all this like you suggested over Slack would be much better. E.g. all places where `err:%v` is logged would be better with `WithError(err)` instead. It would also cut down on all the repetition, since you could initialize the logger with all fields upfront (in a constructor or at the start of the method) and then simply call `Debug()` with just the message. But I don't think this is a requirement for this PR and we can improve on that later.
>
> * Another related thing I was thinking last week would be to abandon our `common.Logger` altogether and use `logrus` directly. Some of the things `Logger.Logf` does are of questionable importance (the `elapsed` field, `debugOverride` is always `false`, the fallback to `fmt.Printf()`...), though standardizing the `category` field and logging the goroutine ID seem useful.

That would be awesome. We can do something like this:

```golang
// common/logger.go
func (l *Logger) WithError(err error) *logrus.Entry {
	return l.logger.WithError(err)
}

// common/browser.go
b.logger.WithError(err).Debugf(
    "Browser:onAttachedToTarget:background_page", "sid:%v tid:%v, returns",
    ev.SessionID, ev.TargetInfo.TargetID)
```

However, as I wrote you on Slack:

* The problem with this approach (`b.logger.WithError`) is that, the returned `*logrus.Entry` no longer uses our custom `common/Logger.Debugf` (and the other custom) methods.

* We can easily use logrus's methods on the returned `*Entry`, but we'll be stuck to Logrus. Also, we won't be able to do custom stuff that we do in our Logger's `Logf` method. 

* `goroutineid` and `elapsed` fields were helpful when I was debugging the context error. For example, `elapsed` showed me the code was stuck 20s and failed (But we can deduct that from the timings as well 🤷 ) Also, I could follow which goroutines were responsible for what. We can add goroutine id at the beginning of each component. But, if we follow this approach, we'll need to get rid of `elapsed`.

>   OTOH, if we ever decide to switch logging libraries, as [we've been planning to do in k6 for years now](https://github.com/grafana/k6/issues/216), having a custom logging API would be very convenient. So maybe we should keep it but simplify it considerably.

Maybe we should write a custom formatter for our use case? That can possibly solve the issues I explained above.

> * If we do decide to keep our wrapper, we might want to abstract away the logging of the calling function names. This adds more work when functions are renamed/moved, and doing it with LSP won't pick it up. We can easily get the caller function name with some `runtime` introspection, see [this SO answer](https://stackoverflow.com/a/58232098). I actually had a patch I never pushed that does this when I was debugging some things last week, and it worked quite well. Obviously it won't pick up all the nuances like adding `:go` for goroutines or custom categories based on `if` branches, but those aren't the biggest concern when refactoring.

Can you send me the patch so I can look at it?

For example, with this code:

```golang
func NewLogger(ctx context.Context, logger *logrus.Logger, debugOverride bool, categoryFilter *regexp.Regexp) *Logger {
	l := Logger{
		ctx:            ctx,
		logger:         logger,
		mu:             sync.Mutex{},
		debugOverride:  debugOverride,
		categoryFilter: categoryFilter,
		lastLogCall:    0,
	}
	logger.SetReportCaller(true)
	logger.SetFormatter(&logrus.JSONFormatter{
		CallerPrettyfier: caller(),
		FieldMap: logrus.FieldMap{
			logrus.FieldKeyFile: "caller",
		},
	})
	return &l
}

// caller returns string presentation of log caller which is formatted as
// `/path/to/file.go:line_number`. e.g. `/internal/app/api.go:25`
func caller() func(*runtime.Frame) (function string, file string) {
	return func(f *runtime.Frame) (function string, file string) {
		return f.Func.Name(), fmt.Sprintf("%s:%d", f.File, f.Line)
	}
}
```

I receive log messages like this:

```
{"caller":"/Users/inanc/grafana/xk6-browser/common/logger.go:124","category":"cdp:recv","elapsed":"0 ms","func":"github.com/grafana/xk6-browser/common.(*Logger).Logf","goroutine":21,"level":"info","msg":"\u003c- {\"id\":6,\"result\":{}}","time":"2021-11-25T14:12:04+03:00"}
{"caller":"/Users/inanc/grafana/xk6-browser/common/logger.go:124","category":"Connection:recvLoop:msg.ID:emit","elapsed":"0 ms","func":"github.com/grafana/xk6-browser/common.(*Logger).Logf","goroutine":21,"level":"info","msg":"method:'\\x06'","time":"2021-11-25T14:12:04+03:00"}
{"caller":"/Users/inanc/grafana/xk6-browser/common/logger.go:124","category":"Browser:Close","elapsed":"1 ms","func":"github.com/grafana/xk6-browser/common.(*Logger).Logf","goroutine":7,"level":"info","msg":"browserProc terminate","time":"2021-11-25T14:12:04+03:00"}
```

It's kind of noisy. Maybe we can strip some parts of it.

imiric:
> However, we use flags for the test runs (like the `debug` flag). So, we might also add a `trace` flag to the launch options and enable CDP messages when it's on? And replace all CDP `Debugf`s with `Tracef`s and update the logger accordingly. WDYT?

But `debug` is not a CLI flag, it's a JS option. I wouldn't want to have to change the script whenever I want to enable trace logs (or use workarounds like we did for `XK6_HEADLESS`), it should be done via the CLI. Since we can't add CLI flags or change the way `--verbose` works, an environment variable for this is the simplest approach, so I would go with that.


> Maybe we should write a custom formatter for using our custom logger on top of Logrus? That can possibly solve the issues I explained above.

:thinking_face: That might work, though then we're really depending even more on Logrus, which we probably want to replace.

> Can you send me the patch so I can look at it?

Sure, it's just a simple change in `debugLog`:

```diff
diff --git a/common/helpers.go b/common/helpers.go
index ec8c20d..851e227 100644
--- a/common/helpers.go
+++ b/common/helpers.go
@@ -52,7 +52,9 @@ func debugLog(msg string, args ...interface{}) {
 		elapsed = 0
 	}
 	magenta := color.New(color.FgMagenta).SprintFunc()
-	args = []interface{}{fmt.Sprintf("[%d] %s", goRoutineID(), fmt.Sprintf(msg, args...))}
+	pc, fn, line, _ := runtime.Caller(1)
+	args = []interface{}{fmt.Sprintf("[%d %s(%s:%d)] %s",
+		goRoutineID(), runtime.FuncForPC(pc).Name(), fn, line, fmt.Sprintf(msg, args...))}
 	args = append(args, fmt.Sprintf("%s ms", magenta(elapsed)))
 	fmt.Println(args...)
 	debugLastCall = now
```

I wasn't aware of Logrus' `SetReportCaller()`, so we should probably use that.

But really, these are all improvements we can do after this PR, including proper structured logging, `WithError`, etc. The only blocker I see for merging this would be to log CDP messages with `Tracef` and enable `TraceLevel` if `XK6_BROWSER_TRACE` is defined, since we can't influence it any other way.

inancgumus:
> Replace all CDP `Debugf`s with `Tracef`s and update the logger accordingly. WDYT?

**First todo:** +1 - Tracef

> But `debug` is not a CLI flag, it's a JS option. I wouldn't want to have to change the script whenever I want to enable trace logs (or use workarounds like we did for `XK6_HEADLESS`), it should be done via the CLI. Since we can't add CLI flags or change the way `--verbose` works, an environment variable for this is the simplest approach, so I would go with that.
> 

**2nd todo:** What do you think about `XK6_DEBUG=1` where 1 can be changed for each level (aligned with the log levels). And it will override the script's launch options when we define it. WDYT?

> > Maybe we should write a custom formatter for using our custom logger on top of Logrus? That can possibly solve the issues I explained above.
> 
> 🤔 That might work, though then we're really depending even more on Logrus, which we probably want to replace.
> 

Let's pass this on for now, then.

> I wasn't aware of Logrus' `SetReportCaller()`, so we should probably use that.

+1 but for later, ok.

> 
> But really, these are all improvements we can do after this PR, including proper structured logging, `WithError`, etc. The only blocker I see for merging this would be to log CDP messages with `Tracef` and enable `TraceLevel` if `XK6_BROWSER_TRACE` is defined, since we can't influence it any other way.

Please see 2nd todo above.

@imiric

imiric:
> What do you think about `XK6_LOG=1` where 1 can be changed for each level (aligned with the log levels). And it will override the script's launch options when we define it.

Sounds good (maybe `XK6_BROWSER_LOG`?), but we don't need to make this sophisticated, or even document it, since we wouldn't want users to become dependent on it. It would be best kept as an internal dev-only workaround until we can integrate it better with the CLI (`-vvv`, etc.).

inancgumus:
I added `XK6_BROWSER_LOG`, and changed CDP messages to Trace level.

The new flag accepts these levels: `panic`, `fatal`, `error`, `warn`, `warning`, `info` (_default_), `debug`, `trace`. So we can use it like this:

```bash
XK6_BROWSER_LOG=warn xk6 run -q examples/colorscheme.js
```

@imiric

common/logger.go
```diff
@@ -84,23 +92,69 @@ func (l *Logger) Logf(level logrus.Level, category string, msg string, args ...i
...
 		}
-		l.lastLogCall = now
 	}
+	l.log.SetFormatter(&logrus.TextFormatter{
+		CallerPrettyfier: caller(),
+		FieldMap: logrus.FieldMap{
+			logrus.FieldKeyFile: "caller",
+		},
+	})
+	l.log.SetReportCaller(true)
```
imiric, line 148: Do we need this custom field and prettifier? From what I see it only changes the field name.

With:
```
func="github.com/grafana/xk6-browser/common.(*Logger).Logf" caller="github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/logger.go:116"
```

Without (just calling `l.log.SetReportCalled(true)`):
```
func="github.com/grafana/xk6-browser/common.(*Logger).Logf" file="github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/logger.go:116"
```

Is the idea to leave it for further customization (cleaning up the `github.com/...` prefix like you mentioned)?

Also, there are a lot of places where the caller is `Logger.Logf`. That might be tricky to resolve, so this is fine as is.

imiric (approved):
Nicely done! I think we should eventually fix how the caller is determined/rendered, but we can do that later.

common/logger.go
inancgumus: Yeah, it's because we're using our custom logger. Maybe we should get rid of it as you said before.

inancgumus: Yeah, it's because we're using our custom logger. Maybe we should get rid of it as you said before.

For now, I can do something like this, WDYT?

```golang
// ReportCaller adds source file and function names to the log entries.
func (l *Logger) ReportCaller() {
	// strip the module informaiton
	strip := func(s string) string {
		s = strings.TrimPrefix(s, "github.com/grafana/xk6-browser")
		s = s[strings.Index(s, "/")+1:]

		return s
	}
	caller := func() func(*runtime.Frame) (string, string) {
		return func(f *runtime.Frame) (fn string, loc string) {
			loc = strip(fmt.Sprintf("%s:%d", f.File, f.Line))
			// find the caller of the log func
			fn = f.Func.Name()

			const stackLevel = 8
			_, file, no, ok := runtime.Caller(stackLevel)
			if ok {
				fn = fmt.Sprintf("%s:%d", strip(file), no)
			}

			return fn, loc
		}
	}
	l.log.SetFormatter(&logrus.TextFormatter{
		CallerPrettyfier: caller(),
		FieldMap: logrus.FieldMap{
			logrus.FieldKeyFile: "caller",
		},
	})
	l.log.SetReportCaller(true)
}
```

Output:
```
time="2021-11-30T12:12:21+03:00" level=debug msg="sid:AFA9A388DA1617DF05688B6718BCD591 tid:045F8A2BB4D7E102A47820CF00E5C9DF" func="common/session.go:75" caller="common/logger.go:116" category="Session:close" elapsed="0 ms" goroutine=68
```

I can easily strip out the `caller` there as well.

@imiric

imiric: Yeah, we wouldn't need `XK6_BROWSER_CALLER` if the caller information could be added by default on levels >= `Debug`.

Can that `stackLevel = 8` be a constant? I mean, will that always be the case? I don't have time to look into it now, but if you think it solves it for all cases, LGTM.

And the `func="common/session.go:75"` doesn't make much sense, there's no function name :) Maybe this should be `caller="common.(*Session).close [common/session.go:75]"`?

In any case, don't spend too much time on this. Feel free to create an issue for it and we can tackle it soon-ish.

inancgumus: Yep, we need function name there as well, but at least it goes to the file and line when I click on it on my terminal 😅 

I created another issue with some rough PRs that solve some of the issues:
* Use logger directly: #137 
* PR: #138

@imiric

inancgumus: > Can that stackLevel = 8 be a constant? I mean, will that always be the case? I don't have time to look into it now, but if you think it solves it for all cases, LGTM.

For this Logrus version, yes!

Issue #137, opened 2021-11-30
Use logger directly


inancgumus:
This issue is a continuation of the previous Proper Logger issue of #54.

Previously we added several log calls in this PR of #131.

**Next todo:**

- [ ] Remove the custom logger
- [ ] Report caller by default
- [ ] Strip the module info from the reporter caller
- [ ] Provide the function name in the caller
- [ ] Colorize logs

Feel free to add more to this list. I worked on some WIP commits to roughly solves some of them, see: x, y.

PR #138, opened 2021-11-30
Use logger directly (improvements)

inancgumus:
See issue #137 for details.

For reference:
https://github.com/grafana/grafana/blob/34f757ba5a970cc5fcffa92f47c5a0d41928f150/pkg/infra/log/log.go


inancgumus:
This diverged from the main branch so that it would be futile to fix the conflicts. It was an experiment, after all.

Issue #147, opened 2021-12-08
Update the chrome flags while launching the browser


inancgumus:
While navigating Playwright code, I noticed [they updated the flags while launching Chrome](https://github.com/microsoft/playwright/blob/43b7e6c26db717f6711e37bf44e9623421fd5df9/packages/playwright-core/src/server/chromium/chromium.ts#L288-L315). We might want to update ours as well to keep the compatibility intact.

Here are the changes:

**disable-features**
Before:
`TranslateUI,BlinkGenPropertyTrees,ImprovedCookieControls,SameSiteByDefaultCookies,LazyFrameLoading`
After:
`ImprovedCookieControls,LazyFrameLoading,GlobalMediaControls,DestroyProfileOnBrowserClose,MediaRouter,AcceptCHFrame`

So they no longer have these ones: `TranslateUI`, `BlinkGenPropertyTrees`, `SameSiteByDefaultCookies`

* `TranslateUI` is expired, see [this](https://github.com/GoogleChrome/lighthouse/discussions/11860#discussioncomment-226193).
* `BlinkGenPropertyTrees` is expired, see [this](https://github.com/kblok/puppeteer/commit/8b49dc62a62282543ead43541316e23d3450ff3c).
* `SameSiteByDefaultCookies`, see [this](https://www.chromium.org/updates/same-site?pli=1).

**enable-use-zoom-for-dsf**
_Only enabled on OS X_

There are some flags that are only enabled for headless mode:
`hide-scrollbars`
`mute-audio`
`blink-settings` with `primaryHoverType=2,availableHoverTypes=2,primaryPointerType=4,availablePointerTypes=4`

@robingustafsson Are you ok with these changes?

PR #148, opened 2021-12-08
Fix/147 update chrome launch flags

inancgumus:
Please see: #147

### DIFF
--- a/chromium/browser_type_test.go
+++ b/chromium/browser_type_test.go
@@ -25,6 +25,8 @@ import (
 	"testing"
 
 	"github.com/grafana/xk6-browser/common"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestBrowserTypeFlags(t *testing.T) {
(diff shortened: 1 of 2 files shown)


chromium/browser_type_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+		var bt BrowserType
+
+		flags := bt.flags(&common.LaunchOptions{
+			Devtools: false,
+		})
+		flagi, ok := flags[devToolsFlag]
+
+		if !ok {
+			t.Fatalf("%q is missing", devToolsFlag)
```
imiric, line 46: We can rely on the `testify` API to do these checks, so this can be `require.True(t, ok)`. You can use `Truef()` if you want the message as well.

I typically use `require` instead of `assert` if the following code depends on the assertion passing.

chromium/browser_type_test.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		flag, ok := flagi.(bool)
+		if !ok {
+			t.Fatalf("%q should be a bool", devToolsFlag)
```
imiric, line 50: Same here, this can be `require.True(t, ok)`.

chromium/browser_type_test.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		if flag {
+			t.Fatalf("%q should be false when launch options Devtools is false, got %t", devToolsFlag, flag)
```
imiric, line 53: `assert.Falsef(t, flag, ...)`

chromium/browser_type_test.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+
+		flags = bt.flags(&common.LaunchOptions{
+			Devtools: true,
+		})
+		flag = flags[devToolsFlag].(bool)
+		if !flag {
+			t.Fatalf("%q should be true when launch options Devtools is true, got %t", devToolsFlag, flag)
```
imiric, line 61: `assert.Falsef(t, flag, ...)`

chromium/browser_type_test.go
```diff
@@ -0,0 +1,98 @@
...
+		t.Parallel()
+		if runtime.GOOS != "darwin" {
+			t.Skip()
+		}
+
+		var bt BrowserType
+		f := bt.flags(&common.LaunchOptions{})
+
+		if _, ok := f["enable-use-zoom-for-dsf"]; !ok {
+			t.Errorf("darwin should enable 'enable-use-zoom-for-dsf'")
```
imiric, line 95: `assert.Contains(t, f, "enable-use-zoom-for-dsf")`

imiric (changes requested):
LGTM, just some testing suggestions.

I haven't looked at Playwright to compare, so @robingustafsson should approve this before merging.

Just for reference, can you link to the Playwright code where these are set? Both in the PR description and in the commit message?

chromium/browser_type_test.go
inancgumus: I'm not a fan 3rd party pkgs while we have `testing.T`. I'm not sure what value it brings. It's like using a pkg for skip writing `if err != nil`. It steals the expressiveness from code.

But.... Sure.

imiric: It standardizes the way assertions are made, and makes tests more readable precisely by avoiding repetitive checks. I thought you were a fan of readability and abstractions? :)

Besides, it's a pretty widely adopted package, there's no harm in depending on it.

inancgumus: `testing.T` is already standardized, and we have good conventions around it. Anybody who knows how to write idiomatic Go code already knows how to use it effectively.

I guess you misunderstood me before. I'm a fan of readability, but not useless abstractions. I believe in abstractions only if they don't take away any expressiveness from code. `if a != b { t.Errorf(...) }` is as expressive as `if err != nil`.

imiric (approved):

imiric:
> @robingustafsson should approve this before merging.

??

inancgumus:
Ah, sorry, I thought it was ready to merge after you approved the changes.

PR #151, opened 2021-12-09
Improve logs 20211203 (cherry picked)

inancgumus:
I cherry-picked this one from #129 as a separate PR.

@imiric So let's merge this one before merging #129.

### DIFF
--- a/common/connection.go
+++ b/common/connection.go
@@ -192,18 +192,18 @@ func (c *Connection) closeSession(sid target.SessionID, tid target.ID) {
 }
 
 func (c *Connection) createSession(info *target.Info) (*Session, error) {
-	c.logger.Debugf("Connection:createSession", "tid:%v bctxid:%v type:%s", info.TargetID, info.BrowserContextID, info.Type)
+	c.logger.Debugf("Connection:createSession", "tid:%v bid::%v type:%s", info.TargetID, info.BrowserContextID, info.Type)
 
 	var sessionID target.SessionID
 	var err error
 	action := target.AttachToTarget(info.TargetID).WithFlatten(true)
 	if sessionID, err = action.Do(cdp.WithExecutor(c.ctx, c)); err != nil {
-		c.logger.Debugf("Connection:createSession", "tid:%v bctxid:%v type:%s err:%v", info.TargetID, info.BrowserContextID, info.Type, err)
+		c.logger.Debugf("Connection:createSession", "tid:%v bid::%v type:%s err:%v", info.TargetID, info.BrowserContextID, info.Type, err)
 		return nil, err
 	}
 	sess := c.getSession(sessionID)
 	if sess == nil {
-		c.logger.Warnf("Connection:createSession", "tid:%v bctxid:%v type:%s sid:%v, session is nil", info.TargetID, info.BrowserContextID, info.Type, sessionID)
+		c.logger.Warnf("Connection:createSession", "tid:%v bid::%v type:%s sid:%v, session is nil", info.TargetID, info.BrowserContextID, info.Type, sessionID)
 	}
 	return sess, nil
 }
@@ -285,9 +285,10 @@ func (c *Connection) recvLoop() {
 			}
 			eva := ev.(*target.EventAttachedToTarget)
 			sid, tid := eva.SessionID, eva.TargetInfo.TargetID
+
 			c.sessionsMu.Lock()
 			session := NewSession(c.ctx, c, sid, tid, c.logger)
-			c.logger.Debugf("Connection:recvLoop:EventAttachedToTarget", "sid:%v tid:%v wsURL:%q, NewSession", sid, tid, c.wsURL)
+			c.logger.Debugf("Connection:recvLoop:EventAttachedToTarget", "sid:%v tid:%v wsURL:%q", sid, tid, c.wsURL)
 			c.sessions[sid] = session
 			c.sessionsMu.Unlock()
 		} else if msg.Method == cdproto.EventTargetDetachedFromTarget {
@@ -299,7 +300,6 @@ func (c *Connection) recvLoop() {
 			evt := ev.(*target.EventDetachedFromTarget)
 			sid := evt.SessionID
 			tid := c.findTargetIDForLog(sid)
-			c.logger.Debugf("Connection:recvLoop:EventDetachedFromTarget", "sid:%v tid:%v wsURL:%q, closeSession", sid, tid, c.wsURL)
 			c.closeSession(sid, tid)
 		}
 
@@ -327,7 +327,7 @@ func (c *Connection) recvLoop() {
 			}
 
 		case msg.Method != "":
-			c.logger.Debugf("Connection:recvLoop:msg.Method:emit", "method:%q", msg.Method)
+			c.logger.Debugf("Connection:recvLoop:msg.Method:emit", "sid:%v method:%q", msg.SessionID, msg.Method)
 			ev, err := cdproto.UnmarshalMessage(&msg)
 			if err != nil {
 				c.logger.Errorf("cdp", "%s", err)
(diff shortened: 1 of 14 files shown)


common/frame.go
```diff
@@ -133,10 +133,24 @@ type Frame struct {
...
 	pendingDocument *DocumentInfo
+
+	log *Logger
 }
 
 // NewFrame creates a new HTML document frame
-func NewFrame(ctx context.Context, m *FrameManager, parentFrame *Frame, frameID cdp.FrameID) *Frame {
+func NewFrame(ctx context.Context, m *FrameManager, parentFrame *Frame, frameID cdp.FrameID, log *Logger) *Frame {
+	if log.DebugMode() {
+		pfid := ""
```
imiric, line 143:
```suggestion
		var pfid, sid string
```

common/frame.go
```diff
@@ -133,10 +133,24 @@ type Frame struct {
...
+		if parentFrame != nil {
+			pfid = parentFrame.ID()
+		}
+		sid := ""
+		if m != nil && m.session != nil {
+			sid = string(m.session.id)
+		}
+		log.Debugf("NewFrame", "sid:%s tid:%s ptid:%s", sid, frameID, pfid)
```
imiric, line 151:
```suggestion
		log.Debugf("NewFrame", "sid:%s fid:%s pfid:%s", sid, frameID, pfid)
```

common/frame.go
```diff
@@ -182,6 +203,8 @@ func (f *Frame) inflightRequestsLen() int {
 }
 
 func (f *Frame) clearLifecycle() {
+	f.log.Debugf("Frame:clearLifecycle", "tid:%s furl:%q", f.id, f.url)
```
imiric, line 206: Am I wrong in correcting `tid`->`fid`? If so, what does `t` stand for?

I won't mention it again, but it's often used to log `Frame.id` in other places.

```suggestion
	f.log.Debugf("Frame:clearLifecycle", "fid:%s furl:%q", f.id, f.url)
```

common/frame_session.go
```diff
@@ -503,7 +502,7 @@ func (fs *FrameSession) onExceptionThrown(event *runtime.EventExceptionThrown) {
 
 func (fs *FrameSession) onExecutionContextCreated(event *runtime.EventExecutionContextCreated) {
 	fs.logger.Debugf("NewFrameSession:onExecutionContextCreated",
-		"sid:%v tid:%v ectxid:%v", fs.session.id, fs.targetID, event.Context.ID)
+		"sid:%v tid:%v ecid:%d", fs.session.id, fs.targetID, event.Context.ID)
```
imiric, line 505: Same here and in other places, `ectxid` is slightly more descriptive.

common/frame_session.go
```diff
@@ -572,7 +571,7 @@ func (fs *FrameSession) onExecutionContextsCleared() {
 
 func (fs *FrameSession) onFrameAttached(frameID cdp.FrameID, parentFrameID cdp.FrameID) {
 	fs.logger.Debugf("NewFrameSession:onFrameAttached",
-		"sid:%v tid:%v fid:%v pfid:%v",
+		"sid:%v tid:%v fid:%v ptid:%v",
```
imiric, line 574: Yeah, still not sure about the `t` :smile:

`pfid` as short for `parentFrameID` makes more sense to me.

common/frame_session.go
```diff
@@ -737,7 +734,7 @@ func (fs *FrameSession) onPageNavigatedWithinDocument(event *cdppage.EventNaviga
 
 func (fs *FrameSession) onAttachedToTarget(event *target.EventAttachedToTarget) {
 	fs.logger.Debugf("NewFrameSession:onAttachedToTarget",
-		"sid:%v tid:%v esid:%v etid:%v ebctxid:%v type:%q",
+		"sid:%v tid:%v esid:%v etid:%v ebid:%v type:%q",
```
imiric, line 737: Here too, `ebctxid` is more descriptive. It's an "event browser context ID", not an "event browser ID".

imiric (changes requested):
Good call for splitting this into a separate PR. It makes reviewing #129 much easier, thanks!

LGTM, just left some minor suggestions and nitpicks.

The short-hand names we use for some values will surely be inconsistent between log calls, but hopefully we can fix that once we switch to proper structured logs.

common/connection.go
```diff
@@ -192,18 +192,18 @@ func (c *Connection) closeSession(sid target.SessionID, tid target.ID) {
 }
 
 func (c *Connection) createSession(info *target.Info) (*Session, error) {
-	c.logger.Debugf("Connection:createSession", "tid:%v bctxid:%v type:%s", info.TargetID, info.BrowserContextID, info.Type)
+	c.logger.Debugf("Connection:createSession", "tid:%v bid::%v type:%s", info.TargetID, info.BrowserContextID, info.Type)
```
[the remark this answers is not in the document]
inancgumus: The reason that I changed back to `bid` is that these lines take a lot of space and become hard to read :(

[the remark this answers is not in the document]
imiric: Sure, but it's only 3 more chars, I doubt it makes it hard to read because of that :smile: These logs usually span more than one line, and `ctx` would add a bit more clarity.

It's super minor though, chances are I won't even notice and would go look directly at the source anyway :man_shrugging:

common/execution_context.go
```diff
@@ -91,95 +117,108 @@ func (e *ExecutionContext) adoptElementHandle(elementHandle *ElementHandle) (*El
...
 			arguments = append(arguments, result)
 		}
 
-		var remoteObject *runtime.RemoteObject
-		var exceptionDetails *runtime.ExceptionDetails
-		action := runtime.CallFunctionOn(functionText + "\n" + suffix + "\n").
+		var (
+			remoteObject     *runtime.RemoteObject
+			exceptionDetails *runtime.ExceptionDetails
+			functionOn       = expression + "\n" + suffix + "\n"
```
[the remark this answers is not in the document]
inancgumus: I have no idea, but I don't want to introduce something unexpected in this PR. Concatenatinations seem like different things here.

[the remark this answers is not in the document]
imiric: Fair enough. The variable rename would be safe though.

common/frame_session.go
inancgumus: `t` means `targetID`. I must have confused it while navigating in thousands of lines of code 😅

common/execution_context.go
```diff
@@ -89,97 +135,120 @@ func (e *ExecutionContext) adoptElementHandle(elementHandle *ElementHandle) (*El
...
-// evaluate will evaluate provided callable within this execution context and return by value or handle
-func (e *ExecutionContext) evaluate(apiCtx context.Context, forceCallable bool, returnByValue bool, pageFunc goja.Value, args ...goja.Value) (res interface{}, err error) {
+// evaluate will evaluate provided callable within this execution context
+// and return by value or handle
+func (e *ExecutionContext) evaluate(
+	apiCtx context.Context,
+	forceCallable bool, returnByValue bool,
+	pageFunc goja.Value, args ...goja.Value,
+) (res interface{}, err error) {
+
```
imiric, line 153:
```suggestion
```

imiric (approved):
LGTM, just a tiny newline nitpick :)

imiric (approved):
This was very tedious to review :sweat_smile:, but it's better as a separate PR than in #129. It's great to have more logging, though I'm slightly concerned we might run into some data races like in the few places you had to add locking.

I mostly glanced at the latest changes and the overall diff, but didn't spot anything wrong. We can fix any issues if they come up later. :)

inancgumus:
It was also tedious for me to meticulously edit/add all these lines. So I can sympathize 😄  Thank you for taking the time to review all this and your patience! ❤️ 

Yeah, making it a separate PR was hard work, but I think it was worth it—both for reviewing it and putting it into a focused PR for referencing later (I hope not! :))

I have the same concerns, and I liked the way you approached it with a pragmatic attitude 🙇 

Merging, it's time for the real work now!

PR #155, opened 2021-12-10
Update dependencies

inancgumus:


go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/grafana/xk6-browser
 go 1.17
 
 require (
-	github.com/chromedp/cdproto v0.0.0-20211025030258-2570df970243
-	github.com/dop251/goja v0.0.0-20211022113120-dc8c55024d06
+	github.com/chromedp/cdproto v0.0.0-20211205231339-d2673e93eee4
+	github.com/dop251/goja v0.0.0-20211207134806-acd374ca9c94
```
imiric, line 7: Could there be any issues with running a different version of goja than k6, which currently uses `v0.0.0-20211115154819-26ebff68a7d5` on `master`?

imiric (approved):
LGTM

Did you need this for something specific or just for maintenance?

go.mod
inancgumus: Hmm, good point. So far, there is no observed behavior change, neither in tests nor when running examples. It was for maintenance (I've been experimenting with some new CDP actions).

PR #167, opened 2021-12-22
Add metrics emission test

imiric:
This adds a missing test for #40, and all other metrics emitted for `page.goto()`.

It uses a slightly modified version of the [`assertRequestMetricsEmittedSingle` helper from k6](https://github.com/grafana/k6/blob/804203bcf574da57cd05d67b39f03d05c54523ab/js/modules/k6/http/request_test.go#L109).

### DIFF
--- a/tests/network_manager_test.go
+++ b/tests/network_manager_test.go
@@ -31,6 +31,7 @@ import (
 )
 
 func TestDataURLSkipRequest(t *testing.T) {
+	t.Parallel()
 	tb := newTestBrowser(t)
 	p := tb.NewPage(nil)
 
@@ -42,6 +43,7 @@ func TestDataURLSkipRequest(t *testing.T) {
 }
 
 func TestMetricsEmission(t *testing.T) {
+	t.Parallel()
 	tb := newTestBrowser(t, withHTTPServer())
 
 	url := tb.URL("/get")
--- a/tests/network_manager_test.go
+++ b/tests/network_manager_test.go
@@ -22,7 +22,6 @@ package tests
 
 import (
 	"testing"
-	"time"
 
 	"github.com/grafana/xk6-browser/common"
 	"github.com/stretchr/testify/assert"
@@ -46,13 +45,9 @@ func TestMetricsEmission(t *testing.T) {
 	tb := newTestBrowser(t, withHTTPServer())
 
 	url := tb.URL("/get")
-	browserLoadedTags := map[string]string{
-		"group": "",
-		"url":   "about:blank",
-	}
 	browserTags := map[string]string{
 		"group": "",
-		"url":   url,
+		"url":   "about:blank",
 	}
 	httpTags := map[string]string{
 		"method":              "GET",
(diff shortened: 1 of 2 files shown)


tests/network_manager_test.go
```diff
@@ -36,3 +41,72 @@ func TestDataURLSkipRequest(t *testing.T) {
...
+			"method": "GET",
+			"url":    url,
+		},
+		k6metrics.HTTPReqsName:              httpTags,
+		k6metrics.HTTPReqDurationName:       httpTags,
+		k6metrics.DataReceivedName:          httpTags,
+		k6metrics.HTTPReqConnectingName:     httpTags,
+		k6metrics.HTTPReqTLSHandshakingName: httpTags,
+		k6metrics.HTTPReqSendingName:        httpTags,
+		k6metrics.HTTPReqReceivingName:      httpTags,
```
inancgumus, line 83: Should we add [other metrics here](https://pkg.go.dev/go.k6.io/k6/lib/metrics#pkg-constants) as well and check against whether they are zero or not? Are they necessary? For example, `HTTPReqFailedName`?

tests/network_manager_test.go
```diff
@@ -36,3 +41,72 @@ func TestDataURLSkipRequest(t *testing.T) {
...
+	p := tb.NewPage(nil)
+	resp := p.Goto(url, tb.rt.ToValue(struct {
+		WaitUntil string `js:"waitUntil"`
+	}{WaitUntil: "networkidle"}))
+	require.NotNil(t, resp)
+
+	// TODO: Remove this sleep. It's only needed to wait for all metrics to be
+	// emitted, but this should be synchronized with waitUntil
+	// load/networkidle/domcontentloaded. Without this the test would be flaky.
+	time.Sleep(100 * time.Millisecond)
```
inancgumus, line 95: Maybe you can this one to check whether the network is idle/waitUntil/load/etc:

https://github.com/grafana/xk6-browser/blob/758d14bb6082e0c8bbea7933522f7f84cff985fc/common/frame.go#L379-L384

It's because metrics are emitted on lifecycle events:

https://github.com/grafana/xk6-browser/blob/758d14bb6082e0c8bbea7933522f7f84cff985fc/common/frame_session.go#L687-L766

inancgumus (commented):
LGTM, I have some comments.

tests/network_manager_test.go
imiric: These are the expected metrics, and if any unexpected ones are emitted the test will fail at
https://github.com/grafana/xk6-browser/blob/48e102154f011da7e7d2b84bd34b3e23a380046d/tests/metrics_test.go#L43

`HTTPReqFailed` is not emitted in this case.

tests/network_manager_test.go
imiric: Good point, I thought `waitUntil` would accomplish the same thing.

Resolved by 3668b38. It seems stable from my local tests.

inancgumus: Actually, yeah, that's weird and it might even be a bug. A new issue? @imiric

PR #173, opened 2022-01-05
Remove outdated CDP call from Browser

inancgumus:
I tested it without this CDP call, and it doesn't affect anything. The bug in Chrome is long gone.

https://chromium-review.googlesource.com/c/chromium/src/+/2885888

### DIFF
--- a/common/browser.go
+++ b/common/browser.go
@@ -185,15 +185,6 @@ func (b *Browser) initEvents() error {
 	if err := action.Do(cdp.WithExecutor(b.ctx, b.conn)); err != nil {
 		return fmt.Errorf("unable to execute %T: %w", action, err)
 	}
-
-	// Target.setAutoAttach has a bug where it does not wait for new Targets being attached.
-	// However making a dummy call afterwards fixes this.
-	// This can be removed after https://chromium-review.googlesource.com/c/chromium/src/+/2885888 lands in stable.
-	action2 := target.GetTargetInfo()
-	if _, err := action2.Do(cdp.WithExecutor(b.ctx, b.conn)); err != nil {
-		return fmt.Errorf("unable to execute %T: %w", action, err)
-	}
-
 	return nil
 }
 


robingustafsson:
How did you determine this was a fixed bug in Chromium?

AFAICT the linked-to fix has not been merged yet (the status is "open" and the patch is currently marked as producing a merge conflict), and to me it seems to be confirmed when looking at the [relevant code in the branch of the latest Chromium release](https://chromium.googlesource.com/chromium/src/+/refs/tags/99.0.4808.1/content/browser/devtools/protocol/target_handler.cc), but perhaps the problem has been solved via some other changeset 🤷‍♂️ (I don't know enough of the Chromium source to tell).

inancgumus:
When I look at the conflicting changeset, I see this:

Screen Shot 2022-01-05 at 14 15 27

So they added a barrier to wait for attachment and called the `callback` only once.

Then there is a merged changeset of the same method. Here they use a protection mechanism similar to a Barrier: `OnceClosure`. It seems that they fixed it with [this changeset](https://chromium.googlesource.com/chromium/src/+/e3ceda740c6cfd1b86b6c43202424505ffcee466).

Screen Shot 2022-01-05 at 14 19 06

robingustafsson:
That seems to be where they introduced the use of `OnceClosure` yes, in 2018, but the changeset linked to from our code is from 2021 so it's not clear to me that they're not fixing related but different bugs with target auto-discovery. I certainly don't know enough of the Chromium source to say for sure though but the dates and the fact that the 2021 changeset seems to not be merged suggests to me that this issue is not solved. Playwright (where I learned about this issue) still has this workaround in place in `main` branch as well: https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/chromium/crBrowser.ts#L71

inancgumus:
Got it, thank you for explaining 🙏 We should better close this issue for now and resurrect it afterwards then.

PR #174, opened 2022-01-05
Improve mainframe handling

inancgumus:
This PR skips setting the mainframe if it is already the same.

It also adds some sanity checks and protects the mainframe field behind a mutex in places that forgot to use it.

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -166,7 +166,7 @@ func (f *Frame) clearLifecycle() {
 	}
 	f.lifecycleEventsMu.Unlock()
 
-	f.page.frameManager.mainFrame.recalculateLifecycle()
+	f.page.frameManager.MainFrame().recalculateLifecycle()
 
 	// keep the request related to the document if present
 	// in f.inflightRequests
@@ -419,7 +419,7 @@ func (f *Frame) position() *Position {
 	if frame == nil {
 		return nil
 	}
-	if frame == f.page.frameManager.mainFrame {
+	if frame == f.page.frameManager.MainFrame() {
 		return &Position{X: 0, Y: 0}
 	}
 	element := frame.FrameElement()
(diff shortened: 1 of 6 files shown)


common/frame.go
```diff
@@ -1107,6 +1107,10 @@ func (f *Frame) IsVisible(selector string, opts goja.Value) bool {
 
 // ID returns the frame id
 func (f *Frame) ID() string {
+	if f == nil {
```
robingustafsson, line 1110: Wouldn't all of these `if f == nil` checks just mask some more serious issue where the existance of frame is expected and the frame isn't actually there? It feels to me like the handling of a `frame == nil` should be done before making a call to `Frame` methods.

common/request.go
```diff
@@ -264,5 +264,8 @@ func (r *Request) Timing() goja.Value {
 
 // URL returns the request URL
 func (r *Request) URL() string {
+	if r == nil {
```
robingustafsson, line 267: Same question as for the `nil` checks in `Frame`.

robingustafsson (changes requested):
LGTM besides the handling of the `nil` checks. Would like to understand the reasoning there before approving.

common/frame.go
inancgumus: They are mostly used for logging but sure I've removed them in 88b2ba9.

common/request.go
inancgumus: https://github.com/grafana/xk6-browser/pull/174/files#r779340558

inancgumus:
We call these methods when we want to log. When entering to or returning, sometimes, Frame or Request may be nil due to the inner workings of a function. Logging shouldn’t crash the extension, so I added some sanity checks.

Fair point; I agree, they might hide possible issues in the future, so I’ve removed them. 

Mainly, because of `Frame.ID()`: we use it also in non-logging logic.

robingustafsson (approved):
LGTM

PR #176, opened 2022-01-06
Improve execution context logs

inancgumus:
This PR improves execution context logs by including `Frame.ID` and `ExecutionContext.ID`. So we can better understand the originating `Frame` if an error ever occurs again.


robingustafsson (approved):
LGTM

PR #178, opened 2022-01-06
Fix frameAbortedNavigation: nil pendingDocument

### LINKED ISSUE #177: frameAbortedNavigation: nil pendingDocument is shared
I noticed that in the [`FrameManager.frameAbortedNavigation` method](https://github.com/grafana/xk6-browser/blob/aadc37495226dd5e8b06e0758f382213bb2a819b/common/frame_manager.go#L125-L152) we share a `nil` `pendingDocument` (see the lines 140 and 149 below):

https://github.com/grafana/xk6-browser/blob/aadc37495226dd5e8b06e0758f382213bb2a819b/common/frame_manager.go#L140-L149

I've checked [the Playwright code](https://github.com/microsoft/playwright/blob/5ba7903ba098586a13745e0d7ac894f1d55d47aa/packages/playwright-core/src/server/frames.ts#L236-L250
) and they share a `non-nil document` and then set it to `nil` _before_ emitting. So the event goes with a `non-nil` pending document, but the frame gets a `nil` one.

So the correct code should look like this:

```go
ne := &NavigationEvent{
	url:         frame.URL(),
	name:        frame.Name(),
	newDocument: frame.pendingDocument,
	err:         errors.New(errorText),
}
frame.pendingDocument = nil
frame.emit(EventFrameNavigation, ne)
```

I've started working on this today, all the tests pass, and I'm working on adding a new test for this.

@robingustafsson WDYT?

inancgumus:
Closes #177

### DIFF
--- a/common/frame_manager.go
+++ b/common/frame_manager.go
@@ -137,18 +137,19 @@ func (m *FrameManager) frameAbortedNavigation(frameID cdp.FrameID, errorText, do
 	if documentID != "" && frame.pendingDocument.documentID != documentID {
 		return
 	}
-	frame.pendingDocument = nil
 
 	m.logger.Debugf("FrameManager:frameAbortedNavigation:emit:EventFrameNavigation",
 		"fmid:%d fid:%v err:%s docid:%s fname:%s furl:%s",
 		m.ID(), frameID, errorText, documentID, frame.Name(), frame.URL())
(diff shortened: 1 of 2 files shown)


robingustafsson (approved):
LGTM, yes this is a bug.

Issue #187, opened 2022-01-11
WaitForFrameNavigation nil pointer issue


inancgumus:
There is a nil pointer panic if a script calls `page.WaitForNavigation`, and while waiting, the context somehow gets canceled.

For example, Tom's script in #182 calls `WaitForNavigation`. You'll see the following panic if you cancel the script while it's waiting for navigation:

```
(*FrameManager).WaitForFrameNavigation(0x140015c20b0, 0x14000503380, {0x0, 0x0})
	common/frame_manager.go:723 +0x690
(*Frame).WaitForNavigation(...)
	common/frame.go:1431
(*Page).WaitForNavigation(0x140029e8780, {0x0, 0x0})
	common/page.go:881 +0xec
```

Suspicious code:

https://github.com/grafana/xk6-browser/blob/54cd29393f66f411e49f7bd2f44d2816998e6163/common/frame_manager.go#L722-L729

PR #188, opened 2022-01-13
Fix WaitForFrameNavigation nil pointer bug

inancgumus:
This PR fixes [the problem](https://github.com/grafana/xk6-browser/issues/187) and adds an [integration test](https://github.com/grafana/xk6-browser/blob/cde9f02d2012356e852e375a03cce2ecddab0285/tests/page_test.go#L106-L113) and [unit test](https://github.com/grafana/xk6-browser/blob/cde9f02d2012356e852e375a03cce2ecddab0285/common/browser_test.go). It also refactors related code for better maintainability.

**Explanation of the fixes:**
`FrameManager.WaitForFrameNavigation` wasn't returning when the context is canceled, and then it was trying to read a `nil` event, resulting in a `nil pointer` error. [Now it returns](https://github.com/grafana/xk6-browser/blob/cde9f02d2012356e852e375a03cce2ecddab0285/common/frame_manager.go#L705-L710) and prevents this problem. There is no need to throw an error because the extension is in the shutting down stage.

After I fixed the problem, `Browser` started to throw errors. Similarly, it was throwing an error when the context is canceled, but it shouldn't do so when the extension is in the shutting down stage. I fixed this behavior [here](https://github.com/grafana/xk6-browser/blob/cde9f02d2012356e852e375a03cce2ecddab0285/common/browser.go#L245-L249) and [here](https://github.com/grafana/xk6-browser/blob/cde9f02d2012356e852e375a03cce2ecddab0285/common/browser.go#L284-L288).

These problems popped up while trying to test the nil pointer error, and I needed to fix them all. While trying to fix things, I was working on the `newPageInContext` method (it is related to these issues) because it was causing some unexpected issues, and it wasn't correctly returning errors. So I also [refactored](https://github.com/grafana/xk6-browser/blob/cde9f02d2012356e852e375a03cce2ecddab0285/common/browser.go#L338-L388) it in the meanwhile.

**Others:**
* Added a new function named `newBrowser` that returns a new `Browser` value without an actual connection to a running browser. This helped me unit test the Browser without actually connecting to a running Browser.
* This commit removes the `conn` field from the `BrowserContext` because `BrowserContext` never actually uses it. It's better to localize value owners. In this case, the `Browser` is the actual owner of the connection.
* `Browser` now has three fields with three different responsibilities for the same connection value for ease of testing, better maintainability, etc.
* We incorrectly named `Browser.connMu` because it doesn't protect the conn field. Instead, it protects the connected state. So this commit changes its name from `connMu` to `connectedMu`.

### DIFF
--- a/common/browser.go
+++ b/common/browser.go
@@ -243,9 +243,9 @@ func (b *Browser) onAttachedToTarget(ev *target.EventAttachedToTarget) {
 			}
 			select {
 			case <-b.ctx.Done():
-				b.logger.Debugf("Browser:onAttachedToTarget:background_page:return",
-					"sid:%v tid:%v context canceled",
-					ev.SessionID, evti.TargetID)
+				b.logger.Debugf("Browser:onAttachedToTarget:background_page:return:<-ctx.Done",
+					"sid:%v tid:%v err:%v",
+					ev.SessionID, evti.TargetID, b.ctx.Err())
 				return // ignore
 			default:
 				k6Throw(b.ctx, "cannot create NewPage for background_page event: %w", err)
@@ -282,9 +282,9 @@ func (b *Browser) onAttachedToTarget(ev *target.EventAttachedToTarget) {
 			}
 			select {
 			case <-b.ctx.Done():
-				b.logger.Debugf("Browser:onAttachedToTarget:page:return",
-					"sid:%v tid:%v context canceled",
-					ev.SessionID, evti.TargetID)
+				b.logger.Debugf("Browser:onAttachedToTarget:page:return:<-ctx.Done",
+					"sid:%v tid:%v err:%v",
+					ev.SessionID, evti.TargetID, b.ctx.Err())
 				return // ignore
 			default:
 				k6Throw(b.ctx, "cannot create NewPage for page event: %w", err)
@@ -381,7 +381,7 @@ func (b *Browser) newPageInContext(id cdp.BrowserContextID) (page *Page, err err
 		page = b.pages[tid]
 		b.pagesMu.RUnlock()
 	case <-ctx.Done():
-		b.logger.Debugf("Browser:newPageInContext:<-ctx.Done", "tid:%v bctxid:%v", tid, id)
+		b.logger.Debugf("Browser:newPageInContext:<-ctx.Done", "tid:%v bctxid:%v err:%v", tid, id, ctx.Err())
 		err = ctx.Err()
 	}
 	return page, err
--- a/common/browser.go
+++ b/common/browser.go
@@ -26,7 +26,6 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
-	"time"
 
 	"github.com/chromedp/cdproto"
 	cdpbrowser "github.com/chromedp/cdproto/browser"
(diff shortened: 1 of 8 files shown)


common/browser.go
```diff
@@ -306,56 +322,63 @@ func (b *Browser) onDetachedFromTarget(ev *target.EventDetachedFromTarget) {
...
+		err = fmt.Errorf("newPageInContext cannot execute %T: %w",
+			action, err)
+		return nil, err
+	}
+	// let the event handler know about the new page's targetID.
+	targetID <- tid
+
 	select {
+	case <-waitForPage:
+		b.logger.Debugf("Browser:newPageInContext:<-ch",
```
inancgumus, line 361: Make this ’<-waitForPage’

common/browser.go
```diff
@@ -306,56 +322,63 @@ func (b *Browser) onDetachedFromTarget(ev *target.EventDetachedFromTarget) {
...
+	// create a new page.
+	action := target.CreateTarget("about:blank").WithBrowserContextID(id)
+	tid, err := action.Do(cdp.WithExecutor(b.ctx, b.conn))
+	if err != nil {
+		err = fmt.Errorf("newPageInContext cannot execute %T: %w",
+			action, err)
+		return nil, err
+	}
+	// let the event handler know about the new page's targetID.
+	targetID <- tid
```
inancgumus, line 357: Send this before the error or the event GR may leak.

inancgumus (commented):
Self TODO

common/browser.go
```diff
@@ -306,56 +333,71 @@ func (b *Browser) onDetachedFromTarget(ev *target.EventDetachedFromTarget) {
...
+	defer removeEventHandler()
+
+	// create a new page.
+	action := target.CreateTarget("about:blank").WithBrowserContextID(id)
+	tid, err := action.Do(cdp.WithExecutor(b.ctx, b.conn))
+	// let the event handler know about the new page's targetID.
+	// signaling here to prevent the event handler GR above not to leak.
+	targetID <- tid
+	if err != nil {
+		err = fmt.Errorf("newPageInContext cannot execute %T: %w",
```
inancgumus, line 376: TODO: Remove newPageInContext from error here.

common/browser_test.go
```diff
@@ -0,0 +1,159 @@
...
+		// newPageInContext will look for this browser context.
+		tc.b.contexts[browserContextID] = tc.bc
+		// newPageInContext will return this page by searching
+		// it by its targetID.
+		tc.b.pages[targetID] = &Page{targetID: targetID}
+
+		done := make(chan struct{})
+		go func() {
+			tc.b.conn = executorTestFunc(func(ctx context.Context, method string, params easyjson.Marshaler, res easyjson.Unmarshaler) error {
+				tc.got.createTargetParams = params.(*target.CreateTargetParams)
```
inancgumus, line 79: TODO: Check for iface, save into got as a failed. Return.

common/browser_test.go
```diff
@@ -0,0 +1,159 @@
...
+			require.Nil(t, tc.got.page)
+			require.NotNil(t, tc.got.err)
+			require.Contains(t, tc.got.err.Error(), wantErr, "wrong error message received")
+		case <-time.After(5 * time.Second):
+			require.FailNow(t, "timed out")
+		}
+	})
+}
+
+type executorTestFunc func(ctx context.Context, method string, params easyjson.Marshaler, res easyjson.Unmarshaler) error
```
inancgumus, line 155: TODO: actionTestFunc?

inancgumus (commented):
TODO

imiric (approved):
Very nice, just a couple of nitpicks. Good job with the tests too! :clap:

common/browser.go
```diff
@@ -306,56 +335,56 @@ func (b *Browser) onDetachedFromTarget(ev *target.EventDetachedFromTarget) {
 	}
 }
 
-func (b *Browser) newPageInContext(id cdp.BrowserContextID) (*Page, error) {
+func (b *Browser) newPageInContext(id cdp.BrowserContextID) (page *Page, err error) {
```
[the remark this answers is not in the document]
inancgumus: 👍  Resolved by: [295182](https://github.com/grafana/xk6-browser/pull/188/commits/2951829494e05feb0f017f498b193892622bff0a)

Issue #170, opened 2021-12-28
Frame.QueryAll does not work


inancgumus:
Given:

```html
<ul>
	<li class="ali">1</li>
	<li class="ali">2</li>
</ul>
```

The following does not work:

```js
...
const lis = page.$$("li.ali");
console.log(len(designOptions) == 2); // false
```

PR #195, opened 2022-01-19
Fix QueryAll cannot return found elements

inancgumus:
This fixes `Page.QueryAll`, `Frame.QueryAll`, and `ElementHandle.QueryAll`. `ElementHandle.QueryAll` wasn't returning elements despite finding them because it was returning a wrong slice.

* Adds: An integration test: `TestElementHandleQueryAll`
* Refactors: `ElementHandle.QueryAll`
* Refactors: `ExecutionContext.getInjectedScript`
* Refactors: `k6Throw` calls
* Moves: `injectedScriptSource` into `ExecutionContext` because the `ExecutionContext` is the only one that uses it.

Fixes: #170


common/element_handle.go
```diff
@@ -1327,51 +1326,56 @@ func (h *ElementHandle) QueryAll(selector string) []api.ElementHandle {
...
 		} else {
-			property.Dispose()
+			elements = append(elements, eh)
 		}
 	}
 	applySlowMo(h.ctx)
 	return elements
 }
 
+// func (h *ElementHandle) queryAll(selector string) []*ElementHandle {
```
inancgumus, line 1376: Remove.

inancgumus (commented):
Self TODO.

inancgumus:
I've been trying to refactor and write a unit test, but it takes more work than anticipated. So I'm going to create another PR for the unit testing work and send this PR for review.

common/element_handle.go
```diff
@@ -1327,45 +1326,47 @@ func (h *ElementHandle) QueryAll(selector string) []api.ElementHandle {
...
-		elementHandle := property.AsElement()
-		if elementHandle != nil {
-			result = append(elements, elementHandle)
+	for _, p := range properties {
+		eh := p.AsElement()
+		if eh == nil {
+			p.Dispose()
 		} else {
-			property.Dispose()
+			elements = append(elements, eh)
```
imiric, line 1369: This is a single line fix, right? If so, can you split it into a separate commit and all the refactors in another? It's difficult to see what the actual fix is this way.

The two current ones (80818cf and 861cea8) can be squashed.

...

Actually, this fix is already on `main` and was inadvertently part of #196. I noticed it when reviewing, but forgot to comment about it :facepalm:

So I guess you can just fix the merge conflicts here and make this an exclusively refactor PR.

Sidenote: this is why it's risky to mix refactors and fixes, in same commits or in same PRs. They have very different implications and refactors often obscure fixes.

imiric (changes requested):
The refactors LGTM, but please update the PR description after you rebase to note that this is a refactor PR now.

common/element_handle.go
inancgumus: I should probably have missed this while separating the other refactoring PR... 🤦 Actually, this is not supposed to be a refactor PR but it has leftover code from the upcoming (not yet exists on Github) unit test PR. There is also a test in this PR for this fix. I'll rebase and see how it goes.

imiric: Right, then I would make a separate PR for adding the integration test, maybe along with the other unit tests you're working on, and make this one a refactor only.

imiric (approved):
LGTM

You don't want to include those unit tests here as well?

inancgumus:
> You don't want to include those unit tests here as well?

No, that's a more significant change and needs a couple of refactoring PRs first.

PR #196, opened 2022-01-19
Refactor ElementHandle evaluate calls

inancgumus:
This PR refactors `ElementHandle`'s evaluate calls that happen through the `ExecutionContext,` which are repetitive. This is not a critical change, but I worked on it while fixing #170 and wanted to create a separate PR.

We now provide two new unexported methods:
* [`evalWithScript`](https://github.com/grafana/xk6-browser/blob/9b773bffd96c063363e3897d06b687ec79f1d64f/common/element_handle.go#L1428-L1442): Runs a javascript code in the current `ElementHandle` scope and provides an ability to use the injected script's functions. In the previous code version, each `ElementHandle` function needs to pass the current `ElementHandle` and call `getInjectedScript` and then pass it to the `ExecutionContext`'s `evaluate` call. This function makes this easier.
* [`eval`](https://github.com/grafana/xk6-browser/blob/9b773bffd96c063363e3897d06b687ec79f1d64f/common/element_handle.go#L1444-L1463): Similar to `evalWithScript` but it runs the given javascript code without the injected script.

Another advantage of doing this is that the methods don't need to call `rt.ToValue` everywhere. However, there is a disadvantage (IDK how critical it is), though: We need to [convert the arguments to `goja.Value`s](https://github.com/grafana/xk6-browser/blob/9b773bffd96c063363e3897d06b687ec79f1d64f/common/element_handle.go#L1451-L1456) in the `eval` functions I explained above.

Naming changes:
* ExecutionContext:
  * Evaluate -> Eval
  * EvaluateHandle -> EvalHandle
  * evaluate -> eval
* evaluateOptions -> evalOptions

### DIFF
--- a/common/element_handle.go
+++ b/common/element_handle.go
@@ -1414,7 +1414,7 @@ func (h *ElementHandle) evalWithScript(
 	return h.eval(ctx, opts, js, args...)
 }
 
-// eval evaluates a given js function in the scope of this ElementHandle.
+// eval evaluates the given js code in the scope of this ElementHandle and returns the result.
 func (h *ElementHandle) eval(
 	ctx context.Context,
 	opts evaluateOptions, js string, args ...interface{},
--- a/common/element_handle.go
+++ b/common/element_handle.go
@@ -231,7 +231,7 @@ func (h *ElementHandle) checkHitTargetAt(apiCtx context.Context, point Position)
 			return injected.checkHitTargetAt(node, point);
 		}
 	`
-	opts := evaluateOptions{
+	opts := evalOptions{
 		forceCallable: true,
 		returnByValue: true,
 	}
@@ -257,7 +257,7 @@ func (h *ElementHandle) checkElementState(apiCtx context.Context, state string) 
 			return injected.checkElementState(node, state);
 		}
 	`
-	opts := evaluateOptions{
+	opts := evalOptions{
 		forceCallable: true,
 		returnByValue: true,
 	}
@@ -378,7 +378,7 @@ func (h *ElementHandle) dispatchEvent(apiCtx context.Context, typ string, eventI
 			injected.dispatchEvent(node, type, eventInit);
 		}
 	`
-	opts := evaluateOptions{
+	opts := evalOptions{
 		forceCallable: true,
 		returnByValue: true,
 	}
@@ -392,7 +392,7 @@ func (h *ElementHandle) fill(apiCtx context.Context, value string) (interface{},
 			return injected.fill(node, value);
 		}
 	`
-	opts := evaluateOptions{
+	opts := evalOptions{
 		forceCallable: true,
 		returnByValue: true,
 	}
@@ -415,7 +415,7 @@ func (h *ElementHandle) focus(apiCtx context.Context, resetSelectionIfNotFocused
 			return injected.focusNode(node, resetSelectionIfNotFocused);
 		}
 	`
-	opts := evaluateOptions{
+	opts := evalOptions{
 		forceCallable: true,
 		returnByValue: true,
 	}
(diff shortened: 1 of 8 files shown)


common/element_handle.go
```diff
@@ -1521,29 +1373,64 @@ func (h *ElementHandle) Uncheck(opts goja.Value) {
...
+	ctx context.Context,
+	opts evaluateOptions, js string, args ...interface{},
+) (interface{}, error) {
+	// get the injected script's handle so that js function can call its functions.
+	script, err := h.execCtx.getInjectedScript(h.ctx)
+	if err != nil {
+		return nil, fmt.Errorf("cannot get injected script: %w", err)
+	}
+	// passing `script` allows `js` to use `script`'s functions.
+	args = append(append([]interface{}(nil), script), args...)
```
inancgumus, line 1413: Simplify.

common/element_handle.go
```diff
@@ -1521,29 +1373,64 @@ func (h *ElementHandle) Uncheck(opts goja.Value) {
...
+	return h.eval(ctx, opts, js, args...)
+}
+
+// eval evaluates a given js function in the scope of this ElementHandle.
+func (h *ElementHandle) eval(
+	ctx context.Context,
+	opts evaluateOptions, js string, args ...interface{},
+) (interface{}, error) {
+	// passing `h` makes it evaluate js code in the element handle's scope.
+	args = append(append([]interface{}(nil), h), args...)
```
inancgumus, line 1423: Simplify.

inancgumus (commented):
Self TODO.

common/element_handle.go
```diff
@@ -1521,29 +1373,64 @@ func (h *ElementHandle) Uncheck(opts goja.Value) {
...
+	script, err := h.execCtx.getInjectedScript(h.ctx)
+	if err != nil {
+		return nil, fmt.Errorf("cannot get injected script: %w", err)
+	}
+	// passing `script` allows `js` to use `script`'s functions.
+	args = append([]interface{}{script}, args...)
+	return h.eval(ctx, opts, js, args...)
+}
+
+// eval evaluates a given js function in the scope of this ElementHandle.
```
imiric, line 1417: Similar to above:
```suggestion
// eval evaluates the given js code in the scope of this ElementHandle.
```

common/element_handle.go
```diff
@@ -1521,29 +1373,64 @@ func (h *ElementHandle) Uncheck(opts goja.Value) {
...
+	// passing `h` makes it evaluate js code in the element handle's scope.
+	args = append([]interface{}{h}, args...)
+	// convert the arguments to goja values.
+	rt := k6common.GetRuntime(ctx)
+	gargs := make([]goja.Value, len(args))
+	for i, arg := range args {
+		gargs[i] = rt.ToValue(arg)
+	}
+	// call the javascript function.
+	result, err := h.execCtx.evaluate(ctx, opts, rt.ToValue(js), gargs...)
```
imiric, line 1431: Same here about the comments being redundant, though the first one about ```passing `h` ...``` is somewhat useful. :)

Also, should we rename `ExecutionContext.evaluate` to `eval` as well, to make it consistent?

imiric (changes requested):
Very nice, this looks much cleaner now! :clap: It's always nice to see a net negative LOC refactor ;) Also, thanks for making this a separate PR, it would be too much to review with other changes.

Just left a few nitpicks about comments and such.

Re: converting arguments to `goja.Value`, that was what we did before anyway, right? So it shouldn't be a problem now.

inancgumus:
> Very nice, this looks much cleaner now! 👏 It's always nice to see a net negative LOC refactor ;) Also, thanks for making this a separate PR, it would be too much to review with other changes.

Sure thing! Thanks for your kind words, @imiric, glad to hear :) Always helpful.

> Re: converting arguments to goja.Value, that was what we did before anyway, right? So it shouldn't be a problem now.

Fair enough!

imiric (approved):
LGTM

Remember to re-request a review when you're done with the changes.

inancgumus:
> Remember to re-request a review when you're done with the changes.

I usually do, but human nature, we're creatures that forget a lot 🤷 :)

PR #201, opened 2022-01-21
Revert "Refactor ElementHandle evaluate calls"

inancgumus:
I'd like to revert #196, remove the `QueryAll` fix and add it to the actual fix PR. I'll also put the other refactorings in #195 to another PR—leave it only with the fix. So we can better point it out in the release and have a better history.


imiric (approved):
Sure :+1: Just fix that linter error.

inancgumus:
@imiric The linter error is about the fix itself! :)

```
common/element_handle.go:1366:4: ineffectual assignment to result (ineffassign)
			result = append(elements, elementHandle)
```

imiric:
Ah, indeed :smile: I would just `nolint` it since it will be fixed in a follow-up PR.

Issue #105, opened 2021-11-12
Make sure we adhere to the k6 options `DNS`, `Blacklist IPs`, `Block hostnames` and `Hosts`


robingustafsson:
Right now xk6-browser doesn't obey the following network related k6 options, but we should (if possible that is):

- [x] [DNS](https://k6.io/docs/using-k6/options/#dns) (N/A, see [comment](https://github.com/grafana/xk6-browser/issues/105#issuecomment-1022093138))
- [x] [Blacklist IPs](https://k6.io/docs/using-k6/options/#blacklist-ips) (PR #205)
- [x] [Block hostnames](https://k6.io/docs/using-k6/options/#block-hostnames) (PR #204)
- [x] [Hosts](https://k6.io/docs/using-k6/options/#hosts) (PR #207)

It should hopefully be possible by using the experimental [`Fetch`](https://chromedevtools.github.io/devtools-protocol/tot/Fetch/) CDP domain and intercepting requests and either aborting request with [`Fetch.failRequest(Network.ErrorReason.BlockedByClient)`](https://chromedevtools.github.io/devtools-protocol/tot/Fetch/#method-failRequest) or modifying the URL and then calling [`Fetch.continueRequest()`](https://chromedevtools.github.io/devtools-protocol/tot/Fetch/#method-continueRequest).

This is also very much related to [`BrowserContext.route()`](https://github.com/grafana/xk6-browser/issues/10) and [`BrowserContext.unroute()`](https://github.com/grafana/xk6-browser/issues/12) as they would also use the Fetch CDP domain APIs.

imiric:
This seems to be possible by handling the `Fetch` events from Go. At least it looks supported by cdproto: https://pkg.go.dev/github.com/chromedp/cdproto/fetch
Here's an example of modifying a request: https://github.com/chromedp/chromedp/issues/805#issuecomment-829806460

Playwright and Puppeteer additionally expose this API to users, so requests can be intercepted from JS. See [this repo](https://github.com/subwaymatch/cdp-modify-response-example) for examples.
This isn't a requirement for this issue, and we won't be able to implement it without an event loop anyway, so we can postpone this for now. I'm just mentioning it as something we might want to eventually expose to JS as well.

I'll experiment with doing this from Go somewhere in `BrowserContext` most likely. :hamster:

imiric:
@robingustafsson I'm working on a PoC for this feature, and noticed there's some previous work in this area:

[...]

Issue #203, opened 2022-01-24
Add HTTP Basic Auth


imiric:
The `httpCredentials` `BrowserContext` option doesn't seem to be working, and HTTP basic auth in general seems to be partially implemented.

Version: de8c9d2

## Steps to reproduce

Run the following script:

```javascript
import launcher from 'k6/x/browser';
import { sleep } from 'k6';

export default function() {
  const browser = launcher.launch('chromium', {
    headless: false,
  });
  const context = browser.newContext({
    httpCredentials: {
      username: 'user',
      password: 'pass',
    }
  });
  const page = context.newPage();

  page.goto('https://authenticationtest.com/HTTPAuth/', { waitUntil: 'load' });
  sleep(3);

  page.close();
  browser.close();
}
```

On de8c9d2 (#202) this fails with:

```
ERRO[0001] Failed to load resource: the server responded with a status of 401 ()  browser_source=network line_number=0 source=browser stacktrace="<nil>" url="chrome-error://chromewebdata/"
ERRO[0001] navigation interrupted by another one
running at reflect.methodValueCall (native)
default at file:///home/ivan/Projects/grafana/xk6-browser/examples/test-105-cdp-fetch.js:16:56(34)
```

Commenting out the [`navigation interrupted by another one` line](https://github.com/grafana/xk6-browser/blob/c613b929657dce22874b621f7dfeff61dd25205b/common/frame_manager.go#L632), shows the browser window and credential prompt, and the above `401` error is logged.

## Implementation suggestions

This seems to require request interception with the `Fetch` domain, which is partially implemented [here](https://github.com/grafana/xk6-browser/blob/c613b929657dce22874b621f7dfeff61dd25205b/common/network_manager.go#L430). What's missing is a call to [`fetch.ContinueWithAuth()`](https://pkg.go.dev/github.com/chromedp/cdproto@v0.0.0-20220122011257-7936d5887bac/fetch#ContinueWithAuth).

unlikelyzero:
Currently this is a blocker for our internal performance testing as everything is (presently, but will not always be) routed through basic auth.

imiric:
Hi @unlikelyzero, basic auth support will be released in the upcoming v0.2.0, scheduled for later this week, or you can use it now by building the extension from `main`.

PR #204, opened 2022-01-25
Add request blocking based on k6 `blockHostnames` option

imiric:
Partially implements #105

I decided to leave most of the previous request interception code intact (`api.Route`, etc.), except for the non-functional parts (47b0121). This will be useful once we expose this API to JS, which seems to have been the original idea, and is [what Playwright and Puppeteer already do](https://github.com/grafana/xk6-browser/issues/105#issuecomment-1016655405). For now, the [`Fetch` CDP domain](https://chromedevtools.github.io/devtools-protocol/tot/Fetch/) is only enabled and used internally.

### DIFF
--- a/common/frame_manager.go
+++ b/common/frame_manager.go
@@ -24,6 +24,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -573,7 +574,8 @@ func (m *FrameManager) NavigateFrame(frame *Frame, url string, opts goja.Value) 
 		"fmid:%d fid:%v furl:%s url:%s", fmid, fid, furl, url)
 
 	rt := k6common.GetRuntime(m.ctx)
-	defaultReferer := m.page.mainFrameSession.getNetworkManager().extraHTTPHeaders["referer"]
+	netMgr := m.page.mainFrameSession.getNetworkManager()
+	defaultReferer := netMgr.extraHTTPHeaders["referer"]
 	parsedOpts := NewFrameGotoOptions(defaultReferer, time.Duration(m.timeoutSettings.navigationTimeout())*time.Second)
 	if err := parsedOpts.Parse(m.ctx, opts); err != nil {
 		k6common.Throw(rt, fmt.Errorf("failed parsing options: %w", err))
@@ -627,10 +629,16 @@ func (m *FrameManager) NavigateFrame(frame *Frame, url string, opts goja.Value) 
 		if err != nil {
 			k6common.Throw(rt, err)
 		}
+
 		event = data.(*NavigationEvent)
 		if event.newDocument.documentID != newDocumentID {
-			k6common.Throw(rt, errors.New("navigation interrupted by another one"))
-		} else if event.err != nil {
+			m.logger.Debugf("FrameManager:NavigateFrame:interrupted",
+				"fmid:%d fid:%v furl:%s url:%s docID:%s newDocID:%s",
+				fmid, fid, furl, url, event.newDocument.documentID, newDocumentID)
+		} else if event.err != nil &&
+			// TODO: A more graceful way of avoiding Throw()?
+			!(netMgr.userReqInterceptionEnabled &&
+				strings.Contains(event.err.Error(), "ERR_BLOCKED_BY_CLIENT")) {
 			k6common.Throw(rt, event.err)
 		}
 	} else {
--- a/common/network_manager.go
+++ b/common/network_manager.go
@@ -23,7 +23,6 @@ package common
 import (
 	"context"
 	"fmt"
-	"net"
 	"strconv"
 	"sync"
 	"time"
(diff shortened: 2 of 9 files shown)


common/network_manager.go
```diff
@@ -235,10 +237,19 @@ func (m *NetworkManager) handleRequestRedirect(req *Request, redirectResponse *n
...
 func (m *NetworkManager) initDomains() error {
-	action := network.Enable()
-	if err := action.Do(cdp.WithExecutor(m.ctx, m.session)); err != nil {
-		return fmt.Errorf("unable to execute %T: %w", action, err)
+	actions := []Action{network.Enable()}
+
+	// Only enable the Fetch domain if necessary, as it has a performance overhead.
+	if m.userReqInterceptionEnabled {
+		actions = append(actions,
+			fetch.Enable().WithPatterns([]*fetch.RequestPattern{{URLPattern: "*"}}))
```
imiric, line 245: I'm wondering if we should disable the network cache here as well, as it's done when setting up the [non-working](https://github.com/grafana/xk6-browser/issues/203) HTTP auth. See: https://github.com/grafana/xk6-browser/blob/ed0de4f92c91dcd848243861fd62a352bbbb1cc4/common/network_manager.go#L437-L443

@robingustafsson Can you comment on if this is required? I suppose it makes sense, just wanted to confirm.

robingustafsson (approved):
LGTM

common/network_manager.go
robingustafsson: I'm not sure tbh. I've not found any clear info on how the `Network` and `Fetch` domains interact. I'd say it's a good precaution though to disable cache when we have to use `Fetch` domain given what we'll use it for.

imiric: Done in 49b6724.

common/frame_manager.go
```diff
@@ -627,10 +629,16 @@ func (m *FrameManager) NavigateFrame(frame *Frame, url string, opts goja.Value)
...
+
 		event = data.(*NavigationEvent)
 		if event.newDocument.documentID != newDocumentID {
-			k6common.Throw(rt, errors.New("navigation interrupted by another one"))
-		} else if event.err != nil {
+			m.logger.Debugf("FrameManager:NavigateFrame:interrupted",
+				"fmid:%d fid:%v furl:%s url:%s docID:%s newDocID:%s",
+				fmid, fid, furl, url, event.newDocument.documentID, newDocumentID)
+		} else if event.err != nil &&
+			// TODO: A more graceful way of avoiding Throw()?
```
inancgumus, line 639: @imiric What do you mean here?

common/session.go
```diff
@@ -32,9 +32,13 @@ import (
...
-// Ensure Session implements the EventEmitter and Executor interfaces
-var _ EventEmitter = &Session{}
-var _ cdp.Executor = &Session{}
+var _ session = &Session{}
+
+type session interface {
+	EventEmitter
+	cdp.Executor
+	ID() target.SessionID
+}
```
inancgumus, line 41: @imiric Suggestion: Instead of making a Java-style preemptive interface, I think we better make the network manager accept only what it needs as I did in the browser:

https://github.com/grafana/xk6-browser/blob/112193c09aa304e6a33dbdee71c10e5776faaa82/common/browser.go#L61-L67

inancgumus: Or, we can create an interface for each component like:

```golang
type ExecuteListener interface {
    EventEmitter
    cdp.Executor
}
```

As far as I see, this is the core behavior they all use.

common/frame_manager.go
imiric: I mean that this stringy check is clunky, but I didn't look too much into it. E.g. could we avoid setting `event.err` somewhere lower level so that this is never reached?

common/session.go
imiric: :man_shrugging: What's wrong with the current approach? Yes, it's mostly a useless interface that only has a single implementation, and it was only introduced so that it could be mocked in tests, but I don't see much harm with it.

I can't say that I'm a fan of defining anonymous types like that outside of tests. We might want to reuse `session` elsewhere where `*Session` is used now, and it would be easy to miss that it's defined inline only for `NetworkManager`.

> As far as I see, this is the core behavior they all use.

We also need `ID()`, though I didn't mean to change it everywhere else. We can get to making cleaner and reusable interfaces when we split `common`?

inancgumus: > We can get to making cleaner and reusable interfaces when we split common?
I believe in starting refactoring here and there, little by little, instead of a significant refactoring which would be more challenging than today or in the future.

> What's wrong with the current approach?
What if another test wants to test something else? Are we going to add more methods to the interface and break all the other tests depending on it? This problem is the first one that comes to my mind right now.

Also, the Session concept has now two interfaces (_I don't mean Go interfaces :)_): `*Session` and `session,` which is confusing to me.

I believe code should change when necessary not for the future stuff that we can't see clearly from this moment. Right now, only the `NetworkManager` uses `session`, and another part of the code depends on `*Session`, increasing the complexity further.

But instead, if code had only accepted what it needs, all of these problems would evaporate, and we can keep refactoring the existing code without all these problems.

common/frame_manager.go
imiric: Yeah, possibly. I didn't spend much time on it, just wanted to avoid calling `Throw()` in this case, but yeah, it needs more investigation, hence the `TODO`. :)

common/session.go
imiric: > I believe code should change when necessary not for the future stuff that we can't see clearly from this moment.

Exactly why I'm against trying to design clean interfaces when we don't have a clear need for them yet. Doing this prematurely without specific requirements is guesswork at best, but it influences future changes once they become necessary.

> Right now, only the `NetworkManager` uses `session`, and another part of the code depends on `*Session`, increasing the complexity further.

Yeah, like I said, it was only added to be able to test `NetworkManager`. I don't want to get into a major refactor when it's not needed.

But I agree that the current `session` interface is sloppy. I'll make some changes based on your feedback and we can discuss it in a PR.

inancgumus: The difference between `session` and `ExecuteEmitter` (or whatever or anonymous one) is that the current concrete implementations are let us discover the latter (`ExecuteEmitter`). I don't believe it's a premature abstraction. Alright, let's discuss afterward.

PR #205, opened 2022-01-26
Add request blocking based on k6 `blacklistIPs` option

imiric:
Partially implements #105

This also fixes an issue with handling URLs with a port, and does a minor refactor to return `error` instead of `string` from `check*` functions.


common/network_manager.go
```diff
@@ -419,14 +430,22 @@ func (m *NetworkManager) failOrContinueRequest(event *fetch.EventRequestPaused,
...
+func checkBlockedHosts(host string, blockedHosts *k6types.HostnameTrie) error {
+	if match, blocked := blockedHosts.Contains(host); blocked {
+		return fmt.Errorf("hostname %s is in a blocked pattern (%s)", host, match)
+	}
+	return nil
+}
+
+func checkBlockedIPs(ip net.IP, blockedIPs []*k6lib.IPNet) error {
+	for _, ipnet := range blockedIPs {
+		if ipnet.Contains(ip) {
```
imiric, line 442: Note that this doesn't do a lookup to block the IP a host resolves to (as [this comment implied](https://github.com/grafana/xk6-browser/commit/47b0121ff97e9fb2d8b5e004f82c785044598924/#diff-30b70a5056bb318589cf41ae368bc837cc814b2f2d7bd0fbd24cb7fc2ed0294fL381)), just does a basic match.

I don't see a way we can get the resolved IP at this stage, so we might have to do an additional lookup if we want this feature. Should we if it's not possible otherwise?

robingustafsson: True, and I think we should do a lookup, yes. Shouldn't add much overhead time-wise. Would perhaps interfere with DNS lookup timings (when the actual requests happens in the browser) if the lookup is cached locally as a result of this blocklist check?

imiric: It could definitely mess with the timings, but the caches would likely be different, since here we'll use Go's built-in resolver (which could also be forced to use `getaddrinfo` via cgo), and the browser has its own internal cache. And I'm not sure what's the status of supporting e.g. `/etc/resolv.conf`, etc.

Since this is an interoperability mess, I'm slightly concerned we might run into a situation where our lookup returns one result, and the browser's returns another. :man_shrugging: I guess we'll cross that bridge when we get there, but it's not out of the question we might have to.

I'll see if there's another way of getting the IP, otherwise I'll add the lookup here.

imiric: This should be _resolved by_ ;) 0cc6e6b. We have to create our own `Resolver` instance since the k6 one is not accessible to extensions :disappointed:, but it should be identical in behavior as the k6 one, since it's built with the same DNS options.

@robingustafsson Let me know if there's anything pending here.

common/network_manager_utils.go
```diff
@@ -0,0 +1,76 @@
+package common
```
imiric, line 1: I know that @inancgumus is not a fan of lib/helpers/utils files and packages, and neither am I if it can be avoided, but I really don't know what to name this one... Any suggestions?

I didn't want to clutter `network_manager.go` even more and these are only relevant to that file, so I'm not against this too much.

robingustafsson (approved):
LGTM now!

inancgumus:
I've started reviewing this PR.

common/network_manager_test.go
```diff
@@ -123,3 +142,69 @@ func TestOnRequestPaused(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			blockedIPs := make([]*k6lib.IPNet, 0, len(tc.blockedIPs))
+			for _, ipcidr := range tc.blockedIPs {
+				ipnet, err := k6lib.ParseCIDR(ipcidr)
+				require.NoError(t, err)
+				blockedIPs = append(blockedIPs, ipnet)
+			}
```
inancgumus, line 193: ?

```suggestion
			blockedIPs := make([]*k6lib.IPNet, len(tc.blockedIPs))
			for i, ipcidr := range tc.blockedIPs {
				ipnet, err := k6lib.ParseCIDR(ipcidr)
				require.NoError(t, err)
				blockedIPs[i] = ipnet
			}
```

common/network_manager_utils.go
```diff
@@ -0,0 +1,76 @@
...
+	k6netext "go.k6.io/k6/lib/netext"
+	k6types "go.k6.io/k6/lib/types"
+)
+
+func checkBlockedHosts(host string, blockedHosts *k6types.HostnameTrie) error {
+	if blockedHosts == nil {
+		return nil
+	}
+	if match, blocked := blockedHosts.Contains(host); blocked {
+		return fmt.Errorf("hostname %s is in a blocked pattern (%s)", host, match)
```
inancgumus, line 18:
```suggestion
		return fmt.Errorf("hostname %q is in a blocked pattern %q", host, match)
```

common/network_manager_utils.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	return nil
+}
+
+func checkBlockedIPs(ip net.IP, blockedIPs []*k6lib.IPNet) error {
+	for _, ipnet := range blockedIPs {
+		if ipnet.Contains(ip) {
+			// TODO: Return netext.BlackListedIPError here once its private
+			// fields are exported, or there's a constructor for it.
+			return fmt.Errorf("IP (%s) is in a blacklisted range (%s)", ip, ipnet)
```
inancgumus, line 28:
```suggestion
			return fmt.Errorf("IP (%q) is in a blacklisted range (%q)", ip, ipnet)
```

inancgumus (approved):
LGTM, excellent work 👏  I can't review anything related to the networking code as I'm not as experienced in it as you are. I have some suggestions, though.

common/network_manager_utils.go
imiric: :-/ Similarly to discussions in #204, I disagree that functions should be defined inline somewhere just because they're used only there. Making them closures inside `onRequestPaused` would make it too long to be readable (probably triggering the function length linter), and the `check*` functions are a lower level of abstraction than `onRequestPaused`, despite being trivial. I do see a point for `failOrContinueRequest` to be there, though, but again, I prefer having shorter, composable and independent functions than everything being defined in a single place.

I'll see if I can move them without the linter complaining. `failOrContinueRequest` if nothing else.

Then what about `newResolver` and `parseTTL`? They're both also used only once, but I definitely don't want those inline anywhere...

common/network_manager.go
```diff
@@ -73,6 +74,12 @@ func NewNetworkManager(
 	ctx context.Context, session *Session, manager *FrameManager, parent *NetworkManager,
 ) (*NetworkManager, error) {
 	state := k6lib.GetState(ctx)
+
+	resolver, err := newResolver(state.Options.DNS)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
imiric: `newResolver` doesn't resolve anything, just creates a new `Resolver` instance. And an error is returned only if `parseTTL()` fails, so as long as that error is clear, this can remain returning `err`.

common/network_manager_utils.go
inancgumus: I understand your concerns, but the `common` is getting bigger and bigger, and its namespace is getting smaller with each identifier (funcs, etc.). If you don't want to put them inside the same function, at least, create a new type and put those functions as methods around that type. Or if you don't want to do that either, please put them in the same file (network_manager.go) or to a file like: `network.go` or something like that but `utils` 🤷 .

common/network_manager_utils.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// Returns a new Resolver.
+// Copied with minor changes from
+// https://github.com/grafana/k6/blob/fb70bc6f3d3f22a40e65f32deea3cea1b6d70a76/js/runner.go#L459
+func newResolver(conf k6types.DNSConfig) (k6netext.Resolver, error) {
+	ttl, err := parseTTL(conf.TTL.String)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
imiric: I thought this would be good, but on second look `parseTTL` already returns a helpful error:
https://github.com/grafana/xk6-browser/blob/0cc6e6b255cdcfffb285a6301824e9a247fe58df/common/network_manager_utils.go#L72

So `cannot parse TTL` is not much more helpful.

common/network_manager.go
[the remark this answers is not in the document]
inancgumus: It's better to follow the error chain on an error, that's why I suggested it. What about?

```suggestion
	resolver, err := newResolver(state.Options.DNS)
	if err != nil {
		return nil, fmt.Errof("newResolver(%+v): %w", state.Options.DNS, err)
	}
```

common/network_manager_utils.go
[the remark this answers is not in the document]
inancgumus: Yes, but `cannot parse` will allow us to follow the error chain.

common/network_manager.go
```diff
@@ -384,31 +392,52 @@ func (m *NetworkManager) onRequestPaused(event *fetch.EventRequestPaused) {
...
 	)
 
-	defer func() { m.failOrContinueRequest(event, failReason) }()
+	defer func() { m.failOrContinueRequest(event, failErr) }()
 
 	purl, err := url.Parse(event.Request.URL)
 	if err != nil {
 		m.logger.Errorf("NetworkManager:onRequestPaused",
-			"error parsing URL: %s", err.Error())
+			"error parsing URL %q: %s", event.Request.URL, err.Error())
```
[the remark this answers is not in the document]
imiric: Ah, for some reason my editor complained when I used it directly before, so I stuck with that, and I think I'm using it elsewhere as well. Will change, thanks.

common/network_manager.go
```diff
@@ -384,31 +392,52 @@ func (m *NetworkManager) onRequestPaused(event *fetch.EventRequestPaused) {
...
 		return
 	}
 
-	failReason = handleBlockedHosts(purl, state.Options.BlockedHostnames.Trie)
+	host, _, err := net.SplitHostPort(purl.Host)
+	if err != nil {
+		// no port, fallback to original
+		host = purl.Host
+	}
```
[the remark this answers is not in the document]
imiric: :thinking: I wasn't aware of `Hostname()`! I'll give it a try, thanks.

common/network_manager_test.go
imiric: Sure, but is there a practical benefit to this, or is it just personal preference?

common/network_manager_utils.go
imiric: I don't see a reason to use both parenthesis and quotes... The parenthesis make sense for the blocked pattern, since it's additional information, but not for the hostname, since the message is "hostname X...".

So I'll change it to `hostname %q is in a blocked pattern (%s)`.

Though I kind of think quotes are not needed even for hostname, since it usually won't include whitespace. `hostname "test.k6.io" ...` looks weird IMO.

So I'd rather leave it as is? :sweat_smile:

common/network_manager_utils.go
imiric: Similar to above, we don't need both parenthesis and quotes in this case. I would actually drop the parenthesis in the first case and change it to `IP %s is in a blacklisted range (%s)`, but note that I left the message the same as the one in `BlackListedIPError`, so maybe it's best if we leave this as is as well.

common/network_manager_utils.go
imiric: > I understand your concerns, but the `common` is getting bigger and bigger, and its namespace is getting smaller with each identifier (funcs, etc.).

`common` will get bigger regardless of where we define functions. It's something we'll have to live with until we start splitting it up.

And is namespace size a practical concern? Is there some limit I'm not aware of? I don't see why we should invent a new type just to avoid polluting the namespace. :man_shrugging: If anything, these could be defined on `NetworkManager` itself, but I chose to keep them independent since they really don't need to be methods.

> if you don't want to do that either, please put them in the same file (`network_manager.go`) or to a file like: network.go or something like that but `utils`.

But they're in `network_manager_utils.go`, i.e. utils for the `NetworkManager`... `network_manager.go` is already 500 lines, but sure, I'll just move them there instead.

You really can't even picture the word `utils` :laughing:

imiric: OK, this is partly resolved in 24b1fce (sorry, couldn't be bothered to split the changes :sweat_smile:).

`failOrContinueRequest` is now gone and is defined inline for the `defer`. But I left the `block*` functions standalone, and moved them to `network_manager.go`, along with `newResolver` and `parseTTL`, so `network_manager_utils.go` is gone too :tada:

common/network_manager_utils.go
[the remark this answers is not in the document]
imiric: I think the error is now too noisy/repetitive, but sure, I changed it in 24b1fce.

common/network_manager.go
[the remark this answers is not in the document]
imiric: This will be extra noisy with the `parseTTL` change, especially with `%+v` of `Options.DNS`, but sure, I changed it in 24b1fce.

common/network_manager.go
```diff
@@ -384,31 +392,52 @@ func (m *NetworkManager) onRequestPaused(event *fetch.EventRequestPaused) {
...
+	if failErr == nil {
+		// Do one last check of the resolved IP
+		ip, err = m.resolver.LookupIP(host)
+		if err != nil {
+			m.logger.Debugf("NetworkManager:onRequestPaused",
+				"error resolving %q: %s", host, err.Error())
+			return
+		}
+		failErr = checkBlockedIPs(ip, state.Options.BlacklistIPs)
+	}
```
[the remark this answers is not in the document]
imiric: Resolved in 24b1fce.

common/network_manager_test.go
inancgumus: 1. Idiomatic (check Go stdlib)
2. Simpler
3. More natural
4. Your code doesn't need to `append` as the size is there, you just want to fill its values
5. You don't need a function call which is designed for appending (and adds unnecessary assembly instructions)
6. ...

common/network_manager_utils.go
inancgumus: So:

```suggestion
		return fmt.Errorf("hostname %s is in a blocked pattern %q", host, match)
```

Then again, never trust what you're going to get in a string.

common/network_manager_utils.go
inancgumus: Fair enough, here you go:

```suggestion
			return fmt.Errorf("IP %s is in a blacklisted range %q", ip, ipnet)
```

common/network_manager_utils.go
inancgumus: By the way, I'd bring those funcs closer to the method that uses them instead of putting them at the end of the file.

common/network_manager_utils.go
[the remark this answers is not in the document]
inancgumus: Feel free to use any error message as long as we can follow the error chain when an error occurs.

common/network_manager.go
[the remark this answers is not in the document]
inancgumus: Same: https://github.com/grafana/xk6-browser/pull/205#discussion_r797333757

inancgumus:
Minor nitpick: I'd name the last commit message as `Address review comments for #205` instead of a vague `Address review comments' even though it properly includes `Resolved`s.

common/network_manager_utils.go
inancgumus: > these could be defined on NetworkManager itself, but I chose to keep them independent since they really don't need to be methods.

Yes, I agree, but if you had done it that way, it'd still be better than `utils` because they are super-related.

imiric: > I'd bring those funcs closer to the method that uses them instead of putting them at the end of the file.

I prefer keeping unexported functions near the bottom, and exported structs and methods towards the top. This is more legible for users of the module since they shouldn't care about unexported functions as much, so the reading flow is not interrupted, and they can/should use `go-to-definition` in their IDE to jump to any function regardless of its position in the file.

But sure, I moved them in b8eef0e.

common/network_manager_utils.go
imiric: > never trust what you're going to get in a string

This doesn't really apply here. There's no risk that quoting the string here would prevent :man_shrugging:

It's a style preference (which I thought we weren't going to bring up in PRs?), and TBH I prefer the previous parenthesis usage, but we've bikeshed about this enough, so it's changed in b8eef0e.

common/network_manager_utils.go
inancgumus: > I prefer keeping unexported functions near the bottom, and exported structs and methods towards the top.

This is the way, I agree, and I follow the same principle, but this time is different.

> the reading flow is not interrupted

This is exactly what worries me. Even though they are implementation details, I doubt anyone who reads the `NewNetworkManager` function won't want to read them.

common/network_manager_utils.go
imiric: Same comment as above, but changed in b8eef0e.

PR #207, opened 2022-01-28
Add support for k6's `hosts` option

### LINKED ISSUE #105: Make sure we adhere to the k6 options `DNS`, `Blacklist IPs`, `Block hostnames` and `Hosts`

imiric:
This translates [k6's `hosts` option](https://k6.io/docs/using-k6/options/#hosts) into [Chrome's `host-resolver-rules` option](https://datacadamia.com/web/browser/chrome#dns_resolver). There's no way that I could see to set this via CDP, so it needs to be done at launch. And of course, re-implemented for Firefox once we support it, though judging by [this issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1523367), there doesn't seem to be an equivalent option (yet?).

This is based on #206, so that should be merged first.

Closes #105

### DIFF
--- a/chromium/browser_type_test.go
+++ b/chromium/browser_type_test.go
@@ -21,20 +21,26 @@
 package chromium
 
 import (
+	"net"
 	"runtime"
 	"testing"
 
 	"github.com/grafana/xk6-browser/common"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	k6lib "go.k6.io/k6/lib"
 )
 
 func TestBrowserTypeFlags(t *testing.T) {
 	t.Parallel()
 
+	host, err := k6lib.NewHostAddress(net.ParseIP("127.0.0.1"), "8000")
+	require.NoError(t, err)
+
 	testCases := []struct {
 		flag                      string
 		changeOpts                *common.LaunchOptions
+		changeK6Opts              *k6lib.Options
 		expInitVal, expChangedVal interface{}
 		pre                       func(t *testing.T)
 		post                      func(t *testing.T, flags map[string]interface{})
@@ -83,6 +89,19 @@ func TestBrowserTypeFlags(t *testing.T) {
 			},
 		},
 		{
+			flag:       "host-resolver-rules",
+			expInitVal: nil,
+			changeOpts: &common.LaunchOptions{Args: []string{
+				`host-resolver-rules="MAP * www.example.com, EXCLUDE *.youtube.*"`}},
+			changeK6Opts: &k6lib.Options{
+				Hosts: map[string]*k6lib.HostAddress{
+					"test.k6.io":         host,
+					"httpbin.test.k6.io": host,
+				}},
+			expChangedVal: "MAP * www.example.com, EXCLUDE *.youtube.*," +
+				"MAP httpbin.test.k6.io 127.0.0.1:8000,MAP test.k6.io 127.0.0.1:8000",
+		},
+		{
 			flag:       "enable-use-zoom-for-dsf",
 			expInitVal: false,
 			pre: func(t *testing.T) {
@@ -115,7 +134,7 @@ func TestBrowserTypeFlags(t *testing.T) {
 
 			var (
 				bt    BrowserType
-				flags = bt.flags(&common.LaunchOptions{})
+				flags = bt.flags(&common.LaunchOptions{}, nil)
 			)
 
 			if tc.expInitVal != nil {
@@ -125,8 +144,8 @@ func TestBrowserTypeFlags(t *testing.T) {
 				require.NotContains(t, flags, tc.flag)
 			}
 
-			if tc.changeOpts != nil {
-				flags = bt.flags(tc.changeOpts)
+			if tc.changeOpts != nil || tc.changeK6Opts != nil {
+				flags = bt.flags(tc.changeOpts, tc.changeK6Opts)
 				if tc.expChangedVal != nil {
 					assert.Equal(t, tc.expChangedVal, flags[tc.flag])
 				}
--- a/chromium/browser_type_test.go
+++ b/chromium/browser_type_test.go
@@ -102,6 +102,13 @@ func TestBrowserTypeFlags(t *testing.T) {
 				"MAP httpbin.test.k6.io 127.0.0.1:8000,MAP test.k6.io 127.0.0.1:8000",
 		},
 		{
+			flag:          "host-resolver-rules",
+			expInitVal:    nil,
+			changeOpts:    &common.LaunchOptions{},
+			changeK6Opts:  &k6lib.Options{},
+			expChangedVal: nil,
+		},
+		{
 			flag:       "enable-use-zoom-for-dsf",
 			expInitVal: false,
 			pre: func(t *testing.T) {
@@ -148,6 +155,8 @@ func TestBrowserTypeFlags(t *testing.T) {
 				flags = bt.flags(tc.changeOpts, tc.changeK6Opts)
 				if tc.expChangedVal != nil {
 					assert.Equal(t, tc.expChangedVal, flags[tc.flag])
+				} else {
+					assert.NotContains(t, flags, tc.flag)
 				}
 			}
 
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"regexp"
 	"runtime"
+	"sort"
 	"strings"
 
 	"github.com/dop251/goja"
@@ -76,9 +77,11 @@ func (b *BrowserType) ExecutablePath() string {
 
 // Launch creates a new client to remote control a Chrome browser.
 func (b *BrowserType) Launch(opts goja.Value) api.Browser {
-	rt := k6common.GetRuntime(b.Ctx)
-
-	launchOpts := common.NewLaunchOptions()
+	var (
+		rt         = k6common.GetRuntime(b.Ctx)
+		state      = k6lib.GetState(b.Ctx)
+		launchOpts = common.NewLaunchOptions()
+	)
 	launchOpts.Parse(b.Ctx, opts)
 
 	b.Ctx = common.WithLaunchOptions(b.Ctx, launchOpts)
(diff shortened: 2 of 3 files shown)


chromium/browser_type_test.go
```diff
@@ -21,20 +21,26 @@
...
 func TestBrowserTypeFlags(t *testing.T) {
 	t.Parallel()
 
+	host, err := k6lib.NewHostAddress(net.ParseIP("127.0.0.1"), "8000")
+	require.NoError(t, err)
+
 	testCases := []struct {
 		flag                      string
 		changeOpts                *common.LaunchOptions
+		changeK6Opts              *k6lib.Options
```
inancgumus, line 43: Since we are aiming for extensible test cases here, each test case should stand on its own without requiring any other initialization steps out of them (which breaks the mental mapping of the test code).

Here's a refactoring that I worked on a little bit:

```golang
func TestBrowserTypeFlags(t *testing.T) {
	t.Parallel()

	type testCase struct {
		flag                      string
		lopts                     *common.LaunchOptions
		k6opts                    *k6lib.Options
		pre                       func(t *testing.T)
		change                    func(t *testing.T, tc *testCase)
		post                      func(t *testing.T, flags map[string]interface{})
		expInitVal, expChangedVal interface{}
	}

	testCases := []*testCase{
		{
			flag:       "auto-open-devtools-for-tabs",
			expInitVal: false,
			change: func(t *testing.T, tc *testCase) {
				tc.lopts = &common.LaunchOptions{Devtools: true}
			},
			expChangedVal: true,
		},
		{
			flag:       "browser-arg",
			expInitVal: nil,
			change: func(t *testing.T, tc *testCase) {
				tc.lopts = &common.LaunchOptions{Args: []string{"browser-arg=value"}}
			},
			expChangedVal: "value",
		},
		{
			flag:       "browser-arg-flag",
			expInitVal: nil,
			change: func(t *testing.T, tc *testCase) {
				tc.lopts = &common.LaunchOptions{Args: []string{"browser-arg-flag"}}
			},
			expChangedVal: "",
		},
		{
			flag:       "browser-arg-trim-double-quote",
			expInitVal: nil,
			change: func(t *testing.T, tc *testCase) {
				tc.lopts = &common.LaunchOptions{Args: []string{
					`   browser-arg-trim-double-quote =  "value  "  `}}
			},
			expChangedVal: "value  ",
		},
		{
			flag:       "browser-arg-trim-single-quote",
			expInitVal: nil,
			change: func(t *testing.T, tc *testCase) {
				tc.lopts = &common.LaunchOptions{Args: []string{
					`   browser-arg-trim-single-quote=' value '`}}
			},
			expChangedVal: " value ",
		},
		{
			flag:       "browser-args",
			expInitVal: nil,
			change: func(t *testing.T, tc *testCase) {
				tc.lopts = &common.LaunchOptions{Args: []string{
					"browser-arg1='value1", "browser-arg2=''value2''", "browser-flag"}}
			},
			post: func(t *testing.T, flags map[string]interface{}) {
				assert.Equal(t, "'value1", flags["browser-arg1"])
				assert.Equal(t, "'value2'", flags["browser-arg2"])
				assert.Equal(t, "", flags["browser-flag"])
			},
		},
		{
			flag:       "host-resolver-rules",
			expInitVal: nil,
			change: func(t *testing.T, tc *testCase) {
				tc.lopts = &common.LaunchOptions{Args: []string{
					`host-resolver-rules="MAP * www.example.com, EXCLUDE *.youtube.*"`}}

				host, err := k6lib.NewHostAddress(net.ParseIP("127.0.0.1"), "8000")
				require.NoError(t, err)

				tc.k6opts = &k6lib.Options{
					Hosts: map[string]*k6lib.HostAddress{
						"test.k6.io":         host,
						"httpbin.test.k6.io": host,
					}}
			},
			expChangedVal: "MAP * www.example.com, EXCLUDE *.youtube.*," +
				"MAP httpbin.test.k6.io 127.0.0.1:8000,MAP test.k6.io 127.0.0.1:8000",
		},
		{
			flag:       "enable-use-zoom-for-dsf",
			expInitVal: false,
			pre: func(t *testing.T) {
				if runtime.GOOS != "darwin" {
					t.Skip()
				}
			},
		},
		{
			flag:       "headless",
			expInitVal: false,
			change: func(t *testing.T, tc *testCase) {
				tc.lopts = &common.LaunchOptions{Headless: true}
			},
			expChangedVal: true,
			post: func(t *testing.T, flags map[string]interface{}) {
				extraFlags := []string{"hide-scrollbars", "mute-audio", "blink-settings"}
				for _, f := range extraFlags {
					assert.Contains(t, flags, f)
				}
			},
		},
	}

	for _, tc := range testCases {
		tc := tc
		t.Run(tc.flag, func(t *testing.T) {
			t.Parallel()
			if tc.pre != nil {
				tc.pre(t)
			}

			var (
				bt    BrowserType
				lopts = &common.LaunchOptions{}
			)

			flags := bt.flags(lopts, nil)

			if tc.expInitVal != nil {
				require.Contains(t, flags, tc.flag)
				assert.Equal(t, tc.expInitVal, flags[tc.flag])
			} else {
				require.NotContains(t, flags, tc.flag)
			}

			if tc.change != nil {
				tc.change(t, tc)
				flags = bt.flags(tc.lopts, tc.k6opts)
				if tc.expChangedVal != nil {
					assert.Equal(t, tc.expChangedVal, flags[tc.flag])
				}
			}

			if tc.post != nil {
				tc.post(t, flags)
			}
		})
	}
}
```

inancgumus (approved):
LGTM. I have some suggestions.

chromium/browser_type_test.go
imiric: I don't see that as an improvement, sorry. Yes, we get rid of the `host` initialization in the outer scope and simplify the change check, but the additional `change` function complicates each test case, since we need to be aware that `tc.lopts` and `tc.k6opts` need to be changed there. We might as well call `flags()` from there and embed the `expChangedVal` check in that function too... :shrug: 

I'd like to keep the amount of functions that must be defined to a minimum, and `pre` and `post` are only there if they optionally need to be used.

imiric:
@inancgumus Please take another look and let me know. I pushed a minor fix in c4ab6ce.

chromium/browser_type_test.go
inancgumus: OK, let's keep it, but it's my responsibility to explain my reasonings :)

* Each test should be self-contained (for parallelism and understandability purposes)
* Doing host initialization only for one test pollutes the test code and makes it harder to understand what's going on (the tests are already clever enough)
* The main reason for this PR's change was about extensibility, but it seems like it cannot do so without adding more hacks. This is why I wouldn't say I like clever tests.

> the additional change function complicates each test case

On the contrary, it makes them self-contained and more understandable for each test case.

> since we need to be aware that tc.lopts and tc.k6opts need to be changed there

Each test case is already doing that and hence it becomes clear.

> We might as well call flags() from there and embed the expChangedVal check in that function too...

WDYM by this?

> I'd like to keep the amount of functions that must be defined to a minimum

I don't think the functions are the problem here. The problem is that this table test is trying to cover everything in a single test function.

inancgumus (approved):
LGTM but this one: https://github.com/grafana/xk6-browser/pull/207#r796385779

Btw, is it better to keep and separate k6 stuff out of the extension somehow? This will make code and tests simpler. Maybe I should make a PR for discussion.

chromium/browser_type_test.go
imiric: > Each test should be self-contained (for parallelism and understandability purposes)

Sure, they're mostly self-contained and already parallel. Understandability is subjective, but I'd argue that a few conditionals are hardly difficult to understand.


> Doing host initialization only for one test pollutes the test code and makes it harder to understand what's going on (the tests are already clever enough)

How does initializing a single variable in the outer scope make the test harder to understand? It's not ideal, sure, but what you're proposing is not an improvement, and is ironically even more clever...

> The main reason for this PR's change was about extensibility, but it seems like it cannot do so without adding more hacks. This is why I wouldn't say I like clever tests.

Agreed, I also prefer simpler tests. But in this case a bit of _simple_ logic gives us a generic test that we can reuse for many flag cases. How is that a "hack"?

I'd prefer to not write or read the same boilerplate for what is essentially the same test logic for all of these.



> > the additional change function complicates each test case

> On the contrary, it makes them self-contained and more understandable for each test case.

I really don't see how adding another function that modifies `testCase` values, where the programmer is meant to know that these values will then be passed to `flags()`, is simpler and more understandable than setting those values on `testCase` to begin with, without a function.

Ideally table tests should only have basic values, not functions. `pre` and `post` go against this, but they're a) not required in most cases, and b) don't modify test state, unlike your `change()` proposal.

All this for avoiding the `host` initialization and pursuing total "self-containment" is not worth it.


> > We might as well call flags() from there and embed the expChangedVal check in that function too...

> WDYM by this?

I mean that we can just shove all test logic in a `testCase` function and call that when we iterate over test cases. That would simplify the actual test logic, but at that point you're just moving things around for the sake of "self-containment".


> > I'd like to keep the amount of functions that must be defined to a minimum

> I don't think the functions are the problem here. The problem is that this table test is trying to cover everything in a single test function. And I fear it'll become more complicated in time.

Functions definitely don't help, and I would get rid of `pre` and `post` if I could. But having a slightly complicated test makes up for being able to test `flags()` in a generic way.

I can't say that it won't get more complicated, that's a possibility. But I don't see what you're proposing as an improvement.


If you feel strongly against this, let's just drop the table tests then. I'll rewrite all of them as subtests with slightly different logic. WDYT?

imiric:
> Btw, is it better to keep and separate k6 stuff out of the extension somehow? This will make code and tests simpler. Maybe I should make a PR for discussion.

Not sure what you mean. Sure, create a PR.

chromium/browser_type_test.go
inancgumus: No, don't bother. Glad to see your opinions as well. Thanks.

PR #212, opened 2022-02-01
Refactor/elementhandle evaluate calls after revert

inancgumus:
This PR is a new one for #196. We had reverted #196 because it included the `QueryAll` (#195) fix by mistake.


imiric (approved):
I didn't take an in-depth look now, but it LGTM in #196.

inancgumus:
> I didn't take an in-depth look now, but it LGTM in #196.

It should be the same except for the fix.

PR #213, opened 2022-02-02
Make session an anonymous interface within NetworkManager

imiric:
Resolves https://github.com/grafana/xk6-browser/pull/204#discussion_r796395070

I prefer the anonymous interface approach rather than coming up with some common one we're not sure we'll have a need for yet.


inancgumus:
Much better ❤️

inancgumus (approved):
Much better ❤️  We'll see whether we can make use of `ExecuteEmitter` later on.

PR #216, opened 2022-02-03
Add golangci-lint configuration

imiric:
This adds a bunch of new linters based on the discussion below, and currently (on 50d0160) passes with `--new-from-rev` enabled. Without it there are 1298 issues... :sob:

See #58

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -69,7 +69,6 @@ linters:
     - unparam
     - unused
     - varcheck
-    - varnamelen
     - wastedassign
     - whitespace
     - wrapcheck
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -88,3 +88,7 @@ linters-settings:
     block-size: 10
   lll:
     maxlength: 120
+  gosec:
+    config:
+      G301: '0755'
+      G306: '0644'
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,90 @@
+run:
+  timeout: 5m
+
+issues:
+  # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
+  max-issues-per-linter: 0
+
+  # Maximum count of issues with the same text. Set to 0 to disable. Default is 3.
+  max-same-issues: 0
+
+  # Disable default exclude rules, see `golangci-lint run --help`.
+  # When true it disables revive's comment warnings, and we want them enabled.
+  exclude-use-default: false
+
+linters:
+  disable-all: true
+  enable:
+    - cyclop
+    - deadcode
+    - dogsled
+    - dupl
+    - durationcheck
+    - errcheck
+    - errchkjson
+    - errname
+    - errorlint
+    - exhaustive
+    - exportloopref
+    - forbidigo
+    - forcetypeassert
+    - funlen
+    - gci
+    - gochecknoglobals
+    - gocognit
+    - goconst
+    - gocritic
+    - godot
+    - gofmt
+    - gofumpt
+    - goimports
+    - gosec
+    - gosimple
+    - govet
+    - ifshort
+    - importas
+    - ineffassign
+    - lll
+    - makezero
+    - misspell
+    - nakedret
+    - nestif
+    - nilerr
+    - nilnil
+    - nlreturn
+    - nolintlint
+    - predeclared
+    - prealloc
+    - revive
+    - staticcheck
+    - structcheck
+    - tagliatelle
+    - thelper
+    - tparallel
+    - typecheck
+    - unconvert
+    - unparam
+    - unused
+    - varcheck
+    - varnamelen
+    - wastedassign
+    - whitespace
+    - wrapcheck
+
+linters-settings:
+  gci:
+    local: github.com/grafana/xk6-browser
+  dupl:
+    threshold: 400
+  gofmt:
+    simplify: true
+  goimports:
+    local-prefixes: github.com/grafana/xk6-browser
+  importas:
+    alias:
+      - pkg: go.k6.io/k6/.*/(\w+)
+        alias: k6$1
+  nlreturn:
+    block-size: 10
+  lll:
+    maxlength: 120
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,3 +1,6 @@
+# v1.44.0
+# Please don't remove the first line. It's used in CI to determine the
+# golangci-lint version. See the lint step in .github/workflows/all.yaml
 run:
   timeout: 5m
 
--- a/common/connection.go
+++ b/common/connection.go
@@ -40,7 +40,7 @@ import (
 
 const wsWriteBufferSize = 1 << 20
 
-// Ensure Connection implements the EventEmitter and Executor interfaces
+// Ensure Connection implements the EventEmitter and Executor interfaces.
 var _ EventEmitter = &Connection{}
 var _ cdp.Executor = &Connection{}
 
@@ -88,7 +88,7 @@ func (f ActionFunc) Do(ctx context.Context) error {
 │Registers with session as a├─────────────■                    │                         │                    │
 │handler for a specific CDP │             │   Event Listener   │      *  *  *  *  *      │   Event Listener   │
 │       Domain event.       │             │                    │                         │                    │
-└───────────────────────────┘             └────────────────────┘                         └────────────────────┘
+└───────────────────────────┘             └────────────────────┘                         └────────────────────┘.
 */
 type Connection struct {
 	BaseEventEmitter
@@ -113,7 +113,7 @@ type Connection struct {
 	encoder jwriter.Writer
 }
 
-// NewConnection creates a new browser
+// NewConnection creates a new browser.
 func NewConnection(ctx context.Context, wsURL string, logger *Logger) (*Connection, error) {
 	var header http.Header
 	var tlsConfig *tls.Config
@@ -467,7 +467,7 @@ func (c *Connection) Close(args ...goja.Value) {
 	_ = c.closeConnection(code)
 }
 
-// Execute implements cdproto.Executor and performs a synchronous send and receive
+// Execute implements cdproto.Executor and performs a synchronous send and receive.
 func (c *Connection) Execute(ctx context.Context, method string, params easyjson.Marshaler, res easyjson.Unmarshaler) error {
 	c.logger.Debugf("connection:Execute", "wsURL:%q method:%q", c.wsURL, method)
 	id := atomic.AddInt64(&c.msgID, 1)
--- a/common/element_handle.go
+++ b/common/element_handle.go
@@ -49,8 +49,9 @@ import (
 	"github.com/chromedp/cdproto/dom"
 	cdppage "github.com/chromedp/cdproto/page"
 	"github.com/dop251/goja"
-	"github.com/grafana/xk6-browser/api"
 	k6common "go.k6.io/k6/js/common"
+
+	"github.com/grafana/xk6-browser/api"
 )
 
 // Ensure ElementHandle implements the api.ElementHandle and api.JSHandle interfaces.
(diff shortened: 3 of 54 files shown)


.golangci.yml
```diff
@@ -0,0 +1,39 @@
...
+    - gosimple
+    - govet
+    - ineffassign
+    - misspell
+    - nakedret
+    - revive
+    - structcheck
+    - unused
+    - varcheck
+    - staticcheck
```
imiric, line 31: Should we add more linters?

The [k6 config](https://github.com/grafana/k6/blob/0dda328d9a7f563607dae4fe89ff8c5138f8dc09/.golangci.yml) enables all and disables some, which I think is an error-prone approach, as new linters are constantly added, which often introduces new failures. So this is a variation of the [helm config](https://github.com/helm/helm/blob/35f6196167f6b90913f3a94237825856b858a215/.golangci.yml) which is a safer approach.

inancgumus: Hey, I've summarized the other linters we might consider and the current number of warnings from them:

| linter | # of warnings |
| --- | --- |
| gocritic | 45 |
| errcheck | 3 |
| typecheck | 0 |
| cyclop | 37 |
| durationcheck | 1 |
| errchkjson | 6 |
| errorlint | 22 |
| exhaustive | 20 |
| exportloopref | 0 |
| forbidigo | 1 |
| funlen | 25 |
| gci | 33 |
| gochecknoglobals | 21 |
| gocognit | 11 |
| goconst | 33 |
| gocyclo | 1 |
| godot | 245 |
| gofumpt | 11 |
| gosec | 6 |
| ifshort | 6 |
| lll | 104 |
| makezero | 1 |
| misspell | 3 |
| nakedret | 0 |
| nestif | 11 |
| nilerr | 0 |
| nilnil | 4 |
| nlreturn | 435 |
| prealloc | 2 |
| thelper | 11 |
| tparallel | 6 |
| unconvert | 4 |
| unparam | 15 |
| wastedassign | 1 |
| whitespace | 4 |
| wrapcheck | 48 |
| forcetypeassert | 76 I'm not sure about this one. |

Should we enable `containedctx` or `ireturn` as well? Just kidding :)

inancgumus:
Hey, I've summarized the other linters we might consider and the current number of warnings from them:

| linter | # of warnings |
| --- | --- |
| gocritic | 45 |
| errcheck | 3 |
| typecheck | 0 |
| cyclop | 37 |
| durationcheck | 1 |
| errchkjson | 6 |
| errorlint | 22 |
| exhaustive | 20 |
| exportloopref | 0 |
| forbidigo | 1 |
| funlen | 25 |
| gci | 33 |
| gochecknoglobals | 21 |
| gocognit | 11 |
| goconst | 33 |
| gocyclo | 1 |
| godot | 245 |
| gofumpt | 11 |
| gosec | 6 |
| ifshort | 6 |
| lll | 104 |
| makezero | 1 |
| misspell | 3 |
| nakedret | 0 |
| nestif | 11 |
| nilerr | 0 |
| nilnil | 4 |
| nlreturn | 435 |
| prealloc | 2 |
| thelper | 11 |
| tparallel | 6 |
| unconvert | 4 |
| unparam | 15 |
| wastedassign | 1 |
| whitespace | 4 |
| wrapcheck | 48 |
| forcetypeassert | 76 I'm not sure about this one. |

Should we enable `containedctx` or `ireturn` as well? Just kidding :)

.golangci.yml
inancgumus: See: https://github.com/grafana/xk6-browser/pull/216#issuecomment-1029766734

imiric:
Seems like a good list, but I'll have to take a closer look at each one. There might be redundant ones there, like `cyclop` which seems to cover all use cases of `gocyclo`.

But how do you want to approach this? Re-enable `--new-from-rev` in CI and run without it locally, while we gradually work on resolving the issues? Maybe some could be automatically resolved, or scripted.

Either way, this is not high priority, so we don't have to start on it now.

inancgumus:
> There might be redundant ones there

I believe others check for different stuff. You may have found the only overlapping one. But sure, take a closer look, that won't hurt, and I might be wrong 🤷 

> But how do you want to approach this?

Agreed, fixing these linter issues isn't a high priority. However, sometimes subtle bugs may have been caused because of these issues and can take more time. Re-enabling `--new-from-rev` seems like a better option, for now, so we can stop introducing more warnings.

I believe we better work on them one day/week for the remaining ones. So we can finish them in a month or so.

imiric:
@inancgumus I enabled most of the linters you listed, and added a few more. I'm concerned we're running way too many, though...

Some comments:
- I relaxed blocks for `nlreturn`, so it only triggers on >10 LOC.
- `staticcheck` seems to be a `go vet` replacement (from `golangci-lint help linters`: `Staticcheck is a go vet on steroids`), so can we disable `govet`?
- Added `importas`, `nolintlint`, `predeclared`, `tagliatelle`, and `varnamelen`. I'm not too sure if we should enable these last two ones, especially `varnamelen`, as it can get annoying. See https://github.com/grafana/k6/pull/2323.

I'll fix the existing errors after `--new-from-rev` in this PR and we can hopefully merge it then.

common/connection.go
```diff
@@ -88,7 +88,7 @@ func (f ActionFunc) Do(ctx context.Context) error {
 │Registers with session as a├─────────────■                    │                         │                    │
 │handler for a specific CDP │             │   Event Listener   │      *  *  *  *  *      │   Event Listener   │
 │       Domain event.       │             │                    │                         │                    │
-└───────────────────────────┘             └────────────────────┘                         └────────────────────┘
+└───────────────────────────┘             └────────────────────┘                         └────────────────────┘.
```
imiric, line 91: :man_facepalming:

common/frame.go
```diff
@@ -1538,6 +1538,7 @@ type frameExecutionContext interface {
 	ID() runtime.ExecutionContextID
 }
 
+//nolint:unparam
```
inancgumus, line 1541: I would rather keep this warning on and refactor these functions, also `framePointerActionFn`.

inancgumus:
> @inancgumus I enabled most of the linters you listed, and added a few more. I'm concerned we're running way too many, though...

I prefer removing or adjusting them until we hit an issue. I diff'ed your changes, you actually added all the linters I suggested but gocyclo as we discussed 👍 

> I relaxed blocks for nlreturn, so it only triggers on >10 LOC.

👍 

> staticcheck seems to be a go vet replacement (from golangci-lint help linters: Staticcheck is a go vet on steroids), so can we disable govet?

We better keep it, see https://staticcheck.io/docs/: _"Staticcheck aims to be trivial to adopt. It behaves just like the official go tool and requires no learning to get started with. Just run staticcheck ./... on your code **in addition to go vet ./....**"_

> Added importas, nolintlint, predeclared, tagliatelle, and varnamelen. I'm not too sure if we should enable these last two ones, especially varnamelen, as it can get annoying. See https://github.com/grafana/k6/pull/2323.

These ones are ok but I'm strongly against varnamelen. It's too noisy and produced non-idiomatic warnings (about 1k).

> I'll fix the existing errors after --new-from-rev in this PR and we can hopefully merge it then.

Are you going to fix every error or do it as a team in a weekly fashion as we discussed?

.golangci.yml
```diff
@@ -1,3 +1,6 @@
+# v1.44.0
+# Please don't remove the first line. It's used in CI to determine the
+# golangci-lint version.
```
inancgumus, line 3:
```suggestion
# Please don't remove the first line. It's used in CI to determine the
# golangci-lint version. See: The lint step in .github/workflows/all.yaml
```

It's good to have this kind of verbose explanation for newcomers to our team.

common/element_handle.go
```diff
@@ -106,6 +106,7 @@ func getElementHandleActionFn(
 	}
 }
 
+//nolint:funlen,gocognit,cyclop,unparam
```
inancgumus, line 109: We better refactor this one in TODO [here](https://github.com/grafana/xk6-browser/issues/58#issue-1044893666) instead of silencing it.

common/element_handle.go
```diff
@@ -64,7 +64,7 @@ type (
 )
 
 func getElementHandleActionFn(
-	h *ElementHandle, states []string, fn elementHandleActionFn, force, noWaitAfter bool, timeout time.Duration,
+	elh *ElementHandle, states []string, ehaFn elementHandleActionFn, force, noWaitAfter bool, timeout time.Duration,
```
inancgumus, line 67: I'm against varnamelen linter and on the side of reverting these changes. There is nothing wrong accepting `h` or `fn`. It is idiomatic to have these as short parameters. Their types make it clear what's going on.

inancgumus (changes requested):
LGTM, nice work, and as usual, I have some suggestions.

common/frame.go
imiric: Sure, but I didn't want to get into a large refactoring now. So we need to keep this `nolint`, and the `TODO` could be "Get rid of all `nolint`s".

common/element_handle.go
imiric: I'm not sure... I see the benefit of `varnamelen`, even in this case. The scope of this function is too large to use vars like `h` and `fn`. This is fine for ~5 LOC functions or `for` loops, but not here.

We definitely abuse this often, as you can see from the 339 errors (not sure how you got 1k?). But I think most of these can be quickly changed using `sed` or [`rf`](https://pkg.go.dev/rsc.io/rf). And we can always [tweak the configuration](https://github.com/blizzy78/varnamelen#golangci-lint-integration) to e.g. relax the scope requirements.

So I'm leaning towards keeping it, as it's valuable with something we're likely to not point out in PRs.

inancgumus: Sorry, I see no benefit here, and I'm firmly against it. This is not something we want to enforce on senior devs. Better use linters for practical purposes instead of forcing non-idiomatic style changes.

imiric:
> Are you going to fix every error or do it as a team in a weekly fashion as we [discussed](https://github.com/grafana/xk6-browser/pull/216#issuecomment-1029797172)?

That's definitely still the plan :sweat_smile: I just had to fix/silence these initial ones as they were getting triggered with this configuration even with `--new-from-rev`.

common/element_handle.go
inancgumus: I just ran this useless linter on Go stdlib and it found 14K warnings. Come on.

inancgumus: I just ran the linter on Go stdlib and it found 14K warnings. Come on.

imiric: It's not an issue of seniority :smile:, but properly naming vars according to the scope they're used in. It's about legibility, avoiding shadowing, `grep`ability, etc.

I disagree, but if you feel strongly about it, sure, I'll remove it.

inancgumus: Please remove it 🙏 

Btw, Rob won't agree. I ran the linter on his ivy project: 300 warnings. It seems he can't write idiomatic code according to this linter...

imiric: Ah, I certainly wouldn't want to be maintaining his ivy project then :stuck_out_tongue:

FWIW, I prefer not blindly sticking to whatever is "idiomatic", but to whatever improves the DX while working on the code. To me it's a no brainer that longer var names _for larger scopes_ are the better approach.

common/element_handle.go
inancgumus: Can you add a TODO to #58? Also for the other one.

common/element_handle.go
inancgumus: > longer var names for larger scopes are the better approach

I agree and but it depends on a blend of a dev's experience + code at hand. An automatic tool cannot decide that. When I recommend something, it's because I know it from my experience, not because I follow people blindly.

common/element_handle.go
imiric: Just added a generic "fill all `nolint`s".

inancgumus (approved):
LGTM 👏

Issue #218, opened 2022-02-04
TestElementHandleClickWithDetachedNode is flaky


inancgumus:
This seems to become flaky after the changes introduced **after** v0.1.3.

On v0.1.3:

```
➜  k6b git:(v0.1.3) go test ./tests -race -timeout 360s -race -count=50 -run=TestElementHandleClickWithDetachedNode -v
=== RUN   TestElementHandleClickWithDetachedNode
--- PASS: TestElementHandleClickWithDetachedNode (0.49s)
=== RUN   TestElementHandleClickWithDetachedNode
--- PASS: TestElementHandleClickWithDetachedNode (0.49s)
...
```

[Now](https://github.com/grafana/xk6-browser/tree/0f2ba32f0ab0131d22cf9a7df15dab214da7e7a0)—happens all the time:

```
➜  k6b git:(main) go test ./tests -race -timeout 360s -race -count=10 -run=TestElementHandleClickWithDetachedNode -v
=== RUN   TestElementHandleClickWithDetachedNode
--- PASS: TestElementHandleClickWithDetachedNode (0.48s)
=== RUN   TestElementHandleClickWithDetachedNode
    element_handle_click_test.go:143:
        	Error Trace:	element_handle_click_test.go:143
        	Error:      	"" does not contain "element is not attached to the DOM"
        	Test:       	TestElementHandleClickWithDetachedNode
        	Messages:   	expected click to result in correct error to be thrown
--- FAIL: TestElementHandleClickWithDetachedNode (0.47s)
=== RUN   TestElementHandleClickWithDetachedNode
    element_handle_click_test.go:143:
        	Error Trace:	element_handle_click_test.go:143
        	Error:      	"" does not contain "element is not attached to the DOM"
        	Test:       	TestElementHandleClickWithDetachedNode
        	Messages:   	expected click to result in correct error to be thrown
--- FAIL: TestElementHandleClickWithDetachedNode (0.48s)
=== RUN   TestElementHandleClickWithDetachedNode
    element_handle_click_test.go:143:
        	Error Trace:	element_handle_click_test.go:143
        	Error:      	"" does not contain "element is not attached to the DOM"
        	Test:       	TestElementHandleClickWithDetachedNode
        	Messages:   	expected click to result in correct error to be thrown
--- FAIL: TestElementHandleClickWithDetachedNode (0.48s)
```

Maybe somewhat related to #164, but I doubt it.

inancgumus:
[...]

PR #219, opened 2022-02-04
Remove k6ThrowOnce and friends

inancgumus:
When you run tests only once, there is no problem, however if you run
them multiple times using the count flag, k6ThrowOnce kicks in and
cause k6ThrowUnsafe not to run. This happens because the Go test runner
compiles the test binary only once and uses it to run the test in the
same process.

I don't know why but somehow we or Goja have fixed the issue in k6Throw
and now this commit entirely removes the k6Throw synchronization.

I tested it several times and I didn't observe any issues.

Fixes: #218


imiric (approved):
This seems safe to remove, but how did you determine `k6ThrowUnsafe` was the root cause for `TestElementHandleClickWithDetachedNode` being flaky?

Issue #215, opened 2022-02-03
Test QueryAll


inancgumus:
This PR is related to #195, and writing a unit test for `QueryAll` has been taking too long, see [this](https://github.com/grafana/xk6-browser/pull/195#issuecomment-1018390899). So I'm planning to do a couple of refactorings in multiple PRs before writing a unit test for it.

**TODO:**

- [x] Integration test: #195
- [x] #212
- [x] #221 
- [x] #224
- [x] #225
- [x] #229
- [ ] Make `ElementHandle.queryAll` a func that takes a context.

~- [ ] Unit test QueryAll (js)~ See: #233

PR #221, opened 2022-02-07
Move injected script to ExecutionContext

inancgumus:
This change moves the injected script only where it is being used. This is part of making `QueryAll` unit testable. See: #215.


common/execution_context.go
```diff
@@ -273,33 +274,48 @@ func (e *ExecutionContext) eval(
...
+	handle, err := e.eval(
+		apiCtx,
+		evalOptions{forceCallable: false, returnByValue: false},
+		rt.ToValue(expressionWithSourceURL),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("cannot get injected script (%q): %w", suffix, err)
+	}
+	if handle == nil {
+		return nil, fmt.Errorf("cannot get handle for inject script (%q): %w", suffix, err)
```
imiric, line 311: `err` will be nil here, so:
```suggestion
		return nil, fmt.Errorf("cannot get handle for injected script (%q)", suffix)
```

Though I'm wondering why we include the `suffix`. Isn't it always `//# sourceURL=__xk6_browser_evaluation_script__`?

imiric (changes requested):
LGTM, just a minor message suggestion.

Though I'm having slight déjà vu with this PR :smile: Didn't you already do this before? I can't find it though :confused:

common/execution_context.go
inancgumus: Yeah, it's not helpful to include it there. It's constant, after all.

inancgumus: Il est normal que vous ayez déjà été vu car cela fait partie d'un grand PR. I mean, don't worry, that's expected because this PR is part of a larger PR :)

inancgumus: Should be resolved in [6717f05](https://github.com/grafana/xk6-browser/pull/221/commits/6717f0592fc2b63175273aee922a13d8d450e06b).

imiric (approved):
:+1:

You can squash+merge this.

PR #225, opened 2022-02-08
Refactor QueryAll JS and Go funcs

inancgumus:
This PR is another step to make `QueryAll` unit testable (#215). This one does something weird and moves the javascript function in the `QueryAll` go method into a new home: `js/query_all.js`. Then it embeds the javascript function in the `js/selectors.go` so that `QueryAll` (go) can get this function's source code.

The reason behind this is separating Go and JS code. So that this change will allow us to test Go and JS code separately. It also has the benefit of code editor support. For example, you can go to the injected script's `querySelectorAll` method from `QueryAll` (js) if your editor supports it.

### DIFF
--- a/common/element_handle.go
+++ b/common/element_handle.go
@@ -52,6 +52,7 @@ import (
 	k6common "go.k6.io/k6/js/common"
 
 	"github.com/grafana/xk6-browser/api"
+	"github.com/grafana/xk6-browser/common/js"
 )
 
 // Ensure ElementHandle implements the api.ElementHandle and api.JSHandle interfaces.
@@ -1202,23 +1203,19 @@ func (h *ElementHandle) Query(selector string) api.ElementHandle {
 	return nil
 }
 
-// QueryAll queries element subtree for matching elements. If no element matches the selector,
-// the return value resolves to "null".
+// QueryAll queries element subtree for matching elements.
+// If no element matches the selector, the return value resolves to "null".
 func (h *ElementHandle) QueryAll(selector string) []api.ElementHandle {
 	parsedSelector, err := NewSelector(selector)
 	if err != nil {
 		k6Throw(h.ctx, "cannot parse selector %q in element query all: %v", selector, err)
 	}
-	fn := `
-		(node, injected, selector) => {
-			return injected.querySelectorAll(selector, node || document, false);
-		}
-	`
-	opts := evalOptions{
-		forceCallable: true,
-		returnByValue: false,
-	}
-	result, err := h.evalWithScript(h.ctx, opts, fn, parsedSelector)
+	result, err := h.evalWithScript(
+		h.ctx,
+		evalOptions{forceCallable: true, returnByValue: false},
+		js.QueryAll,
+		parsedSelector,
+	)
 	if err != nil {
 		k6Throw(h.ctx, "cannot evaluate selector %q: %v", selector, err)
 	}
@@ -1226,20 +1223,25 @@ func (h *ElementHandle) QueryAll(selector string) []api.ElementHandle {
 		return nil
 	}
 
-	arrayHandle := result.(api.JSHandle)
-	defer arrayHandle.Dispose()
-	properties := arrayHandle.GetProperties()
-	elements := make([]api.ElementHandle, 0, len(properties))
-	for _, property := range properties {
-		elementHandle := property.AsElement()
-		if elementHandle != nil {
-			elements = append(elements, elementHandle)
+	handles, ok := result.(api.JSHandle)
+	if !ok {
+		k6Throw(h.ctx, "cannot get selector (%q) handle: %w", selector, ErrJSHandleInvalid)
+	}
+	defer handles.Dispose()
+	var (
+		props = handles.GetProperties()
+		els   = make([]api.ElementHandle, 0, len(props))
+	)
+	for _, prop := range props {
+		if el := prop.AsElement(); el != nil {
+			els = append(els, el)
 		} else {
-			property.Dispose()
+			prop.Dispose()
 		}
 	}
 	applySlowMo(h.ctx)
-	return elements
+
+	return els
 }
 
 func (h *ElementHandle) Screenshot(opts goja.Value) goja.ArrayBuffer {
--- a/common/js/query_all.js
+++ b/common/js/query_all.js
@@ -1,0 +1,11 @@
+/**
+ * Finds all elements in a given scope.
+ * @param {Node} scope - The scope of searching. It can be a node.
+ *                       By default, it is document.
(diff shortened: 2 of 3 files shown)


imiric (commented):
I don't mind this, just a few questions/comments:

1. How does this make `QueryAll` more unit testable? You say that it allows us to test Go and JS code separately, but why would we want to test these one-line wrapping functions separately? It seems to me that we'd want our "unit" to be the Go `QueryAll` method itself, not the JS code.

   I remember from previous discussions that the issue with unit testing `QueryAll` was initializing `ElementHandle` and all its dependencies. How does this change help us with that?

2. Do we want to embed all other inline `fn` definitions for the other methods?

   Having JSDocs for them is nice I guess, but these are one-line wrapping functions. We'd be better off documenting `injected_script.js` itself, though I'd argue that is not high on our priority list.

3. I'm just noticing this, and it's out of scope for this PR, but whenever we call `evalWithScript()` we evaluate the entire 28kB `injected_script.js` _on each call_.

   I don't know if `cdproto/runtime` has some optimizations for this (doubtful), but we could at least ensure that we evaluate the script only once per frame. Should I create an issue for this?

   Maybe we don't need to evaluate the wrapping functions on each call either, but do it with `sync.Once` and just call it without evaluation after the first call.

inancgumus:
> How does this make QueryAll more unit testable?
> I remember from previous discussions that the issue with unit testing QueryAll was initializing ElementHandle and all its dependencies. How does this change help us with that?

You might have an idea what I'm planning to do in #229. Separating the actual Javascript code from the `QueryAll` allows us to test whether it uses the correct function. Please ask if you have any questions.

> You say that it allows us to test Go and JS code separately, but why would we want to test these one-line wrapping functions separately? It seems to me that we'd want our "unit" to be the Go QueryAll method itself, not the JS code.

Embedding JS funcs into the Go file makes us forget something or make mistakes, as you've also noticed [here](https://github.com/grafana/xk6-browser/pull/225#discussion_r802792683). Putting it in an editor-controlled environment allowed me to find that mistake.

Regarding testing. I believe, in time, we should also unit test our Javascript functions along with the injected script to understand whether they work correctly and ensure their quality, etc. One unit test for the Go function and another for the javascript code. With our integration tests in the tests folder, we'll ensure that everything works correctly and has a high quality. Btw, I'm planning to test the javascript itself, but I have no good idea for it yet.

> Do we want to embed all other inline fn definitions for the other methods?

Yes, in time, I want to embed all other inline Javascript functions to a proper Javascript file because of the reasons I explained above.

> Having JSDocs for them is nice I guess, but these are one-line wrapping functions. We'd be better off documenting injected_script.js itself, though I'd argue that is not high on our priority list.

We can do it step by step like I've been doing instead of a giant refactoring.

> I'm just noticing this, and it's out of scope for this PR, but whenever we call evalWithScript() we evaluate the entire 28kB injected_script.js on each call.

TBH, I noticed it before but I didn't tell you 😳  

> I don't know if cdproto/runtime has some optimizations for this (doubtful), but we could at least ensure that we evaluate the script only once per frame. Should I create an issue for this?
> Maybe we don't need to evaluate the wrapping functions on each call either, but do it with sync.Once and just call it without evaluation after the first call.

Yeah, good idea. I think we can create an issue for it with an optimization label. Then we can decide when to prioritize it.

imiric:
> You might have an idea what I'm planning to do in #229.

I'm still not seeing it, sorry. In #229 you're introducing an unexported method, which I agree is easier to test since it avoids the `k6Throw()` panic and `applySlowMo()`, but you still have to initialize `ElementHandle`, no?

> Separating the actual Javascript code from the QueryAll allows us to test whether it uses the correct function.

How so? And why would that be important? If you're asserting on what `QueryAll` (or `queryAll`) returns, why does it matter which JS function it uses?

> Embedding JS funcs into the Go file makes us forget something or make mistakes, as you've also noticed here. Putting it in an editor-controlled environment allowed me to find that mistake.

Sure, but that's a very minor benefit. If we're treating these units as black boxes, we should only care about their correct behavior. An additional argument in a JS function call is practically irrelevant.

> Regarding testing. I believe, in time, we should also unit test our Javascript functions along with the injected script to understand whether they work correctly and ensure their quality, etc.

How do you envision this happening if `injected_script.js` depends on a JS environment and the DOM? By definition `injected_script.js` is not unit testable since it depends on the browser.

This is why I think our smallest units should be at the Go method level.

> I want to embed all other inline Javascript functions to a proper Javascript file because of the reasons I explained above.

Sure, that makes sense, and we could only evaluate that single file once.

> We can do it step by step like I've been doing instead of a giant refactoring.

I appreciate that, but again, this is very low priority right now.

> TBH, I noticed it before but I didn't tell you

:facepalm: Please mention these things in the future... I missed it in #196 and #212, and it seems like a pretty major optimization issue.

I'll create the GH issue for it today/tomorrow.

imiric (approved):
Don't get me wrong, I'm not against this change. I just don't see how it helps with unit testing `QueryAll`, and think you're anticipating how and if we'll test JS in isolation. But there's no reason to not merge it :)

inancgumus:
Thanks! Here's the general idea: I'll return an `ElementHandle` stub since the `evalFn` returns an `api.ElementHandle`. Then I'm going to check how `QueryAll` uses the value. This way we can test the function in isolation instead of checking what it does with the session through the execution context (which I believe will unnecessarily complicate the test).

It's because it involves a lot of other types in the system to check what QueryAll returns, which will complicate the test and fix the current design into the tests. I don't want to bind the test to the QueryAll js's contents. Since this is a unit test, at this level, I only want to see what it calls (whether it is `js.QueryAll`), but I don't want to check the actual content. 

Regarding black-box testing: As I explained, with a relatively legacy code, it's hard to treat these units barely as black boxes until we have a proper design that allows us to test black-box only.

> An additional argument in a JS function call is practically irrelevant.

It was an example. It could have been a subtle bug that happens but rarely.

> This is why I think our smallest units should be at the Go method level.

Ideally, yes, I agree. But like I said, it'll take some time to do so.

> > We can do it step by step like I've been doing instead of a giant refactoring.

> I appreciate that, but again, this is very low priority right now.

I think a little bit of betterment of the code at hand is never a low priority, especially if it helps with testing, and especially we have 10% of code coverage.

> it seems like a pretty major optimization issue.

We never had a problem with it yet. Is there a performance problem that you observe and users complain about?

imiric:
OK, let's discuss the test in #229.

> Is there a performance problem that you observe and users complain about?

Just because it's unreported and we haven't noticed it doesn't mean it doesn't exist and that it won't become or already is a problem for someone. #231 is to explore what the actual impact is and how we should prioritize it.

Issue #200, opened 2022-01-21
waitForXXX are not working


inancgumus:
One of our users [reported](https://community.k6.io/t/auto-waits-and-page-waits-not-working-v0-1-3/2768/3) that the following sometimes not working:

* `page.waitForNavigation()`
* `page.waitForSelector()`
* `page.waitForLoadState('networkidle');`

imiric:
This issue is blocking reliable tests for #40 and #90 (#117). A test for #40 was added in #167 but we reverted it in e6db54c9 because of flakiness.

So just a reminder here to bring that test back after this is fixed, since both of those issues are now closed.

imiric:
I've been looking into this issue and here's what's happening and the timeline of events of the current [fillform.js](https://github.com/grafana/xk6-browser/blob/05f8b5fc8b5ae278296537bd490282c814a7ef07/examples/fillform.js) example:

1. The initial page is loaded (`/my_messages.php`), and the `networkidle` state is set to true (`Frame.lifecycleEvents`).
2. The script does a `page.$().click()`, which triggers a form submission and the eventual page navigation (`Page.frameRequestedNavigation` is received).
3. Immediately after the script calls `page.waitForLoadState('networkidle')`.

The issue is that by the time the `hasLifecycleEventFired()` call is done here:
https://github.com/grafana/xk6-browser/blob/05f8b5fc8b5ae278296537bd490282c814a7ef07/common/frame.go#L1412-L1422

... we haven't yet started processing the `Page.frameRequestedNavigation` event and the previous `networkidle` state is still true. I.e. we return right away here ignoring the `waitForLoadState()` call because we still have the state of the previous page load.

So this is a timing problem: we can't reliably know that a click will trigger a form submission for which we'll receive a navigation event (at which point we might call `Frame.clearLifecycle()`), since the `page.waitForLoadState()` call happens so quickly.

Since we can't reliably determine this in the background, the solution is simple: wait for the navigation event in the script before waiting for the load state. We already have an API for this: `page.waitForNavigation()`. :tada:

So the "fix" is in the script itself. Here is a working `fillform.js` with no `slowMo` and no sleeps:

```javascript
import launcher from 'k6/x/browser';
import { check } from 'k6';

export default function() {
  const browser = launcher.launch('chromium', {
    headless: __ENV.XK6_HEADLESS ? true : false,
  });
  const context = browser.newContext();
  const page = context.newPage();

  // Goto front page, find login link and click it
[...]

PR #230, opened 2022-02-09
Fix `fillform.js` example

imiric:
This removes the `sleep()` and `slowMo` option which were workarounds for the `waitForNavigation()` functionality we already have.

There's no issue for this, but see [this comment](https://github.com/grafana/xk6-browser/issues/200#issuecomment-1032454173).

### DIFF
--- a/examples/fillform.js
+++ b/examples/fillform.js
@@ -1,10 +1,9 @@
 import launcher from 'k6/x/browser';
-import { check, sleep } from 'k6';
+import { check } from 'k6';
 
 export default function() {
   const browser = launcher.launch('chromium', {
     headless: __ENV.XK6_HEADLESS ? true : false,
-	slowMo: '500ms' // slow down by 500ms
   });
   const context = browser.newContext();
   const page = context.newPage();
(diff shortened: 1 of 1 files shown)


examples/fillform.js
```diff
@@ -19,10 +18,9 @@ export default function() {
...
   page.$('input[type="submit"]').click();
 
-  sleep(5);
-  // Wait for next page to load
-  // FIXME: This doesn't work without the above sleep.
-  // Same with state 'domcontentloaded'.
+  // We expect the above form submission to trigger a navigation, so wait for it.
+  page.waitForNavigation();
+  // Wait for the next page to load
   page.waitForLoadState('networkidle');
```
inancgumus, line 24: Do we still need this? `waitForNavigation` already listens for a `networkidle` event. I tested locally by running the script multiple times and there wasn't any error.

inancgumus (approved):
LGTM with one question.

examples/fillform.js
imiric: Good point, it's not needed. Unless specified, `waitForNavigation()` defaults to waiting for a `load` event, which in this case serves the same purpose as `waitForLoadState('networkidle')`.

https://github.com/grafana/xk6-browser/blob/c8a58063eeb522ca999c9a6186057c5766221cd1/common/frame_options.go#L653-L659

Fixed in latest push.

PR #238, opened 2022-02-15
Fix contextWithDoneChan possible goroutine leak

inancgumus:
The goroutine may leak if the passed context is canceled before the done channel, even if it is closed afterward.

This doesn't add a test for checking whether the function hangs on the `done` chan if the context gets canceled. I couldn't find a clean way to do that. Please let me know if you have an idea.

Related: #237

### DIFF
--- a/common/context.go
+++ b/common/context.go
@@ -70,8 +70,11 @@ func GetProcessID(ctx context.Context) int {
 func contextWithDoneChan(ctx context.Context, done chan struct{}) context.Context {
 	ctx, cancel := context.WithCancel(ctx)
 	go func() {
-		<-done
-		cancel()
+		defer cancel()
+		select {
+		case <-done:
+		case <-ctx.Done():
+		}
 	}()
 	return ctx
 }
--- a/common/context_test.go
+++ b/common/context_test.go
@@ -1,0 +1,20 @@
+package common
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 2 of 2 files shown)


imiric (approved):
This definitely fixes that scenario, but I'm not sure how to test it either :confused: If we pass a context that we cancel in the test, then the child context returned by `contextWithDoneChan` will be done as well, so we can't assert on it.

I think this is fine as is.

inancgumus:
> If we pass a context that we cancel in the test, then the child context returned by `contextWithDoneChan` will be done as well, so we can't assert on it.

Exactly... One solution that I worked on was sending a message to the `done` channel and seeing whether it's been received if the context is canceled. However, this required me to change the `contextWithDoneChan` code unnecessarily (I needed to check `<-ctx.Done` first then again with `<-done`). So I excluded it from the PR.

common/context.go
```diff
@@ -70,8 +70,11 @@ func GetProcessID(ctx context.Context) int {
 func contextWithDoneChan(ctx context.Context, done chan struct{}) context.Context {
```
imiric, line 70: One thing I forgot to mention (and I can't comment on directly): should we remove the "The context will leak if the done channel is never closed" sentence from the comment, since we'll now cancel it if the the main context is done?

inancgumus: Right 🤦 I'll remove it and directly push it to `main`.

inancgumus: See: https://github.com/grafana/xk6-browser/commit/917b83ef14b3e2075f92c092163c47e359883a6a

PR #244, opened 2022-02-16
Add test short skip for slow mo tests

inancgumus:
We can skip slow motion tests when we give the `-short` flag. The rest of the tests become 10 seconds faster to run. I think we don't need to run slow-motion tests all the time. I'd like to merge this. WDYT, @imiric?

```
➜  k6b git:(test/skip-slowmo) ✗ go test ./tests -failfast -race
ok  	github.com/grafana/xk6-browser/tests	26.945s
➜  k6b git:(test/skip-slowmo) ✗ go test ./tests -failfast -race -short
ok  	github.com/grafana/xk6-browser/tests	17.964s
```

### DIFF
--- a/tests/launch_options_slowmo_test.go
+++ b/tests/launch_options_slowmo_test.go
@@ -32,6 +32,10 @@ import (
 )
 
 func TestLaunchOptionsSlowMo(t *testing.T) {
+	if testing.Short() {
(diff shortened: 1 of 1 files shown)


imiric (approved):
Sure, makes sense :+1:

PR #255, opened 2022-02-25
Fix binary signing on Windows

imiric:
This should hopefully fix the recent failure to sign the binary, see https://github.com/grafana/xk6-browser/runs/5332928690.

I confirmed that it [works on `windows-2019`](https://github.com/grafana/xk6-browser/runs/5334303581), but not on [`windows-latest`](https://github.com/grafana/xk6-browser/runs/5334286611). The strange thing is that [`signtool` is still in the same path](https://github.com/grafana/xk6-browser/runs/5333157519) on both versions, but I'm guessing there's an issue with GH Actions and the way `GITHUB_PATH` works, but it hasn't been documented/reported yet.

### DIFF
--- a/.github/workflows/all.yaml
+++ b/.github/workflows/all.yaml
@@ -18,7 +18,7 @@ env:
 
 jobs:
   deps:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - name: Checkout code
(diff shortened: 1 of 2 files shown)


inancgumus (approved):
Let's give it a try 🤞

Issue #258, opened 2022-02-28
[Page|Frame].WaitForSelector is flaky


inancgumus:
_This issue is valid for [v0.2.0](https://github.com/grafana/xk6-browser/releases/tag/v0.2.0)._

While working on #171, I noticed that `waitForSelector` is fragile and may randomly fail if the script doesn't sleep or slow-motion between steps.

<details>
<summary>The flaky script.</summary>

```js
import { group } from 'k6';
import launcher from 'k6/x/browser';

export default function () {
  const browser = launcher.launch('chromium', {
    headless: false,
    // slowMo: '500ms' // slow down by 500ms
  });

  let context = browser.newContext({
    // hack:
    // Without setting viewport, the script cannot find the add-to-cart
    // element. We might want to improve this.

    // screen: { width: 1920, height: 1080 },
    // viewport: { width: 1920, height: 1080 },
  });

  const page = context.newPage();

  group("Navigate to Homepage", () => {
    page.goto('http://ecommerce.test.k6.io/', { waitUntil: 'networkidle' });

    page.waitForSelector('.woocommerce-result-count');
  });

  group("Click Product", () => {
    page.click('.woocommerce-loop-product__title', { waitUntil: 'networkidle' });
  });

  group("Click Add to Cart", () => {
    console.log("Clicking Add to Cart...");

    // wait for the element to be attached
    let b = page.waitForSelector('button[name="add-to-cart"]');

    // hack: scroll the element into view
    page.evaluate(() => {
      const el = document.querySelector('button[name="add-to-cart"]');
      el.scrollIntoView();
    });
    b.click();

    console.log("Waiting for .woocommerce-message...");

    page.waitForSelector('.woocommerce-message');
  });

  group("View Cart", () => {
    page.click('.cart-contents', { waitUntil: 'networkidle' });

    page.waitForSelector('.woocommerce-cart-form__cart-item');
  });

  page.close();
  browser.close();
}
```

</details>

However, this may be the tip of the iceberg because the problem is more profound: Stale execution contexts #259.

**Explanation:**

* Frame calls `Runtime.callFunctionOn(...)` using execution context: X.
* Frame navigates.
* New execution context gets created: Y.
* FrameSession updates the new execution context Y for the frame.
* `Runtime.callFunctionOn` returns an error because it evaluated the function call in the execution context X instead of Y.

[...]

PR #260, opened 2022-02-28
Fix/258 wait for selector flaky

inancgumus:
See issue #258 for details.

Fixes: #258

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -36,6 +36,9 @@ import (
 	"github.com/grafana/xk6-browser/api"
 )
 
+// maxRetry controls how many times to retry if an action fails.
+const maxRetry = 1
+
 // Ensure frame implements the Frame interface.
 var _ api.Frame = &Frame{}
 
@@ -546,6 +549,18 @@ func (f *Frame) waitForFunction(
 		return nil, fmt.Errorf("frame cannot wait for function: %w", err)
 	}
 	return result, nil
+}
+
+func (f *Frame) waitForSelectorRetry(
+	selector string, opts *FrameWaitForSelectorOptions, retry int,
+) (h *ElementHandle, err error) {
+	for ; retry >= 0; retry-- {
+		if h, err = f.waitForSelector(selector, opts); err == nil {
+			return h, nil
+		}
+	}
+
+	return nil, err
 }
 
 func (f *Frame) waitForSelector(selector string, opts *FrameWaitForSelectorOptions) (*ElementHandle, error) {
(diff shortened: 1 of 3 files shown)


common/frame.go
```diff
@@ -548,7 +551,17 @@ func (f *Frame) waitForFunction(
 	return result, nil
 }
 
-func (f *Frame) waitForSelector(selector string, opts *FrameWaitForSelectorOptions) (*ElementHandle, error) {
+func (f *Frame) waitForSelector(selector string, opts *FrameWaitForSelectorOptions) (h *ElementHandle, err error) {
```
inancgumus, line 554: Better to name this `waitForSelectorWithRetry(selector string, opts *FrameWaitForSelectorOptions, retry int)` and leave the original `waitForSelector`'s name as it was.

inancgumus:
Actually, I noticed that a better solution might be adding retry functionality to [frameAction](https://github.com/grafana/xk6-browser/blob/87162345e798faf90b116b2acede278dda3ad132/common/frame.go#L1542-L1545) and [framePointerAction](https://github.com/grafana/xk6-browser/blob/87162345e798faf90b116b2acede278dda3ad132/common/frame.go#L1570-L1573) because all (?) the actions use them.

common/frame.go
imiric: Yeah, makes sense. Would you want to expose the `retry` argument to JS instead of having it constant?

common/frame.go
```diff
@@ -548,7 +551,17 @@ func (f *Frame) waitForFunction(
...
+	for retry := maxRetry; retry > 0; retry-- {
+		if h, err = f.waitForSelectorOnce(selector, opts); err == nil {
+			return h, nil
```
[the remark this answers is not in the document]
inancgumus: @imiric, It allows the frame to use the updated execution context. Please let me know if you have another idea, but this did the work :-)

A better solution can be: When an execution context gets changed, detect it, cancel all the actions, and retry them. However, I can anticipate how much refactoring work it's going to need 😱 

I checked out the PW code now, and it turned out they have a [similar retry mechanism](https://github.com/microsoft/playwright/blob/b6c2b32b530e88bb8b468e84d26929d06a15af5f/tests/page/page-click.spec.ts#L809).

imiric (approved):
I'm finding it difficult to reproduce the issue this fixes on `main` (24a8ffd), but I did get one failure after ~20 attempts running the script in #258:

```
ERRO[0036] waitForFrameNavigation timed out after 30s
        at file:///home/ivan/Projects/grafana/xk6-browser/examples/258-waitForSelector-flaky.js:29:27(102)
        at native  executor=per-vu-iterations scenario=default source=stacktrace
```

Is this the error you're seeing?

I'm not fond of this workaround, but it's better than retrying 5 times, and hopefully we can fix it for good in #200. The fact it's difficult to reproduce makes it difficult to fix.

inancgumus:
@imiric

Here you can see the difference between running the script on `main` and after this PR.
https://asciinema.org/a/cclB8uuJOaZp5VQ5W0L9kLCDa

Then, please try it with the updated script (#258) 🤦 Thank you for your understanding!

PR #268, opened 2022-03-08
WIP Implement `Browser.on(event)`

### LINKED ISSUE #95: Implement `Browser.on(event)`
Add support for `Browser.on(event)`, to allow listening and reacting to browser events.

**Relevant links**:
- Playwright docs: https://playwright.dev/docs/api/class-browser#browser-event-disconnected

imiric:
This is an initial implementation of [`Browser.on(event)`](https://playwright.dev/docs/api/class-browser#browser-event-disconnected), using the recently released [event loop support in k6](https://github.com/grafana/k6/pull/2228). The only supported event is `disconnected`, so it can be used as an alternative to `teardown()`.

It also simplifies the `Browser.isConnected` implementation, and properly closes the WS connection when `Browser.close()` is called, which should minimize the [`close 1006 (abnormal closure): unexpected EOF` we sometimes see on shutdown](https://community.k6.io/t/erro-0006-unexpected-eof-error-during-run-of-example-script/3136).

I'm opening it as a draft to get initial feedback since I'm unsure about several things I'll point out in the comments.

Closes #95


examples/browser_on.js
```diff
@@ -0,0 +1,41 @@
+import launcher from 'k6/x/browser';
```
imiric, line 1: Currently it's only possible to add E2E tests for this feature, since [the mock `VU` doesn't have a `RegisterCallback` implementation yet](https://github.com/grafana/k6/blob/c31e840793b392e16316d35b83e08345629b2fd5/js/modulestest/modulestest.go#L62-L66), so we can't have a Go integration test for it.

If this is a blocker for merging this, then we probably should contribute that to k6... :sweat_smile:

common/browser.go
```diff
@@ -464,6 +464,36 @@ func (b *Browser) NewPage(opts goja.Value) api.Page {
...
+		case <-b.browserProc.lostConnection:
+			cb(func() error {
+				resolve(true)
+				return nil
+			})
+		case <-b.ctx.Done():
+			cb(func() error {
+				reject(fmt.Errorf("browser.on promise rejected: %w", b.ctx.Err()))
+				return nil
+			})
```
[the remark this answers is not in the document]
MStoykov: > Another question is about the value passed to reject(). I wasn't sure how to create a JS Error object or whether Goja will properly convert this Go error, and whether we should even return ctx.Err() in this case...

It doesn't need to be error so you can pass w/e. I guess making an error is best and will try to look into it. 

> I haven't been able yet to trigger this failure handler externally..

Does it trigger the other case at all? Do you hit ^C twice?

common/browser.go
```diff
@@ -59,15 +60,16 @@ type Browser struct {
...
 	// A *Connection is saved to this field, see: connect().
-	conn         cdp.Executor
+	conn interface {
+		cdp.Executor
+		Close(...goja.Value)
+	}
 	connEvents   EventEmitter
 	connSessions interface {
 		getSession(target.SessionID) *Session
 	}
```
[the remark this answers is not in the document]
inancgumus: Somewhat agree. Please take a look: https://github.com/grafana/xk6-browser/pull/269

common/browser.go
```diff
@@ -464,6 +464,36 @@ func (b *Browser) NewPage(opts goja.Value) api.Page {
...
 
+// On returns a Promise that is resolved when the browser process is disconnected.
+// The only accepted event value is "disconnected".
+func (b *Browser) On(event string) *goja.Promise {
+	if event != EventBrowserDisconnected {
+		k6Throw(b.ctx, "unknown browser event: %q, must be %q", event, EventBrowserDisconnected)
+	}
+
+	rt := k6common.GetRuntime(b.ctx) //nolint: staticcheck // SA1019 (deprecated)
+	rcb := GetRegisterCallback(b.ctx)
```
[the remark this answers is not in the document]
inancgumus: > I know that abusing the context for this is an anti-pattern

What could be the more challenging and maybe longer way? Any ideas?

> these are functions that will be used extension-wide, and passing them everywhere they're needed would be very cumbersome.

Yeah, but anything that can change behavior is dangerous to put in context.

common/types.go
```diff
@@ -33,6 +33,9 @@ import (
 	"github.com/grafana/xk6-browser/api"
 )
 
+// RegisterCallbackFn represents the function used to register an event loop callback.
+type RegisterCallbackFn func() func(func() error)
+
```
inancgumus, line 38: `RegisterCallbackFunc`: I didn't want to bring this up but can we at least follow the idiomatic Go naming conventions here? For example: https://pkg.go.dev/bufio@go1.17.8#SplitFunc 🙏

common/context.go
```diff
@@ -65,6 +66,23 @@ func GetProcessID(ctx context.Context) int {
...
+	v := ctx.Value(ctxKeyCallback)
+	if v == nil {
+		return nil
+	}
+	if rcb, ok := v.(RegisterCallbackFn); ok {
+		return rcb
+	}
+	return nil
+}
+
```
inancgumus, line 85: We can simplify this with:

```suggestion
func GetRegisterCallback(ctx context.Context) RegisterCallbackFn {
	v := ctx.Value(ctxKeyCallback)
	if v, ok := v.(RegisterCallbackFn); ok {
		return v
	}
	return nil
}
```

examples/browser_on.js
inancgumus: :) I think we can merge this first and create another issue for the E2E test (if mock VU `RegisterCallback` won't be available yet).

examples/browser_on.js
```diff
@@ -0,0 +1,41 @@
...
+export default function() {
+  const browser = launcher.launch('chromium', {
+    headless: __ENV.XK6_HEADLESS ? true : false,
+  });
+
+  check(browser, {
+    'should be connected after launch': browser.isConnected(),
+  });
+
+  sleep(2);
```
inancgumus, line 13: Is this necessary here? Since this is an example, we'll guide the users to use `sleep` calls that we don't recommend.

common/browser.go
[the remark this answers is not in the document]
MStoykov: I did add some `fmt.Println` in this go code and a `sleep` before the closing the `browser.close()` and it works as expected:
on success I get: 
```
here
hereing (00m02.2s), 1/1 VUs, 0 complete and 0 interrupted iterations
lost connection

running (00m04.3s), 0/1 VUs, 1 complete and 0 interrupted iterations
default ✓ [======================================] 1 VUs  00m04.3s/10m0s  1/1 iters, 1 per VU

     ✓ should be connected after launch
     ✓ should be connected before ending iteration
     ✓ should be disconnected on event
     ✓ the browser.on success handler should be called

     checks...............: 100.00% ✓ 4        ✗ 0
     data_received........: 0 B     0 B/s
     data_sent............: 0 B     0 B/s
     iteration_duration...: avg=4.28s min=4.28s med=4.28s max=4.28s p(90)=4.28s p(95)=4.28s
     iterations...........: 1       0.233322/s
     vus..................: 1       min=1      max=1
     vus_max..............: 1       min=1      max=1
```

on failure you get: 
```
here
hereing (00m02.2s), 1/1 VUs, 0 complete and 0 interrupted iterations
^Cctx.Done

running (00m02.9s), 0/1 VUs, 0 complete and 1 interrupted iterations
default ✗ [--------------------------------------] 1 VUs  00m02.9s/10m0s  0/1 iters, 1 per VU
WARN[0003] No script iterations finished, consider making the test duration longer

     ✓ should be connected after launch
     ✓ should be connected before ending iteration

     checks..........: 100.00% ✓ 2   ✗ 0
     data_received...: 0 B     0 B/s
     data_sent.......: 0 B     0 B/s
     vus.............: 1       min=1 max=1
     vus_max.........: 1       min=1 max=1

ERRO[0003] err:websocket: close 1006 (abnormal closure): unexpected EOF  category="Connection:handleIOError" elapsed="0 ms" goroutine=98
```

As I explained in the event loop PR - while we wait for all register callbacks to be called that doesn't mean that you will be able to execute code. For now the same context is used for everything and k6 interrupts the goja VM once it's done, so while there is *some* chance for race where you manage to execute the `catch` in this particular case it's not big ;)

This will likely be worked on in future improvements

common/browser.go
```diff
@@ -429,11 +424,16 @@ func (b *Browser) Contexts() []api.BrowserContext {
...
+// IsConnected returns whether the WebSocket connection to the browser process
+// is active or not.
 func (b *Browser) IsConnected() bool {
-	b.connectedMu.RLock()
-	defer b.connectedMu.RUnlock()
-
-	return b.connected
+	var ok bool
+	select {
+	case _, ok = <-b.browserProc.lostConnection:
```
inancgumus, line 432: Instead of checking it like this, we can hand over this responsibility to the browser process:

```go
Browser.IsConnnected() bool {
  return b.browserProc.connected()
}
```

common/browser.go
```diff
@@ -414,6 +408,7 @@ func (b *Browser) Close() {
 	// will stop emitting events.
 	b.browserProc.GracefulClose()
 	b.browserProc.Terminate()
+	b.conn.Close()
```
inancgumus, line 411: It makes sense to put this here since the browser is the one that opens up a new connection.

inancgumus (commented):
Here are my suggestions.

common/browser.go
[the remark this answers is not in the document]
MStoykov: In a addition currently even if you decide to clear the interrupt k6 has *probably* already gotten an error and is only waiting for the register callbacks to callback and will *not* even try to execute them. 

Again this in practice requires that we have separate mechanism to say:
- an iteration is ending (due to an exception for example)
- the scenario is ending 
- the process is ending
- (possibly) gracefulShutdown/Rampdown has been entered. In practice for gracefulRampDown we might actually also want to signal that it's no longer relevant. There is  a case where going up to 10 down to 0 and back up to 10 VUs in cycles might mean that gracefulRampDown doesn't actually stop a VU outright as by the time it has run out we are on the next "up to 10 VUs". 

In all of the above cases we can potentially let javascript code and extensions to do ... something before we completely shutdown the system (as we do now). But we will still need a way to interrupt the goja interpreter as otherwise malicious code can always block k6. Figuring out if goja is executing malicious code or not is also in practice very hard. I am not even certain we can even guess if it's executing code *at all*. 

Arguably a "quick" fix for your case might be to add a bit of a wait on the interrupt :thinking: . Although this will still not work for if I have add `sleep(2)` before the `browser.close()` as then the `sleep` will *not* stop so even if you queue something it will still not get executed ... 

This also kind of illustrates the general problem - we likely want to abort some code in this situation ... but not others. And while most of the code is running synchronously this means the synchronous code blocks any execution of any asynchronous possibly trying to do something. 

So yeah this is an open problem and one that won't be fixed easily. 

Currently waiting on `ctx.Done` is generally useful if you guys want to clean something up without the use of javascript code. For example closing browser or flushing metrics - although that last one usually depends on the context not being close so ... yeah.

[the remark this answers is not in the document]
inancgumus: As always, the wisdom of words, thanks 🙇 

What about this?

```golang
func (b *Browser) On(event string) *goja.Promise {
	...
	go func() {
		a := func() error {
			resolve(true)
			return nil
		}
		r := func() error {
			reject(fmt.Errorf("browser.on promise rejected: %w", b.ctx.Err()))
			return nil
		}
		select {
		case <-b.browserProc.lostConnection:
			select {
			case <-b.ctx.Done():
				cb(r)
				return
			default:
				cb(a)
			}
		case <-b.ctx.Done():
			cb(r)
		}
	}()
```

*RESULT:*

```shell
ERRO[0002] promise rejected: browser.on promise rejected: context canceled  source=console

     ✓ should be connected after launch
     ✓ should be connected before ending iteration
     ✗ the browser.on success handler should be called
      ↳  0% — ✓ 0 / ✗ 1
```

This is most probably not a solution to this problem but it might give us an idea.

inancgumus:
@imiric, by the way, we don't return here even though the connection is closed:

```golang
func (b *Browser) initEvents() error {
	...
	go func() {
		for {
			select {
			case <-cancelCtx.Done():
				return
			case event := <-chHandler:
				...
				} else if event.typ == EventConnectionClose {
					...
					b.browserProc.didLoseConnection()
					b.cancelFn()

					// LOOK HERE: it makes sense to return here
				}
			}
		}
	}()
```

This might be a bug. I wanted to let you know.

common/browser.go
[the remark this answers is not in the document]
imiric: Hhmm the thing is that whatever approach we take with `RegisterCallback`, it should be consistent with what we do for [the other `VU` methods](https://github.com/grafana/k6/blob/a3d7fda089e00bf04ce3d70d4e3cdca43b81e018/js/initcontext.go#L168-L186). Since the context based getters were deprecated in k6, I don't see it as premature optimization, and we should change this soon.

Since these are essentially static within the context of a browser `Launch()`/`Close()`, WDYT if we just attach the `VU` itself to the context? That way we don't have to keep adding methods if/when they change, and can have access to any of the methods throughout the extension?

The problem I see with passing them is that we'll have to change a bunch of signatures, especially with `Runtime()`.  `rg '\.GetRuntime' | wc -l` outputs `158`.

common/browser.go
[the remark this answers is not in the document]
inancgumus: I noticed something:

`Browser.On`'s `case <-b.ctx.Done():` is selected at the end of the script when we press CTRL+C before the test run ends:

```js
export default function() {
  ...
  sleep(5); // hit CTRL+C here
  browser.close();
}
```

Even though it executes the rejection callback in the `Browser.On`, nothing happens on the k6 output as if it was not rejected at all. Without CTRL+C, the accept callback runs (and context is closed).

This brings us to:

> As I explained in the event loop PR - while we wait for all register callbacks to be called that doesn't mean that you will be able to execute code.

common/types.go
imiric: I find it funny that Go is usually in favor of terseness, but the convention is `Func` instead of the widely known `Fn`. :man_shrugging:

In any case, I'll probably remove this [once I attach `VU` to the context](https://github.com/grafana/xk6-browser/pull/268#discussion_r822537491).

examples/browser_on.js
imiric: We discussed this previously, but we shouldn't suggest that `sleep()` is wrong for _all_ use cases. After all, if you genuinely want the script to wait for whatever reason (user pause, or in this case just waiting some time so that I could test `^C`), it should be fine to use. The "wrong" use case is the one where the script is waiting for an element to appear on the page, or any other browser event, for which we have the `waitFor*` methods, `waitUntil`, now `on(event)`, etc.

common/browser.go
imiric: Good idea, will do :+1:

common/browser.go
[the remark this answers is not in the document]
inancgumus: > k6common.GetRuntime() was deprecated, but I chose to ignore it for now since we use this everywhere and we'll have to figure out a way of exposing vu.Runtime in a similar way...

This is precisely the problem that we better stop exposing any VU mechanics in the extension code.

> WDYT if we just attach [the VU itself](https://github.com/grafana/xk6-browser/blob/36f44841b3242f1f924eb0b44775b5356b749481/main.go#L71) to the context

I think we shouldn't clutter code with VU mechanics and start refactoring VU-related stuff away as much as we can. So, instead of passing it in a Context, I'm on the side of passing and embedding VU in the types.

> The problem I see with passing them is that we'll have to change many signatures, especially with Runtime(). rg '\.GetRuntime' | wc -l outputs 158.

`Runtime` is most needed because of Goja conversions. If we can provide a translation layer between Goja and the extension code, we can greatly simplify it. For example:

```go
func (b *BrowserContext) WaitForEvent(event string, optsOrPredicate goja.Value) interface{} {
	rt := k6common.GetRuntime(b.ctx)

	if optsOrPredicate != nil && !goja.IsUndefined(optsOrPredicate) && !goja.IsNull(optsOrPredicate) {
		switch optsOrPredicate.ExportType() {
		case reflect.TypeOf(goja.Object{}):
			opts := optsOrPredicate.ToObject(rt)
			for _, k := range opts.Keys() {
				switch k {
				case "predicate":
					predicateFn, isCallable = goja.AssertFunction(opts.Get(k))
					if !isCallable {
						k6Throw(b.ctx, "expected callable predicate")
					}
				case "timeout":
					timeout = time.Duration(opts.Get(k).ToInteger()) * time.Millisecond
				}
			}
```

This kind of cluttering happens because of internally exposing Goja, whereas we could do:

```go
// Extension mechanics package (actual logic here)

package browser

func (c *Context) WaitForEvent(event string, predicate func()) (Event, error) {
	...
	return predicate(c.page)
}

// Translation layer (goja -> extension)

package common

type BrowserContext struct {
	...
	bc browser.Context
	...
}

func (b *BrowserContext) WaitForEvent(event string, optsOrPredicate goja.Value) interface{} {
	// goja stuff here
	r, err := b.bc.WaitForEvent(event predicate)
	if err != nil {
		k6Throw(...)
	}   
}
```

Or something along these lines. But I think this is too much for this discussion right now :) At least, we don't keep cluttering other stuff and get VU-related stuff in an embedded field in each type as a way to pave the future refactoring work. WDYT?

examples/browser_on.js
inancgumus: I think people can see it and use it everywhere without thinking of that kind of stuff. But you know. I was wondering if it's essential there?

common/browser.go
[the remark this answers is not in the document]
inancgumus: Since you think not putting it in the context will be costly, I guess I have a practical idea that combines both approaches:

* Pass it in the context to the `BrowserType`.
* Let it initialize its own VU field.
* Let it have methods that access the VU methods through the VU field.
* Each other type would do the same. They'll extract VU from the context in their `NewX` function. And then fill their own fields.

This can be greatly simplified if we embed VU-accessor in each type.

common/browser.go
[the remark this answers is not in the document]
imiric: @inancgumus We should definitely do that, but from my tests it doesn't help with the interruption. I wasn't able to get a Promise rejection on `^C` even after liberally sprinkling `time.Sleep()` everywhere. Not sure why it works for you... :confused:

Ah, it works if I remove `sleep()` from the script like Mihail mentioned, even without `time.Sleep()` in Go. But that's a general context cancellation, not what happens with `^C`. Though it might be easy to write a test for that case, and if we attach VU to the context and hackily implement the `GetRegisterCallback()` mock, even a unit test should be easy to write. :tada: Will explore this a bit...

But yeah, we need some changes in k6 for the `^C` case to work. @MStoykov A solution might be to pass extensions a cancelable context [here](https://github.com/grafana/k6/blob/fb320f6df8a1694477d96d2f57d982fae1c3d2d1/js/runner.go#L153), and simply cancel it when appropriate. Additionally/alternatively, we could allow extensions to specify a callback on registration (for "deregistration"/cleanup), and run that when appropriate. Or, if we don't want to allow extensions to block k6 (though this shouldn't be a concern; extensions can already do that in many different ways), we could pass them a cancelable context and a "shutdown" channel exposed on `VU` with which they can signal when they're ready to shutdown, and k6 could timeout and do the current forced shutdown after a (user configurable?) amount of time.

examples/browser_on.js
imiric: People should read documentation, and I can add a comment here why it's used, but _we_ should be clear about what we recommend to users. Again, `sleep()` is perfectly valid in some use cases, and yes, it's essential here because I'm not quick enough to press `^C`. ;)

common/browser.go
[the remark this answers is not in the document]
imiric: I agree that we should separate the JS layer from our "business" logic, but that's an entirely separate discussion.

I don't entirely follow your last suggestion, but I'll push a change with what I'm thinking and we can discuss it.

common/browser.go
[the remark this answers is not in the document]
inancgumus: +1 for this one:

> allow extensions to specify a callback on registration (for "deregistration"/cleanup), and run that when appropriate

examples/browser_on.js
inancgumus: I'm not arguing whether it's valid or not, for ^C, it's OK, I was asking for the merged example :-) No problemo. And this would be a good addition, +1:

> I can add a comment here why it's used

common/browser.go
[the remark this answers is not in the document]
inancgumus: Let me explain what I mean.

First of all, let's make an internal VU struct:

```go
type VU struct {
	k6modules.VU
	// Makes sense to make it a field since we use it often
	rt *goja.Runtime
}

// WithVU returns a new context based on ctx with the k6 VU instance attached.
func WithVU(ctx context.Context, vu k6modules.VU) context.Context {
	return context.WithValue(ctx, ctxKeyVU, vu)
}

// GetVU returns the attached k6 VU instance from ctx, which can be used to
// retrieve the goja runtime and other k6 objects relevant to the currently
// executing VU.
// See https://github.com/grafana/k6/blob/v0.38.0/js/initcontext.go#L168-L186
func GetVU(ctx context.Context) *VU {
	v := ctx.Value(ctxKeyVU)
	vu, ok := v.(k6modules.VU)
	if !ok {
		return nil
	}

	return &VU{
		VU: vu,
		rt: vu.Runtime(),
	}
}
```

Then:

1. Pass the VU itself in the context (like you did: ✔️):

```go
ctx := common.WithVU(m.vu.Context(), m.vu)

if browserName == "chromium" {
	bt := chromium.NewBrowserType(ctx)
	...
}
```

2. Register the VU in the `Browser`'s field:

```go
type Browser struct {
	...
	vu       *VU
	...
}
```

3. Get VU:

```go
func newBrowser(ctx context.Context, ...) *Browser {
	return &Browser{
		...,
		vu:                  GetVU(ctx),
		...,
	}
}

4. Use VU:

```go
func (b *Browser) On(event string) *goja.Promise {
	...
	cb := b.vu.RegisterCallback()
	p, resolve, reject := b.vu.rt.NewPromise()
	...
}
```

Goal: Not carrying over into the unknown places in the system, for now and the future. And do it in a controlled way.

common/browser.go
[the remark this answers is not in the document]
MStoykov: @imiric please open an issue with your suggestions :)

common/browser.go
imiric: Resolved by 0f468ea.

examples/browser_on.js
imiric: I decided to remove `sleep()` from this script (7c332cb), since it was only useful for interactive testing the `^C` behavior, but since that's not useful for this yet, we should try to add integration tests for it by canceling the context, which will need to wait until https://github.com/grafana/k6/pull/2399 is merged.

Unfortunately, this exposed a nil panic issue which I tentatively fixed in a0e8d14. Let me know if you want a separate PR for that (and 02976ae and ce47855 I guess?). I'd rather keep them here if you don't mind; they're isolated commits, very small and easy to review.

And now I ran into #272... https://github.com/grafana/xk6-browser/runs/5498886146

common/context.go
imiric: Resolved by e5076d5.

common/types.go
imiric: Resolved by e5076d5.

common/browser.go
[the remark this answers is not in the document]
imiric: The issue is https://github.com/grafana/k6/issues/2432.

@inancgumus I decided against implementing [this suggestion](https://github.com/grafana/xk6-browser/pull/268#discussion_r822478618), since it fails the current `browser_on.js` test by rejecting the promise (i.e. what we wanted to reproduce with `^C`). I think that if the `lostConnection` case happens, we don't need to do another `ctx.Done()` check to reject the promise. If the connection is closed it should be a successful scenario, and the promise should be rejected only if that's not the case.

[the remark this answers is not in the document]
inancgumus: It's OK. It wasn't a suggestion, but hopefully, a meant to be a way to give you (us) an idea of what was happening :)

examples/browser_on.js
inancgumus: Is it better to check whether it's nil here by failing fast instead of silencing it? Besides, we now have a zombie FrameSession GR.

```go
func (fs *FrameSession) initFrameTree() error {
	...
	if frameTree, err = action2.Do(cdp.WithExecutor(fs.ctx, fs.session)); err != nil {
		return fmt.Errorf("unable to get page frame tree: %w", err)
	} else if frameTree == nil {
		...
	}

```

imiric: I considered that, but returning an error here might have unintended side effects up the stack. Even though [the current test looks fine](https://github.com/grafana/xk6-browser/runs/5512995232), we know these are not exhaustive to be reliable.

And what zombie FrameSession GR? The only one launched is in `initEvents()`, which happens before this. Presumably they would all exit once `fs.ctx` is done, or just be reaped by Go on shutdown. We probably have dozens of those right now.

Anyway, since the test passes with that change, I left it in 51de7f9.

examples/browser_on.js
imiric: I'll make a note in #95 and leave the issue open so that we don't forget to add tests for it once https://github.com/grafana/k6/pull/2399 is merged.

But I think we can merge this PR with this single E2E test.

inancgumus (approved):
Nice job.

Issue #266, opened 2022-03-08
Improve detection of non-clickable elements


inancgumus:
This issue is a spin-off of issue #171, and issue #264 depends on this one.

When we run the script in issue #171, it results in:

```shell
"wait for selector %s did not result in any nodes error".
```

It's because we think the element we're looking for is clickable when it's not (another element gets above the target element and makes it not clickable in practice). Scroll retrying needs this fix. Otherwise, we would think that we can properly scroll to the target element and expect we can click on it—resulting in a timeout error (with the error message I explained above).

PR #270, opened 2022-03-10
Fix improve detection of non clickable elements

inancgumus:
See issue #266 for an explanation.

#### First fix:
`checkHitTargetAt` [now returns with an error](https://github.com/grafana/xk6-browser/pull/270#discussion_r823564738) if the target element is obscured by another element; instead of waiting for the target element to become clickable. This fix is needed before improving the scrolling retry mechanism. Otherwise, we would think that we properly scroll to the target element (and find it clickable) but cannot click on it! Timeout.

#### Second fix:
`cdppage.GetLayoutMetrics` now uses CSS layout metric when calculating the client's inner width and height (viewport) instead of depending on the first layout parameter ([deprecated](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-getLayoutMetrics) and was returning incorrect results).

```diff
- if layoutViewport, _, _, _, _, _, err = action2.Do(cdp.WithExecutor(h.ctx, h.session)); err != nil
+ if _, _, _, layoutViewport, _, _, err = action2.Do(cdp.WithExecutor(h.ctx, h.session)); err != nil
```

#### Nice-to-have fix:
We now count the target element as clickable if its surface area is visible >0.99px. It was 1.01px before.

```diff
- if math.Abs(area) > 1 {
+ // Count any area above 0.99px is clickable
+ if math.Abs(area) > 0.99 {
	filteredQuads = append(filteredQuads, q)
}
```

### DIFF
--- a/common/element_handle.go
+++ b/common/element_handle.go
@@ -227,7 +227,7 @@ func (h *ElementHandle) boundingBox() (*Rect, error) {
 func (h *ElementHandle) checkHitTargetAt(apiCtx context.Context, point Position) (bool, error) {
 	frame := h.ownerFrame(apiCtx)
 	if frame != nil && frame.parentFrame != nil {
-		element := h.frame.FrameElement().(*ElementHandle)
+		element := h.frame.FrameElement().(*ElementHandle) //nolint:forcetypeassert
 		box, err := element.boundingBox()
 		if err != nil {
 			return false, err
@@ -236,8 +236,8 @@ func (h *ElementHandle) checkHitTargetAt(apiCtx context.Context, point Position)
 			return false, errors.New("cannot get bounding box of element")
 		}
 		// Translate from viewport coordinates to frame coordinates.
-		point.X = point.X - box.X
-		point.Y = point.Y - box.Y
+		point.X -= box.X
+		point.Y -= box.Y
 	}
 	fn := `
 		(node, injected, point) => {
@@ -253,15 +253,19 @@ func (h *ElementHandle) checkHitTargetAt(apiCtx context.Context, point Position)
 		return false, err
 	}
 
+	// Either we're done or an error happened (returned as "error:..." from JS)
+	const done = "done"
+	//nolint:forcetypeassert
 	value := result.(goja.Value)
-	switch value.ExportType().Kind() {
-	case reflect.String: // Either we're done or an error happened (returned as "error:..." from JS)
-		if value.String() == "done" {
-			return true, nil
-		}
+	if value.ExportType().Kind() != reflect.String {
+		// We got a { hitTargetDescription: ... } result
+		// Meaning: Another element is preventing pointer events.
+		return false, errorFromDOMError("error:intercept")
+	} else if value.String() != done {
 		return false, errorFromDOMError(value.String())
 	}
-	return true, nil // We got a { hitTargetDescription: ... } result
+
+	return true, nil
 }
 
 func (h *ElementHandle) checkElementState(apiCtx context.Context, state string) (*bool, error) {
@@ -296,12 +300,12 @@ func (h *ElementHandle) click(p *Position, opts *MouseClickOptions) error {
 }
 
 func (h *ElementHandle) clickablePoint() (*Position, error) {
-	var quads []dom.Quad
-	var layoutViewport *cdppage.LayoutViewport
-	var err error
-
-	action := dom.GetContentQuads().
-		WithObjectID(h.remoteObject.ObjectID)
+	var (
+		layoutViewport *cdppage.LayoutViewport
+		quads          []dom.Quad
+		err            error
+	)
+	action := dom.GetContentQuads().WithObjectID(h.remoteObject.ObjectID)
 	if quads, err = action.Do(cdp.WithExecutor(h.ctx, h.session)); err != nil {
 		return nil, fmt.Errorf("cannot request node content quads %T: %w", action, err)
 	}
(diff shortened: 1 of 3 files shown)


common/element_handle.go
```diff
@@ -253,15 +253,19 @@ func (h *ElementHandle) checkHitTargetAt(apiCtx context.Context, point Position)
...
+	if value.ExportType().Kind() != reflect.String {
+		// We got a { hitTargetDescription: ... } result
+		// Meaning: Another element is preventing pointer events.
+		return false, errorFromDOMError("error:intercept")
+	} else if value.String() != done {
 		return false, errorFromDOMError(value.String())
 	}
-	return true, nil // We got a { hitTargetDescription: ... } result
+
+	return true, nil
```
inancgumus, line 268: The fix is this.

common/element_handle.go
```diff
@@ -296,85 +300,88 @@ func (h *ElementHandle) click(p *Position, opts *MouseClickOptions) error {
...
 	}
 
-	action2 := cdppage.GetLayoutMetrics()
-	if layoutViewport, _, _, _, _, _, err = action2.Do(cdp.WithExecutor(h.ctx, h.session)); err != nil {
-		return nil, fmt.Errorf("cannot get page layout metrics %T: %w", action, err)
+	// Filter out quads that have too small area to click into.
+	var layoutViewport *cdppage.LayoutViewport
+	getLayoutMetrics := cdppage.GetLayoutMetrics()
+	if _, _, _, layoutViewport, _, _, err = getLayoutMetrics.Do(cdp.WithExecutor(h.ctx, h.session)); err != nil {
+		return nil, fmt.Errorf("cannot get page layout metrics %T: %w", getLayoutMetrics, err)
```
inancgumus, line 319: The second fix.

common/element_handle.go
```diff
@@ -296,85 +300,88 @@ func (h *ElementHandle) click(p *Position, opts *MouseClickOptions) error {
...
 		for i := 0; i < len(q); i += 2 {
 			p2 := (i + 2) % (len(q) / 2)
 			area += (q[i]*q[p2+1] - q[p2]*q[i+1]) / 2
 		}
-
-		if math.Abs(area) > 1 {
+		// Count any area above 0.99px is clickable
+		if math.Abs(area) > 0.99 {
 			filteredQuads = append(filteredQuads, q)
 		}
```
inancgumus, line 343: Nice-to-have fix.

common/element_handle.go
```diff
@@ -253,15 +253,19 @@ func (h *ElementHandle) checkHitTargetAt(apiCtx context.Context, point Position)
...
 	value := result.(goja.Value)
-	switch value.ExportType().Kind() {
-	case reflect.String: // Either we're done or an error happened (returned as "error:..." from JS)
-		if value.String() == "done" {
-			return true, nil
-		}
+	if value.ExportType().Kind() != reflect.String {
+		// We got a { hitTargetDescription: ... } result
+		// Meaning: Another element is preventing pointer events.
+		return false, errorFromDOMError("error:intercept")
```
imiric, line 263: Would it be safer to explicitly check for this result instead of assuming it's always an intercept error?

common/element_handle.go
```diff
@@ -296,85 +300,88 @@ func (h *ElementHandle) click(p *Position, opts *MouseClickOptions) error {
...
 		// https://en.wikipedia.org/wiki/Polygon#Area
-		var area float64 = 0
+		var area float64
 		for i := 0; i < len(q); i += 2 {
 			p2 := (i + 2) % (len(q) / 2)
 			area += (q[i]*q[p2+1] - q[p2]*q[i+1]) / 2
 		}
-
-		if math.Abs(area) > 1 {
+		// Count any area above 0.99px is clickable
```
imiric, line 340:
```suggestion
		// Count any area above 0.99px as clickable
```

<hr>

I can infer what the comment says from the code, but not _why_ 0.99 is better than 1. Can you expand on that?

Also, are the units here really pixels?

imiric (commented):
LGTM, just minor suggestions.

And a meta one: you don't need to detail each change in the PR description and include diffs for them, or point them out as comments; that's what commits are for :slightly_smiling_face: It's fine to just add a short overall description and a link to the relevant issue(s).

inancgumus:
> LGTM, just minor suggestions.

Thanks for your review.

common/element_handle.go
inancgumus: But it's always an intercept error :)

imiric: Is that based on your tests or on some documentation? In general it's better to be explicit, but I don't mind it, just checking.

common/element_handle.go
inancgumus: Yes, it's in pixels. The viewport is in pixels (what we compare to), and we're also returning a `Position` in pixels.

Do we want to click on something less than a pixel? :)

common/element_handle.go
inancgumus: It's based on the JS code we use ;)

https://github.com/grafana/xk6-browser/blob/3ad187ce3f3cd07fd572bb504f8a69d28cf29d4f/common/js/injected_script.js#L279-L312

common/element_handle.go
inancgumus: Better?

```suggestion
		// We don't want to click on something less than a pixel.
```

common/element_handle.go
inancgumus: Better?

```suggestion
	if value.ExportType().Kind() != reflect.String {
		// We got a { hitTargetDescription: ... } result
		// Meaning: Another element is preventing pointer events.
		//
		// It's safe to count an object return as an interception.
		// We just don't interpret what is intercepting with the target element
		// because we don't need any more functionality from this JS function
		// right now. 
		return false, errorFromDOMError("error:intercept")
```

imiric: OK, clear.

common/element_handle.go
```diff
@@ -296,85 +300,88 @@ func (h *ElementHandle) click(p *Position, opts *MouseClickOptions) error {
...
+
+func filterQuads(viewportWidth, viewportHeight int64, quads []dom.Quad) (*Position, error) {
 	var filteredQuads []dom.Quad
 	for _, q := range quads {
 		nq := q
 		for i := 0; i < len(q); i += 2 {
-			nq[i] = math.Min(math.Max(q[i], 0), float64(clientWidth))
-			nq[i+1] = math.Min(math.Max(q[i+1], 0), float64(clientHeight))
+			nq[i] = math.Min(math.Max(q[i], 0), float64(viewportWidth))
+			nq[i+1] = math.Min(math.Max(q[i+1], 0), float64(viewportHeight))
```
imiric, line 331: Do you know if this actually does anything? `nq` is never used after this. Does it also modify `q`?

common/element_handle.go
imiric: I see. I suppose if it's an area the units would be `px²` :smile:, but sure, that's fine.

inancgumus: The "pixel" word in "We don't want to click on something less than a pixel." is `px².` It's clear from the context, isn't it? Otherwise, we'd need to have a 1d-mouse 😄

common/element_handle.go
inancgumus: Yep... They are different slices with the same backing array. Good catch, nevertheless. Eagle eyes you have :)

That algorithm makes the points within the viewport and positive.

imiric (approved):
Changing the 0.99 comment and that other one you mentioned would be good, otherwise LGTM.

Issue #279, opened 2022-03-16
Upgrade to Go 1.19


inancgumus:
Go 1.19 is released: The [announcement](https://go.dev/blog/go1.19) and [release notes](https://go.dev/doc/go1.19).

Tasks:
- [x] Update the modules: `go get -u && go mod tidy`
- [x] Replace `interface{}` with `any` for less code: `gofmt -r 'interface{} -> any' -w **/*.go`
- [x] Fix the linter issues
- [x] Bump Go version in [workflows/all.yaml](https://github.com/grafana/xk6-browser/blob/main/.github/workflows/all.yaml)
  - [x] Bump prev to 1.17 and current to 1.18

Example:
```diff
- func(_ context.Context, opts evalOptions, jsFunc string, args ...interface{}) (interface{}, error) {
+ func(_ context.Context, opts evalOptions, jsFunc string, args ...any) (any, error) {
```

<details>
<summary>Go versions used by our dependencies:</summary>

```shell
1.5 github.com/josharian/intern
1.9 github.com/golang/protobuf
1.9 google.golang.org/protobuf
1.12 github.com/andybalholm/brotli
1.12 github.com/gorilla/websocket
1.12 github.com/mailru/easyjson
1.12 github.com/mattn/go-isatty
1.12 github.com/oxtoacart/bpool
1.12 github.com/tidwall/gjson
1.13 github.com/PuerkitoBio/goquery
1.13 github.com/fatih/color
1.13 github.com/hashicorp/go-multierror
1.13 github.com/mattn/go-colorable
1.13 github.com/sirupsen/logrus
1.13 github.com/spf13/afero
1.13 github.com/stretchr/testify
1.14 github.com/chromedp/cdproto
1.14 github.com/dop251/goja
1.14 google.golang.org/grpc
1.15 github.com/chromedp/sysutil
1.15 github.com/klauspost/compress
1.15 github.com/tidwall/match
1.15 google.golang.org/genproto
1.16 github.com/andybalholm/cascadia
1.16 github.com/tidwall/pretty
1.17 github.com/grafana/xk6-browser
1.17 golang.org/x/crypto
1.17 golang.org/x/net
1.17 golang.org/x/sys
1.17 golang.org/x/text
1.17 go.k6.io/k6
```

</details>

---

# Update:

One of the linters is not yet ready for Go 1.19 (due to generics):

```log
[...]

PR #280, opened 2022-03-16
Update Go from 1.17 to 1.18

inancgumus:
This was a 10-minutes work, and I was curious whether it would work on CI.

* I tested the changes with our tests and example scripts. No issues.
* Renamed `interface{}` to `any` for less code.
* Note: This also updates other dependencies like k6 v0.37.0 from k6 v0.36.1.


imiric (changes requested):
Yeah, like mentioned [here](https://github.com/grafana/xk6-browser/issues/279#issuecomment-1069133638), we really don't need to do this now.

Issue #272, opened 2022-03-10
`Page.Goto()` times out in CI tests


inancgumus:
Related: #254: Data race in TestWaitForFrameNavigationWithinDocument.

We started seeing frequent timeouts from the same test after fixing the data race issue (possible 🤞).

See: https://github.com/grafana/xk6-browser/issues/254#issuecomment-1060385306 and https://github.com/grafana/xk6-browser/issues/254#issuecomment-1063818639.

Issue #271, opened 2022-03-10
Separate JS layer from logic


inancgumus:
See https://github.com/grafana/xk6-browser/pull/268#discussion_r822566026.

imiric:
This is also important to avoid data races from sharing `goja.Runtime`. See https://github.com/grafana/xk6-browser/issues/254#issuecomment-1060380139 and https://github.com/grafana/xk6-browser/pull/263#pullrequestreview-905621976. The quickest relief would be to remove it from `ElementHandle.eval()`.

inancgumus:
> quickest relief would be to remove it from ElementHandle.eval()

What's on your mind doing so? :)

imiric:
Well, we currently pass `goja.Value`s to this function, and to `ExecutionContext.eval()`, and do the conversion inside it:

https://github.com/grafana/xk6-browser/blob/3ad187ce3f3cd07fd572bb504f8a69d28cf29d4f/common/execution_context.go#L180

https://github.com/grafana/xk6-browser/blob/3ad187ce3f3cd07fd572bb504f8a69d28cf29d4f/common/helpers.go#L55

Instead, this should happen much higher up in the stack, as close as possible or inside the methods we already expose to JS, via this JS->Go translation layer. This way we should(?) ensure a single goroutine has access to the `goja.Runtime`, and then when used internally we can worry about our own data races.

inancgumus:
+1 That is exactly what I had in my mind: Goja should not enter into the extension realm:

```
                      Test script
                           │
                           │
┌──────────────────────────▼────────────────────────────┐
│            Exposed methods the test script            │
└──────────────────────────┬────────────────────────────┘
                           │
┌──────────────────────────▼────────────────────────────┐
│      JS -> Goja -> Extension translation layer        │
└──────────────────────────┬────────────────────────────┘
                           │
┌──────────────────────────┴────────────────────────────┐
│                          ▼                            │
│                                                       │
│               Extension logic (Goja-free)             │
│                                                       │
│                                                       │
│                          │                            │
└──────────────────────────┼────────────────────────────┘
                           │
                           │
             ┌─────────────┼─────────────┐
             │             │             │
             │             ▼             │
             │                           │
             │                           │
             │          Browser          │
             │                           │
             │                           │
             │                           │
             │                           │
             └───────────────────────────┘
```

inancgumus:
Moved this to the v0.9.0 milestone to help the planning, but we can decide otherwise in the meeting.

PR #281, opened 2022-03-16
WIP Fix frame navigation timeout

imiric:
This is a _very_ WIP fix for #272, that evolved into a larger refactor of some of the navigation methods.

The commits are very messy, but here's the gist of the changes:

- #272 happens because of a race condition where we sometimes receive the `EventFrameNavigation` before we have a chance to run `waitForEvent`, i.e. what you pointed out [here](https://github.com/grafana/xk6-browser/issues/272#issuecomment-1066804247).

  The fix is relatively straightforward: instead of using `waitForEvent()` (which we should probably deprecate because of issues like this), call `FrameSession.navigateFrame` concurrently while setting up `createWaitForEventHandler`, see [`Frame.navigate`](https://github.com/grafana/xk6-browser/blob/49052ed7f3387205cda33e05517a3f91b24d47b4/common/frame.go#L939).

  This doesn't eliminate the race condition entirely, but does make it very unlikely, and I haven't run into the timeout since that change. In theory we could setup the wait before calling `FrameSession.navigateFrame` to completely avoid it, but because of the refactors I did, it would be a bit cumbersome.

The refactors include:

- To simplify, instead of waiting for `EventFrameNavigation` for same-document and cross-document navigations, wait for a single event by blocking until we get the `documentID` (confusingly [also referred to as `loaderID`](https://github.com/grafana/xk6-browser/blob/12eeeb59caa75cc2818172b096debf1aea153848/common/request.go#L70-L73)).

- To DRY and simplify, have a single `waitForNavigation` implementation and reuse it where possible. There are still places where we repeat it (e.g. `Page.Reload`), but we might want to eventually incorporate that scenario as well.

- Move the frame related navigation methods from `FrameManager` to `Frame`. I'm not sure why this was on `FrameManager` to begin with, they really belong on `Frame`. Though I'm concerned if the change of using `Frame.ctx` instead of `FrameManager.ctx` might have something to do with the current failures...

- Initial attempt at doing #271, where the public methods accessible from JS are responsible for doing the JS->Go conversion (for e.g. options), and then we use internal methods to avoid depending on `goja.Runtime`. Though we still use `k6Throw()` liberally, so not sure how much of an improvement this is.

Anyway, sorry for the wall of text and messy draft PR. I'm opening it as I'm struggling with the current `TestBrowserContextOptionsExtraHTTPHeaders` failure, where for some reason `resp.Body()` throws `error fetching response body: No data found for resource with given identifier (-32000)`. And there's another issue with `TestPageGoto` where `resp` is sometimes nil. I've only left `TestBrowserContextOptionsExtraHTTPHeaders` enabled in CI with `-count=100` so we can stress test it, but it also happens when running the full suite. Once I fix that one, I'll enable `TestPageGoto`.

I'd appreciate if you could take a look and see if I missed something obvious, but if we can't figure it out, I'm tempted to fix the #272 race with a simple fix and forget about the refactor for now. Thanks!


common/helpers.go
```diff
@@ -142,26 +142,25 @@ func createWaitForEventHandler(
...
-			case <-evCancelCtx.Done():
-				return
-			case ev := <-chEvHandler:
-				if stringSliceContains(events, ev.typ) {
-					if predicateFn != nil {
-						if predicateFn(ev.data) {
-							ch <- ev.data
-						}
-					} else {
-						ch <- nil
```
inancgumus, line 156: Any reason for removing this loop here? This function may receive other events as well (besides the ones `predicateFn` is looking for)? If so, might it return if it finds an irrelevant event first and misses the target event that the `predicateFn` is looking for?

inancgumus (commented):
Agree on moving `Frame` related stuff from `FrameManager` into `Frame`. Other than a comment I made in the changes, I couldn't see anything problematic at this moment.

common/helpers.go
imiric: Hhmm you're right, nice catch :+1: My initial thought was that it returned in both cases, so the loop wasn't needed, but on closer look it definitely should keep running :facepalm:

Thanks for looking into it!

inancgumus: Sure thing, glad I could help :)

inancgumus:
> I'm opening it as I'm struggling with the current `TestBrowserContextOptionsExtraHTTPHeaders` failure, where for some reason `resp.Body()` throws `error fetching response body: No data found for resource with given identifier (-32000)`

From the logs, it seems like the browser is getting closed before we had a chance to grab the response (or both are happening at the same time?). IDK whether this is the issue.

```
time="2022-03-17T10:23:07Z" level=debug msg="bctxid:34083D60BA7B851D889E36A598562099" category="BrowserContext:Close" elapsed="0 ms" goroutine=20
time="2022-03-17T10:23:07Z" level=debug msg="bctxid:34083D60BA7B851D889E36A598562099" category="Browser:disposeContext" elapsed="1 ms" goroutine=20
time="2022-03-17T10:23:07Z" level=debug msg="wsURL:\"ws://127.0.0.1:39045/devtools/browser/d6bcbc04-e1a3-4d13-abfb-381b935beaad\" method:\"Target.disposeBrowserContext\"" category="connection:Execute" elapsed="0 ms" goroutine=20
time="2022-03-17T10:23:07Z" level=debug msg="url:http://127.0.0.1:38429/get method:GET rid:09B7B63E4EDBB36D93ADB471F2F61773" category="Response:fetchBody" elapsed="13 ms" goroutine=123
```

imiric:
Yeah, it's something with that, but I haven't found what's causing the premature close yet. I'll keep looking into it, don't spend more time on this.

imiric:
So the issue with `TestBrowserContextOptionsExtraHTTPHeaders` turned out not to be because of a premature close, but the fact that _sometimes_ we were calling `Response.Body()` in the test before we had received the `Network.loadingFinished` event.

The thing is that `waitForNavigation()` returns an `api.Response`, and the user should expect that it's fully usable and that not only navigation happened, but that the response was fully received as well. Since our API is still fully sync (eventually `waitForNavigation()` should return a Promise, but I'm not doing that as well in this PR :sweat_smile:), we have no other option but to block until both `Page.frameNavigated` _and_ `Network.loadingFinished` are received.

The fix I just pushed is embarassingly hacky, but I'll clean this up next week, and move on to the `Page.Goto()` failure where we sometimes get a nil `Response`, which still happens judging from [the previous CI failure](https://github.com/grafana/xk6-browser/actions/runs/2005565337/attempts/1). :disappointed:

imiric:
So it turned out to be rather difficult to come up with a fix for #272 that handles all edge cases correctly and doesn't introduce other issues. Since we don't have extensive tests for this functionality, I'm not confident that we can rely on any fix even if it doesn't break existing tests.

To fix the current `TestBrowserContextOptionsExtraHTTPHeaders` failure, in 4b2c3b83 I added a hack to wait for [`Network.loadingFinished`](https://chromedevtools.github.io/devtools-protocol/tot/Network/#event-loadingFinished), which is when the response is fully received and we can get its body, but that approach has a major issue: it fails for same-doc, internal (`about:blank`) and data URI (`data:text/html,hello`) navigations, since no responses will be received in those cases, and the returned `api.Response` from `waitForNavigation` will be nil. The thing is that `Page.frameNavigated` might be recieved before or after `Network.loadingFinished` depending on the response size, but we currently have no way of knowing whether it's a same-doc or cross-doc navigation until we get `NavigationEvent`. The value of `NavigationEvent.newDocument` is also racy since it depends on the received lifecycle events (this is why we sometimes get a nil `Response`...), but that's a rabbit hole I briefly looked at but didn't go all the way into. :sweat_smile:

Waiting on `Page.loadingFinished` is also not necessary, since this should already be possible by waiting on the `load` or `domcontentloaded` lifecycle events that can be passed as `waitForNavigation` options. But alas, this is racy as well. :disappointed:

So tldr; since this is only an issue we've seen in CI tests (which doesn't discard us working on it; it's still an issue that can happen to users), I'm putting this on hold for a bit. At least to focus on other issues in this cycle, and I can get back to it later, or if someone else has time to dig into it, be my guest! :pray:

I'm of the opinion that this entire internal event based system is inherently racy with CDP events, which causes many issues we've seen, and likely others that haven't been reported yet. [I proposed a rearchitecture](https://github.com/grafana/xk6-browser/discussions/142#discussioncomment-2416943) to address it and would love to read your comments.

imiric:
I'm closing this as the fix for #272 was done differently in #480, and this won't be merged in its current state. Though I'll try to take some refactoring ideas from here for an upcoming PR.

PR #283, opened 2022-03-25
Fix keyboard press of special keys

### LINKED ISSUE #239: `ElementHandle.Press()` panic: Invalid 'text' parameter
On current `main` (917b83e) the following test:

```go
func TestElementHandlePress(t *testing.T) {
	tb := newTestBrowser(t)
	p := tb.NewPage(nil)
	p.SetContent(`<input type="text">`, nil)
	el := p.Query("input")
	el.Press("Backspace", nil)
}
```

... panics with:

```
--- FAIL: TestElementHandlePress (0.34s)
panic: cannot handle element key ("Backspace") press: cannot do key down: cannot execute dispatch key event down: Invalid 'text' parameter (-32602) [recovered]
        panic: cannot handle element key ("Backspace") press: cannot do key down: cannot execute dispatch key event down: Invalid 'text' parameter (-32602)

goroutine 9 [running]:
testing.tRunner.func1.2({0x17ac1e0, 0xc0001c3530})
        /home/ivan/.local/share/go/src/testing/testing.go:1209 +0x36c
testing.tRunner.func1()
        /home/ivan/.local/share/go/src/testing/testing.go:1212 +0x3b6
panic({0x17ac1e0, 0xc0001c3530})
        /home/ivan/.local/share/go/src/runtime/panic.go:1047 +0x266
go.k6.io/k6/js/common.Throw(0xc0006a5c88, {0x19c9680, 0xc000121ad0})
        /home/ivan/Projects/grafana/xk6-browser/vendor/go.k6.io/k6/js/common/util.go:36 +0x79
github.com/grafana/xk6-browser/common.k6Throw({0x19e6a60, 0xc000692060}, {0x1810dc4, 0x28}, {0xc0006a5e10, 0x2, 0x2})
        /home/ivan/Projects/grafana/xk6-browser/common/helpers.go:214 +0x22f
github.com/grafana/xk6-browser/common.(*ElementHandle).Press(0xc00017aa80, {0x17e6cee, 0x5}, {0x0, 0x0})
        /home/ivan/Projects/grafana/xk6-browser/common/element_handle.go:1169 +0x3d6
github.com/grafana/xk6-browser/tests.TestElementHandlePress(0x0)
        /home/ivan/Projects/grafana/xk6-browser/tests/element_handle_test.go:40 +0xd7
testing.tRunner(0xc000183380, 0x183cbd0)
        /home/ivan/.local/share/go/src/testing/testing.go:1259 +0x230
created by testing.(*T).Run
        /home/ivan/.local/share/go/src/testing/testing.go:1306 +0x727
FAIL    github.com/grafana/xk6-browser/tests    0.411s
```

From initial investigation, it seems like `text` shouldn't be set to `key` in some cases, as done here:
https://github.com/grafana/xk6-browser/blob/917b83ef14b3e2075f92c092163c47e359883a6a/common/keyboard.go#L198-L201

imiric:
Previously pressing some keys like Backspace, Delete, ArrowLeft and AltGraph would panic because we were sending an unsupported `text` argument in the [`Input.dispatchKeyEvent` command](https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchKeyEvent). The fix simply does this only for single character keys, and replicates [what Puppeteer does](https://github.com/puppeteer/puppeteer/blob/7659bf92e6f2e4efe8edef6d326cbeca358a995b/src/common/Input.ts#L159).

I added some tests for currently missing functionality that Playwright supports, and will create issues for fixing them (see #284 and #285).

Fixes #239

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,6 +15,11 @@ issues:
   # When true it disables revive's comment warnings, and we want them enabled.
   exclude-use-default: false
 
+  exclude-rules:
+    - linters:
+        - funlen
+      source: "^func Test"
+
 linters:
   disable-all: true
   enable:
--- a/tests/keyboard_test.go
+++ b/tests/keyboard_test.go
@@ -1,0 +1,64 @@
+package tests
+
+import (
+	_ "embed"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/xk6-browser/common"
+	"github.com/grafana/xk6-browser/keyboardlayout"
+)
+
+func TestKeyboardPress(t *testing.T) {
+	tb := newTestBrowser(t)
+
+	t.Run("all_keys", func(t *testing.T) {
+		p := tb.NewPage(nil)
+		cp, ok := p.(*common.Page)
+		require.True(t, ok)
+		kb := cp.Keyboard
+		layout := keyboardlayout.GetKeyboardLayout("us")
+
+		assert.NotPanics(t, func() {
+			for k := range layout.Keys {
+				kb.Press(string(k), nil)
+			}
+		})
+	})
+
+	t.Run("backspace", func(t *testing.T) {
+		p := tb.NewPage(nil)
+		cp, ok := p.(*common.Page)
+		require.True(t, ok)
+		kb := cp.Keyboard
+
+		p.SetContent(`<input>`, nil)
+		el := p.Query("input")
+		p.Focus("input", nil)
+
+		kb.Type("Hello World!", nil)
+		require.Equal(t, "Hello World!", el.InputValue(nil))
+
+		kb.Press("Backspace", nil)
+		assert.Equal(t, "Hello World", el.InputValue(nil))
+	})
+
+	t.Run("newline", func(t *testing.T) {
+		p := tb.NewPage(nil)
+		cp, ok := p.(*common.Page)
+		require.True(t, ok)
+		kb := cp.Keyboard
+
+		p.SetContent(`<textarea>`, nil)
+		el := p.Query("textarea")
+		p.Focus("textarea", nil)
+
+		kb.Type("Hello", nil)
+		kb.Press("Enter", nil)
+		kb.Press("Enter", nil)
+		kb.Type("World!", nil)
+		assert.Equal(t, "Hello\n\nWorld!", el.InputValue(nil))
+	})
+}
--- a/tests/keyboard_test.go
+++ b/tests/keyboard_test.go
@@ -2,6 +2,7 @@ package tests
 
 import (
 	_ "embed"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -45,6 +46,32 @@ func TestKeyboardPress(t *testing.T) {
 		assert.Equal(t, "Hello World", el.InputValue(nil))
 	})
 
+	t.Run("combo", func(t *testing.T) {
+		t.Skip("FIXME") // See https://github.com/grafana/xk6-browser/issues/285
+		p := tb.NewPage(nil)
+		cp, ok := p.(*common.Page)
+		require.True(t, ok)
+		kb := cp.Keyboard
+
+		p.SetContent(`<input>`, nil)
+		el := p.Query("input")
+		p.Focus("input", nil)
+
+		kb.Press("Shift+KeyA", nil)
+		kb.Press("Shift+b", nil)
+		kb.Press("C", nil)
+		kb.Press("d", nil)
+		require.Equal(t, "AbCd", el.InputValue(nil))
+
+		metaKey := "Control"
+		if runtime.GOOS == "darwin" {
+			metaKey = "Meta"
+		}
+		kb.Press(metaKey+"+A", nil)
+		kb.Press("Delete", nil)
+		assert.Equal(t, "", el.InputValue(nil))
+	})
+
 	t.Run("newline", func(t *testing.T) {
 		p := tb.NewPage(nil)
 		cp, ok := p.(*common.Page)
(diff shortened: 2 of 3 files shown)


tests/keyboard_test.go
```diff
@@ -0,0 +1,111 @@
...
+		p.SetContent(`<input type="textarea">`, nil)
+		el := p.Query("input")
+		p.Focus("input", nil)
+
+		kb.Type("Hello", nil)
+		kb.Press("Enter", nil)
+		kb.Press("Enter", nil)
+		kb.Type("World!", nil)
+		assert.Equal(t, "Hello\n\nWorld!", el.InputValue(nil))
+	})
```
imiric, line 110: Strange... This test actually works as expected in JS, but for some reason fails in Go.

This works fine:

```javascript
page.goto('https://mdbootstrap.com/docs/standard/extended/textarea/', { waitUntil: 'networkidle' });
const el = page.$('#textAreaExample1');
el.focus();

page.keyboard.type('Hello');
page.keyboard.press('Enter')
page.keyboard.press('Enter')
page.keyboard.type('World!')

check(el, {
  'value': el.inputValue() == 'Hello\n\nWorld!',
});
```

But this Go test fails with:

```
expected: "Hello\n\nWorld!"
actual  : "HelloWorld!"
```

Any ideas why? :thinking:

olegbespalov: @imiric actually the `https://mdbootstrap.com/docs/standard/extended/textarea/` uses `textarea` and in our test we use an `input` 

```go
p.SetContent(`<input type="textarea">`, nil)
```

It seems like if we switch to it, the test will pass

.golangci.yml
```diff
@@ -15,6 +15,11 @@ issues:
   # When true it disables revive's comment warnings, and we want them enabled.
   exclude-use-default: false
 
+  exclude-rules:
+    - linters:
+        - funlen
```
olegbespalov, line 20:
```suggestion
        - funlen
        - forcetypeassert
```

WDYT? Instead of using `nolint` every time

tests/keyboard_test.go
imiric: :facepalm: Thanks, I overlooked that and for some reason thought they would be equivalent.

I'm wondering if there's any value in the test, but I guess it doesn't hurt leaving it.

olegbespalov: >I'm wondering if there's any value in the test, but I guess it doesn't hurt leaving it.

Well, when I checked this locally, it seems like this is the first and the only usage of `Enter,` so since it's an important key, I'd say that this test is an important one for now at least :sweat_smile:

.golangci.yml
imiric: Hhmm I think the linter is useful in general, it's just that in these cases we know it must be this type, so I'd rather `nolint` it.

We could also just do:
```golang
cp, _ := p.(*common.Page)
kb := cp.Keyboard
```

Or to be safe:
```golang
cp, ok := p.(*common.Page)
require.True(t, ok)
kb := cp.Keyboard
```

But that's just noise for tests.

I don't have a strong preference, but the `nolint` doesn't bother me. :man_shrugging: Let me know if you strongly prefer another approach and I'll change it.

olegbespalov: I considered that this linter is essential for the application code, but in the tests, I can't imagine the cases for that.

The issue with the `nolint` is that it's also a kind of code smell, so if there is a simple workaround like you mentioned before, it's better to use it to avoid `nolint`.

Regarding the suggested options, I'll vote for the safer one (with the usage of `require`). Why? Simply because even if this happens (during refactoring or so), we'll save some time for the future generations of the developers (or us) because they don't need investigating that `p` is not more the `*common.Page` or read panic output (as it is right now).

tests/keyboard_test.go
imiric: Resolved in latest push (032f45b).

.golangci.yml
imiric: Sure, a couple more lines are worth avoiding the panic. Resolved in latest force push (hope you don't mind, wanted to keep this clean and ready for merging).

olegbespalov: Jus a small nitpick :see_no_evil: 

If we're changing tests to using `require.True(t, ok)` I believe there should not be `forcetypeassert` in the `exclude-rules`

imiric: Yeah, it's not, I didn't want to disable it globally.

olegbespalov: :man_facepalming: still not here with the Github UI. I was looking for my suggestion and thought that it was committed. Shame on me :sweat_smile:

olegbespalov (approved):

inancgumus (commented):
LGTM. The following test didn't work:

```go
t.Run("backspace_then_enter", func(t *testing.T) {
	p := tb.NewPage(nil)
	cp, ok := p.(*common.Page)
	require.True(t, ok)
	kb := cp.Keyboard

	p.SetContent(`<input>`, nil)
	el := p.Query("input")
	p.Focus("input", nil)

	kb.Type("Hello World!", nil)
	require.Equal(t, "Hello World!", el.InputValue(nil))

	kb.Press("Backspace", nil)
	kb.Press("Backspace", nil)
	kb.Press("Backspace", nil)
	kb.Press("Backspace", nil)
	kb.Press("Backspace", nil)
	kb.Press("Backspace", nil)
	kb.Press("Backspace", nil)
	kb.Press("Enter", nil)
	kb.Type("World!", nil)
	assert.Equal(t, "Hello\nWorld!", el.InputValue(nil))
})
```

Output:

```
Error:      	Not equal: 
				expected: "Hello\nWorld!"
				actual  : "HelloWorld!"
```

imiric:
@inancgumus That test doesn't work because it's using an `input` element, which doesn't allow newlines. If you change it to `textarea` it works.

I had the same confusion a few weeks ago :smile:

inancgumus:
Haha, ok :)

PR #286, opened 2022-03-29
Upgrade k6 dependency, use new JS API

imiric:
This drops all usage of context-based getter functions (removed from k6 in https://github.com/grafana/k6/commit/64f17c7e8dbdf283ed4729bf0c2967eb278071ad) in favor of methods on the VU instance passed to the extension. We attached this VU instance to the internal context in #268 to keep a similar API, but in the future this should be refactored to avoid the use of context.

It also uses `k6Throw()` in a lot of places instead of `k6common.Throw()` to avoid depending on `goja.Runtime` by the caller and refactors the way `k6modulestest.VU` is initialized. See the commits for details.


common/frame.go
```diff
@@ -1340,27 +1313,26 @@ func (f *Frame) SetContent(html string, opts goja.Value) {
 		forceCallable: true,
 		returnByValue: true,
 	}
+	rt := f.vu.Runtime()
 	if _, err := f.evaluate(f.ctx, utilityWorld, eopts, rt.ToValue(js), rt.ToValue(html)); err != nil {
-		k6common.Throw(rt, err)
+		k6Throw(f.ctx, "error setting content: %w", err)
```
olegbespalov, line 1318: That is a nice one, maybe "wrap" with the details in other places (I see that you did it in some places :muscle: ) in this file? 

I mean to avoid this:
```go
k6Throw(f.ctx, "%s", err).
``` 

or maybe, as the minimal change, make it more flexible with:

```go
k6Throw(f.ctx, "%w", err).
``` 

What do you think?

olegbespalov (commented):
In general looks good, left a minor comment

common/context.go
```diff
@@ -78,7 +78,7 @@ func WithVU(ctx context.Context, vu k6modules.VU) context.Context {
 // GetVU returns the attached k6 VU instance from ctx, which can be used to
 // retrieve the goja runtime and other k6 objects relevant to the currently
 // executing VU.
-// See https://github.com/grafana/k6/blob/v0.38.0/js/initcontext.go#L168-L186
+// See https://github.com/grafana/k6/blob/v0.37.0/js/initcontext.go#L168-L186
```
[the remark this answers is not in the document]
imiric: Yeah, I forgot to do this in a separate commit.

I added the URL before `v0.37.0` was released so I changed it manually, but I messed up the version :smile:

common/frame.go
imiric: Heh I started doing this in some places, but gave up as it was a lot of work to come up with appropriate messages :sweat_smile: Plus I think @inancgumus and me disagree on the phrasing (he tends to use "unable to" and "cannot" whereas I think "error" is clearer), so I didn't want to add more discussion points.

Re: `%s` vs `%w`, sure, I'll change it, though there's no need to wrap these since `k6Throw` doesn't unwrap them and simply panics with the error, so `%s` would be fine as well.

common/context.go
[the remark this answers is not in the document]
olegbespalov: No, all good :smile: In that case, eyes are just "nice catch" :sweat_smile:, no further actions required

common/frame.go
imiric: Did the minimal change in 934d913.

olegbespalov (approved):
:muscle: :crossed_fingers:

inancgumus (commented):
LGTM. However, this new `mockVU` type is noisy. I'd prefer using the following:

```go
vu := newMockVU(t)
fm := NewFrameManager(vu.ctx, nil, nil, nil, log)
opts := vu.V(map[string]interface{}{ // or verbosely: vu.toGojaValue
	"timeout":   "1000",
	"waitUntil": "networkidle",
})
vu.state.TLSConfig = ...
```

Instead of:

```go
mockVU := newMockVU(t)
fm := NewFrameManager(mockVU.CtxField, nil, nil, nil, log)
opts := mockVU.RuntimeField.ToValue(map[string]interface{}{
	"timeout":   "1000",
	"waitUntil": "networkidle",
})
state = mockVU.StateField
state.TLSConfig = ...
```

inancgumus:
LGTM. The new `mockVu` type is much better than before. Otherwise, vu-related code gets noisy fast and fogs the test itself. I'm on the side of keeping it concise.

So I'd like to use it as follows:

```go
vu := newMockVU(t)
fm := NewFrameManager(vu.ctx, nil, nil, nil, log)
opts := vu.V(map[string]interface{}{ // or verbosely: vu.toGojaValue
	"timeout":   "1000",
	"waitUntil": "networkidle",
})
vu.state.TLSConfig = ...
```

Instead of:

```go
mockVU := newMockVU(t)
fm := NewFrameManager(mockVU.CtxField, nil, nil, nil, log)
opts := mockVU.RuntimeField.ToValue(map[string]interface{}{
	"timeout":   "1000",
	"waitUntil": "networkidle",
})
state = mockVU.StateField
state.TLSConfig = ...
```

PR #288, opened 2022-04-01
Fix negative metric emission

### LINKED ISSUE #191: Negative `browser_first_meaningful_paint` metric
While testing #117 with the script from [this comment](https://github.com/grafana/xk6-browser/pull/117#issuecomment-1002521620) on `main` (54cd293), the `browser_first_meaningful_paint` metric was reported with negative `avg`, `min` and `med` values:

```
browser_first_meaningful_paint...: avg=-671428500ns min=-2167962000ns med=-671428500ns max=825.1ms  p(90)=525.79ms p(95)=675.45ms
```

The JSON output shows the following:

```javascript
{"type":"Point","data":{"time":"2022-01-14T12:07:03.583311545+01:00","value":-2167.962,"tags":{"group":"","scenario":"default","url":"<snipped>"}},"metric":"browser_first_meaningful_paint"}
{"type":"Point","data":{"time":"2022-01-14T12:07:16.499621629+01:00","value":825.105,"tags":{"group":"::Click Continue (Recommendations)","scenario":"default","url":"<snipped>"}},"metric":"browser_first_meaningful_paint"}
```

imiric:
This issue happened when receiving `Page.lifecycleEvent`s for different frames which overwrote the global `FrameSession.initTime` value on which the metric value calculation was being done.

The fix is simple: store the `init` time on `Frame` instead and calculate the elapsed time per frame, since the metric emission is done per frame anyway. Unfortunately this exposes another issue worked around by 47d9e82, but should ultimately be fixed by fixing the racy way CDP events are handled; some discussion about this [here](https://github.com/grafana/xk6-browser/discussions/142).

I tried to explain both behaviors in [this comment](https://github.com/grafana/xk6-browser/issues/191#issuecomment-1086016760).

This also does some refactoring to DRY things.

I didn't add any tests since the issue is only reproducible (~1/10 runs) with the script from [this comment](https://github.com/grafana/xk6-browser/pull/117#issuecomment-1002521620), which tests a very complex site, and reproducing it in an integration or unit test would be very difficult, and likely flaky since we would have to depend on sleep. That said, I've run the script dozens of times and couldn't reproduce the issue anymore. **Please test it yourself and report any weirdness.**

Closes #191

### DIFF
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -28,7 +28,6 @@ import (
 	"runtime"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/chromedp/cdproto"
 	"github.com/chromedp/cdproto/browser"
@@ -67,8 +66,6 @@ type FrameSession struct {
 	targetID target.ID
 	windowID browser.WindowID
 
-	initTime *cdp.MonotonicTime
-
 	// To understand the concepts of Isolated Worlds, Contexts and Frames and
 	// the relationship betwween them have a look at the following doc:
 	// https://chromium.googlesource.com/chromium/src/+/master/third_party/blink/renderer/bindings/core/v8/V8BindingDesign.md
@@ -100,7 +97,6 @@ func NewFrameSession(
 		parent:               parent,
 		manager:              p.frameManager,
 		targetID:             tid,
-		initTime:             &cdp.MonotonicTime{},
 		contextIDToContextMu: sync.Mutex{},
 		contextIDToContext:   make(map[cdpruntime.ExecutionContextID]*ExecutionContext),
 		isolatedWorlds:       make(map[string]bool),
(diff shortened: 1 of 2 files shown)


olegbespalov (approved):
So, I finally managed to do tests locally. It seems to work. However, it's essential to notice that even if I run the test script on the `main` branch, I faced no issue :man_shrugging: 

So :crossed_fingers:

imiric:
Thanks @olegbespalov. Indeed, it's difficult to reproduce, but it happens for me every 1/10 or 1/20 runs.

common/frame_session.go
```diff
@@ -700,82 +696,34 @@ func (fs *FrameSession) onLogEntryAdded(event *log.EventEntryAdded) {
...
-		fs.session.ID(), fs.targetID, event.FrameID, event.Name)
+		"sid:%v tid:%v fid:%v event:%s eventTime:%q",
+		fs.session.ID(), fs.targetID, event.FrameID, event.Name, event.Timestamp.Time())
 
-	state := fs.vu.State()
-	if event.Name == "init" || event.Name == "commit" {
-		fs.initTime = event.Timestamp
+	frame := fs.manager.getFrameByID(event.FrameID)
+	if frame == nil {
+		return
```
inancgumus, line 704: It would help later if we add a log statement here that says `frame` was nil.

inancgumus (commented):
LGTM. I read all your previous comments, and your current approach seems the simplest and most effective one for now. Also, nice way of using maps to DRY code.

PR #298, opened 2022-04-22
Move `Page` and `ElementHandle` tests into a single file

imiric:
Just a minor `tests` refactoring. There's no reason there should be a separate test file for each method, and the reasoning behind for which there should or shouldn't be isn't clear.

Also fix some linter issues.


inancgumus (approved):
Makes sense, LGTM!

tests/element_handle_test.go
```diff
@@ -1,12 +1,271 @@
...
+}
+
+func TestElementHandleBoundingBoxSVG(t *testing.T) {
+	tb := newTestBrowser(t)
+	p := tb.NewPage(nil)
+
+	p.SetContent(`
+        <svg xmlns="http://www.w3.org/2000/svg" width="500" height="500">
+            <rect id="theRect" x="30" y="50" width="200" height="300"></rect>
+        </svg>`, nil)
```
inancgumus, line 73: I would do the htmls as following (there are others as well in the PR):

```suggestion
	p.SetContent(`
		<svg xmlns="http://www.w3.org/2000/svg" width="500" height="500">
			<rect id="theRect" x="30" y="50" width="200" height="300"></rect>
		</svg>`,
	nil)
```

I know this PR is not about changing the existing tests. But, while on it, why not?

inancgumus (commented):
One minor suggestion.

tests/element_handle_test.go
imiric: :man_shrugging: Sure, changed in fe52721. I didn't see other instances of this, and I wouldn't want to change this for cases that fit on a single line. Since this used spaces instead of tabs, it makes some sense.

Also, let's not bikeshed about style issues, please.

inancgumus: Yeah, it was just a suggestion, not a requirement. Sorry.

PR #303, opened 2022-05-06
Do `waitForFunction` predicate truthy check inside `injected_script.js`

imiric:
This avoids having to export `continuePolling` and doing the check from the wrapper function in Go, since it's really a JS implementation detail. It also hopefully avoids [the need for a comment explaining internal details of injected_script](https://github.com/grafana/xk6-browser/pull/294/files#r866750362).


inancgumus (approved):
Makes sense!

PR #315, opened 2022-05-18
Refactor k6 VU and tests

inancgumus:
This PR refactors the VU usage in the extension.

* `newMockVU` was duplicated. This commit moves it into a new package named `k6`.
* Replaces `mockVU` with `vu` for brevity.
* Instead of crowding the common package (which is already big), moves the handling of VUs in contexts to the k6 package. And testing into the k6test package.
* Provides helper methods for the functions we need to use frequently.

Removes `rt`, `state`, and `samples` fields from the test browser.

Renamings:
* `newMockVU(t)` -> `k6.NewMockVU(t)`
* `mockVu` -> `vu`
* `mockVU.RuntimeField.ToValue` -> `vu.ToGojaValue`
* `GetVU(ctx).Runtime()` -> `k6.Runtime(ctx)`
* `vu.CtxField` -> `vu.Context()`
* `common.WithVU` -> `k6.WithVU`
* `common.GetVU` -> `k6.GetVU`
* `test_context.go mockVU/newMockVU` -> `k6test.VU` `k6test.NewVU`
* `common/helpers_test.go mockVU/newMockVU` -> `k6test.VU` `k6test.NewVU`
* `tests/test_browser.go runtime|toGojaValue`

PR #316, opened 2022-05-18
Fix gci linter config

inancgumus:
The current (ac82660fe746dcab8b819d53b94389086e591f9f) `local` setting for the gci linter is defunct. There is no local setting for the gci linter (see [an example good config](https://github.com/daixiang0/gci/blob/73cde2312b7818a42a260f2d49085e3a3cd456a0/pkg/gci/internal/testdata/already-good.cfg.yaml)).

This change makes each Go file to import as follows:
* The stdlib pkgs
* xk6-browser browser pkgs
* k6 pkgs
* The other 3rd-party packages.

To make the change work, I also updated the linter version to v1.46.2. We should merge this PR after merging #315 (or better, this PR first :)). In #315, I explained that gci was problematic. The cause was the incorrect config.

---

The reference command to fix imports:

```bash
$ gci write -s Standard -s "Prefix(github.com/grafana/xk6-browser)" -s "Prefix(go.k6.io)" -s Default . 
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,4 +1,4 @@
-# v1.44.0
+# v1.46.2
 # Please don't remove the first line. It's used in CI to determine the
 # golangci-lint version. See the lint step in .github/workflows/all.yaml
 run:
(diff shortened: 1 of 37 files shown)


ankur22 (approved):

inancgumus:
Thanks for the review @ankur22 ❤️

Issue #311, opened 2022-05-16
locator.click([options])


inancgumus:
The [`locator.click([options])`](https://playwright.dev/docs/api/class-locator#locator-click) method clicks on an element using the locator's selector with the `strict` option.

[Locators are strict](https://playwright.dev/docs/locators#strictness). All operations on locators that imply some target DOM element will throw an exception if more than one element matches the given selector.

PR #318, opened 2022-05-20
Add locator.click([options])

inancgumus:
This PR adds the ability to click on an element with strict mode via `Locator`. In the test, that is why it expects a panic when the test tries to select multiple link elements. The test also verifies whether the locator can click on an element using the saved selector in the locator.

Pitfalls: Although `locator.Click` forwards the click call to `frame.Click`, I wished there was a way to set the incoming goja value's `strict` property to `true`, instead of (re)parsing the options in `locator.Click`. Do let me know if you know a way to do so.

Closes: #311

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -717,24 +717,28 @@ func (f *Frame) ChildFrames() []api.Frame {
 func (f *Frame) Click(selector string, opts goja.Value) {
 	f.log.Debugf("Frame:Click", "fid:%s furl:%q sel:%q", f.ID(), f.URL(), selector)
 
-	parsedOpts := NewFrameClickOptions(f.defaultTimeout())
-	err := parsedOpts.Parse(f.ctx, opts)
-	if err != nil {
+	popts := NewFrameClickOptions(f.defaultTimeout())
+	if err := popts.Parse(f.ctx, opts); err != nil {
 		k6Throw(f.ctx, "%w", err)
 	}
-
-	fn := func(apiCtx context.Context, handle *ElementHandle, p *Position) (interface{}, error) {
-		return nil, handle.click(p, parsedOpts.ToMouseClickOptions())
-	}
-	actFn := f.newPointerAction(
-		selector, DOMElementStateAttached, parsedOpts.Strict, fn, &parsedOpts.ElementHandleBasePointerOptions,
-	)
-	_, err = callApiWithTimeout(f.ctx, actFn, parsedOpts.Timeout)
-	if err != nil {
+	if err := f.click(selector, popts); err != nil {
 		k6Throw(f.ctx, "%w", err)
 	}
-
 	applySlowMo(f.ctx)
+}
+
+func (f *Frame) click(selector string, opts *FrameClickOptions) error {
+	click := func(apiCtx context.Context, handle *ElementHandle, p *Position) (interface{}, error) {
+		return nil, handle.click(p, opts.ToMouseClickOptions())
+	}
+	act := f.newPointerAction(
+		selector, DOMElementStateAttached, opts.Strict, click, &opts.ElementHandleBasePointerOptions,
+	)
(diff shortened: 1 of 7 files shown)


common/frame.go
```diff
@@ -737,24 +717,28 @@ func (f *Frame) ChildFrames() []api.Frame {
...
 
-	parsedOpts := NewFrameClickOptions(f.defaultTimeout())
-	err := parsedOpts.Parse(f.ctx, opts)
-	if err != nil {
+	popts := NewFrameClickOptions(f.defaultTimeout())
+	if err := popts.Parse(f.ctx, opts); err != nil {
+		k6Throw(f.ctx, "%w", err)
+	}
+	if err := f.click(selector, popts); err != nil {
 		k6Throw(f.ctx, "%w", err)
```
ankur22, line 725: Not a fan of this naming tbh, it looks like it's a wrapper around panic. Can't do much about this though

common/frame.go
```diff
@@ -737,24 +717,28 @@ func (f *Frame) ChildFrames() []api.Frame {
...
 	}
+	applySlowMo(f.ctx)
+}
 
-	fn := func(apiCtx context.Context, handle *ElementHandle, p *Position) (interface{}, error) {
-		return nil, handle.click(p, parsedOpts.ToMouseClickOptions())
+func (f *Frame) click(selector string, opts *FrameClickOptions) error {
```
ankur22, line 730: If you exported this method (and renamed it too), couldn't you call this method from locator's `click` method?

common/locator.go
```diff
@@ -23,3 +25,26 @@ func NewLocator(ctx context.Context, selector string, f *Frame, l *Logger) *Loca
...
+	}
+	if err = l.click(copts); err != nil {
+		return
+	}
+}
+
+// click is like Click but takes parsed options and neither throws an
+// error, or applies slow motion.
+func (l *Locator) click(opts *FrameClickOptions) error {
+	opts.Strict = true
```
ankur22, line 48: Using `locators` forces the user to work in `strict` mode?

common/frame.go
inancgumus: What do you have in mind? Tell me more :)

common/locator.go
inancgumus: Yep, that's the whole idea. Well, mostly.

common/frame.go
inancgumus: We might discuss it in a new issue if you have some other ideas? 💡

ankur22: When i saw `k6Throw`, i thought "Go doesn't have throw and catch like Java", and so I think naming it as `k6Panic` might be a little better. If someone new to Go sees this then they might get confused and expect to be able to `catch` errors... which is possible with `recover` but something we should avoid to reduce the complexity

common/frame.go
ankur22: Oh, never mind, I've just noticed that the call from `locators.go` is already calling this unexported method `return l.frame.click(l.selector, opts)`.

ankur22 (approved):

common/frame.go
inancgumus: That makes a lot of sense. Would you like to open it up in a new discussion? :)

common/frame.go
inancgumus: I'll tell you more about it in the next meeting :)

common/frame.go
ankur22: Yeah sure, what's the process for starting a new discussion?

inancgumus: Here: https://github.com/grafana/xk6-browser/discussions ;)

inancgumus: @ankur22, I made a refactoring according to your suggestion, see: https://github.com/grafana/xk6-browser/pull/319

PR #319, opened 2022-05-20
Refactor k6 facilities

inancgumus:
This PR refactors the k6-related code into the k6 package for a better organization:

* Moves `k6Throw` to `k6.Panic`
* Moves `common.WithProcessID` and friends to `k6` pkg.
* Move `CustomK6Metrics` and friends to `k6.CustomMetrics`.

Inspired by this comment: https://github.com/grafana/xk6-browser/pull/318#discussion_r877971730


k6/panic.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+	"fmt"
+	"os"
+
+	k6common "go.k6.io/k6/js/common"
+)
+
+// Panic throws a k6 error, and before throwing the error, it finds the
+// browser process from the context and kills it if it still exists.
+// TODO: test.
```
ankur22, line 13: I think it will be useful to refactor the comment a bit, maybe something along the lines of:

```suggestion
// Panic will cause a panic with the given error which will shut
// the application down. Before panicking, it will find the
// browser process from the context and kill it if it still exists.
// TODO: test.
```

k6/panic.go
```diff
@@ -0,0 +1,37 @@
...
+	p, err := os.FindProcess(pid)
+	if err != nil {
+		// optimistically return and don't kill the process
+		return
+	}
+	// no need to check the error for waiting the process to release
+	// its resources or whether we could kill it as we're already
+	// dying.
+	_ = p.Release()
+	_ = p.Kill()
```
ankur22, line 36: I personally feel that all errors should be handled even during shutdown, we can at least log those errors if there are any which could help us diagnose unknown issues

ankur22 (approved):

k6/panic.go
inancgumus: Make sense, but the thing is, we don't have a logger at the stage. Do you have a suggestion? Or can you make another PR?

ankur22: Yeah, it would be good to revisit logging and find a simpler way of passing it around... possibly through the context or still passed in when we call `New...`.

In my previous place we placed the logger in the context, and the context was always the first param of any function/method so it was easy to log with the correct context/meta data.

PR #323, opened 2022-05-24
Fix Chrome data store directory is not cleaned up 

### LINKED ISSUE #278: Chrome data store directory is not cleaned up
Currently (12eeeb5) the temporary Chrome data storage directory is not cleaned up after each test run.

This is especially visible when running tests:

```sh
$ find /tmp -type d -name 'xk6-browser-data-*' | wc -l
0
$ GOMAXPROCS=3 go test -p 3 -v -race -count=1 ./...
[...]
$ find /tmp -type d -name 'xk6-browser-data-*' | wc -l
40
```

This doesn't seem like a regression and is the way it's always worked, at least going back to v0.1.1.

## Suggested solution

To avoid leaving files behind, the extension should by default remove this directory when the test is finished. This should be easier now after the recent [`DataStore` refactor](https://github.com/grafana/xk6-browser/blob/12eeeb59caa75cc2818172b096debf1aea153848/chromium/data_store.go#L43-L54).

Unless this behavior is wanted, in which case we might consider an env var to control it (`XK6_BROWSER_CLEANUP=0`?) and maybe a new `Browser` option.

Integration tests might be good to have as well, since the current `DataStore` unit tests missed this.

ankur22:
This fix resolves https://github.com/grafana/xk6-browser/issues/278 in majority of times the extension is ran. As outlined in the issue itself, the problem is:

1. The cleanup of the temp dir is done too early, but it does work;
2. Another process (chrome ran with `exec`) creates the temp dir if it's not present -- the cleanup occurs prior to the `exec` being called.

The fix is:

1. To move the cleanup call from `BrowserType` to the `Browser` downstream process and call it from `Close`.
2. To try to call the `cleanupFunc` when the context is closed or the browser sub process exits.

There is an edge case that remains which is that the cleanup is not guaranteed because `k6` can shut down at any point during runtime, and currently there is no way for `k6` to wait for extensions to terminate cleanly. This issue is being tracked in https://github.com/grafana/k6/issues/4446.


chromium/browser_type.go
```diff
@@ -177,7 +177,7 @@ func (b *BrowserType) allocate(
 	opts *common.LaunchOptions,
 	flags map[string]interface{},
 	env []string,
-) (_ *common.BrowserProcess, rerr error) {
+) (_ *common.BrowserProcess, rErr error) {
```
inancgumus: I'd keep this as it was, as it's a common convention in Go to name short error variables as `rerr` instead of `rErr`. For example, see https://github.com/golang/go/search?q=rerr.

inancgumus (commented):
Nice work, I haven't finished my review yet :) I'll do so today.

By the way, you can squash [this commit](https://github.com/grafana/xk6-browser/pull/323/commits/3be1ce2c8045166c6adb032cf746d48e3150e073).

chromium/browser_type.go
ankur22: 👍

imiric (changes requested):
Nice work! Initially I liked the approach of doing this in `Browser.close()`, but after thinking about it some more, I think we need to ensure that the directory is cleaned up regardless if the script calls `close()` or not. It might be because of a script error or a browser crash, but we should ensure the directory is removed in all of these cases if we created it initially. This is because it could potentially have sensitive information we shouldn't leave lying around, especially when run in the cloud.

The only cases where it should remain is if the k6 process itself crashes, in which case there's nothing we can do outside of an external cleanup process, or if the user sets a new option or env var to keep it.

tests/browser_test.go
```diff
@@ -50,6 +52,23 @@ func TestBrowserNewPage(t *testing.T) {
...
+	defer func() {
+		err = os.Unsetenv("TMPDIR")
+		assert.NoError(t, err)
+	}()
+
+	b := newTestBrowser(t)
+	_ = b.NewPage(nil)
+
+	_, err = os.Stat(tmpDirPath + "xk6-browser-data-*")
+	assert.True(t, errors.Is(err, os.ErrNotExist), tmpDirPath+"xk6-browser-data-* should have been deleted")
```
inancgumus, line 69: This test does not seem to check whether the temporary directory is removed.

I just returned `nil` from the `DataStore.Cleanup` method, and it still worked 😊

common/browser_test.go
```diff
@@ -21,7 +21,7 @@ func TestBrowserNewPageInContext(t *testing.T) {
 	}
 	newTestCase := func(id cdp.BrowserContextID) *testCase {
 		ctx := context.Background()
-		b := newBrowser(ctx, nil, nil, NewLaunchOptions(), NewNullLogger())
+		b := newBrowser(ctx, nil, nil, NewLaunchOptions(), NewNullLogger(), nil)
```
inancgumus: I think the cleanup func should be a field on the browser instead of enlarging the already large input parameters list. WDYT?

chromium/browser_type.go
```diff
@@ -133,26 +134,34 @@ func (b *BrowserType) Launch(opts goja.Value) api.Browser {
...
 	}
 
+	logger, err := makeLogger(b.Ctx, launchOpts)
+	if err != nil {
+		k6common.Throw(rt, fmt.Errorf("cannot make logger: %w", err))
+	}
+
+	flags := prepareFlags(launchOpts, &state.Options)
+
+	cleanupFunc, err := b.createTempDir(flags, logger)
```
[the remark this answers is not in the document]
ankur22: Yeah, i've kind of done that. Now we pass it a `dirPath`, and it returns the `finalDirPath`, so the flag is read and set outside of `createTempDir`

inancgumus (changes requested):
Nice! I like to see ditching more bugs 🐞!

A little nitpick: Could you do refactorings in one commit, and fixes in another one? As it was kind of difficult for me to review the PR :) There were others, but, as an example, extracting to `createTempDir` was refactoring and could be done in its own commit.

ankur22:
> Nice! I like to see ditching more bugs 🐞!
> 
> A little nitpick: Could you do refactorings in one commit, and fixes in another one? As it was kind of difficult for me to review the PR :) There were others, but, as an example, extracting to `createTempDir` was refactoring and could be done in its own commit.

Yep, that makes sense. Please just keep point this out when i miss it 😅 I'll learn eventually #BadHabits

datastore/data_store_test.go
```diff
@@ -1,4 +1,4 @@
-package chromium
+package datastore
```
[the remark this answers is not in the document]
ankur22: Yeah, sounds good

tests/browser_test.go
ankur22: Just realised that `os.Stat` doesn't work with patterns so now working with `filepath.Glob` which does

inancgumus: Great! Please let me know by re-requesting my review when you finish the whole PR so I can look at the whole thing in one Go.

chromium/browser_type.go
```diff
@@ -173,37 +195,52 @@ func (b *BrowserType) Name() string {
...
+		return "", nil, fmt.Errorf("cannot make temp data directory: %w", err)
+	}
+
+	finalDirPath = b.storage.Dir
+
+	m := sync.Mutex{}
+	return finalDirPath, func() {
+		m.Lock()
+		defer m.Unlock()
+		if err := b.storage.Cleanup(); err != nil {
```
imiric, line 213: What is this mutex protecting against? Concurrent calls to `Cleanup()`? What would actually fail if that were to happen?

I tried removing it and calling this lambda concurrently, and couldn't reproduce a failure. But if it can fail, then maybe it should be an implementation detail in `Cleanup()` instead?

chromium/browser_type.go
```diff
@@ -134,26 +136,46 @@ func (b *BrowserType) Launch(opts goja.Value) api.Browser {
...
+	flags["user-data-dir"] = finalDirPath
+
+	go func(ctx context.Context) {
+		defer cleanupFunc()
+		// There's a small chance that this might be called
+		// if the context is closed by the k6 runtime. To
+		// guarantee the cleanup we would need to orchestrate
+		// it correctly which https://github.com/grafana/k6/issues/2432
+		// will enable once it's complete.
+		<-ctx.Done()
```
imiric, line 159: Good point. From my tests it seemed to be called pretty consistently, as there's apparently enough time between [when the context is cancelled](https://github.com/grafana/k6/blob/b725f03be5873b9b0d933b94a83924ff7c1f15e4/js/runner.go#L746-L748) and k6 shutdown for `cleanupFunc()` to finish running. And it worked fine with multiple iterations and VUs. But it's probably a "works fine on my machine" scenario :sweat_smile:

tests/browser_test.go
```diff
@@ -50,6 +52,26 @@ func TestBrowserNewPage(t *testing.T) {
...
+	}()
+
+	b := newTestBrowser(t, withSkipClose())
+	p := b.NewPage(nil)
+	p.Close(nil)
+	b.Close()
+
+	matches, err := filepath.Glob(tmpDirPath + "xk6-browser-data-*")
+	assert.NoError(t, err)
+	assert.Empty(t, matches, "a file/dir exists still which matches the pattern `xk6-browser-data-*`")
```
imiric, line 72: Can you also do a check before calling `b.Close()` that the temp dir does in fact exist?

imiric (commented):
Nice job with this! :clap: And sorry again it was more complicated than we originally thought, but hopefully it was a good learning experience for you. :sweat_smile:

Unfortunately, as you explained in that comment, I don't think we can do better than this until https://github.com/grafana/k6/issues/2432 is implemented. From my local tests it does get rid of the temp directory consistently, even without calling `Browser.close()`, but of course, we can't guarantee this will happen all the time. I'm fine with having both solutions in place until we can add a single one after 2432 is implemented. And for Cloud tests, we'll still have to add an external cleanup to make sure it's removed even if the process crashes for some reason.

Can you please squash and rebase this on `main`, and address the minor comments I had? I'll take another look when that's done.

common/browser_test.go
ankur22: Agreed, it does seem a bit too long. I think this and a few other files need to be abstracted and refactored a bit 😅 

I think for now I'd prefer to leave it as an input param on `NewBrowser` and `newBrowser` since to me the `cleanupFunc` is a required field (even though there's a check for `nil` if it wasn't assigned, and not really needed in tests but only during normal runtime).

If you have a reason why it shouldn't be a field param, I'd be happy to change it though.

chromium/browser_type.go
ankur22: It is protecting an implementation detail within `Cleanup()`, so i will move it into `Cleanup()`.

Odd that you weren't able to find the race condition when running it concurrently. I ran all the tests with the `--race` flag to help me spot it. It's because of:

```
	if d.fsRemoveAll == nil {
		d.fsRemoveAll = os.RemoveAll
	}

        return d.fsRemoveAll(d.Dir)
```

I could just make it:

```
	return os.RemoveAll(d.Dir)
```

I don't know the context behind why we're doing the way we're doing it.

storage/storage.go
```diff
@@ -0,0 +1,63 @@
...
+// Dir manages data storage for the extension and user specific data
+// on the local filesystem.
+type Dir struct {
+	Dir    string // path to the data storage directory
+	remove bool   // whether to remove the temporary directory in cleanup
+
+	// FS abstractions
+	fsMkdirTemp       func(dir, pattern string) (string, error)
+	fsRemoveAll       func(path string) error
+	fsRemovalAllMutex sync.Mutex
```
inancgumus, line 20: You mean `fsRemoveAllMutex`—`Removal`? It's a little style thing, but you can shorten `Mutex` to `Mu`. It's a widely accepted convention, so people can see what's it about.

chromium/browser_type.go
imiric: Ah, I see.

> I don't know the context behind why we're doing the way we're doing it.

I'm not sure about it myself. In general, it's a good idea to abstract away FS function calls so that you can mock them out in tests and avoid touching the FS (or use a virtual FS like afero), but from what I've seen we're not using them in our tests, or anywhere else.

@inancgumus originally wrote this, so WDYT? Should we remove `fsMkdirTemp` and `fsRemoveAll` and just use `os` directly?

inancgumus: Can we use `sync.Once` here since `fsRemoveAll` and friends are initialized only in the constructor.

ankur22: Oh, good point on virtual filesystems. I will give that a go as that will make things easier to deal with if the tests fails. Currently, if the directory removal test fails, it will cause future runs of the tests to fail as it only deletes the directory that's created in the current run.

I will use `sync.Once` to init the functions.

chromium/browser_type.go
ankur22: Created an issue to track the fix to make deletion deterministic i.e. use the fix from grafana/k6#2432: grafana/xk6-browser#403

chromium/browser_type.go
ankur22: I will work on this refactor in issue grafana/xk6-browser#404

imiric (approved):

inancgumus (approved):

Issue #284, opened 2022-03-25
Keyboard `Up`/`Down` doesn't hold key


imiric:
On current `main` (12eeeb5) the following test fails:

https://github.com/grafana/xk6-browser/blob/db7a4ceff10f68fa93a950b577ff290a49806277/tests/keyboard_test.go#L45-L70

I'm not sure if this is an issue with the `Keyboard.pressedKeys` implementation, only with the Shift key and keeping the selection, or something else.

PR #325, opened 2022-05-26
Fix hold modifier key down and up

ankur22:
This fixes https://github.com/grafana/xk6-browser/issues/284

The modifier key (in the original skipped test it is shift), wasn't
being set correctly as it was using the AND bitwise operator against
0, which would always be 0. It also doesn't need the NOT operator when
press is called as we're not trying to negate away from the modifier.


imiric (approved):
Nice catch! Love 1-line fixes :smile:

PR #326, opened 2022-05-26
Add combo key press

### LINKED ISSUE #285: Pressing keys in combination is not supported
On current `main` (12eeeb5) the following test:

https://github.com/grafana/xk6-browser/blob/61fb0c26e29d1b95209ab1080e49f08875b8e8d6/tests/keyboard_test.go#L72-L94

... fails with:
```
panic: cannot press key: cannot do key down: "Shift+KeyA" is not a valid key for layout "us" [recovered]
        panic: cannot press key: cannot do key down: "Shift+KeyA" is not a valid key for layout "us"

goroutine 67 [running]:
testing.tRunner.func1.2({0x1850c60, 0xc00059fef0})
        /home/ivan/.local/share/go/src/testing/testing.go:1209 +0x36c
testing.tRunner.func1()
        /home/ivan/.local/share/go/src/testing/testing.go:1212 +0x3b6
panic({0x1850c60, 0xc00059fef0})
        /home/ivan/.local/share/go/src/runtime/panic.go:1047 +0x266
go.k6.io/k6/js/common.Throw(0xc0005d7cc0, {0x1a7c8c0, 0xc0005ba520})
        /home/ivan/Projects/grafana/xk6-browser/vendor/go.k6.io/k6/js/common/util.go:36 +0x79
github.com/grafana/xk6-browser/common.k6Throw({0x1a9a780, 0xc00057c090}, {0x189cd1d, 0x14}, {0xc0005d7de8, 0x1, 0x1})
        /home/ivan/Projects/grafana/xk6-browser/common/helpers.go:214 +0x22f
github.com/grafana/xk6-browser/common.(*Keyboard).Press(0xc0003561e0, {0x188ec4a, 0xa}, {0x0, 0x0})
        /home/ivan/Projects/grafana/xk6-browser/common/keyboard.go:91 +0x1e5
github.com/grafana/xk6-browser/tests.TestKeyboardPress.func4(0x0)
        /home/ivan/Projects/grafana/xk6-browser/tests/keyboard_test.go:81 +0x12e
testing.tRunner(0xc0001f21a0, 0xc0001ee140)
        /home/ivan/.local/share/go/src/testing/testing.go:1259 +0x230
created by testing.(*T).Run
        /home/ivan/.local/share/go/src/testing/testing.go:1306 +0x727
```

This is more of a feature request than a bug, since [Playwright supports this](https://playwright.dev/docs/api/class-keyboard).

ankur22:
This fixes https://github.com/grafana/xk6-browser/issues/285

Shift+X when passed to press() and used in conjunction with down() and up() now will change the casing of the character, as long as the character is in the format `KeyX`. A key press that is `x` or `X` will ignore the shift modifier key. This fix matches the playwright behaviour.

E.g.

```
Shift+KeyA -> A
KeyA -> a
A -> A
a -> a
Shift+A -> A
Shift+a -> a
```

The Shift modifier will only change the case of a character if KeyA - KeyZ is used.

### DIFF
--- a/common/keyboard.go
+++ b/common/keyboard.go
@@ -23,6 +23,7 @@ package common
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/grafana/xk6-browser/api"
@@ -88,7 +89,8 @@ func (k *Keyboard) Press(key string, opts goja.Value) {
 	if err := kbdOpts.Parse(k.ctx, opts); err != nil {
 		k6ext.Panic(k.ctx, "parsing keyboard options: %w", err)
 	}
-	if err := k.press(key, kbdOpts); err != nil {
+
+	if err := k.comboPress(key, kbdOpts); err != nil {
 		k6ext.Panic(k.ctx, "pressing key: %w", err)
 	}
 }
@@ -190,9 +192,13 @@ func (k *Keyboard) keyDefinitionFromKey(key keyboardlayout.KeyInput) keyboardlay
 		srcKeyDef, ok = k.layout.KeyDefinition(key)
 	}
 	// Try to find with the shift key value
+	// e.g. for `@`, the shift modifier needs to
+	// be used.
+	var foundInShift bool
 	if !ok {
 		srcKeyDef = k.layout.ShiftKeyDefinition(key)
 		shift = k.modifiers | ModifierKeyShift
+		foundInShift = true
 	}
 
 	var keyDef keyboardlayout.KeyDefinition
@@ -217,7 +223,18 @@ func (k *Keyboard) keyDefinitionFromKey(key keyboardlayout.KeyInput) keyboardlay
 	if srcKeyDef.Text != "" {
 		keyDef.Text = srcKeyDef.Text
 	}
-	if shift != 0 && srcKeyDef.ShiftKey != "" {
+	// Shift is only used on keys which are `KeyX`` (where X is
+	// A-Z), or on keys which require shift to be pressed e.g.
+	// `@`, and shift must be pressed as well as a shiftKey
+	// text value present for the key.
+	// Not all keys have a text value when shift is pressed
+	// e.g. `Control`.
+	// When a key such as `2` is pressed, we must ignore shift
+	// otherwise we would type `@`.
+	isKeyXOrOnShiftLayerAndShiftUsed := (strings.HasPrefix(string(key), "Key") || foundInShift) &&
+		shift != 0 &&
+		srcKeyDef.ShiftKey != ""
+	if isKeyXOrOnShiftLayerAndShiftUsed {
 		keyDef.Key = srcKeyDef.ShiftKey
 		keyDef.Text = srcKeyDef.ShiftKey
 	}
@@ -242,13 +259,56 @@ func (k *Keyboard) modifierBitFromKeyName(key string) int64 {
 	return 0
 }
 
+func (k *Keyboard) comboPress(keys string, opts *KeyboardOptions) error {
+	if opts.Delay > 0 {
+		if err := wait(k.ctx, opts.Delay); err != nil {
+			return err
+		}
+	}
+
+	kk := split(keys)
+	for _, key := range kk {
+		if err := k.down(key); err != nil {
+			return fmt.Errorf("cannot do key down: %w", err)
+		}
+	}
+
+	for i := range kk {
+		key := kk[len(kk)-i-1]
+		if err := k.up(key); err != nil {
+			return fmt.Errorf("cannot do key up: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// This splits the string on `+`.
+// If `+` on it's own is passed, it will return ["+"].
+// If `++` is passed in, it will return ["+", ""].
+// If `+++` is passed in, it will return ["+", "+"].
+func split(keys string) []string {
+	var (
+		kk = make([]string, 0)
+		s  strings.Builder
+	)
+	for _, r := range keys {
+		if r == '+' && s.Len() > 0 {
+			kk = append(kk, s.String())
+			s.Reset()
+		} else {
+			s.WriteRune(r)
+		}
+	}
+	kk = append(kk, s.String())
+
+	return kk
+}
+
 func (k *Keyboard) press(key string, opts *KeyboardOptions) error {
-	if opts.Delay != 0 {
-		t := time.NewTimer(time.Duration(opts.Delay) * time.Millisecond)
-		select {
-		case <-k.ctx.Done():
-			t.Stop()
-		case <-t.C:
+	if opts.Delay > 0 {
+		if err := wait(k.ctx, opts.Delay); err != nil {
+			return err
 		}
 	}
 	if err := k.down(key); err != nil {
@@ -260,12 +320,9 @@ func (k *Keyboard) press(key string, opts *KeyboardOptions) error {
 func (k *Keyboard) typ(text string, opts *KeyboardOptions) error {
 	layout := keyboardlayout.GetKeyboardLayout(k.layoutName)
 	for _, c := range text {
-		if opts.Delay != 0 {
-			t := time.NewTimer(time.Duration(opts.Delay) * time.Millisecond)
-			select {
-			case <-k.ctx.Done():
-				t.Stop()
-			case <-t.C:
+		if opts.Delay > 0 {
+			if err := wait(k.ctx, opts.Delay); err != nil {
+				return err
 			}
 		}
 		keyInput := keyboardlayout.KeyInput(c)
@@ -279,5 +336,19 @@ func (k *Keyboard) typ(text string, opts *KeyboardOptions) error {
 			return fmt.Errorf("inserting text: %w", err)
 		}
 	}
+	return nil
+}
+
+func wait(ctx context.Context, delay int64) error {
+	t := time.NewTimer(time.Duration(delay) * time.Millisecond)
+	select {
+	case <-ctx.Done():
+		if !t.Stop() {
+			<-t.C
+		}
+		return fmt.Errorf("%w", ctx.Err())
+	case <-t.C:
+	}
+
 	return nil
 }
--- a/tests/keyboard_test.go
+++ b/tests/keyboard_test.go
@@ -47,7 +47,32 @@ func TestKeyboardPress(t *testing.T) {
 	})
 
 	t.Run("combo", func(t *testing.T) {
-		t.Skip("FIXME") // See https://github.com/grafana/xk6-browser/issues/285
+		p := tb.NewPage(nil)
+		cp, ok := p.(*common.Page)
+		require.True(t, ok)
+		kb := cp.Keyboard
+
+		p.SetContent(`<input>`, nil)
+		el := p.Query("input")
+		p.Focus("input", nil)
+
+		kb.Press("Shift++", nil)
+		kb.Press("Shift+=", nil)
+		kb.Press("Shift+@", nil)
+		kb.Press("Shift+6", nil)
+		kb.Press("Shift+KeyA", nil)
+		kb.Press("Shift+b", nil)
+		kb.Press("Shift+C", nil)
+
+		kb.Press("Control+KeyI", nil)
+		kb.Press("Control+J", nil)
+		kb.Press("Control+k", nil)
+
+		require.Equal(t, "+=@6AbC", el.InputValue(nil))
+	})
+
+	t.Run("meta", func(t *testing.T) {
+		t.Skip("FIXME") // See https://github.com/grafana/xk6-browser/issues/424
 		p := tb.NewPage(nil)
 		cp, ok := p.(*common.Page)
 		require.True(t, ok)
@@ -59,9 +84,9 @@ func TestKeyboardPress(t *testing.T) {
 
 		kb.Press("Shift+KeyA", nil)
 		kb.Press("Shift+b", nil)
-		kb.Press("C", nil)
-		kb.Press("d", nil)
-		require.Equal(t, "AbCd", el.InputValue(nil))
+		kb.Press("Shift+C", nil)
+
+		require.Equal(t, "AbC", el.InputValue(nil))
 
 		metaKey := "Control"
 		if runtime.GOOS == "darwin" {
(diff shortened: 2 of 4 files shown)


common/keyboard.go
```diff
@@ -115,58 +116,68 @@ func (k *Keyboard) Type(text string, opts goja.Value) {
...
-	keyDef := k.keyDefinitionFromKey(keyInput)
-	k.modifiers &= ^k.modifierBitFromKeyName(keyDef.Key)
-	text := keyDef.Text
-	_, autoRepeat := k.pressedKeys[keyDef.KeyCode]
-	k.pressedKeys[keyDef.KeyCode] = true
+	for _, key := range kk {
+		keyInput := keyboardlayout.KeyInput(key)
+		if _, ok := k.layout.ValidKeys[keyInput]; !ok {
+			return fmt.Errorf("%q is not a valid key for layout %q", key, k.layoutName)
+		}
```
imiric, line 127: I think you should do the `+` split and this key validation in `press()` instead, so that you don't have to repeat it, and `down()`/`up()` can continue handling single keys. It shouldn't be valid to call `down()`/`up()` from JS on a key combo anyway; that is what `press()` is for.

Also, if I'm reading this right, it will accept combos like `A+B+C`, which don't make sense. Maybe we should add validation that there should be only one non-modifier key pressed? Or should we just assume that `A+B+C` equals `ABC`? I guess we can fallback to whatever Playwright accepts, minus the `Shift+B`/`Shift+b` behavior.

imiric (changes requested):
Nicely done, I just think we should move it to `press()` instead.

And re: the `Shift+b` behavior, I don't think Playwright makes sense in that case. `Shift+b` or `Shift+B` should produce `B` in both cases. A lowercase `b` can be typed without `Shift`. It's fine if we diverge with them to improve things. :)

common/keyboard.go
inancgumus: > And re: the `Shift+b` behavior, I don't think Playwright makes sense in that case. `Shift+b` or `Shift+B` should produce `B` in both cases. A lowercase `b` can be typed without `Shift`. It's fine if we diverge with them to improve things. :)

I'm mostly OK with what you propose @imiric, but here are my two cents.

Is diverging PW, in this case, worth it? If I were a tester, I'd at least expect similar behavior.  `Shift+KeyA+B+C` means holding down the shift key while pressing the other keys. For example: `Shift+KeyA+KeyB+B+b` is `ABBb`, which makes sense to me.

Puppeteers also work with `KeyA`, `KeyX`, etc. style.

So instead of `Shift+B`, it should be `Shift+KeyB`, not `Shift+B`. `KeyX` is also in tandem with JS key codes: https://javascript.info/keyboard-events. As I said, I'm also OK with what you proposed.

inancgumus:
> And re: the `Shift+b` behavior, I don't think Playwright makes sense in that case. `Shift+b` or `Shift+B` should produce `B` in both cases. A lowercase `b` can be typed without `Shift`. It's fine if we diverge with them to improve things. :)

I'm mostly OK with what you propose @imiric, but here are my two cents.

Is diverging PW, in this case, worth it? If I were a tester, I'd at least expect similar behavior.  `Shift+KeyA+B+C` means holding down the shift key while pressing the other keys. For example: `Shift+KeyA+KeyB+B+b` is `ABBb`, which makes sense to me.

Puppeteers also work with `KeyA`, `KeyX`, etc. style.

So instead of `Shift+B`, it should be `Shift+KeyB`, not `Shift+B`. `KeyX` is also in tandem with JS key codes: https://javascript.info/keyboard-events.

As I said, I'm also OK with what you proposed. But, I'm not sure whether we're improving things or making people's life harder 😅

inancgumus (commented):
As @imiric also noticed, splitting a given key should be done in `press`. `down` and `up` are only responsible for a single key, and that key can be a modifier as well. For more info: https://github.com/grafana/xk6-browser/pull/326/files#r885768387.

ankur22:
I'm unable to get the `Meta` and `Alt` modifier keys to work with any of the existing functions e.g.

```go
		kb.Press("Shift+KeyA", nil)
		kb.Press("Shift+b", nil)
		kb.Press("Shift+C", nil)

		metaKey := "Control"
		if runtime.GOOS == "darwin" {
			metaKey = "Meta"
		}
		kb.Press(metaKey+"+A", nil)
		kb.Press("Delete", nil)
```

OR

```go
		kb.Press("Shift+KeyA", nil)
		kb.Press("Shift+b", nil)
		kb.Press("Shift+C", nil)

		kb.Press("Alt+ArrowLeft", nil)
		kb.Press("Delete", nil)
```

This should delete the text that has been written into the input field. To continue to track and resolve this a new ticket has been opened: #?

common/keyboard.go
ankur22: Done

common/keyboard.go
```diff
@@ -89,7 +89,8 @@ func (k *Keyboard) Press(key string, opts goja.Value) {
 	if err := kbdOpts.Parse(k.ctx, opts); err != nil {
 		k6ext.Panic(k.ctx, "cannot parse keyboard options: %w", err)
 	}
-	if err := k.press(key, kbdOpts); err != nil {
+
+	if err := k.comboPress(key, kbdOpts); err != nil {
```
inancgumus, line 93: Nit: No need to put an empty line here, but your call :)

common/keyboard.go
```diff
@@ -253,6 +251,36 @@ func (k *Keyboard) modifierBitFromKeyName(key string) int64 {
...
 }
 
+func (k *Keyboard) comboPress(keys string, opts *KeyboardOptions) error {
+	if opts.Delay != 0 {
+		t := time.NewTimer(time.Duration(opts.Delay) * time.Millisecond)
+		select {
+		case <-k.ctx.Done():
+			t.Stop()
+		case <-t.C:
+		}
```
inancgumus, line 261: This timer will leak if `t.C` matches.

There are similar leaks in the code (I guess we missed them before). Maybe it's time to fix them all.

common/keyboard.go
```diff
@@ -253,6 +251,36 @@ func (k *Keyboard) modifierBitFromKeyName(key string) int64 {
...
+	}
+
+	kk := []string{"+"}
+	if keys != "+" {
```
inancgumus, line 265: I couldn't understand what is this for? Should it be `strings.Contains`?

common/keyboard.go
```diff
@@ -253,6 +251,36 @@ func (k *Keyboard) modifierBitFromKeyName(key string) int64 {
...
+		kk = strings.Split(keys, "+")
+	}
+	for _, key := range kk {
+		if err := k.down(key); err != nil {
+			return fmt.Errorf("cannot do key down: %w", err)
+		}
+	}
+
+	for i := len(kk) - 1; i >= 0; i-- {
```
inancgumus, line 274: Is there a reason why this is not a `range` as the above `down` loop?

inancgumus (changes requested):
I'll continue looking into this tomorrow, but it failed on my machine as follows:

```
--- FAIL: TestPageInputSpecialCharacters (0.68s)
    page_test.go:341: 
                Error Trace:    page_test.go:341
                Error:          Not equal: 
                                expected: "test@k6.io"
                                actual  : "test@k^>io"
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -test@k6.io
                                +test@k^>io
                Test:           TestPageInputSpecialCharacters
    page_test.go:341: 
                Error Trace:    page_test.go:341
                Error:          Not equal: 
                                expected: "<hello WoRlD \\/>"
                                actual  : "<hello WoRlD |?>"
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -<hello WoRlD \/>
                                +<hello WoRlD |?>
                Test:           TestPageInputSpecialCharacters
    page_test.go:341: 
                Error Trace:    page_test.go:341
                Error:          Not equal: 
                                expected: "¯\\_(ツ)_/¯"
                                actual  : "¯|_(ツ)_?¯"
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -¯\_(ツ)_/¯
                                +¯|_(ツ)_?¯
                Test:           TestPageInputSpecialCharacters
```

Update: CI has reported the failure as well: https://github.com/grafana/xk6-browser/runs/7114936484?check_suite_focus=true

common/keyboard.go
ankur22: It's going backwards:

a->b->c
c->b->a

common/keyboard.go
ankur22: Yeah, this is an interesting edge case, so in playwright:

1. `press("+")` -> types `+`
2. `press("++")` -> errors with `Unknown key: ""`
3. `press("+++")` -> types `++`
4. `press("Sift++")` -> types `+`

I'm trying to code the same behaviour, any ideas how this could be done?

common/keyboard.go
inancgumus: The timer creates a goroutine, but `t.Stop` is only called to stop the goroutine in the first `case`.

common/keyboard.go
```diff
@@ -310,6 +311,7 @@ func split(keys string) []string {
 	var ss string
 	justAdded := true
 	for _, s := range keys {
+		//nolint:gocritic
 		if string(s) == "+" && ss != "" {
 			kk = append(kk, ss)
 			ss = ""
```
inancgumus, line 317: The linter is right here. It's better to convert the if-else chain into a switch.

```go
for _, k := range keys {
	switch {
	case k == '+' && ss != "":
		...
...
```

And, you can use [`strings.Builder`](https://stackoverflow.com/questions/1760757/how-to-efficiently-concatenate-strings-in-go/47798475#47798475) to append runes without converting them.

common/keyboard.go
```diff
@@ -43,6 +43,7 @@ const (
 	ModifierKeyShift
 )
 
+//nolint:gochecknoglobals
```
inancgumus, line 46: You can put `keyAToKeyZ` into a function and ask the function instead of this `nolint` directive. The Go compiler will most probably inline the function.

Besides, I'm not sure whether we need the map. You can simply create a function and return `true`/`false` depending on the pattern `Key[A-Z]`.

inancgumus (changes requested):
I looked more into this. Here are my new findings.

common/keyboard.go
```diff
@@ -219,14 +249,7 @@ func (k *Keyboard) keyDefinitionFromKey(key keyboardlayout.KeyInput) keyboardlay
...
 	}
-
-	strKey := string(key)
-	if strings.ToUpper(strKey) == strKey && srcKeyDef.ShiftKey != "" {
-		keyDef.Key = srcKeyDef.ShiftKey
-		keyDef.Text = srcKeyDef.ShiftKey
-	}
-
-	if shift != 0 && srcKeyDef.ShiftKey != "" && strings.Contains(string(key), "Key") {
+	if _, ok := keyAToKeyZ[string(key)]; (ok || foundInShift) && shift != 0 && srcKeyDef.ShiftKey != "" {
```
inancgumus, line 253: Can you move this logic (`(ok || foundInShift) && shift != 0 && srcKeyDef.ShiftKey != "")`) into a closure? It's harder to see what it does.

common/keyboard.go
ankur22: Ok, as far as i can tell, the timer doesn't need to be stopped since we've read off of the channel. Although we should be reading off of the channel when we call `Stop` in the case above. Sound correct to you? Source: https://pkg.go.dev/time#NewTimer

common/keyboard.go
ankur22: Yeah it is the same for PW, here's the script I used to test this:

```js
// @ts-check
const { test, chromium } = require('@playwright/test');

test.describe('Editing', () => {
  test('Type login creds', async () => {
    const browser = await chromium.launch({ headless: false, slowMo: 500 });
    await page.goto('https://www.bing.com/');
    var elm = await page.$('input[name="q"]');

    await elm.press("+++"); // types in ++
    await elm.press("+"); // types in +
    await elm.press("++"); // will error and exit with `Unknown key: ""`

    await new Promise(resolve => setTimeout(resolve, 5000));

    await browser.close();
  });
});
```

inancgumus: Interesting, thanks!

common/keyboard.go
ankur22: Yeah sure, nice little trick 🙂

common/keyboard.go
ankur22: Nice, I hadn't realised that it was idiomatic to write them as switch statements. Source: https://go.dev/doc/effective_go#switch

common/keyboard.go
inancgumus: You're right, sorry 🤦

inancgumus: Ah, another idea. Why don't we just create a child timeout context? And block there?

```go
if opts.Delay != 0 {
	ctx, _ := context.WithTimeout(k.ctx, time.Duration(opts.Delay) * time.Millisecond)
	<-ctx.Done()
}
```

common/keyboard.go
inancgumus: Isn't it more like this?

```go
func split(keys string) []string {
	var (
		kk = make([]string, 0, len(keys))
		s  strings.Builder
	)
	for _, r := range keys {
		if r == '+' && s.Len() > 0 {
			kk = append(kk, s.String())
			s.Reset()
		} else {
			s.WriteRune(r)
		}
	}
	kk = append(kk, s.String())

	return kk
}
```

See: https://go.dev/play/p/fJoqZIMTBF5

common/keyboard.go
```diff
@@ -242,6 +292,74 @@ func (k *Keyboard) modifierBitFromKeyName(key string) int64 {
...
+			justAdded = false
+		}
+	}
+	if ss != "" || justAdded {
+		kk = append(kk, ss)
+	}
+
+	return kk
+}
+
```
inancgumus, line 362: Isn't it more like this?

```go
func split(keys string) []string {
	var (
		kk = make([]string, 0, len(keys))
		s  strings.Builder
	)
	for _, r := range keys {
		if r == '+' && s.Len() > 0 {
			kk = append(kk, s.String())
			s.Reset()
		} else {
			s.WriteRune(r)
		}
	}
	kk = append(kk, s.String())

	return kk
}
```

See: https://go.dev/play/p/fJoqZIMTBF5

ankur22: Nice! This is much easier to understand. Can you link me to the PW code please?

common/keyboard.go
imiric: That trick is not intuitive at all IMO. The previous C-style `for` was clear and straightforward.

Please let's not point out style issues in reviews, unless the reviewer can point out an objective benefit of the change.

imiric (changes requested):
Nice work :+1: Only the test comment seems like a blocker to me, the others are minor suggestions.

It's strange that meta keys don't seem to work. That's a pretty important issue, but we can tackle it later.

Also, remember to mark the PR "Ready for review" when it's done being a draft.

common/keyboard.go
```diff
@@ -228,6 +245,39 @@ func (k *Keyboard) keyDefinitionFromKey(key keyboardlayout.KeyInput) keyboardlay
 	return keyDef
 }
 
+func isKeyX(key string) bool {
+	_, ok := map[string]interface{}{
+		"KeyA": nil,
+		"KeyB": nil,
+		"KeyC": nil,
```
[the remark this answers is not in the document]
ankur22: Resolved in: 8b684872b7bdf97135eae6fffd8074c13f16d6b8

common/keyboard.go
```diff
@@ -242,6 +292,58 @@ func (k *Keyboard) modifierBitFromKeyName(key string) int64 {
...
+		t := time.NewTimer(time.Duration(opts.Delay) * time.Millisecond)
+		select {
+		case <-k.ctx.Done():
+			if !t.Stop() {
+				<-t.C
+			}
+			return context.Canceled
+		case <-t.C:
+		}
+	}
```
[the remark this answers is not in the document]
ankur22: Resolved in: 20908e3c5310b6392d96a9768fd84bc1c12b97d0

common/keyboard.go
```diff
@@ -242,6 +292,58 @@ func (k *Keyboard) modifierBitFromKeyName(key string) int64 {
...
+	return nil
+}
+
+// This splits the string on `+`.
+// If `+` on it's own is passed, it will return ["+"].
+// If `++` is passed in, it will return ["+", ""].
+// If `+++` is passed in, it will return ["+", "+"].
+func split(keys string) []string {
+	var (
+		kk = make([]string, 0, len(keys))
```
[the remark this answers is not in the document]
ankur22: Resolved in: 8ad3b446b98b3f1c69ed7a43000d8aebc56b3828

tests/keyboard_test.go
```diff
@@ -72,6 +97,95 @@ func TestKeyboardPress(t *testing.T) {
 		assert.Equal(t, "", el.InputValue(nil))
 	})
 
+	t.Run("type does not split on +", func(t *testing.T) {
+		p := tb.NewPage(nil)
+		cp, ok := p.(*common.Page)
+		require.True(t, ok)
+		kb := cp.Keyboard
```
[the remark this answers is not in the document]
ankur22: I've tried to refactor a couple of the tests to unit tests. Some of them I can't really move to unit tests even though they're kind of related to split or not being split.

Resolved in: 71ba9f56f23908eceda6bd93e13911489f93f51f

imiric (approved):
LGTM, nice work :+1:

I would just move `split()` to `common/helpers.go` and its tests to `helpers_test.go`. On second thought, it might be a uniquely keyboard helper function, so it makes sense where it is. Up to you.

The CI failures don't seem to be related to these changes, though that [`TestPageGotoWaitUntilDOMContentLoaded`](https://github.com/grafana/xk6-browser/runs/7636154407?check_suite_focus=true) is a new one I haven't seen before. :confused: Please squash this into a single commit and push it so the tests run again before merging.

PR #327, opened 2022-05-30
Fix `Page.content()`

### LINKED ISSUE #322: page.content() not working
image
page.contet() not working
 else if (env == 'perf') {
    url = 'https://www.google.com/';
  }

    headless: false,
    debug: false,
    slowMo: '500ms',
  });

  page.goto(url, { waitUntil: 'load' });
   //page.waitForSelector(Locator.userName);
   let newvalue = page.content();
   console.log(newvalue);

imiric:
The `Unexpected identifier` error was because of the `return` used outside of a function context. This seemed like a regression introduced by #293, but it also happens on `v0.2.0`, so :man_shrugging:. Casting the `Evaluate()` result to `string` would've failed anyway, so this addresses both issues.

I wanted to add a `Frame.Content()` unit test, but mocking the `ExecutionContext` became troublesome, so I just added this `Page.Content()` integration test instead.

Closes #322

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -754,7 +754,7 @@ func (f *Frame) Content() string {
 			if (document.documentElement) {
 				content += document.documentElement.outerHTML;
 			}
-			return document.documentElement.outerHTML;
+			return content;
 		}`
 
 	c := f.Evaluate(rt.ToValue(js))
--- a/common/frame.go
+++ b/common/frame.go
@@ -758,10 +758,6 @@ func (f *Frame) Content() string {
 		}`
 
 	c := f.Evaluate(rt.ToValue(js))
-	if c == nil {
-		return ""
-	}
-
 	content, ok := c.(goja.Value)
 	if !ok {
 		k6.Panic(f.ctx, "unexpected content() value type: %T", c)
--- a/common/frame.go
+++ b/common/frame.go
@@ -746,15 +746,28 @@ func (f *Frame) Content() string {
 	f.log.Debugf("Frame:Content", "fid:%s furl:%q", f.ID(), f.URL())
 
 	rt := f.vu.Runtime()
-	js := `let content = '';
-		if (document.doctype) {
-			content = new XMLSerializer().serializeToString(document.doctype);
-		}
-		if (document.documentElement) {
-			content += document.documentElement.outerHTML;
-		}
-		return content;`
-	return f.Evaluate(rt.ToValue(js)).(string)
+	js := `() => {
+			let content = '';
+			if (document.doctype) {
+				content = new XMLSerializer().serializeToString(document.doctype);
+			}
+			if (document.documentElement) {
+				content += document.documentElement.outerHTML;
+			}
+			return document.documentElement.outerHTML;
+		}`
+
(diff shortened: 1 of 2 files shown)


common/frame.go
```diff
@@ -746,15 +746,28 @@ func (f *Frame) Content() string {
...
-	return f.Evaluate(rt.ToValue(js)).(string)
+	js := `() => {
+			let content = '';
+			if (document.doctype) {
+				content = new XMLSerializer().serializeToString(document.doctype);
+			}
+			if (document.documentElement) {
+				content += document.documentElement.outerHTML;
+			}
+			return document.documentElement.outerHTML;
```
ankur22, line 757: Should this be `return content;`?

ankur22: Maybe it would be worth creating two tests that traverses both branches of the `if` statement?

imiric: Whoops, nice catch! I was experimenting with something and this slipped through. :man_facepalming:

Good idea about the test, will add.

imiric: Fixed in e311e44 and e8d78b4.

I opted for just adding a doctype to the content in the same test, as that hits both branches.

ankur22 (approved):

common/frame.go
```diff
@@ -746,15 +746,28 @@ func (f *Frame) Content() string {
...
+			if (document.documentElement) {
+				content += document.documentElement.outerHTML;
+			}
+			return content;
+		}`
+
+	c := f.Evaluate(rt.ToValue(js))
+	if c == nil {
+		return ""
+	}
```
inancgumus, line 763: Why is this `nil` check necessary here? Can it mask errors?

I was suspicious of it. Then, I removed it, and the test still passed.

tests/page_test.go
```diff
@@ -74,6 +74,18 @@ func TestPageEmulateMedia(t *testing.T) {
 	assert.True(t, res.ToBoolean(), "expected reduced motion setting to be 'reduce'")
 }
 
+func TestPageContent(t *testing.T) {
+	t.Parallel()
+
+	tb := newTestBrowser(t)
+	p := tb.NewPage(nil)
+
+	content := `<!DOCTYPE html><html><head></head><body><h1>Hello</h1></body></html>`
```
inancgumus, line 83: Is there any reason for having `<!DOCTYPE html>` here?

inancgumus (commented):
I have some questions.

tests/page_test.go
imiric: It's to test both branches of the `if` statement. See https://github.com/grafana/xk6-browser/pull/327#discussion_r884842527.

common/frame.go
imiric: It follows the same pattern as `InnerHTML()` and `InnerText()`. I suppose there are cases where it could be `nil`, so this is just a safeguard.

inancgumus: I don't think it's necessary and can't think of how it may occur. If you insist, would it be better to add a log if it's nil (error level)? Or get rid of it until a user complains so we can find why it happens?

inancgumus (approved):
One less bug 🐞!

PR #328, opened 2022-05-30
Rename k6 pkg to k6ext

inancgumus:
Due to a bug in xk6 (https://github.com/grafana/xk6/issues/45), we can't have a directory or a file with the name k6.

As discussed internally, @imiric suggested renaming the k6 pkg to k6ext: "k6ext follows the k6 convention of doing the same: errext, netext, etc., so let's stick to that.".


ankur22 (approved):

imiric (approved):
Thanks! It just needs to be rebased to fix the conflicts.

Issue #100, opened 2021-11-12
Implement `Locator` handling


robingustafsson:
Add support for `Locator`, a way to capture a pointer to DOM element(s) without attaching to an actual DOM node like `ElementHandle` does. Sort of a just-in-time version of querying the DOM and getting an `ElementHandle` back to do actions on.

**Relevant links**:
- Playwright docs: https://playwright.dev/docs/api/class-locator and https://playwright.dev/docs/api/class-page#page-locator

inancgumus:
I checked out the Locator API on Playwright, and it seems like we need to do this for every ElementHandle action, etc. It looks more involved. Maybe we should separate this issue into sub-issues and handle the main issue together? Could you also check out the PW docs and tell me WDYT, @imiric?

inancgumus:
Check out my findings of this feature in [the description](https://github.com/grafana/xk6-browser/issues/100#issue-1051701572).

inancgumus:
I'm noticing a lot of repetitions. I'll refactor this out after we upgrade to Go 1.18 (#279) and use generics/interface somehow (depending on which one is clearer).

PR #332, opened 2022-06-01
Add locator.dblclick([options])

### LINKED ISSUE #331: locator.dblclick([options])
The [`locator.dblclick([options])`](https://playwright.dev/docs/api/class-locator#locator-dblclick) method double-clicks on an element using the locator's selector with the `strict` option.

inancgumus:
This PR closes #331.

* Extracts `Frame.dblclick` from `Frame.Dblclick` so that we can use `dblclick` from `Locator.Dblclick`.
* Adds `locator.Dblclick` and a test.

### DIFF
--- a/api/locator.go
+++ b/api/locator.go
@@ -6,4 +6,6 @@ import "github.com/dop251/goja"
 type Locator interface {
 	// Click on an element using locator's selector with strict mode on.
 	Click(opts goja.Value)
+	// Dblclick double clicks on an element using locator's selector with strict mode on.
+	Dblclick(opts goja.Value)
 }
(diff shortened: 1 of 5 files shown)


imiric (approved):
LGTM, nice work!

One thing I'm not clear about regarding Locators is how they work across navigations. Is it because the selector is looked up on each Locator method call? Should we add a test for that scenario? Maybe not for every method, but at least that it happens for one.

inancgumus:
> LGTM, nice work!

Thanks ❤️ 

> One thing I'm not clear about regarding Locators is how they work across navigations. Is it because the selector is looked up on each Locator method call? Should we add a test for that scenario? Maybe not for every method, but at least that it happens for one.

They don't work across navigation as I explained [here](https://github.com/grafana/xk6-browser/issues/100):

> A locator is created with a selector for a specific frame. When you invoke one of the Locator methods, they query the frame and run the action. Since they query the frame, each call can resolve to a different DOM element.

The reason is that:

> Each locator belongs to a single frame.

🙄

PR #362, opened 2022-06-02
Add locator.Check

### LINKED ISSUE #334: locator.check([options])
The [`locator.check([options])`](https://playwright.dev/docs/api/class-locator#locator-check) method checks the element using the locator's selector with the `strict` option.

inancgumus:
This PR closes #334 .

* Extracts `Frame.check` from `Frame.Check` so that we can use `check` from `Locator.Check`.
* Adds `locator.Check` and a test.
* Renames `strict_link.html` to `locators.html` so that we can use it for all locator tests.


inancgumus:
GH is experiencing a glitch (although they say all their systems are OK). I hope it's why we get CI failures.

https://www.githubstatus.com/incidents/8xxms3q3qzgp

tests/locator_test.go
```diff
@@ -45,10 +45,36 @@ func TestLocatorDblclick(t *testing.T) {
...
+func TestLocatorCheck(t *testing.T) {
+	tb := newTestBrowser(t, withFileServer())
+	p := tb.NewPage(nil)
+	require.NotNil(t, p.Goto(tb.staticURL("/locators.html"), nil))
+
+	// Selecting a single element and checking it is OK.
+	t.Run("ok", func(t *testing.T) {
+		check := func() bool {
+			cr := p.Evaluate(tb.toGojaValue(`() => window.check`))
+			return cr.(goja.Value).ToBoolean() //nolint:forcetypeassert
```
imiric, line 65: To get rid of the `nolint` (same for other tests in this file):
```suggestion
			cb, ok := cr.(goja.Value)
			require.True(t, ok)
			return cb.ToBoolean()
```

tests/static/locators.html
```diff
@@ -8,16 +8,22 @@
...
+        window.check = false;
 
         document.querySelector('#link').addEventListener(
             'click', e => { result = true; }, false
         );
         document.querySelector('#link').addEventListener(
             'dblclick', e => { dblclick = true; }, false
         );
+        document.querySelector('#input').addEventListener(
+            'change', e => { window.check = event.currentTarget.checked; }, false
```
imiric, line 25: How come `event` works here? Shouldn't it be `e`?

imiric (approved):

tests/locator_test.go
inancgumus: Alright, I'll address this in one Go for all tests after merging `Check` and `Uncheck`.

tests/static/locators.html
inancgumus: Good catch! It works, but to not create confusion in the future, I changed it to `e`: https://github.com/grafana/xk6-browser/pull/362/commits/b1aece296a1db114735651d14dc93b7ccb1122c6

PR #365, opened 2022-06-02
Add locator.Uncheck

### LINKED ISSUE #359: locator.uncheck([options])
The [`locator.uncheck([options])`](https://playwright.dev/docs/api/class-locator#locator-uncheck) method unchecks the element using the locator's selector with the `strict` option.

inancgumus:
This PR closes #359.

* Extracts `Frame.uncheck` from `Frame.Uncheck` so that we can use `uncheck` from `Locator.Uncheck`.
* Adds `locator.Uncheck` and a test.

These other commits from #362 will disappear when we merge #362. I'll rebase this PR if necessary. I based off this branch from #362 because otherwise, a lot of conflicts happen, which is hair pulling 🤦

### DIFF
--- a/tests/locator_test.go
+++ b/tests/locator_test.go
@@ -78,3 +78,29 @@ func TestLocatorCheck(t *testing.T) {
 		)
 	})
 }
+
+func TestLocatorUncheck(t *testing.T) {
+	tb := newTestBrowser(t, withFileServer())
+	p := tb.NewPage(nil)
+	require.NotNil(t, p.Goto(tb.staticURL("/locators.html"), nil))
+
+	// Selecting a single element and checking it is OK.
+	t.Run("ok", func(t *testing.T) {
+		uncheck := func() bool {
+			cr := p.Evaluate(tb.toGojaValue(`() => window.uncheck`))
+			return cr.(goja.Value).ToBoolean() //nolint:forcetypeassert
+		}
+		input := p.Locator("#checkedInput", nil)
+		input.Uncheck(nil)
+		require.True(t, uncheck(), "could not uncheck the input box")
+	})
+	// There are two checked input boxes in the document (locators.html).
+	// The strict mode should disallow selecting multiple elements.
+	t.Run("strict", func(t *testing.T) {
+		input := p.Locator("input", nil)
+		require.Panics(t,
+			func() { input.Uncheck(nil) },
+			"should not select multiple elements",
+		)
+	})
+}
--- a/tests/locator_test.go
+++ b/tests/locator_test.go
@@ -58,15 +58,20 @@ func TestLocatorCheck(t *testing.T) {
 	p := tb.NewPage(nil)
 	require.NotNil(t, p.Goto(tb.staticURL("/locators.html"), nil))
 
-	// Selecting a single element and checking it is OK.
-	t.Run("ok", func(t *testing.T) {
-		check := func() bool {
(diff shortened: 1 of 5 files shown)


tests/static/locators.html
```diff
@@ -24,6 +27,9 @@
         document.querySelector('#input').addEventListener(
             'change', e => { window.check = event.currentTarget.checked; }, false
         );
+        document.querySelector('#checkedInput').addEventListener(
+            'change', e => { window.uncheck = !event.currentTarget.checked; }, false
+        );
```
imiric, line 32: Could you merge the `Check` and `Uncheck` tests into one, so that we use the same `checkbox` element and `window.check` property for both tests?

imiric (approved):
It's great that you got into a rhythm now with adding `Locator` methods :clap:

> These other commits from https://github.com/grafana/xk6-browser/pull/362 will disappear when we merge https://github.com/grafana/xk6-browser/pull/362. I'll rebase this PR if necessary. I based off this branch from https://github.com/grafana/xk6-browser/pull/362 because otherwise, a lot of conflicts happen, which is hair pulling

It's fine to base branches on other WIP branches, but you can also set the merge target in this PR to be the base branch instead of `main`. That will avoid the repeated commits making it easier to review, and once the base PR is merged, GH will automatically retarget this PR to `main`. :tada:

Maybe mention in the description the order they should be reviewed and merged in, though it's obvious if we go by the PR number.

inancgumus:
> It's fine to base branches on other WIP branches, but you can also set the merge target in this PR to be the base branch instead of `main`. That will avoid the repeated commits making it easier to review, and once the base PR is merged, GH will automatically retarget this PR to `main`. 🎉

Thanks, I didn't know about that :) 🙇

tests/static/locators.html
inancgumus: Thanks, makes sense 🙏

https://github.com/grafana/xk6-browser/pull/365/commits/8e7f57b509a7dd3e3e07996595f34bd9189c83d5

tests/locator_test.go
```diff
@@ -58,16 +58,21 @@ func TestLocatorCheck(t *testing.T) {
...
+	t.Run("check", func(t *testing.T) {
+		checkbox := p.Locator("#inputCheckbox", nil)
+		checkbox.Check(nil)
 		require.True(t, check(), "could not check the input box")
 	})
+	t.Run("uncheck", func(t *testing.T) {
+		checkbox := p.Locator("#inputCheckbox", nil)
+		checkbox.Uncheck(nil)
+		require.False(t, check(), "could not uncheck the input box")
+	})
```
imiric, line 75: My comment about merging check/uncheck tests was about calling both methods in the same test, one after the other. Otherwise you can't ensure that this test is actually changing the `window.check` state, and that it's not just returning the default `false` value.

inancgumus: Hmm, got it. It is similar to what you suggest. In this case, it first checks the element in the first test; since the element is checked, the 2nd test tries to uncheck it. Dependency between tests is not good, so I will change it as you suggested. Thanks for letting me know 🙂

inancgumus: And, what you suggested is also better in terms of testing the Locator behavior, so 👍  See: #367

imiric: > Dependency between tests is not good

Right, and you can't ensure the order they'll run in.

PR #369, opened 2022-06-03
Add locator.IsChecked

### LINKED ISSUE #346: locator.isChecked([options])
The [`locator.isChecked([options])`](https://playwright.dev/docs/api/class-locator#locator-is-checked) method returns `true` if the element matches the locator's selector (with strict mode on) and is checked. Otherwise, returns `false`.

inancgumus:
This PR closes #346.

* Extracts `Frame.isChecked` from `Frame.IsChecked` so that we can use `isChecked` from `Locator.IsChecked`.
* Adds `locator.IsChecked` and a test.
* Removes applying slow motion in the `Frame.IsChecked` as discussed in #234.

### DIFF
--- a/common/element_handle.go
+++ b/common/element_handle.go
@@ -585,31 +585,6 @@ func (h *ElementHandle) selectText(apiCtx context.Context) error {
 	return nil
 }
 
-func (h *ElementHandle) setChecked(apiCtx context.Context, checked bool, p *Position) error {
-	state, err := h.checkElementState(apiCtx, "checked")
-	if err != nil {
-		return err
-	}
-	if checked == *state {
-		return nil
-	}
-
-	err = h.click(p, NewMouseClickOptions())
-	if err != nil {
-		return err
-	}
-
-	state, err = h.checkElementState(apiCtx, "checked")
-	if err != nil {
-		return err
-	}
-	if checked != *state {
-		return errors.New("clicking the checkbox did not change its state")
-	}
-
-	return nil
-}
-
 func (h *ElementHandle) tap(apiCtx context.Context, p *Position) error {
 	return h.frame.page.Touchscreen.tap(p.X, p.X)
 }
@@ -731,11 +706,6 @@ func (h *ElementHandle) BoundingBox() *api.Rect {
 		return nil // Don't panic here, just return nil
 	}
 	return bbox.toApiRect()
-}
-
-// Check scrolls element into view and clicks in the center of the element.
-func (h *ElementHandle) Check(opts goja.Value) {
-	h.SetChecked(true, opts)
 }
 
 // Click scrolls element into view and clicks in the center of the element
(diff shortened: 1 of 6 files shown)


common/frame.go
```diff
@@ -770,6 +770,47 @@ func (f *Frame) uncheck(selector string, opts *FrameUncheckOptions) error {
...
+		v, err := handle.isChecked(apiCtx, 0) // Zero timeout when checking state
+		if errors.Is(err, ErrTimedOut) {      // We don't care about timeout errors here!
+			return v, nil
+		}
+		return v, err
+	}
+	act := f.newAction(
+		selector, DOMElementStateAttached, opts.Strict, isChecked, []string{}, false, true, opts.Timeout,
+	)
+	v, err := callApiWithTimeout(f.ctx, act, opts.Timeout)
```
imiric, line 801: If the timeout is not wanted when checking state, should we still pass `opts.Timeout` to `newAction` and `callApiWithTimeout`? Does it make sense to call `callApiWithTimeout` at all?

This seems like an awful lot of code for what can essentially be done in a single line with `Page.evaluate()`... I wonder if we need all of this. :thinking:

common/frame.go
```diff
@@ -770,6 +770,47 @@ func (f *Frame) uncheck(selector string, opts *FrameUncheckOptions) error {
...
+	}
+
+	return checked
+}
+
+func (f *Frame) isChecked(selector string, opts *FrameIsCheckedOptions) (bool, error) {
+	isChecked := func(apiCtx context.Context, handle *ElementHandle) (interface{}, error) {
+		v, err := handle.isChecked(apiCtx, 0) // Zero timeout when checking state
+		if errors.Is(err, ErrTimedOut) {      // We don't care about timeout errors here!
+			return v, nil
```
imiric, line 794: Can you add a test for this scenario as well?

common/element_handle.go
```diff
@@ -1106,6 +1076,60 @@ func (h *ElementHandle) queryAll(selector string, eval evalFunc) ([]api.ElementH
 	return els, nil
 }
 
+// SetChecked checks or unchecks an element.
+func (h *ElementHandle) SetChecked(checked bool, opts goja.Value) {
```
imiric, line 1080: Will `SetChecked` also be exposed as a `Locator` method? [Playwright has it](https://playwright.dev/docs/api/class-locator#locator-set-checked), so did you plan to add it later? If so, this refactor would be better suited for that PR.

common/frame.go
inancgumus: 🤔 Hmm, good point. We need to make a refactoring for all the related methods in the `Frame` as well if we are going to follow this route. I created a new issue for this https://github.com/grafana/xk6-browser/issues/374.

common/element_handle.go
```diff
@@ -1106,6 +1076,60 @@ func (h *ElementHandle) queryAll(selector string, eval evalFunc) ([]api.ElementH
...
+	}
+	pointerFn := h.newPointerAction(fn, &parsedOpts.ElementHandleBasePointerOptions)
+	_, err = callApiWithTimeout(h.ctx, pointerFn, parsedOpts.Timeout)
+	if err != nil {
+		k6ext.Panic(h.ctx, "cannot check element: %w", err)
+	}
+	applySlowMo(h.ctx)
+}
+
+// Uncheck scrolls element into view and clicks in the center of the element.
```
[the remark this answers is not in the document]
inancgumus: Resolved by 677e460f2

common/element_handle.go
```diff
@@ -1106,6 +1076,60 @@ func (h *ElementHandle) queryAll(selector string, eval evalFunc) ([]api.ElementH
...
+func (h *ElementHandle) Uncheck(opts goja.Value) {
+	h.SetChecked(false, opts)
+}
+
+// Check scrolls element into view and clicks in the center of the element.
```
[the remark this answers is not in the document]
inancgumus: Resolved by 677e460f2

common/frame.go
imiric: Sure, a `Frame` test would make more sense, which would be out of scope for this PR, so I guess it can wait.

inancgumus: 👍  I updated the main functional tests issue: https://github.com/grafana/xk6-browser/issues/241.

imiric (approved):

Issue #366, opened 2022-06-03
Refactor Locator API tests


inancgumus:
The tests are repetitive, and some refactoring would be good.

- [x] Frame Element state methods and tests are repetitive ([PR](https://github.com/grafana/xk6-browser/blob/4e0b7e659ece7fec48f36714064bfd3d48938be2/tests/locator_test.go#L97-L153))
- [x] Parallelize the tests where it makes sense ([PR](https://github.com/grafana/xk6-browser/pull/389)).
- [x] Goja bool check ([PR](https://github.com/grafana/xk6-browser/pull/410))
- [x] Sanity checks (strict and timeout) ~Check for empty matches and add timeout tests~ ([PR 1](https://github.com/grafana/xk6-browser/pull/411) & [PR 2](https://github.com/grafana/xk6-browser/pull/412))
- [x] Convert to table-tests ([PR](https://github.com/grafana/xk6-browser/pull/413))

Somewhat related: #141

inancgumus:
Completed.

PR #372, opened 2022-06-06
Add locator.IsEditable

### LINKED ISSUE #348: locator.isEditable([options])
The [`locator.isEditable([options])`](https://playwright.dev/docs/api/class-locator#locator-is-editable) method returns `true` if the element matches the locator's selector (with strict mode on) and is editable. Otherwise, returns `false`.

inancgumus:
This PR closes #348.

* Extracts `Frame.isEditable` from `Frame.IsEditable` so that we can use `isEditable` from `Locator.IsEditable`.
* Adds `locator.IsEditable` and a test.
* Removes applying slow motion in the `Frame.IsEditable` as discussed in #234.

### DIFF
--- a/common/locator.go
+++ b/common/locator.go
@@ -3,10 +3,10 @@ package common
 import (
 	"context"
 
-	"github.com/dop251/goja"
-
 	"github.com/grafana/xk6-browser/k6ext"
 	"github.com/grafana/xk6-browser/log"
+
+	"github.com/dop251/goja"
 )
 
 // Locator represent a way to find element(s) on the page at any moment.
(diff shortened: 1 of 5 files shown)


common/locator.go
```diff
@@ -3,10 +3,10 @@ package common
 import (
 	"context"
 
-	"github.com/dop251/goja"
-
 	"github.com/grafana/xk6-browser/k6ext"
 	"github.com/grafana/xk6-browser/log"
+
+	"github.com/dop251/goja"
```
imiric, line 9: Nit, but why is `goja` moved below internal imports?

tests/locator_test.go
```diff
@@ -91,3 +91,21 @@ func TestLocatorCheck(t *testing.T) {
 		require.Panics(t, func() { input.IsChecked(nil) }, "should not select multiple elements")
 	})
 }
+
+func TestLocatorIsEditable(t *testing.T) {
+	tb := newTestBrowser(t, withFileServer())
```
imiric, line 96: Why is the `tparallel` linter broken again? :facepalm:

This test should have `t.Parallel()`, while the subtests probably shouldn't because they're sharing the same browser, `goja.Runtime`, etc., but then `tparallel` should complain about both cases, and we'd have to `nolint` it for the subtests... Ugh.

imiric (approved):
LGTM besides the couple minor things.

And the same comment about `timeout` applies here as well. Maybe we should mention and track these in #374?

tests/locator_test.go
inancgumus: No idea 🙄 You know we had many problems because of parallel tests back then (#141) and decided not to use them. Are you that we should enable it?

inancgumus:
> LGTM besides the couple minor things.

Thanks for the review!

> And the same comment about `timeout` applies here as well. Maybe we should mention and track these in #374?

Yup, updated #374.

common/locator.go
imiric: Ah right, my bad, external deps should indeed be last.

tests/locator_test.go
imiric: I guess I'm just venting about `golangci-lint` being unreliable, as usual... If `tparallel` is not reporting, how can we be sure that other linters are? :disappointed:

It's fine if you want to not parallelize these tests; they run quick enough and we won't notice them. But I think by default we should add `t.Parallel()` anywhere where it makes sense, and deal with flakiness once we run into it. Otherwise we won't know it's there, and we might as well disable the `tparallel` linter.

inancgumus: Right... I updated #366 for parallel tests. I'll make the tests parallel when it makes sense.

PR #373, opened 2022-06-06
Add locator.isEnabled

### LINKED ISSUE #349: locator.isEnabled([options])
The [`locator.isEnabled([options])`](https://playwright.dev/docs/api/class-locator#locator-is-enabled) method returns `true` if the element matches the locator's selector (with strict mode on) and is enabled. Otherwise, returns `false`.

inancgumus:
This PR closes #349.

* Extracts `Frame.isEnabled` from `Frame.IsEnabled` so that we can use `isEnabled` from `Locator.IsEnabled`.
* Adds `locator.IsEnabled` and a test.
* Refactors the locator state tests.
* Removes applying slow motion in the `Frame.IsEnabled` as discussed in #234.

### DIFF
--- a/api/locator.go
+++ b/api/locator.go
@@ -18,4 +18,7 @@ type Locator interface {
 	// IsEditable returns true if the element matches the locator's
 	// selector and is editable. Otherwise, returns false.
 	IsEditable(opts goja.Value) bool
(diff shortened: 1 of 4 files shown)


imiric (approved):
LGTM

:+1: for table tests.

PR #375, opened 2022-06-07
Add locator.isDisabled

### LINKED ISSUE #347: locator.isDisabled([options])
The [`locator.isDisabled([options])`](https://playwright.dev/docs/api/class-locator#locator-is-disabled) method returns `true` if the element matches the locator's selector (with strict mode on) and is disabled. Otherwise, returns `false`.

inancgumus:
This PR closes #347.

* Extracts `Frame.isDisabled` from `Frame.IsDisabled` so that we can use `isDisabled` from `Locator.IsDisabled`.
* Adds `locator.IsDisabled` and a test.
* Removes applying slow motion in the `Frame.IsDisabled` as discussed in #234.

### DIFF
--- a/api/locator.go
+++ b/api/locator.go
@@ -21,4 +21,7 @@ type Locator interface {
 	// IsEnabled returns true if the element matches the locator's
 	// selector and is enabled. Otherwise, returns false.
 	IsEnabled(opts goja.Value) bool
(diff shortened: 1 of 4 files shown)


imiric (approved):

inancgumus:
Thx, @imiric ❤️

PR #376, opened 2022-06-07
Add locator.isVisible

### LINKED ISSUE #351: locator.isVisible([options])
The [`locator.isVisible([options])`](https://playwright.dev/docs/api/class-locator#locator-is-visible) method returns `true` if the element matches the locator's selector (with strict mode on) and is visible. Otherwise, returns `false`.

Also, note PW deprecated the timeout property.

inancgumus:
This PR closes #351.

* Extracts `Frame.isVisible` from `Frame.IsVisible` so that we can use `isVisible` from `Locator.IsVisible`.
* Adds `locator.IsVisible` and a test.
* Removes applying slow motion in the `Frame.IsVisible` as discussed in #234.

### DIFF
--- a/api/locator.go
+++ b/api/locator.go
@@ -24,4 +24,7 @@ type Locator interface {
 	// IsDisabled returns true if the element matches the locator's
 	// selector and is disabled. Otherwise, returns false.
 	IsDisabled(opts goja.Value) bool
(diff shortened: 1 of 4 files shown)


imiric (approved):

inancgumus:
Thx, again, @imiric ❤️

PR #377, opened 2022-06-07
Add locator. isHidden

### LINKED ISSUE #350: locator.isHidden([options])
The [`locator.isHidden([options])`](https://playwright.dev/docs/api/class-locator#locator-is-hidden) method returns `true` if the element matches the locator's selector (with strict mode on) and is hidden. Otherwise, returns `false`.

Also, note PW deprecated the timeout property.

inancgumus:
This PR closes #350.

* Extracts `Frame.isHidden` from `Frame.IsHidden` so that we can use `isHidden` from `Locator.IsHidden`.
* Adds `locator.IsHidden` and a test.
* Removes applying slow motion in the `Frame.IsHidden` as discussed in #234.

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -1270,31 +1270,45 @@ func (f *Frame) isDisabled(selector string, opts *FrameIsDisabledOptions) (bool,
 	return bv, nil
 }
 
+// IsHidden returns true if the first element that matches the selector
+// is hidden. Otherwise, returns false.
 func (f *Frame) IsHidden(selector string, opts goja.Value) bool {
 	f.log.Debugf("Frame:IsHidden", "fid:%s furl:%q sel:%q", f.ID(), f.URL(), selector)
 
-	parsedOpts := NewFrameIsHiddenOptions(f.defaultTimeout())
-	if err := parsedOpts.Parse(f.ctx, opts); err != nil {
+	popts := NewFrameIsHiddenOptions(f.defaultTimeout())
+	if err := popts.Parse(f.ctx, opts); err != nil {
 		k6ext.Panic(f.ctx, "%w", err)
 	}
-
-	fn := func(apiCtx context.Context, handle *ElementHandle) (interface{}, error) {
-		value, err := handle.isHidden(apiCtx, 0) // Zero timeout when checking state
-		if err == ErrTimedOut {                  // We don't care about timeout errors here!
-			return value, nil
-		}
-		return value, err
-	}
-	actFn := f.newAction(
-		selector, DOMElementStateAttached, parsedOpts.Strict, fn, []string{}, false, true, parsedOpts.Timeout,
-	)
-	value, err := callApiWithTimeout(f.ctx, actFn, parsedOpts.Timeout)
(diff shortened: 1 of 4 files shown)


common/frame.go
```diff
@@ -1270,31 +1270,45 @@ func (f *Frame) isDisabled(selector string, opts *FrameIsDisabledOptions) (bool,
...
+		return v, err
 	}
-	actFn := f.newAction(
-		selector, DOMElementStateAttached, parsedOpts.Strict, fn, []string{}, false, true, parsedOpts.Timeout,
+	act := f.newAction(
+		selector, DOMElementStateAttached, opts.Strict, isHidden, []string{}, false, true, opts.Timeout,
 	)
-	value, err := callApiWithTimeout(f.ctx, actFn, parsedOpts.Timeout)
+	v, err := callApiWithTimeout(f.ctx, act, opts.Timeout)
 	if err != nil {
```
ankur22, line 1302: Should we catch the context timeout and return `false, nil` like we above on line 1293?

tests/locator_test.go
```diff
@@ -119,6 +119,11 @@ func TestLocatorElementState(t *testing.T) {
 			`() => document.getElementById('inputText').style.visibility = 'hidden'`,
 			func(l api.Locator) bool { return l.IsVisible(nil) },
 		},
+		{
+			"hidden",
```
ankur22, line 123: I think we should have a test for when `isHidden` return `true` as well as one for when `isHidden` returns `false`.

ankur22:
In the play wright docs in `isHidden` it mentions that:

> Returns whether the element is hidden, the opposite of [visible](https://playwright.dev/docs/actionability#visible).

So could we not have just wrapped the call `isVisible` and returned the opposite i.e.:

```go
func isHidden() {
    return !isVisible()
}
```

common/frame.go
inancgumus: No, we don't need to because the `callApiWithTimeout` already does that for us.

tests/locator_test.go
inancgumus: Hmm, but we already have a test for it:

https://github.com/grafana/xk6-browser/blob/4e0b7e659ece7fec48f36714064bfd3d48938be2/tests/locator_test.go#L138-L139

inancgumus:
Thanks for your review, @ankur22.

> So could we not have just wrapped the call `isVisible` and returned the opposite i.e.:
> 
> ```go
> func isHidden() {
>     return !isVisible()
> }
> ```

Fair question; however, due to [the actionability checks](https://playwright.dev/docs/actionability), we can't do that. We fait for a visible element to be attached. However, for a hidden element, we don't need to wait for it. So they need to be two different checks.

https://github.com/grafana/xk6-browser/blob/4e0b7e659ece7fec48f36714064bfd3d48938be2/common/js/injected_script.js#L235-L242

As you can see above, even if the element is not attached, we can check whether it's hidden or not.

ankur22 (approved):

Issue #338, opened 2022-06-01
locator.fill(value[, options])


inancgumus:
The [`locator.fill(value, [options])`](https://playwright.dev/docs/api/class-locator#locator-fill) method fills the input element matches the locator's selector (with strict mode on).

PR #388, opened 2022-06-10
Add locator.focus

### LINKED ISSUE #339: locator.focus([options])
The [`locator.focus([options])`](https://playwright.dev/docs/api/class-locator#locator-focus) method focuses on the input element matches the locator's selector (with strict mode on).

inancgumus:
* Extracts `Frame.focus` from `Frame.Focus` so that we can use `focus` from `Locator.Focus`.
* Adds `locator.Focus` and a test.

Closes #339.

----

Note: Incorrectly merged #385 onto #380. I should have merged #338 first.

### DIFF
--- a/tests/locator_test.go
+++ b/tests/locator_test.go
@@ -169,3 +169,28 @@ func TestLocatorFill(t *testing.T) {
 		require.Panics(t, func() { link.Fill(value, nil) }, "should not select multiple elements")
 	})
 }
+
+func TestLocatorFocus(t *testing.T) {
+	tb := newTestBrowser(t, withFileServer())
+	p := tb.NewPage(nil)
+	require.NotNil(t, p.Goto(tb.staticURL("/locators.html"), nil))
+
+	t.Run("ok", func(t *testing.T) {
+		focused := func() bool {
+			ok := p.Evaluate(tb.toGojaValue(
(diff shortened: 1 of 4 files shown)


ankur22 (approved):

tests/locator_test.go
```diff
@@ -169,3 +169,28 @@ func TestLocatorFill(t *testing.T) {
 		require.Panics(t, func() { link.Fill(value, nil) }, "should not select multiple elements")
 	})
 }
+
+func TestLocatorFocus(t *testing.T) {
+	tb := newTestBrowser(t, withFileServer())
```
imiric, line 174: I think we'll need to start adding `t.Parallel()` to at least these top-level tests. One or two without it is OK, but with the amount we're adding for `Locator`, it's increasing the total test runtime enough to be noticeable.

On `v0.3.0`, `GOMAXPROCS=3 go test -p 3 -v -race -count=1 ./...` finishes in 41.9s on my machine. The same command on `main` (before this was merged) takes 48.2s.

I know we run the risk of flakiness with more tests running in parallel, but I'd rather run that risk than incur the duration penalty every time I run the test suite.

inancgumus: See: https://github.com/grafana/xk6-browser/pull/389

PR #390, opened 2022-06-10
Add locator.getAttribute

### LINKED ISSUE #341: locator.getAttribute(name[, options])
The [`locator.getAttribute(name, [options])`](https://playwright.dev/docs/api/class-locator#locator-get-attribute) method gets an attribute of the input element matches the locator's selector (with strict mode on).

inancgumus:
* Extracts `Frame.getAttribute` from `Frame.GetAttribute` so that we can use `getAttribute` from `Locator.GetAttribute`.
* Adds `locator.GetAttribute` and a test.

Closes #341

### DIFF
--- a/api/locator.go
+++ b/api/locator.go
@@ -34,4 +34,6 @@ type Locator interface {
 	Fill(value string, opts goja.Value)
 	// Focus on the element using locator's selector with strict mode on.
 	Focus(opts goja.Value)
+	// GetAttribute of the element using locator's selector with strict mode on.
+	GetAttribute(name string, opts goja.Value) goja.Value
 }
--- a/api/locator.go
+++ b/api/locator.go
@@ -2,6 +2,12 @@ package api
 
 import "github.com/dop251/goja"
 
+// Strict mode:
+// All operations on locators throw an exception if more
+// than one element matches the locator's selector.
+//
+// See Issue #100 for more details.
+
 // Locator represents a way to find element(s) on a page at any moment.
 type Locator interface {
 	// Click on an element using locator's selector with strict mode on.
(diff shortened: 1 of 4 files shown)


api/locator.go
```diff
@@ -34,4 +34,6 @@ type Locator interface {
 	Fill(value string, opts goja.Value)
 	// Focus on the element using locator's selector with strict mode on.
 	Focus(opts goja.Value)
+	// GetAttribute of the element using locator's selector with strict mode on.
```
imiric, line 37: This is a late comment, but we mention "strict mode" a lot in this file, but it's never explained what that really means. Initially I confused it with JS strict mode, which is entirely different, and it wasn't clear until I dug into `injected_script.js`.

Can you maybe expand the `Locator` comment, or at the first occurrence explain what strict mode is? No need to have it repeated everywhere, but it would be helpful to mention it once.

tests/locator_test.go
```diff
@@ -206,3 +206,24 @@ func TestLocatorFocus(t *testing.T) {
 		require.Panics(t, func() { link.Focus(nil) }, "should not select multiple elements")
 	})
 }
+
+// Skip adding t.Parallel to subtests because they might cause flakiness.
```
imiric, line 210: In this case I'm pretty sure there would be a consistent data race either in goja or in our own code, so not really flakiness.

imiric (approved):

api/locator.go
inancgumus: Please see: 9fd4fc1

tests/locator_test.go
inancgumus: Please see: 9fd4fc1

PR #395, opened 2022-06-14
Add locator.inputValue

### LINKED ISSUE #345: locator.inputValue([options])
The [`locator.inputValue([options])`](https://playwright.dev/docs/api/class-locator#locator-input-value) method returns the element's input value that matches the locator's selector (with strict mode on).

inancgumus:
Closes #345.

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -1146,27 +1146,41 @@ func (f *Frame) innerText(selector string, opts *FrameInnerTextOptions) (string,
 	return gv.ToString().String(), nil
 }
 
+// InputValue returns the input value of the first element found
+// that matches the selector.
 func (f *Frame) InputValue(selector string, opts goja.Value) string {
 	f.log.Debugf("Frame:InputValue", "fid:%s furl:%q sel:%q", f.ID(), f.URL(), selector)
 
-	parsedOpts := NewFrameInputValueOptions(f.defaultTimeout())
-	if err := parsedOpts.Parse(f.ctx, opts); err != nil {
-		k6ext.Panic(f.ctx, "%w", err)
+	popts := NewFrameInputValueOptions(f.defaultTimeout())
+	if err := popts.Parse(f.ctx, opts); err != nil {
+		k6ext.Panic(f.ctx, "parse: %w", err)
+	}
+	v, err := f.inputValue(selector, popts)
+	if err != nil {
+		k6ext.Panic(f.ctx, "InputValue: %w", err)
 	}
 
-	fn := func(apiCtx context.Context, handle *ElementHandle) (interface{}, error) {
+	return v
+}
+
+func (f *Frame) inputValue(selector string, opts *FrameInputValueOptions) (string, error) {
+	inputValue := func(apiCtx context.Context, handle *ElementHandle) (interface{}, error) {
 		return handle.inputValue(apiCtx)
 	}
-	actFn := f.newAction(
-		selector, DOMElementStateAttached, parsedOpts.Strict, fn, []string{}, false, true, parsedOpts.Timeout,
+	act := f.newAction(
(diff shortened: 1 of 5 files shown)


common/frame.go
```diff
@@ -1146,27 +1146,41 @@ func (f *Frame) innerText(selector string, opts *FrameInnerTextOptions) (string,
...
-	parsedOpts := NewFrameInputValueOptions(f.defaultTimeout())
-	if err := parsedOpts.Parse(f.ctx, opts); err != nil {
-		k6ext.Panic(f.ctx, "%w", err)
+	popts := NewFrameInputValueOptions(f.defaultTimeout())
+	if err := popts.Parse(f.ctx, opts); err != nil {
+		k6ext.Panic(f.ctx, "parse: %w", err)
+	}
+	v, err := f.inputValue(selector, popts)
+	if err != nil {
+		k6ext.Panic(f.ctx, "InputValue: %w", err)
```
imiric, line 1160: This should be camelCase, not PascalCase, as that's how it's exposed in JS.

```suggestion
		k6ext.Panic(f.ctx, "inputValue: %w", err)
```

common/frame.go
```diff
@@ -1146,27 +1146,41 @@ func (f *Frame) innerText(selector string, opts *FrameInnerTextOptions) (string,
...
 	)
-	value, err := callApiWithTimeout(f.ctx, actFn, parsedOpts.Timeout)
+	v, err := callApiWithTimeout(f.ctx, act, opts.Timeout)
 	if err != nil {
-		k6ext.Panic(f.ctx, "%w", err)
+		return "", err
+	}
+	gv, ok := v.(goja.Value)
+	if !ok {
+		return "", fmt.Errorf("got %T; want goja value", v)
```
imiric, line 1180: Same here as in the other PRs: this error message should be user friendlier.

To summarize why:
- we shouldn't expose "goja value", as that's an internal implementation detail.
- the "got"/"want" phrasing is fine for unit tests, but not for end users.
- we should mention which method failed.

The message that was used previously in other PRs is fine for me: `unexpected <method name> value type: %T`.

imiric (changes requested):
Just some error message changes here as well. LGTM otherwise.

common/frame.go
inancgumus: I understand your concern, however, the type-mismatch error would be meaningless to users in any case. They can do nothing to solve it, can they?

Regarding the method name, we still print it in the upper layer: `InputValue` method, so it will look as follows:

`textContent: got *Something; want goja value`

imiric: Sorry, I think all of those are worse than `unexpected <method name> value type: %T`.

It's true that if this returns it's probably an internal bug for us to fix, but again, we definitely shouldn't mention `goja`, and the `got`/`want` phrasing is fine for unit tests, but not for user-facing code.

inancgumus: The upper layer adds the `inputValue:` prefix:

https://github.com/grafana/xk6-browser/blob/62a68854eea2adc7dd0bd0e9907cb5b902d3368d/common/frame.go#L1158-L1161

The inner layer describes the error:

https://github.com/grafana/xk6-browser/blob/62a68854eea2adc7dd0bd0e9907cb5b902d3368d/common/frame.go#L1174-L1181

I'm trying to find a solution that does what you want with this code. I'll omit `goja`, but I really couldn't understand what's the difference between `invalid` and `unexpected` :(

What about something like the following, then?

`inputValue: unexpected type: *Something`

Does it sound better?

imiric: Sure, let's go with that.

imiric (approved):

common/frame.go
inancgumus: #396 will resolve this.

PR #401, opened 2022-06-15
Add locator.tap

### LINKED ISSUE #356: locator.tap([options])
The [`locator.tap([options])`](https://playwright.dev/docs/api/class-locator#locator-tap) method taps on the element that matches the locator's selector (with strict mode).

inancgumus:
Closes #356.

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -1594,27 +1594,30 @@ func (f *Frame) SetInputFiles(selector string, files goja.Value, opts goja.Value
 	// TODO: needs slowMo
 }
 
+// Tap the first element that matches the selector.
 func (f *Frame) Tap(selector string, opts goja.Value) {
 	f.log.Debugf("Frame:Tap", "fid:%s furl:%q sel:%q", f.ID(), f.URL(), selector)
 
-	parsedOpts := NewFrameTapOptions(f.defaultTimeout())
-	err := parsedOpts.Parse(f.ctx, opts)
-	if err != nil {
-		k6ext.Panic(f.ctx, "%w", err)
+	popts := NewFrameTapOptions(f.defaultTimeout())
+	if err := popts.Parse(f.ctx, opts); err != nil {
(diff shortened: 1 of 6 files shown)


common/frame.go
```diff
@@ -1594,27 +1594,30 @@ func (f *Frame) SetInputFiles(selector string, files goja.Value, opts goja.Value
 	// TODO: needs slowMo
 }
 
+// Tap the first element matches the selector.
```
imiric, line 1597:
```suggestion
// Tap the first element that matches the selector.
```

tests/static/locators.html
```diff
@@ -32,7 +32,10 @@
         );
         document.querySelector('#inputText').addEventListener(
             'mousemove', e => { result = true; }, false
-        );        
+        );
+        document.addEventListener(
+            'touchstart', e => { result = true; }, false
```
imiric, line 37:
```suggestion
            'touchstart', e => { window.result = true; }, false
```

imiric (changes requested):
LGTM, just minor things.

Also, I think there's a bug here:
https://github.com/grafana/xk6-browser/blob/3ea1fdcfe4526151c7a4e126358001de9249f17b/common/element_handle.go#L593-L595

Shouldn't this be `tap(p.X, p.Y)`? Can you fix this in another PR?

tests/static/locators.html
inancgumus: Thanks.

inancgumus:
> Also, I think there's a bug here:
> 
> https://github.com/grafana/xk6-browser/blob/3ea1fdcfe4526151c7a4e126358001de9249f17b/common/element_handle.go#L593-L595
> 
> Shouldn't this be `tap(p.X, p.Y)`? Can you fix this in another PR?

Nice catch! Thanks. I fixed it in this PR as the tap test needs the fix to work.

ankur22 (approved):

PR #405, opened 2022-06-20
Add locator.dispatchEvent

### LINKED ISSUE #335: locator.dispatchEvent(type[, eventInit, options])
The [`locator.dispathEvent(type[, eventInit, options])`](https://playwright.dev/docs/api/class-locator#locator-dispatch-event) method dispatches an event for the element matches the locator's selector with strict mode on.

inancgumus:
The `Frame` and `ElementHandle` refactoring for this feature took a little more time. It's because `Frame` and `ElementHandle` were using `DblClickOptions` instead of `DispatchEventOptions`. Please see the commits for more details.

I tested `dispatchEvent` by dispatching a click event in the `example/fillform.js` instead of calling `click`, and it worked. 

However, `Frame` and `ElementHandle` should have proper dispatch event tests in the future. The locator test only tests whether we can route the `dispatchEvent` call to `Frame`, and it does not test the event dispatching behavior to the full extent.

_PS: I forgot to name the branch `feat/335-locator-dispatchEvent` 🤦_

Closes #335.

### DIFF
--- a/common/element_handle_options.go
+++ b/common/element_handle_options.go
@@ -495,3 +495,33 @@ func (o *ElementHandleWaitForElementStateOptions) Parse(ctx context.Context, opt
 	}
 	return nil
 }
+
+// ElementHandleDispatchEventOptions are options for ElementHandle.dispatchEvent.
+type ElementHandleDispatchEventOptions struct {
+	ElementHandleBasePointerOptions
+	Timeout time.Duration `json:"timeout"`
+}
+
+// NewElementHandleDispatchEventOptions returns a new ElementHandleDispatchEventOptions.
+func NewElementHandleDispatchEventOptions(defaultTimeout time.Duration) *ElementHandleDispatchEventOptions {
+	return &ElementHandleDispatchEventOptions{
+		Timeout: defaultTimeout,
+	}
+}
+
+// Parse the ElementHandleDispatchEvent options from a goja value.
+func (o *ElementHandleDispatchEventOptions) Parse(ctx context.Context, opts goja.Value) error {
+	if !isGojaValueExists(opts) {
+		return nil
+	}
+
+	gopts := opts.ToObject(k6ext.Runtime(ctx))
+	for _, k := range gopts.Keys() {
+		if k == "timeout" {
+			o.Timeout = time.Duration(gopts.Get(k).ToInteger()) * time.Millisecond
+			break
+		}
+	}
+
+	return nil
+}
--- a/common/element_handle_options.go
+++ b/common/element_handle_options.go
@@ -137,20 +137,21 @@ func NewElementHandleBaseOptions(defaultTimeout time.Duration) *ElementHandleBas
 }
 
 func (o *ElementHandleBaseOptions) Parse(ctx context.Context, opts goja.Value) error {
-	rt := k6ext.Runtime(ctx)
-	if opts != nil && !goja.IsUndefined(opts) && !goja.IsNull(opts) {
-		opts := opts.ToObject(rt)
-		for _, k := range opts.Keys() {
-			switch k {
-			case "force":
-				o.Force = opts.Get(k).ToBoolean()
-			case "noWaitAfter":
-				o.NoWaitAfter = opts.Get(k).ToBoolean()
-			case "timeout":
-				o.Timeout = time.Duration(opts.Get(k).ToInteger()) * time.Millisecond
-			}
+	if !gojaValueExists(opts) {
+		return nil
+	}
+	gopts := opts.ToObject(k6ext.Runtime(ctx))
+	for _, k := range gopts.Keys() {
+		switch k {
+		case "force":
+			o.Force = gopts.Get(k).ToBoolean()
+		case "noWaitAfter": //nolint:goconst
+			o.NoWaitAfter = gopts.Get(k).ToBoolean()
+		case "timeout":
+			o.Timeout = time.Duration(gopts.Get(k).ToInteger()) * time.Millisecond
 		}
 	}
+
 	return nil
 }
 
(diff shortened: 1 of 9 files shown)


common/element_handle_options.go
```diff
@@ -495,3 +495,33 @@ func (o *ElementHandleWaitForElementStateOptions) Parse(ctx context.Context, opt
...
+func (o *ElementHandleDispatchEventOptions) Parse(ctx context.Context, opts goja.Value) error {
+	if !isGojaValueExists(opts) {
+		return nil
+	}
+
+	gopts := opts.ToObject(k6ext.Runtime(ctx))
+	for _, k := range gopts.Keys() {
+		if k == "timeout" {
+			o.Timeout = time.Duration(gopts.Get(k).ToInteger()) * time.Millisecond
+			break
```
imiric, line 522: `timeout` is already parsed in `ElementHandleBaseOptions`:
https://github.com/grafana/xk6-browser/blob/f97ca063a8a7f483faed0d2d01251d9d370c9af8/common/element_handle_options.go#L149-L150

So we could just return the error from `o.ElementHandleBasePointerOptions.Parse(ctx, opts)` here.

Though like I mentioned in a previous PR, that method doesn't actually return an error. So we should validate the received options and return an error on failure.

common/frame_options.go
```diff
@@ -717,3 +717,36 @@ func (o *FrameWaitForSelectorOptions) Parse(ctx context.Context, opts goja.Value
...
+		return err
+	}
+	if !isGojaValueExists(opts) {
+		return nil
+	}
+	gopts := opts.ToObject(k6ext.Runtime(ctx))
+	for _, k := range gopts.Keys() {
+		if k == "strict" {
+			o.Strict = gopts.Get(k).ToBoolean()
+			break
```
imiric, line 747: AFAICS we only need the `timeout` and `strict` options, for which we have [`FrameBaseOptions`](https://github.com/grafana/xk6-browser/blob/f97ca063a8a7f483faed0d2d01251d9d370c9af8/common/frame_options.go#L34-L37), where [we already parse both of them](https://github.com/grafana/xk6-browser/blob/f97ca063a8a7f483faed0d2d01251d9d370c9af8/common/frame_options.go#L174-L179). So if you embed `FrameBaseOptions` instead of `ElementHandleDispatchEventOptions`, we should be able to just return the error from `o.FrameBaseOptions.Parse(ctx, opts)` here (after proper validation as mentioned above).

Embedding `ElementHandleDispatchEventOptions` also allows passing all the `ElementHandleBasePointerOptions` like `position` and `trial`, which I think are irrelevant for `Frame.dispatchEvent`.

common/helpers.go
```diff
@@ -215,3 +215,9 @@ func TrimQuotes(s string) string {
 	}
 	return s
 }
+
+// isGojaValueExists returns true if a given value is not nil and exists
+// (defined and not null) in the goja runtime.
+func isGojaValueExists(v goja.Value) bool {
```
imiric, line 221: `isGojaValueExists` is a bit awkwardly named, so maybe `isValidGojaValue` or `gojaValueExists`?

examples/fillform.js
```diff
@@ -1,29 +1,30 @@
-import { check } from 'k6';
-import launcher from 'k6/x/browser';
+import { check } from "k6";
+import launcher from "k6/x/browser";
```
imiric, line 2: These changes are out of place for this PR, so please leave them out and create an issue so that we can standardize this across all JS files.

tests/locator_test.go
```diff
@@ -410,3 +410,24 @@ func TestLocatorTap(t *testing.T) {
...
+	p := tb.NewPage(nil)
+	require.NotNil(t, p.Goto(tb.staticURL("/locators.html"), nil))
+
+	t.Run("ok", func(t *testing.T) {
+		result := func() bool {
+			ok := p.Evaluate(tb.toGojaValue(`() => window.result`))
+			return ok.(goja.Value).ToBoolean() //nolint:forcetypeassert
+		}
+		p.Locator("#link", nil).DispatchEvent("click", tb.toGojaValue("mouseevent"), nil)
+		require.True(t, result(), "could not dispatch event")
```
imiric, line 428: Can you assert on the value of `result()` before dispatching the event?

imiric (changes requested):

common/element_handle_options.go
inancgumus: Thanks, nice! I tried this and it worked. I'll be sending a PR shortly.

Regarding the validation, see #408.

common/frame_options.go
inancgumus: The problem here is that `FrameBaseOptions` don't have `Force` and `NoWaitAfter` options (unlike the currently embedded `ElementHandleDispatchEventOptions`). So it won't work if I simply change `ElementHandleDispatchEventOptions` to `FrameBaseOptions`.

`position` and `trial` can still be valuable here to change the behavior of the `dispatchEvent` method as what the `Frame` only does is to route the call to `ElementHandle`.

How should I proceed next?

tests/locator_test.go
inancgumus: Makes sense.

common/frame_options.go
imiric: But why do we need `Force` and `NoWaitAfter`? Those aren't options for `Frame.dispatchEvent` [according to Playwright](https://playwright.dev/docs/api/class-frame#frame-dispatch-event). Neither are `position` and `trial`.

My main issue is with repeating the parsing of `strict`, which, given that it applies to all(?) `Frame` methods, doesn't belong here.

And if we want to differ from Playwright's API, that's fine I suppose, but it adds more things to document, so I wonder if we need to.

inancgumus: We need `Force` and `NoWaitAfter` in `Frame.newAction`. I've provided the options' default values and called the method. Doing so allows us only to embed `FrameBaseOptions`.

Please take a look at [the commit]([7b4df2a](https://github.com/grafana/xk6-browser/pull/405/commits/7b4df2a5f5037da827af0284366bda7b91d6a6b9)) and let me know how I proceed next.

imiric (approved):

Issue #254, opened 2022-02-25
Data race in TestWaitForFrameNavigationWithinDocument: NavigateFrame


inancgumus:
[This CI job](https://github.com/grafana/xk6-browser/runs/5331842835?check_suite_focus=true) has failed with a data race error. It seems sharing of a `Request` concurrently is causing the issue.

<details>
<summary>Click here to see the log output.</summary>

```shell
Read at 0x00c0075f7318 by goroutine 2552:
  github.com/grafana/xk6-browser/common.(*FrameManager).NavigateFrame()
      /home/runner/work/xk6-browser/xk6-browser/common/frame_manager.go:677 +0x151d
  github.com/grafana/xk6-browser/common.(*Frame).Goto()
      /home/runner/work/xk6-browser/xk6-browser/common/frame.go:843 +0x6c
  github.com/grafana/xk6-browser/common.(*Page).Goto()
      /home/runner/work/xk6-browser/xk6-browser/common/page.go:555 +0x1d3
  github.com/grafana/xk6-browser/tests.TestWaitForFrameNavigationWithinDocument.func1()
      /home/runner/work/xk6-browser/xk6-browser/tests/frame_manager_test.go:29 +0x137
  testing.tRunner()
      /home/runner/sdk/gotip/src/testing/testing.go:1439 +0x213
  testing.(*T).Run.func1()
      /home/runner/sdk/gotip/src/testing/testing.go:1486 +0x47

Previous write at 0x00c0075f7318 by goroutine 2667:
  github.com/grafana/xk6-browser/common.(*NetworkManager).handleRequestRedirect()
      /home/runner/work/xk6-browser/xk6-browser/common/network_manager.go:311 +0x95
  github.com/grafana/xk6-browser/common.(*NetworkManager).onRequest()
      /home/runner/work/xk6-browser/xk6-browser/common/network_manager.go:442 +0x114
  github.com/grafana/xk6-browser/common.(*NetworkManager).handleEvents()
      /home/runner/work/xk6-browser/xk6-browser/common/network_manager.go:377 +0x29e
  github.com/grafana/xk6-browser/common.(*NetworkManager).initEvents.func1()
      /home/runner/work/xk6-browser/xk6-browser/common/network_manager.go:356 +0x3e

Goroutine 2552 (running) created at:
  testing.(*T).Run()
      /home/runner/sdk/gotip/src/testing/testing.go:1486 +0x724
[...]

PR #411, opened 2022-06-22
Fix timeout detection (callApiWithTimeout)

inancgumus:
Before this fix, we couldn't detect the timeouts properly (v0.3.0~now). For example, in the following test, we assert that the result of the `TextContent` method is empty even if the element we're looking for does not exist. The frame action should have timed out instead of returning an empty string (`p` tag doesn't exist in `sampleHTML`).

See the `err_wrong_selector` subtest.

https://github.com/grafana/xk6-browser/blob/eeab22ebfdc5c6d2863eab9a744ea1205d6d9309/tests/page_test.go#L256-L289

This PR fixes the problem and detects the timeout error properly by using `context.DeadlineExceeded` instead of `context.Canceled`. The current system is still unstable because this fix re-surfaced a data race bug that happens from time to time (happens on frame navigation). Previously reported on #254's description.

I've also fixed a few page tests vulnerable to this bug by requiring that they panic (one of them is above).

This bug has also prevented me from testing the locator timeout tests (#366). Then again, ~I'll send a PR~I sent #412 for the locator timeout tests ~shortly~ (based on this branch). For example, the following test wasn't panicking because of the bug that this PR fixes:

```go
tb := newTestBrowser(t)
p := tb.NewPage(nil)
p.SetContent("<html></html>", nil)
assert.Panics(t, func() { p.Locator("NOTEXIST", nil).Tap(tb.toGojaValue(jsFrameBaseOpts{Timeout: "100"})) })
```

### DIFF
--- a/common/helpers.go
+++ b/common/helpers.go
@@ -23,6 +23,7 @@ package common
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math"
 	"time"
@@ -121,7 +122,8 @@ func callApiWithTimeout(ctx context.Context, fn func(context.Context, chan inter
 
 	select {
 	case <-apiCtx.Done():
-		if apiCtx.Err() == context.Canceled {
+		err = apiCtx.Err()
+		if errors.Is(err, context.DeadlineExceeded) {
 			err = ErrTimedOut
 		}
 	case result = <-resultCh:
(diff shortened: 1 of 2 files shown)


imiric (approved):
Nice catch!

I suppose nobody had bothered reporting the previous behavior, but yeah, it should definitely throw an error on timeout.

inancgumus:
> Nice catch!
> 
> I suppose nobody had bothered reporting the previous behavior, but yeah, it should definitely throw an error on timeout.

Thanks! I started seeing fewer CI failures. I hope some of them were happening because of this bug 🤞

Issue #415, opened 2022-06-23
Browser process close errors when running more than 1 iteration


imiric:
Currently (on eb13672) the following command:

```
XK6_BROWSER_LOG=info XK6_HEADLESS=1 xk6 run -u 1 -i 5 examples/fillform.js
```

Produces these errors:

```
ERRO[0002] cmd.Wait: signal: killed                      category="BrowserType:execute" elapsed="0 ms" goroutine=22
ERRO[0002] stdout.Close: close |0: file already closed   category="BrowserType:execute" elapsed="0 ms" goroutine=22
```

It seems to happen on each iteration end.

imiric:
This seems to be a recent regression that's not present in v0.3.0.

`git bisect` points to it being introduced in #409 (72b006d).

@ankur22 Could you take a look? We should fix this before releasing v0.4.0.

ankur22:
> This seems to be a recent regression that's not present in v0.3.0.
> 
> `git bisect` points to it being introduced in #409 ([72b006d](https://github.com/grafana/xk6-browser/commit/72b006dfab0d698faff183b8ba8b41b5af340c73)).
> 
> @ankur22 Could you take a look? We should fix this before releasing v0.4.0.

Yep, the change (to log any errors during `cmd.Wait()` and `stdout.Close()`) has surfaced a couple of bugs:
1. `stdout.Close: close |0: file already closed` -- we don't need to call `stdout.Close()` since the `cmd.Wait()` will close the `stdout` pipe for us (Source: https://pkg.go.dev/os/exec#Cmd.StdoutPipe)
2. `cmd.Wait: signal: killed ` -- :
    1. Is it because a process is killing it (only thing i can find is if there is a panic, then it will get the PID of the browser process and kill it)
    2. Is it due to the context closing (Source: https://pkg.go.dev/os/exec#CommandContext). Unfortunately we can't unwrap the error and look for a context closed error so instead we could just do a simple `err.Error() != "signal: killed"` and only print errors if it's not been killed.

inancgumus:
> 1. `stdout.Close: close |0: file already closed` -- we don't need to call `stdout.Close()` since the `cmd.Wait()` will close the `stdout` pipe for us (Source: https://pkg.go.dev/os/exec#Cmd.StdoutPipe)
[...]

PR #419, opened 2022-06-27
Fix errors on Close and Wait when browser closes

ankur22:
- Close on the stdout pipe is already dealt with by Wait.
- Kill signal is sent when the context is closed so only log other non kill related errors.
Fixes: #415


chromium/browser_type.go
```diff
@@ -387,11 +387,11 @@ func execute(
...
-			logger.Errorf("BrowserType:execute", "cmd.Wait: %v", err)
-		}
-
-		if err := stdout.Close(); err != nil {
-			logger.Errorf("BrowserType:execute", "stdout.Close: %v", err)
+			if err.Error() != "signal: killed" {
+				logger.Errorf("BrowserType:execute", "cmd.Wait: %v", err)
+			} else {
+				logger.Debug("BrowserType:execute", "cmd.Wait: process killed")
+			}
```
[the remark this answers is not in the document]
ankur22: 1. Yeah sure, how do you go about testing on Windows?
2. Yep, I was thinking the same tbh. I would prefer to create a new task to track that bug though as I don't think it'll be a quick fix, as I have a feeling that the context is being closed in k6 which is in turn killing the browser processor and not allowing us to gracefully shut it down.
  An approach could be to use a new context when creating the browser process; catch the context closing; and finally attempt to close browser process. But whatever approach we take will depend on https://github.com/grafana/k6/issues/2432 :/
3. Yep, I'll do that change with the log message.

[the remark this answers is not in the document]
imiric: 1. Discussed over Slack.

2. Sounds good about the new issue. [The context is cancelled after each iteration](https://github.com/grafana/k6/blob/be4fcb8cb4e30eebaa40d2c356e0be5cc1b21d2c/js/runner.go#L746-L748), besides the `RunContext` being cancelled at shutdown. So we should be able to stop the process gracefully [here](https://github.com/grafana/xk6-browser/blob/eb136726e37282fffad0c642062a04a6680a6adc/chromium/browser_type.go#L163), and ensure it works when running with multiple VUs/iterations, even though we can't ensure that it will work when k6 is stopped abruptly or at test end because of #2432.

chromium/browser_type.go
```diff
@@ -387,11 +387,11 @@ func execute(
 		}()
 
 		if err := cmd.Wait(); err != nil {
-			logger.Errorf("BrowserType:execute", "cmd.Wait: %v", err)
-		}
-
-		if err := stdout.Close(); err != nil {
-			logger.Errorf("BrowserType:execute", "stdout.Close: %v", err)
+			if err.Error() != "signal: killed" {
+				logger.Errorf("BrowserType:execute", "browser process with PID %d ended unexpectedly %v", cmd.Process.Pid, err)
```
imiric, line 391:
```suggestion
				logger.Errorf("BrowserType:execute", "browser process with PID %d ended unexpectedly: %v", cmd.Process.Pid, err)
```

chromium/browser_type.go
```diff
@@ -387,11 +387,11 @@ func execute(
...
+			} else {
+				logger.Debug("BrowserType:execute", "browser process with PID %d killed, context cancelled", cmd.Process.Pid)
```
imiric, line 393: I think we can use the same log message in both cases, just change the method to call. And we don't need to mention `context cancelled`; we can't know whether that's always the case, so it might be confusing.

So something like:
```go
if err := cmd.Wait(); err != nil {
	log := logger.Errorf
	if err.Error() == "signal: killed" {
		// The browser process is killed when the context is cancelled
		// after a k6 iteration ends, so silence the log message until
		// we can stop it gracefully. See #new-issue-number
		log = logger.Debugf
	}
	log("BrowserType:execute", "browser process with PID %d ended unexpectedly: %v", cmd.Process.Pid, err)
}
```

imiric (commented):
LGTM, assuming it works on Windows. :crossed_fingers:

Just left minor suggestions.

chromium/browser_type.go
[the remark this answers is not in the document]
ankur22: I've tested it on Windows and I get `exit status 1` from `err.Error()`. So I think it's:
1. Best to revert the logging for `cmd.Wait()` and move it into the new ticket where I will attempt to gracefully shutdown the browser, or otherwise log an error.
2. The `stdout.Close()` is definitely not needed in Windows either, so that can safely be deleted.

WDYT?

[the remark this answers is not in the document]
ankur22: New issue to try to gracefully shutdown the browser process is here: https://github.com/grafana/xk6-browser/issues/423

chromium/browser_type.go
ankur22: Resolved in: 85046600c42120189084442e6e56a3caa5096bd1

chromium/browser_type.go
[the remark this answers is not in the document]
ankur22: Resolved in: 5222abb4d1d92ec43ce28fe9295f147fce1f61e2

inancgumus (approved):

chromium/browser_type.go
```diff
@@ -387,11 +387,14 @@ func execute(
...
-			logger.Errorf("BrowserType:execute", "cmd.Wait: %v", err)
-		}
-
-		if err := stdout.Close(); err != nil {
-			logger.Errorf("BrowserType:execute", "stdout.Close: %v", err)
+			log := logger.Errorf
+			if err.Error() == "signal: killed" || err.Error() == "exit status 1" {
+				// The browser process is killed when the context is cancelled
+				// after a k6 iteration ends, so silence the log message until
+				// we can stop it gracefully. See #new-issue-number
```
imiric, line 394: :smile:

```suggestion
				// we can stop it gracefully. See https://github.com/grafana/xk6-browser/issues/423
```

ankur22: 🤦 Thanks :)

ankur22: Resolved in: b87e9711395eb5bc473212c73ee02e27479d016c

imiric: The hash prefix is not needed, but it's fine :)

ankur22: Ah, i was avoiding "Commit suggestion" as I didn't know about the batch feature. I will use it next time, thanks

Issue #417, opened 2022-06-27
panic: interface conversion: interface is nil, not goja.Value


sniku:
xk6-browser panics with the attached script. 
version: latest main eb13672

The panic is reproducible on my machine. It happened 4 times in the row.

```
➜  Downloads xk6-browser3 run reproducible-websocket-EOF-errors.js

          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io

  execution: local
     script: reproducible-websocket-EOF-errors.js
     output: -

  scenarios: (100.00%) 1 scenario, 10 max VUs, 10m30s max duration (incl. graceful stop):
           * default: 10 looping VUs for 10m0s (gracefulStop: 30s)

[...]
ERRO[0030] err:websocket: close 1006 (abnormal closure): unexpected EOF  category="Connection:handleIOError" elapsed="0 ms" goroutine=209
ERRO[0030] cmd.Wait: signal: killed                      category="BrowserType:execute" elapsed="2 ms" goroutine=23
ERRO[0030] stdout.Close: close |0: file already closed   category="BrowserType:execute" elapsed="0 ms" goroutine=23
panic: interface conversion: interface is nil, not goja.Value
running (00m30.0s), 10/10 VUs, 22 complete and 0 interrupted iterations
goroutine 4320 [running]:------------------------] 10 VUs  00m30.0s/10m0s
github.com/grafana/xk6-browser/common.(*ElementHandle).waitForElementState(0xc000deb060?, {0x1687188, 0xc00372d500}, {0xc003b31530, 0x3, 0x3}, 0x6fc23ac00)
	github.com/grafana/xk6-browser@v0.3.1-0.20220622164257-eb136726e372/common/element_handle.go:660 +0x145
github.com/grafana/xk6-browser/common.(*ElementHandle).newPointerAction.func1({0x1687188, 0xc00372d500}, 0xc000deb060)
	github.com/grafana/xk6-browser@v0.3.1-0.20220622164257-eb136726e372/common/element_handle.go:1371 +0x105
github.com/grafana/xk6-browser/common.retryPointerAction({0x1687188, 0xc00372d500}, 0xc00022e460, 0xc00372d4a0)
[...]

PR #425, opened 2022-06-30
Fix type assertions and remove nolint directives for good

inancgumus:
This commit properly type-asserts to remove `nolint:forcetypeassert` directives both from the code and tests.

There are new helpers to reduce code bloat and possible error messages refactoring work:
+ `asGojaValue(v)`: returns `v` as a `goja` value or panics.
+ `gojaValueToString(v)`: returns `v` as `string` or panics.

Fixes: #417
Related: #58

### DIFF
--- a/common/element_handle.go
+++ b/common/element_handle.go
@@ -63,7 +63,13 @@ func (h *ElementHandle) boundingBox() (*Rect, error) {
 func (h *ElementHandle) checkHitTargetAt(apiCtx context.Context, point Position) (bool, error) {
 	frame := h.ownerFrame(apiCtx)
 	if frame != nil && frame.parentFrame != nil {
-		element := h.frame.FrameElement().(*ElementHandle) //nolint:forcetypeassert
+		var (
+			el          = h.frame.FrameElement()
+			element, ok = el.(*ElementHandle)
+		)
+		if !ok {
+			return false, fmt.Errorf("unexpected type %T", el)
+		}
 		box, err := element.boundingBox()
 		if err != nil {
 			return false, err
@@ -91,9 +97,11 @@ func (h *ElementHandle) checkHitTargetAt(apiCtx context.Context, point Position)
 
 	// Either we're done or an error happened (returned as "error:..." from JS)
 	const done = "done"
-	//nolint:forcetypeassert
-	value := result.(goja.Value)
-	if value.ExportType().Kind() != reflect.String {
+	v, ok := result.(goja.Value)
+	if !ok {
+		return false, fmt.Errorf("unexpected type %T", result)
+	}
+	if v.ExportType().Kind() != reflect.String {
 		// We got a { hitTargetDescription: ... } result
 		// Meaning: Another element is preventing pointer events.
 		//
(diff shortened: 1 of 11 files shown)


common/element_handle.go
```diff
@@ -263,11 +268,7 @@ func (h *ElementHandle) fill(_ context.Context, value string) error {
 	if err != nil {
 		return err
 	}
-	r, ok := result.(goja.Value)
-	if !ok {
-		return fmt.Errorf("expected goja value; got %T", result)
-	}
-	if s := r.String(); s != resultDone {
+	if s := gojaValueToString(h.ctx, result); s != resultDone {
```
ankur22, line 271: Are we changing the behaviour here? It used to return an error, but now panics

inancgumus: Ah, nice catch! You might be right, I'll check.

common/frame.go
```diff
@@ -1033,12 +1027,8 @@ func (f *Frame) getAttribute(selector, name string, opts *FrameBaseOptions) (goj
 	if err != nil {
 		return nil, err
 	}
-	gv, ok := v.(goja.Value)
-	if !ok {
-		return nil, fmt.Errorf("unexpected type %T", v)
-	}
 
-	return gv, nil
+	return asGojaValue(f.ctx, v), nil
```
ankur22, line 1031: Same here, maybe it's ok to panic instead of erroring, since i believe all errors are eventually turned into a panic

inancgumus: Yep, I noticed it and fixing it now, thanks.

common/element_handle.go
inancgumus: I sent the updates, please let me know 🙇

common/element_handle.go
```diff
@@ -63,7 +63,13 @@ func (h *ElementHandle) boundingBox() (*Rect, error) {
 func (h *ElementHandle) checkHitTargetAt(apiCtx context.Context, point Position) (bool, error) {
 	frame := h.ownerFrame(apiCtx)
 	if frame != nil && frame.parentFrame != nil {
-		element := h.frame.FrameElement().(*ElementHandle) //nolint:forcetypeassert
+		var (
+			el          = h.frame.FrameElement()
+			element, ok = el.(*ElementHandle)
+		)
+		if !ok {
+			return false, fmt.Errorf("unexpected type %T", el)
```
imiric, line 71: Should we `k6ext.Panic()` in these cases? It would be consistent with the `FrameSession.onTargetCrashed` change, and it seems like something we should panic on.

Or will this panic higher up, as @ankur22 mentioned?

common/frame.go
```diff
@@ -1113,7 +1107,7 @@ func (f *Frame) innerHTML(selector string, opts *FrameInnerHTMLOptions) (string,
 		return "", fmt.Errorf("unexpected type %T", v)
 	}
 
-	return gv.ToString().String(), nil
+	return gv.String(), nil
```
imiric, line 1110: Are we sure these are equivalent? The tests pass, so likely yes, but I'm wondering why the previous usage is common.

common/element_handle.go
inancgumus: Yes, it should be safe. It will eventually panic in the higher layers. For example, [`checkHitTargetAt`](https://github.com/grafana/xk6-browser/blob/fc34711f5ea6e1178bb10a059e77aa544eddccfc/common/element_handle.go#L63) returns err => [`newPointerAction`](https://github.com/grafana/xk6-browser/blob/fc34711f5ea6e1178bb10a059e77aa544eddccfc/common/element_handle.go#L1435) returns err => [`Click`](https://github.com/grafana/xk6-browser/blob/fc34711f5ea6e1178bb10a059e77aa544eddccfc/common/element_handle.go#L745) panics on the error.

imiric: OK. Still, I think we should panic as close as possible to where the issue happens, so here, but it's fine, we can leave it as is.

imiric (approved):

common/frame.go
inancgumus: I made a little research on this and here are my findings. For example, `Frame.Content()` uses `ExecutionContext.eval` to get a goja value:

```
// remote_object.go
func valueFromRemoteObject(...) {
    ...
    return k6ext.Runtime(ctx).ToValue(val)
}
```

The above function returns the value as `goja.asciiString`. `asciiString.ToString()` returns itself. Whereas, `asciiString.String()` returns itself as `string` as its underlying type is `string`. So `.ToString().String()` == `String()`.

PR #429, opened 2022-06-30
Make log output more concise

### LINKED ISSUE #304: Keep log output concise
**Note:** I suggested [an approach](https://github.com/grafana/xk6-browser/discussions/157) earlier.

We've been receiving complaints from our users that the extension's log output is somewhat noisy. Excessive log output makes users worry about something wrong with their scripts, even if their scripts are totally fine.

I find our current user-facing error messages unnecessarily verbose and noisy. It would be much better if we could display separate error messages and debug logs. For example, we show the following error to the users:

`panic: unable to selectOption: element handle cannot evaluate: cannot call function on expression ("\n\t\t(node, injected, selector, strict, state, timeout, ...args) => {\n\t\t\treturn injected.waitForSelector(selector, node, strict, state, 'raf', timeout, ...args);\n\t\t}\n\t\n//# sourceURL=__xk6_browser_evaluation_script__\n") in execution context (3) in frame (39730BFD4B4E60220EF0B0A7816E7905) with session (BF9F73D22785325E395258F9B99CC591): exception "Uncaught (in promise)" (0:0): error:strictmodeviolation`

We might do the following instead:

`panic: selectOption: selector "divs" cannot select multiple elements while the strict mode is on.`

While the other details can go to the debug output.

Note: The error message I'm suggesting is just an example. The strict mode is not an implementation detail according to Playwright and we offer the strict option to the users as well. Our mileage may vary. We can discuss and decide to omit it from the user-facing error message.

**Related:**
* We received a lot of these complaints on our Slack channels.
* https://community.k6.io/t/is-there-a-way-to-suppress-warnings-in-the-console/3636/2
* https://community.k6.io/t/erro-0006-unexpected-eof-error-during-run-of-example-script/3136/4
* https://github.com/grafana/xk6-browser/issues/232 (somewhat related)
* https://github.com/grafana/xk6-browser/issues/250 (somewhat related)
* https://github.com/grafana/xk6-browser/pull/396#discussion_r896676441
* https://github.com/grafana/xk6-browser/pull/286#discussion_r866890577

imiric:
This changes many of our log messages to make them more concise and user friendlier, according to the discussion in #304, and particularly following some of the conventions mentioned [here](https://github.com/grafana/xk6-browser/issues/304#issuecomment-1165579869). See the commits for details. It might be easier to review by commit.

I added a [temporary script](https://github.com/grafana/xk6-browser/blob/eef08a9f7fa95b0eb7dc77b347254f21499e510d/examples/304-log-errors.js) you can use to check some of the error messages yourself, or see the before/after comparison comments. I'll remove the script before merging this, as it's not important.

This is probably not an exhaustive improvement of all error messages, and some will still be confusing, but I think we can close #304 with this, and improve any leftover ones later.


examples/304-log-errors.js
```diff
@@ -0,0 +1,178 @@
...
+  page.goto('https://test.k6.io/', { waitUntil: 'networkidle' });
+
+// ===== BEFORE =====
+// ERRO[0002] click: element handle cannot evaluate: cannot call function on expression ("\n\t\t(node, injected, selector, strict, state, timeout, ...args) => {\n\t\t\treturn injected.waitForSelector(selector, node, strict, state, 'raf', timeout, ...args);\n\t\t}\n\t\n//# sourceURL=__xk6_browser_evaluation_script__\n") in execution context (3) in frame (F229F0E745BC94879FA7D9EA9AC1CA63) with session (DEE3FBD05BF682C55E68A2D972EA215F): exception "Uncaught (in promise)" (0:0): error:strictmodeviolation
+//         at reflect.methodValueCall (native)
+//         at locatorStrictMode (file:///home/ivan/Projects/grafana/xk6-browser/examples/304-log-errors.js:144:15(41))
+//         at native  executor=shared-iterations scenario=err source=stacktrace
+//
+// ===== AFTER =====
+// TODO: Should we make this error:strictmodeviolation friendlier?
```
ankur22, line 35: I remember that there was some confusion as to what strict mode actually was (https://github.com/grafana/xk6-browser/pull/390#discussion_r895472111), so maybe it would be good to clarify that it it found more than one element that matched the selector?

inancgumus:
Nice work, much better than before :clap: It would be good for users to see Playwright-like
error messages (for further compatibility and at-home feeling).

I haven't finished my review yet. Here are my first observations and suggestions. @imiric 

### page.Goto

Using the script (`examples/304-log-errors.js`), I see the following error when I run `pageGotoTimeout0`:

```
ERRO[0001] frame navigation to "https://test.k6.io/my_messages.php": timed out after 0s
        at pageGotoTimeout0 (file:///Users/inanc/grafana/xk6-browser/examples/304-log-errors.js:63:50(32))
        at native  executor=shared-iterations scenario=err source=stacktrace
```

Could we drop `frame` from the error message since navigation happens
because of a frame (as explained [here](https://github.com/grafana/xk6-browser/issues/304#issuecomment-1167245835)). Following the present continuous tense style, we can
replace `navigation to` with `navigating to` as you suggested
in #304.

Can we get rid of the following? It doesn't provide any value to users.
Probably we cannot do so because they are coming from the k6 itself.

```
        at native  executor=shared-iterations scenario=err source=stacktrace
```

Here's what happens in Playwright when `page.goto` times out. Note that I'm not saying we should do exactly what PW does. I'm only
comparing the differences so that we may get inspired.

```
page.goto: Timeout 1ms exceeded.
=========================== logs ===========================
navigating to "https://demo.playwright.dev/todomvc", waiting until "load"
============================================================

    3 | test.describe('page', () => {
    4 |   test('goto', async ({ page }) => {
> 5 |     await page.goto('https://demo.playwright.dev/todomvc', { timeout: 1 });
    |                ^
    6 |   });
    7 | })

    at /Users/inanc/grafana/playwright-tests/tests/concise.spec.ts:5:16
```

### pageQuery

When running `page.$('h5', { timeout: 0 });`, the following error occurs:

```
querying selector "h5": execution context with ID 3 not found
```

However, with `page.$('h5');`, no error occurs. This could be not very clear
for users because it happens because of a timeout, but the users might
not have an idea what an execution context is.

What do you think about the following?

```
querying selector "h5": not found on the frame.
```

Maybe we can provide a link to a document about the possible causes of
this error:

```
querying selector "h5": not found on the frame. see: http://link for more details.
```

### locatorStrictMode

Agree on making this error more user-friendly:

```
ERRO[0002] click: error:strictmodeviolation
        at locatorStrictMode (file:///Users/inanc/grafana/xk6-browser/examples/304-log-errors.js:40:15(41))
        at native  executor=shared-iterations scenario=err source=stacktrace
```

Maybe something like:

```
ERRO[0002] click: "selector" selected multiple elements. it should select a single element while the strict mode is on
        at locatorStrictMode (file:///Users/inanc/grafana/xk6-browser/examples/304-log-errors.js:40:15(41))
        at native  executor=shared-iterations scenario=err source=stacktrace
```

examples/304-log-errors.js
```diff
@@ -0,0 +1,178 @@
...
+//         at reflect.methodValueCall (native)
+//         at pageWaitForSelector (file:///home/ivan/Projects/grafana/xk6-browser/examples/304-log-errors.js:132:29(70))
+//         at native  executor=shared-iterations scenario=err source=stacktrace
+//
+// ===== AFTER =====
+// ERRO[0008] querying selector "h5": execution context with ID 3 not found
+//         at reflect.methodValueCall (native)
+//         at pageQuery (file:///home/ivan/Projects/grafana/xk6-browser/examples/304-log-errors.js:125:15(70))
+//         at native  executor=shared-iterations scenario=err source=stacktrace
+  page.$('h5', { timeout: 0 });
```
ankur22, line 127: The new error message doesn't mention that `h5` couldn't be found, and instead mentions a mysterious `execution context with ID 3`. Would it be better to just say something like `timeout while looking for element h5: element not found`?

chromium/browser_type.go
```diff
@@ -371,15 +371,16 @@ func execute(
 	// can run into a data race.
 	err = cmd.Start()
 	if os.IsNotExist(err) {
-		return nil, nil, fmt.Errorf("does not exist: %s", path)
+		return nil, nil, fmt.Errorf("file does not exist: %s", path)
```
ankur22, line 374: Is this when it can't find the browser exec? I think it was ok as `does not exists`

common/frame.go
```diff
@@ -1859,7 +1859,7 @@ func (f *Frame) evaluate(
 	}
 	eh, err := ec.eval(apiCtx, opts, pageFunc.ToString().String(), evalArgs...)
 	if err != nil {
-		return nil, fmt.Errorf("frame cannot evaluate: %w", err)
+		return nil, fmt.Errorf("%w", err)
```
ankur22, line 1862:
```suggestion
		return nil, err
```

chromium/browser_type.go
```diff
@@ -371,15 +371,16 @@ func execute(
...
 	}
 	if err != nil {
 		return nil, nil, fmt.Errorf("%w", err)
```
ankur22, line 377:
```suggestion
		return nil, nil, err
```

There are several of these, but i think we should just be able to return the error without wrapping it, no?

common/keyboard.go
```diff
@@ -166,7 +166,7 @@ func (k *Keyboard) up(key string) error {
 		WithCode(keyDef.Code).
 		WithLocation(keyDef.Location)
 	if err := action.Do(cdp.WithExecutor(k.ctx, k.session)); err != nil {
-		return fmt.Errorf("cannot execute dispatch key event up: %w", err)
+		return fmt.Errorf("executing dispatch key event up: %w", err)
```
inancgumus, line 169:
```suggestion
		return fmt.Errorf("dispatching key event up: %w", err)
```

common/keyboard.go
```diff
@@ -175,7 +175,7 @@ func (k *Keyboard) up(key string) error {
 func (k *Keyboard) insertText(text string) error {
 	action := input.InsertText(text)
 	if err := action.Do(cdp.WithExecutor(k.ctx, k.session)); err != nil {
-		return fmt.Errorf("cannot execute insert text: %w", err)
+		return fmt.Errorf("executing insert text: %w", err)
```
inancgumus, line 178:
```suggestion
		return fmt.Errorf("inserting text: %w", err)
```

common/mouse.go
```diff
@@ -153,53 +153,53 @@ func (m *Mouse) up(x float64, y float64, opts *MouseDownUpOptions) error {
 func (m *Mouse) Click(x float64, y float64, opts goja.Value) {
 	mouseOpts := NewMouseClickOptions()
 	if err := mouseOpts.Parse(m.ctx, opts); err != nil {
-		k6ext.Panic(m.ctx, "failed parsing options: %w", err)
+		k6ext.Panic(m.ctx, "parsing click options: %w", err)
 	}
 	if err := m.click(x, y, mouseOpts); err != nil {
-		k6ext.Panic(m.ctx, "unable to mouse click: %w", err)
+		k6ext.Panic(m.ctx, "mouse click: %w", err)
```
inancgumus, line 159:
```suggestion
		k6ext.Panic(m.ctx, "clicking with mouse: %w", err)
```

common/mouse.go
```diff
@@ -153,53 +153,53 @@ func (m *Mouse) up(x float64, y float64, opts *MouseDownUpOptions) error {
...
 
 func (m *Mouse) DblClick(x float64, y float64, opts goja.Value) {
 	mouseOpts := NewMouseDblClickOptions()
 	if err := mouseOpts.Parse(m.ctx, opts); err != nil {
-		k6ext.Panic(m.ctx, "failed parsing options: %w", err)
+		k6ext.Panic(m.ctx, "parsing dblclick options: %w", err)
 	}
 	if err := m.dblClick(x, y, mouseOpts); err != nil {
-		k6ext.Panic(m.ctx, "unable to mouse double click: %w", err)
+		k6ext.Panic(m.ctx, "mouse double click: %w", err)
```
inancgumus, line 169:
```suggestion
		k6ext.Panic(m.ctx, "double clicking with mouse: %w", err)
```

common/mouse.go
```diff
@@ -153,53 +153,53 @@ func (m *Mouse) up(x float64, y float64, opts *MouseDownUpOptions) error {
...
 // Down will trigger a MouseDown event in the browser.
 func (m *Mouse) Down(x float64, y float64, opts goja.Value) {
 	mouseOpts := NewMouseDownUpOptions()
 	if err := mouseOpts.Parse(m.ctx, opts); err != nil {
-		k6ext.Panic(m.ctx, "failed parsing options: %w", err)
+		k6ext.Panic(m.ctx, "parsing down options: %w", err)
 	}
 	if err := m.down(x, y, mouseOpts); err != nil {
-		k6ext.Panic(m.ctx, "unable to mouse down: %w", err)
+		k6ext.Panic(m.ctx, "mouse down: %w", err)
```
inancgumus, line 180:
```suggestion
		k6ext.Panic(m.ctx, "pressing mouse button: %w", err)
```

common/mouse.go
```diff
@@ -153,53 +153,53 @@ func (m *Mouse) up(x float64, y float64, opts *MouseDownUpOptions) error {
...
 // Move will trigger a MouseMoved event in the browser.
 func (m *Mouse) Move(x float64, y float64, opts goja.Value) {
 	mouseOpts := NewMouseDownUpOptions()
 	if err := mouseOpts.Parse(m.ctx, opts); err != nil {
-		k6ext.Panic(m.ctx, "failed parsing options: %w", err)
+		k6ext.Panic(m.ctx, "parsing move options: %w", err)
 	}
 	if err := m.down(x, y, mouseOpts); err != nil {
-		k6ext.Panic(m.ctx, "unable to move mouse: %w", err)
+		k6ext.Panic(m.ctx, "mouse move: %w", err)
```
inancgumus, line 191:
```suggestion
		k6ext.Panic(m.ctx, "moving mouse to x:%d y:%y: %w", x, y, err)
```

common/mouse.go
```diff
@@ -153,53 +153,53 @@ func (m *Mouse) up(x float64, y float64, opts *MouseDownUpOptions) error {
...
 // Up will trigger a MouseUp event in the browser.
 func (m *Mouse) Up(x float64, y float64, opts goja.Value) {
 	mouseOpts := NewMouseDownUpOptions()
 	if err := mouseOpts.Parse(m.ctx, opts); err != nil {
-		k6ext.Panic(m.ctx, "failed parsing options: %w", err)
+		k6ext.Panic(m.ctx, "parsing up options: %w", err)
 	}
 	if err := m.up(x, y, mouseOpts); err != nil {
-		k6ext.Panic(m.ctx, "unable to mouse up: %w", err)
+		k6ext.Panic(m.ctx, "mouse up: %w", err)
```
inancgumus, line 202:
```suggestion
		k6ext.Panic(m.ctx, "releasing mouse button: %w", err)
```

common/network_manager.go
```diff
@@ -330,7 +330,7 @@ func (m *NetworkManager) initDomains() error {
 	}
 	for _, action := range actions {
 		if err := action.Do(cdp.WithExecutor(m.ctx, m.session)); err != nil {
-			return fmt.Errorf("unable to execute %T: %w", action, err)
+			return fmt.Errorf("executing %T: %w", action, err)
```
inancgumus, line 333:
```suggestion
			return fmt.Errorf("initializing CDP domain %T: %w", action, err)
```

common/page.go
```diff
@@ -149,7 +149,7 @@ func NewPage(
 
 	action := target.SetAutoAttach(true, true).WithFlatten(true)
 	if err := action.Do(cdp.WithExecutor(p.ctx, p.session)); err != nil {
-		return nil, fmt.Errorf("cannot execute %T: %w", action, err)
+		return nil, fmt.Errorf("executing CDP action %T: %w", action, err)
```
inancgumus, line 152:
```suggestion
		return nil, fmt.Errorf("auto-attaching to page frames: %w", err)
```

common/worker.go
```diff
@@ -76,7 +76,7 @@ func (w *Worker) initEvents() error {
 	}
 	for _, action := range actions {
 		if err := action.Do(cdp.WithExecutor(w.ctx, w.session)); err != nil {
-			return fmt.Errorf("unable to execute %T: %w", action, err)
+			return fmt.Errorf("executing %T: %w", action, err)
```
inancgumus, line 79:
```suggestion
			return fmt.Errorf("protocol error while initializing worker %T: %w", action, err)
```

log/logger.go
```diff
@@ -179,7 +179,7 @@ func goRoutineID() int {
 	idField := strings.Fields(strings.TrimPrefix(string(buf[:n]), "goroutine "))[0]
 	id, err := strconv.Atoi(idField)
 	if err != nil {
-		panic(fmt.Sprintf("cannot get goroutine id: %v", err))
+		panic(fmt.Sprintf("getting goroutine ID: %v", err))
```
inancgumus, line 182:
```suggestion
		panic(fmt.Sprintf("internal error while getting goroutine ID: %v", err))
```

tests/page_test.go
```diff
@@ -115,13 +115,13 @@ func TestPageEvaluate(t *testing.T) {
 			{
 				"promise",
 				`async () => { return await new Promise((res, rej) => { rej('rejected'); }); }`,
-				`exception "Uncaught (in promise)" (0:0): rejected`,
+				"evaluating JS: rejected",
```
inancgumus, line 118:
```suggestion
				"promise rejected",
```

common/browser.go
```diff
@@ -205,15 +205,15 @@ func (b *Browser) initEvents() error {
 
 	action := target.SetAutoAttach(true, true).WithFlatten(true)
 	if err := action.Do(cdp.WithExecutor(b.ctx, b.conn)); err != nil {
-		return fmt.Errorf("unable to execute %T: %w", action, err)
+		return fmt.Errorf("executing setAutoAttach: %w", err)
```
inancgumus, line 208: Would "executing setAutoAttach" be meaningful to users? Could we say:

```suggestion
		return fmt.Errorf("auto-attaching to browser frames: %w", err)
```

inancgumus (changes requested):
I tried to test each error message locally. Here are some suggestions I could spot 🙇 

I hope most should be pretty easy to update. Since we're about to release, we might forego some of them if you think they will take a lot of your time.

PS: As you might already know, you might want to just ["batch-commit"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request) most directly on Github as ~they contain suggestion tags.

chromium/browser_type.go
inancgumus: ? 🤔 

```suggestion
		return nil, nil, fmt.Errorf("browser executable does not exist: %s", path)
```

chromium/browser_type.go
inancgumus: @ankur, that would result in a linter error `wrapCheck`.

inancgumus:
> * Is there a need to distinguish between an "internal" error vs an "user" error? Maybe internal errors could be `fatal` and user errors could be `error`?

👍 🤔 

> * Can we add the function name at the beginning of the error message so it helps the user find what function was last called when the error occurred and to give even more context to the error?

Yeah, I was asking about it [here](https://github.com/grafana/xk6-browser/issues/304#issuecomment-1167245835) 😅 Ivan fairly thinks users can find the problematic line in the stack trace.

inancgumus:
We can change the followings as well:

```go
func (h *ElementHandle) QueryAll(selector string) []api.ElementHandle {
	...
	if err != nil {
		k6ext.Panic(h.ctx, "querying all with selector %q: %w", selector, err)
	}
	...
}
```

```go
func (m *FrameManager) WaitForFrameNavigation(frame *Frame, opts goja.Value) api.Response {
	...
	if err := parsedOpts.Parse(m.ctx, opts); err != nil {
		k6ext.Panic(m.ctx, "parsing waitForNavigation options: %v", err)
	}
	...
	// there are a few ones to update here in this method
}
```

chromium/browser_type.go
ankur22: Ah, of course. Do we need that check?

ankur22:
> Yeah, I was asking about it [here](https://github.com/grafana/xk6-browser/issues/304#issuecomment-1167245835) 😅 Ivan fairly thinks users can find the problematic line in the stack trace.

Yep, that is true, although as an end user I think looking through stack traces isn't the best user experience (reminds me of debugging issues with Ansible 😭). Then again if there are multiple uses of a function in a test (e.g. `go.To`), how would the user know which one to look at, unless they looked at the stack trace 🤔

chromium/browser_type.go
imiric: Yeah, that's why the empty wrapping. I think we should leave that linter enabled, yes.

chromium/browser_type.go
imiric: :thinking: But we don't know `path` refers to a browser executable at this point. `os.IsNotExist()` simply checks for the existance of a file. Whether that points to a browser executable and whether we can ultimately run it should be mentioned in a different error message IMO.

common/frame.go
imiric: Yeah, as mentioned below, we're wrapping it to avoid a `wrapcheck` linter error.

examples/304-log-errors.js
imiric: The `execution context with ID 3 not found` is a race condition that we should fix, so it's not the same scenario as a timeout. It's a slight improvement over the previous `Cannot find context with specified id (-32000)` message, and since there's no friendlier way of logging this, we have to expose the internal [CDP concept of "execution contexts"](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#type-ExecutionContextDescription).

That said, I was wrong to mention that this can actually fail with a timeout. Notice that our [`Page.Query` method](https://github.com/grafana/xk6-browser/blob/f6b0363b12fdca200bb943a92a75c0dfacf91067/common/page.go#L673) doesn't actually receive an options object, and the [Playwright docs](https://playwright.dev/docs/api/class-page#page-query-selector) don't mention it either. So timeouts just aren't implemented for this method.

inancgumus: If our goal is to make error messages user-friendly, even if we don't say "timeout", I [agree](https://github.com/grafana/xk6-browser/pull/429#issuecomment-1173665476) with @ankur22 that we might print a user-friendly message such as "not found" (or something in that sense) instead of the cryptic "execution context".

imiric: But it's not cryptic :smile: Cryptic would be hiding the actual error to avoid confusing the user. This is something we need to fix so that the issue never happens, but saying anything else than "execution context not found" would be even more confusing.

If you have a specific suggestion let me know, but we can't just say "not found".

inancgumus: The execution context error means that we couldn't find the element. Of course, we shouldn't hide the actual error; but I believe we should *translate it into the user's language*. The user shouldn't care whether we have an underlying race condition. "execution context with ID x not found" is confusing to users 😒

Like I explained [here](https://github.com/grafana/xk6-browser/pull/429#issuecomment-1173665476), we can tell the user we could not find the element and link to a document (in our repository) that explains the possible causes of this error _until we fix it_. By the way, there is no guarantee that it will never happen.

My two cents.

chromium/browser_type.go
```diff
@@ -138,14 +138,14 @@ func (b *BrowserType) Launch(opts goja.Value) api.Browser {
...
-		k6common.Throw(rt, fmt.Errorf("cannot make logger: %w", err))
+		k6common.Throw(rt, fmt.Errorf("setting up logger: %w", err))
 	}
 
 	flags := prepareFlags(launchOpts, &state.Options)
 
 	dataDir := b.storage
 	if err := dataDir.Make("", flags["user-data-dir"]); err != nil {
-		k6common.Throw(rt, fmt.Errorf("cannot make temp data directory: %w", err))
+		k6common.Throw(rt, fmt.Errorf("making user data directory: %w", err))
```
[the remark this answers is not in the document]
imiric: Take a look at 8f3905e.

The message is now:
```
making browser data directory "somedir/xk6-browser-data-1283455864": permission denied
```

examples/304-log-errors.js
inancgumus: I can't think of another cause other than a race condition 🤔 What we've seen so far, is this error only occurred because of an execution context switch, and we look for the element in a stale frame. Either the script code races with our internals, or we race with ourselves/CDP.

We can tell users in a linked issue/wiki that it might happen because of navigation and suggest possible solutions. If the proposed solutions won't fix their problem, we can advise them to open an issue or use the community forum.

chromium/browser_type.go
```diff
@@ -422,14 +423,14 @@ func parseWebsocketURL(ctx context.Context, rc io.Reader) (wsURL string, _ error
 			}
 		}
 		if err := scanner.Err(); err != nil {
-			c <- result{"", fmt.Errorf("scanner err: %w", err)}
+			c <- result{"", err}
```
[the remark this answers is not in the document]
inancgumus: Yeah, I know :)

common/browser.go
imiric: Most of the errors relating to CDP will be confusing to users, so I don't think "auto-attaching to browser frames" would be an improvement. I think what this actually does is [enable auto-attaching to related targets](https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-setAutoAttach), whatever that means, but even that would be confusing.

Maybe we should translate all of these errors into a generic "initializing events" error, but we can discuss that in a new issue.

common/frame.go
```diff
@@ -567,7 +567,7 @@ func (f *Frame) waitForFunction(
 
 	execCtx := f.executionContexts[world]
 	if execCtx == nil {
-		return nil, fmt.Errorf("cannot find execution context: %q", world)
+		return nil, fmt.Errorf("execution context %q not found", world)
```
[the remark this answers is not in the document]
imiric: "Execution context" is a CDP concept, so yes, it's confusing to users, but "function not found on the frame" is an incorrect message in this case. As a user, I'd rather see confusing than wrong error messages.

Since we'll have to agree on the right approach for this, let's discuss it in a new issue.

common/browser.go
inancgumus: But, in that code's context, it's auto-attaching to browser pages (_not frames, sorry_). It sounds better to me than "executing setAutoAttach" or "initializing setAutoAttach" event because users can understand that we're tracking the browser pages. Anyway, let's discuss it later then.

common/mouse.go
imiric: I'm on the fence about these `mouse` ones... "pressing" and "releasing" requires a mental mapping to `down`/`up`, whereas mentioning the method is immediately clear.

So I don't think we gain much from translating these into prose and would rather go sligthly against the convention and use `mouse <method>` in these errors. But we can discuss it in that new issue.

tests/page_test.go
imiric: Makes sense, but I'll leave it for the new issue.

chromium/browser_type.go
```diff
@@ -217,12 +217,12 @@ func (b *BrowserType) allocate(
...
 	if err != nil {
-		return nil, fmt.Errorf("cannot start browser: %w", err)
+		return nil, err
 	}
 
 	wsURL, err := parseWebsocketURL(ctx, stdout)
 	if err != nil {
-		return nil, fmt.Errorf("cannot parse websocket url: %w", err)
+		return nil, fmt.Errorf("getting DevTools URL: %w", err)
 	}
```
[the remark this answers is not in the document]
imiric: Sure, I'll add this to the items in the new issue.

chromium/browser_type.go
```diff
@@ -371,15 +371,16 @@ func execute(
...
-		return nil, nil, fmt.Errorf("does not exist: %s", path)
+		return nil, nil, fmt.Errorf("file does not exist: %s", path)
 	}
 	if err != nil {
 		return nil, nil, fmt.Errorf("%w", err)
 	}
 	if ctx.Err() != nil {
-		return nil, nil, fmt.Errorf("context err: %w", ctx.Err())
+		return nil, nil, fmt.Errorf("%w", ctx.Err())
 	}
```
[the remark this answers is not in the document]
imiric: Sure, there are a lot of cases like this. This PR wasn't an attempt to fix all of them, but I'll mention this in the new issue as well.

tests/page_test.go
imiric: Actually, we should probably keep "evaluating JS", since that's what's happening, and removing it wouldn't make sense for the below cases (`SyntaxError`, `ReferenceError`, etc.). But adding "promise" would be clearer.

Issue #416, opened 2022-06-24
Allow specifying log level per component


imiric:
Allow specifying log level per component

A long time ago, in ~a galaxy~ #131, we introduced the undocumented `XK6_BROWSER_LOG` environment variable, used to control the log level globally. At the time, this was only useful for developers, but since then we've received numerous feedback that our log output is too noisy and that users would like to control it in some way.

We've been [suggesting that env var to some users](https://community.k6.io/t/is-there-a-way-to-suppress-warnings-in-the-console/3636/2), but it's clear that even that wasn't enough.

Part of the problem is the noisiness of our log output, which we're addressing in #304, so this issue is to implement a way of specifying the log level per "component".

## Proposed Solution

There's [a discussion here](https://github.com/grafana/xk6-browser/discussions/157) about possible approaches.

I think we agree that we should continue to use an environment variable to configure this, so we don't need any JS options or CLI flags (which we currently couldn't implement as a k6 extension either).

But there are a few open questions there we should agree on:

1. What should the environment variable be named?
   Inanc suggested using `DEBUG` as Playwright does, or `K6B`.
   Ivan prefers keeping it as `XK6_BROWSER_LOG`.

2. What are "components" exactly? They surely wouldn't map to specific Go structs. So can they be "categories", [as we currently pass to all log events](https://github.com/grafana/xk6-browser/blob/eb136726e37282fffad0c642062a04a6680a6adc/log/logger.go#L84)? We even have a `categoryFilter` regex, though it's currently not in use.

    It makes sense to have `cdp` as that's a current category, and maybe `browser` or `console`, as in "JS console".

    Which others? We preferably shouldn't have more than a handful of these, so we probably shouldn't separate them in internal categories like `page` or `frame`. Maybe use `internal` for all these "common" categories?

3. Should we have sub-components like Playwright's `DEBUG` has? E.g. `command`, `response`, etc. I think we don't need this level of granularity yet.

In any case, we should predefine a list of all components we want to have. All logged events should be part of a category. So maybe consider making that first `Logger` parameter an enum.

inancgumus:
From a practicality angle, I'd love to see what a component does and what CDP messages it sends and receives. For example, I can filter by the `Page` component and the `click` action and see something like:

```
Page.click
    -> NewFrameClickOptions.Parse (...parameters here...)
    -> ElementHandle.click(...parameters here...)
         -> Mouse.click (...parameters here...)
              -> ...
              -> CDP:Send <json here>
[...]

PR #438, opened 2022-07-07
Improve more log messages and auto detect timeout errors

### LINKED ISSUE #434: Improve more log messages
In #429 we made some log messages more concise and user-friendlier, but as shown by the comments by @inancgumus, there are many more errors we can improve.

Here's a summary of the leftover work:

- [x] [`parseWebsocketURL`](https://github.com/grafana/xk6-browser/pull/429/files#r912904231)
- [x] [`parseArgs`](https://github.com/grafana/xk6-browser/pull/429/files#r912905056)
- [x] Context errors: [1](https://github.com/grafana/xk6-browser/pull/429/files#r912908037), [2](https://github.com/grafana/xk6-browser/pull/429/files#r912908898)
- [x] [`makeLogger`](https://github.com/grafana/xk6-browser/pull/429/files#r912905641)
- [x] CDP errors: [1](https://github.com/grafana/xk6-browser/pull/429/files#r912943624), [2](https://github.com/grafana/xk6-browser/pull/429/files#r912912428), [3](https://github.com/grafana/xk6-browser/pull/429/files#r912932733), [4](https://github.com/grafana/xk6-browser/pull/429/files#r912939997)
- [x] `Mouse` errors: [1](https://github.com/grafana/xk6-browser/pull/429/files#r912934624), [2](https://github.com/grafana/xk6-browser/pull/429/files#r912935111), [3](https://github.com/grafana/xk6-browser/pull/429/files#r912935808), [4](https://github.com/grafana/xk6-browser/pull/429/files#r912936157), [5](https://github.com/grafana/xk6-browser/pull/429/files#r912936838)
- [x] `Mouse` up/down: [1](https://github.com/grafana/xk6-browser/pull/429#discussion_r913857307)
- [x] [`Promise` rejection](https://github.com/grafana/xk6-browser/pull/429/files#r912943265)
- [x] Execution context errors [1](https://github.com/grafana/xk6-browser/pull/429#issuecomment-1173665476), [2](https://github.com/grafana/xk6-browser/pull/429#discussion_r913637581), [3](https://github.com/grafana/xk6-browser/pull/429#discussion_r912928070)
- [x] CDP "executing" logs [1](https://github.com/grafana/xk6-browser/pull/429#discussion_r912912428), [2](https://github.com/grafana/xk6-browser/pull/429#discussion_r912932733)
- [x] [Evaluating JS](https://github.com/grafana/xk6-browser/pull/429#discussion_r913861266)
- [x] [Other left-overs](https://github.com/grafana/xk6-browser/pull/429#issuecomment-1173774129)

We might need to discuss some of these and reach an agreement about how to handle them.

inancgumus:
I added an "internal error" prefix to some messages as per @ankur22's [suggestion](https://github.com/grafana/xk6-browser/pull/429#issuecomment-1173700059). Not a category per se, but the prefix indicates that something went internally wrong. So users can hopefully see whether the problem is because of their scripts or ours.

> Is there a need to distinguish between an "internal" error vs. a "user" error? Maybe internal errors could be fatal, and user errors could be just errors?

**Update:**
This PR improves the auto-detection of timeout errors and makes them user-friendly as well.

### DIFF
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -447,8 +447,11 @@ func makeLogger(ctx context.Context, launchOpts *common.LaunchOptions) (*log.Log
 		_ = logger.SetLevel("debug")
 	}
 	if el, ok := os.LookupEnv("XK6_BROWSER_LOG"); ok {
-		if err := logger.SetLevel(el); err != nil {
-			return nil, fmt.Errorf("setting logger level to %q: %w", el, err)
+		if logger.SetLevel(el) != nil {
+			return nil, fmt.Errorf(
+				"invalid log level %q, should be one of: panic, fatal, error, warn, warning, info, debug, trace",
+				el,
+			)
 		}
 	}
 	if _, ok := os.LookupEnv("XK6_BROWSER_CALLER"); ok {
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -241,7 +241,7 @@ func parseArgs(flags map[string]interface{}) ([]string, error) {
 				args = append(args, fmt.Sprintf("--%s", name))
 			}
 		default:
-			return nil, fmt.Errorf(`invalid browser command line flag: "%s=%s"`, name, value)
+			return nil, fmt.Errorf(`invalid browser command line flag: "%s=%v"`, name, value)
 		}
 	}
 	if _, ok := flags["no-sandbox"]; !ok && os.Getuid() == 0 {
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -220,7 +220,7 @@ func (b *BrowserType) allocate(
 		return nil, err
 	}
 
-	wsURL, err := parseWebsocketURL(ctx, stdout)
+	wsURL, err := parseDevToolsURL(ctx, stdout)
 	if err != nil {
 		return nil, fmt.Errorf("getting DevTools URL: %w", err)
 	}
@@ -402,11 +402,11 @@ func execute(
 	return cmd, stdout, nil
 }
 
-// parseWebsocketURL grabs the websocket address from chrome's output and returns it.
-func parseWebsocketURL(ctx context.Context, rc io.Reader) (wsURL string, _ error) {
+// parseDevToolsURL grabs the websocket address from chrome's output and returns it.
+func parseDevToolsURL(ctx context.Context, rc io.Reader) (wsURL string, _ error) {
 	type result struct {
-		wsURL string
-		err   error
+		devToolsURL string
+		err         error
 	}
 	c := make(chan result, 1)
 	go func() {
@@ -428,7 +428,7 @@ func parseWebsocketURL(ctx context.Context, rc io.Reader) (wsURL string, _ error
 	}()
 	select {
 	case r := <-c:
-		return r.wsURL, r.err
+		return r.devToolsURL, r.err
 	case <-ctx.Done():
 		return "", fmt.Errorf("%w", ctx.Err())
 	}
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -129,6 +129,10 @@ func (b *BrowserType) Launch(opts goja.Value) api.Browser {
 
 	bp, err := b.launch(launchOpts)
 	if err != nil {
+		err = &k6ext.UserFriendlyError{
+			Err:     err,
+			Timeout: launchOpts.Timeout,
+		}
 		k6ext.Panic(b.Ctx, "%w", err)
 	}
 
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -152,7 +152,7 @@ func (b *BrowserType) Launch(opts goja.Value) api.Browser {
 	go func(ctx context.Context) {
 		defer func() {
 			if err := dataDir.Cleanup(); err != nil {
-				logger.Errorf("BrowserType:Launch", "%v", err)
+				logger.Errorf("BrowserType:Launch", "cleaning up the user data directory: %v", err)
 			}
 		}()
 		// There's a small chance that this might be called
@@ -383,19 +383,22 @@ func execute(
 		// TODO: How to handle these errors?
 		defer func() {
 			if err := dataDir.Cleanup(); err != nil {
-				logger.Errorf("BrowserType:execute", "%v", err)
+				logger.Errorf("BrowserType:Close", "cleaning up the user data directory: %v", err)
 			}
 		}()
 
 		if err := cmd.Wait(); err != nil {
-			log := logger.Errorf
-			if err.Error() == "signal: killed" || err.Error() == "exit status 1" {
+			logErr := logger.Errorf
+			if s := err.Error(); strings.Contains(s, "signal: killed") || strings.Contains(s, "exit status 1") {
 				// The browser process is killed when the context is cancelled
 				// after a k6 iteration ends, so silence the log message until
 				// we can stop it gracefully. See #https://github.com/grafana/xk6-browser/issues/423
-				log = logger.Debugf
+				logErr = logger.Debugf
 			}
-			log("BrowserType:execute", "browser process with PID %d ended unexpectedly: %v", cmd.Process.Pid, err)
+			logErr(
+				"browser", "process with PID %d unexpectedly ended: %v",
+				cmd.Process.Pid, err,
+			)
 		}
 	}()
 
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -121,31 +121,37 @@ func (b *BrowserType) ExecutablePath() (execPath string) {
 // Launch allocates a new Chrome browser process and returns a new api.Browser value,
 // which can be used for controlling the Chrome browser.
 func (b *BrowserType) Launch(opts goja.Value) api.Browser {
-	var (
-		rt         = b.vu.Runtime()
-		state      = b.vu.State()
-		launchOpts = common.NewLaunchOptions()
-	)
+	launchOpts := common.NewLaunchOptions()
 	if err := launchOpts.Parse(b.Ctx, opts); err != nil {
-		k6common.Throw(rt, fmt.Errorf("parsing launch options: %w", err))
+		k6ext.Panic(b.Ctx, "parsing launch options: %w", err)
 	}
 	b.Ctx = common.WithLaunchOptions(b.Ctx, launchOpts)
 
-	envs := make([]string, 0, len(launchOpts.Env))
-	for k, v := range launchOpts.Env {
+	bp, err := b.launch(launchOpts)
+	if err != nil {
+		k6ext.Panic(b.Ctx, "%w", err)
+	}
+
+	return bp
+}
+
+func (b *BrowserType) launch(opts *common.LaunchOptions) (*common.Browser, error) {
+	envs := make([]string, 0, len(opts.Env))
+	for k, v := range opts.Env {
 		envs = append(envs, fmt.Sprintf("%s=%s", k, v))
 	}
 
-	logger, err := makeLogger(b.Ctx, launchOpts)
+	logger, err := makeLogger(b.Ctx, opts)
 	if err != nil {
-		k6common.Throw(rt, fmt.Errorf("setting up logger: %w", err))
+		return nil, fmt.Errorf("setting up logger: %w", err)
 	}
 
-	flags := prepareFlags(launchOpts, &state.Options)
-
-	dataDir := b.storage
+	var (
+		flags   = prepareFlags(opts, &(b.vu.State()).Options)
+		dataDir = b.storage
+	)
 	if err := dataDir.Make("", flags["user-data-dir"]); err != nil {
-		k6common.Throw(rt, err)
+		return nil, fmt.Errorf("%w", err)
 	}
 	flags["user-data-dir"] = dataDir.Dir
 
@@ -163,9 +169,9 @@ func (b *BrowserType) Launch(opts goja.Value) api.Browser {
 		<-ctx.Done()
 	}(b.Ctx)
 
-	browserProc, err := b.allocate(launchOpts, flags, envs, dataDir, logger)
+	browserProc, err := b.allocate(opts, flags, envs, dataDir, logger)
 	if browserProc == nil {
-		k6ext.Panic(b.Ctx, "launching browser: %s", err)
+		return nil, fmt.Errorf("launching browser: %w", err)
 	}
 
 	browserProc.AttachLogger(logger)
@@ -174,12 +180,12 @@ func (b *BrowserType) Launch(opts goja.Value) api.Browser {
 	// so that we can kill it afterward if it lingers
 	// see: k6ext.Panic function.
 	b.Ctx = k6ext.WithProcessID(b.Ctx, browserProc.Pid())
-	browser, err := common.NewBrowser(b.Ctx, b.CancelFn, browserProc, launchOpts, logger)
+	browser, err := common.NewBrowser(b.Ctx, b.CancelFn, browserProc, opts, logger)
 	if err != nil {
-		k6common.Throw(rt, err)
+		return nil, fmt.Errorf("launching browser: %w", err)
 	}
 
-	return browser
+	return browser, nil
 }
 
 // LaunchPersistentContext launches the browser with persistent storage.
--- a/common/browser.go
+++ b/common/browser.go
@@ -382,7 +382,7 @@ func (b *Browser) newPageInContext(id cdp.BrowserContextID) (*Page, error) {
 	action := target.CreateTarget("about:blank").WithBrowserContextID(id)
 	tid, err := action.Do(cdp.WithExecutor(ctx, b.conn))
 	if err != nil {
-		return nil, fmt.Errorf("%T: %w", action, err)
+		return nil, fmt.Errorf("creating a new blank page: %w", err)
 	}
 	// let the event handler know about the new page.
 	targetID <- tid
--- a/common/browser.go
+++ b/common/browser.go
@@ -404,7 +404,7 @@ func (b *Browser) newPageInContext(id cdp.BrowserContextID) (*Page, error) {
 func (b *Browser) Close() {
 	defer func() {
 		if err := b.browserProc.userDataDir.Cleanup(); err != nil {
-			b.logger.Errorf("Browser:Close", "%v", err)
+			b.logger.Errorf("Browser:Close", "cleaning up the user data directory: %v", err)
 		}
 	}()
 
--- a/common/browser.go
+++ b/common/browser.go
@@ -260,7 +260,7 @@ func (b *Browser) onAttachedToTarget(ev *target.EventAttachedToTarget) {
 					ev.SessionID, evti.TargetID, b.ctx.Err())
 				return // ignore
 			default:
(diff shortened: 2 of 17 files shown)


common/execution_context.go
```diff
@@ -152,10 +152,10 @@ func (e *ExecutionContext) adoptElementHandle(eh *ElementHandle) (*ElementHandle
 		efid, esid)
 
 	if eh.execCtx == e {
-		panic("cannot adopt handle that already belongs to this execution context")
+		return nil, errors.New("already belongs to the execution context")
```
inancgumus, line 155: 🤔 

```suggestion
			return nil, errors.New("already belongs to the same execution context")
```

common/execution_context.go
```diff
@@ -152,10 +152,10 @@ func (e *ExecutionContext) adoptElementHandle(eh *ElementHandle) (*ElementHandle
...
 	}
 	if e.frame == nil {
-		panic("cannot adopt handle without frame owner")
+		return nil, errors.New("does not have a frame owner")
```
inancgumus, line 158: Note: The higher layer already says about "adopting".

chromium/browser_type.go
```diff
@@ -383,19 +383,19 @@ func execute(
...
 		if err := cmd.Wait(); err != nil {
-			log := logger.Errorf
 			if err.Error() == "signal: killed" || err.Error() == "exit status 1" {
 				// The browser process is killed when the context is cancelled
 				// after a k6 iteration ends, so silence the log message until
 				// we can stop it gracefully. See #https://github.com/grafana/xk6-browser/issues/423
-				log = logger.Debugf
+				logger.Errorf("closing the browser", "process with PID %d unexpectedly ended: %v", cmd.Process.Pid, err)
+			} else {
+				logger.Debugf("BrowserType:Execute", "process with PID %d unexpectedly ended: %v", cmd.Process.Pid, err)
```
inancgumus, line 397: Indecisive about this done. I thought `BrowserType:Execute` would be meaningless to users.

common/connection.go
```diff
@@ -234,7 +235,7 @@ func (c *Connection) createSession(info *target.Info) (*Session, error) {
 }
 
 func (c *Connection) handleIOError(err error) {
-	c.logger.Errorf("Connection:handleIOError", "err:%v", err)
+	c.logger.Errorf("Connection:handleIOError", "internal error communicating over WebSocket: %v", err)
```
inancgumus, line 238: This is a super common error. So we should better think about this more.

common/frame_session.go
```diff
@@ -933,7 +933,7 @@ func (fs *FrameSession) updateGeolocation(initial bool) error {
 			WithLongitude(geolocation.Longitude).
 			WithAccuracy(geolocation.Accurracy)
 		if err := action.Do(cdp.WithExecutor(fs.ctx, fs.session)); err != nil {
-			return fmt.Errorf("overriding geolocation: %w", err)
+			return fmt.Errorf("%w", err)
```
inancgumus, line 936: The upper layer says already.

common/element_handle.go
```diff
@@ -687,7 +687,7 @@ func (h *ElementHandle) waitForElementState(
 	}
 
 	return false, fmt.Errorf(
-		"checking states %v of element: %q", states, reflect.TypeOf(result))
+		"waiting states %v of element %q", states, reflect.TypeOf(result))
```
inancgumus, line 690: 🤔 

```suggestion
		"waiting for states %v of element %q", states, reflect.TypeOf(result))
```

common/element_handle.go
```diff
@@ -1079,7 +1079,7 @@ func (h *ElementHandle) queryAll(selector string, eval evalFunc) ([]api.ElementH
 		parsedSelector,
 	)
 	if err != nil {
-		return nil, fmt.Errorf("querying selector %q: %w", selector, err)
+		return nil, fmt.Errorf("querying all selectors %q: %w", selector, err)
```
inancgumus, line 1082: Oops. It might be:

```suggestion
		return nil, fmt.Errorf("querying all selector %q: %w", selector, err)
```

common/connection.go
```diff
@@ -378,20 +379,17 @@ func (c *Connection) send(ctx context.Context, msg *cdproto.Message, recvCh chan
...
-		return err
+		return fmt.Errorf("sending message to browser over WebSocket: %w", err)
 	case code := <-c.closeCh:
 		c.logger.Debugf("Connection:send:<-c.closeCh", "wsURL:%q sid:%v, websocket code:%v", c.wsURL, msg.SessionID, code)
 		_ = c.closeConnection(code)
-		return &websocket.CloseError{Code: code}
+		return fmt.Errorf("closing the WebSocket communication with the browser: %w", &websocket.CloseError{Code: code})
+	case <-ctx.Done():
+		c.logger.Debugf("Connection:send:<-ctx.Done", "wsURL:%q sid:%v err:%v", c.wsURL, msg.SessionID, c.ctx.Err())
+		return nil
```
inancgumus, line 389: This will probably lift some errors. There is no need for an error after closing.

chromium/browser_type.go
```diff
@@ -448,7 +448,7 @@ func makeLogger(ctx context.Context, launchOpts *common.LaunchOptions) (*log.Log
 	}
 	if el, ok := os.LookupEnv("XK6_BROWSER_LOG"); ok {
 		if err := logger.SetLevel(el); err != nil {
-			return nil, fmt.Errorf("setting logger level to %q: %w", el, err)
+			return nil, fmt.Errorf("%w", err)
```
ankur22, line 451: I think it make sense to notify the user of what was wrong and what they can do to fix it too (as asked in the commit comment).

ankur22, line 451: I think it make sense to notify the user of what was wrong and what they can do to fix it too (as asked in the commit comment).

k6ext/panic.go
```diff
@@ -36,3 +44,20 @@ func Panic(ctx context.Context, format string, a ...interface{}) {
...
+func (e *userFriendlyError) Unwrap() error { return e.err }
+
+func (e *userFriendlyError) Error() string {
+	switch {
+	default:
+		return e.err.Error()
+	case e.err == nil:
+		return ""
+	case errors.Is(e.err, context.DeadlineExceeded):
+		return strings.ReplaceAll(e.err.Error(), context.DeadlineExceeded.Error(), "timed out")
```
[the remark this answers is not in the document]
inancgumus: Sorry, I didn't get what you meant. This code only converts a `context deadline exceeded` with `timeout` if the context is canceled because of a timeout to print a user-friendly message.

chromium/browser_type.go
```diff
@@ -152,7 +152,7 @@ func (b *BrowserType) Launch(opts goja.Value) api.Browser {
 	go func(ctx context.Context) {
 		defer func() {
 			if err := dataDir.Cleanup(); err != nil {
-				logger.Errorf("BrowserType:Launch", "%v", err)
+				logger.Errorf("closing the browser", "%v", err)
```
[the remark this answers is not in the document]
ankur22: LGTM

k6ext/panic.go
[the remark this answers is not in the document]
ankur22: 🤦 I was getting confused with `ContextClosed`. LGTM 👍

common/element_handle.go
```diff
@@ -915,7 +915,7 @@ func (h *ElementHandle) InputValue(opts goja.Value) string {
 func (h *ElementHandle) IsChecked() bool {
 	result, err := h.isChecked(h.ctx, 0)
 	if err != nil && err != ErrTimedOut { // We don't care anout timeout errors here!
-		k6ext.Panic(h.ctx, "element isChecked: %w", err)
+		k6ext.Panic(h.ctx, "getting element is checked: %w", err)
```
imiric, line 918: `"getting element is checked"` sounds awkward. Same for the other `ElementHandle` errors.

Maybe:

```suggestion
		k6ext.Panic(h.ctx, "getting element checked value: %w", err)
```

common/execution_context.go
```diff
@@ -152,10 +152,10 @@ func (e *ExecutionContext) adoptElementHandle(eh *ElementHandle) (*ElementHandle
 		efid, esid)
 
 	if eh.execCtx == e {
-		panic("cannot adopt handle that already belongs to this execution context")
+		return nil, errors.New("already belongs to the same execution context")
```
imiric, line 155:
```suggestion
		return nil, errors.New("element handle already belongs to the same execution context")
```

common/execution_context.go
```diff
@@ -152,10 +152,10 @@ func (e *ExecutionContext) adoptElementHandle(eh *ElementHandle) (*ElementHandle
...
 	}
 	if e.frame == nil {
-		panic("cannot adopt handle without frame owner")
+		return nil, errors.New("does not have a frame owner")
```
imiric, line 158:
```suggestion
		return nil, errors.New("element handle does not have a frame owner")
```

common/execution_context.go
```diff
@@ -224,7 +224,7 @@ func (e *ExecutionContext) eval(
 	if remoteObject, exceptionDetails, err = action.Do(cdp.WithExecutor(apiCtx, e.session)); err != nil {
 		var cdpe *cdproto.Error
 		if errors.As(err, &cdpe) && cdpe.Code == -32000 {
-			err = fmt.Errorf("execution context with ID %d not found", e.id)
+			err = fmt.Errorf("execution context with ID %d not found; most likely because of a navigation", e.id)
```
imiric, line 227: This is better, but maybe we can remove the `"execution context not found"` part altogether. Something like:
```suggestion
			err = errors.New("execution context changed, most likely because of a navigation")
```

common/frame.go
```diff
@@ -567,7 +567,7 @@ func (f *Frame) waitForFunction(
 
 	execCtx := f.executionContexts[world]
 	if execCtx == nil {
-		return nil, fmt.Errorf("execution context %q not found", world)
+		return nil, fmt.Errorf("execution context %q not found; most likely because of a navigation", world)
```
imiric, line 570: I don't think this is right. A navigation wouldn't cause us to delete a `"main"` or `"utility"` context. They're replaced with contexts with different IDs, not outright deleted.

common/frame.go
```diff
@@ -656,14 +656,14 @@ func (f *Frame) waitForSelector(selector string, opts *FrameWaitForSelectorOptio
 
 	ec := f.executionContexts[mainWorld]
 	if ec == nil {
-		return nil, fmt.Errorf("wait for selector %q cannot find execution context: %q", selector, mainWorld)
+		return nil, fmt.Errorf("wait for selector %q execution context is missing: %q", selector, mainWorld)
```
imiric, line 659: The message is not grammatically correct now...

I think it should be consistent with the one in `waitForFunction`, so:
```suggestion
		return nil, fmt.Errorf("execution context %q not found", mainWorld)
```

Or `"execution context %q is missing"`, but it's the same lookup, so the messages should match.

And maybe this should be a panic. If it happens it's probably a bug on our side.

common/mouse.go
```diff
@@ -199,7 +199,7 @@ func (m *Mouse) Up(x float64, y float64, opts goja.Value) {
 		k6ext.Panic(m.ctx, "parsing up options: %w", err)
 	}
 	if err := m.up(x, y, mouseOpts); err != nil {
-		k6ext.Panic(m.ctx, "mouse up: %w", err)
+		k6ext.Panic(m.ctx, "releasing mouse up: %w", err)
```
imiric, line 202: `"releasing mouse up"` sounds redundant. If we're releasing a mouse click, it's a "mouse up" event. So maybe this should be just `"releasing mouse"`.

But I'm slightly against these mouse error changes. E.g. `"moving mouse"` is slightly wrong; the cursor/pointer is what moves. So I'd prefer keeping these as they were, since they just describe the actual `Mouse` method that was called, and are less confusing. Even if it goes against our new conventions.

common/page.go
```diff
@@ -149,7 +149,7 @@ func NewPage(
 
 	action := target.SetAutoAttach(true, true).WithFlatten(true)
 	if err := action.Do(cdp.WithExecutor(p.ctx, p.session)); err != nil {
-		return nil, fmt.Errorf("executing CDP action %T: %w", action, err)
+		return nil, fmt.Errorf("internal error while auto attaching to browser pages: %w", err)
```
imiric, line 152: Same comment here as the previous one about auto-attach.

k6ext/panic.go
imiric, line 59: As you mentioned in the commit message, it would be useful if we could log the deadline time as well. But if it's a lot of work, don't bother. This is an improvement already.

k6ext/panic.go
```diff
@@ -36,3 +44,20 @@ func Panic(ctx context.Context, format string, a ...interface{}) {
...
 }
+
+type userFriendlyError struct{ err error }
+
+func (e *userFriendlyError) Unwrap() error { return e.err }
+
+func (e *userFriendlyError) Error() string {
+	switch {
+	default:
+		return e.err.Error()
```
imiric, line 55: I'm curious why `default` is the first case. Is there a functional reason for it?

imiric (changes requested):
Good improvements overall, but I think we should standardize how we handle categories, and make some other minor changes.

I like the `userFriendlyError` and handling that in a single place :+1:

chromium/browser_type.go
inancgumus: > in which case BrowserType:Execute is fine

I was talking about the category of `Errorf`.

> "browser" makes sense to me.

OK, that makes sense to me too.

> Note that we are potentially losing the method that logged the message, but hopefully that should be clear from the message itself.

Most are there. We can search and find them on an error.

common/connection.go
inancgumus: `"cdp"` makes sense.

common/connection.go
```diff
@@ -378,20 +379,17 @@ func (c *Connection) send(ctx context.Context, msg *cdproto.Message, recvCh chan
 	case c.sendCh <- msg:
 	case err := <-c.errorCh:
 		c.logger.Debugf("Connection:send:<-c.errorCh", "wsURL:%q sid:%v, err:%v", c.wsURL, msg.SessionID, err)
-		return err
+		return fmt.Errorf("sending message to browser over WebSocket: %w", err)
```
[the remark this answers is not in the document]
inancgumus: Sure, I too agree with removing it. What's the difference between having it in our message? Users will see "websocket" at the end in any case.

common/browser.go
```diff
@@ -205,15 +205,15 @@ func (b *Browser) initEvents() error {
 
 	action := target.SetAutoAttach(true, true).WithFlatten(true)
 	if err := action.Do(cdp.WithExecutor(b.ctx, b.conn)); err != nil {
-		return fmt.Errorf("executing setAutoAttach: %w", err)
+		return fmt.Errorf("internal error while auto-attaching to browser pages: %w", err)
```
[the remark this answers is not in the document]
inancgumus: 👎  This doesn't sound right to me: `"enabling auto-attaching to related browser targets"`. As I said before, that CDP command attaches to browser pages. "Target" is a general term. It applies to browser pages here.

common/frame.go
inancgumus: So are you suggesting [this](https://github.com/grafana/xk6-browser/pull/438#discussion_r916713586)?

k6ext/panic.go
inancgumus: No functional reason. Only as a hint that most of the time the default one will match.

common/mouse.go
inancgumus: Earlier, I suggested using "releasing mouse", then you were against it. That's why I combined both. I can make it "releasing the mouse button" again. It's universal that everyone can understand that releasing the mouse button means mouse up.

We can say "moving the mouse pointer" instead of ditching them all.

common/browser.go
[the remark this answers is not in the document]
imiric: OK, let's keep "internal error" then.

> that CDP command attaches to browser pages

We are interpreting [the CDP docs](https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-setAutoAttach) differently then. To me it reads that it _enables_ auto-attaching to related targets, not that any attaching is being done in this command. I.e. when a new `Target.attachToTarget` command is received, then the auto-attach rule takes place. So in that case "while auto-attaching to browser pages" would be wrong.

But this is one of the many confusing parts of the CDP docs, and I don't want to spend a lot of time arguing about semantics, so if you feel strongly about this, let's leave it as is. I just think it's important that our log messages are correct.

common/connection.go
imiric: OK, let's leave `Debugf` logs as is then in this PR.

But I think we should at least standardize the category regardless of the log level we're using. Putting method names in the category field seems wrong, and we'll probably need to standardize it for #416.

common/frame.go
imiric: No, that's a different suggestion, where "most likely because of a navigation" probably makes sense. In this case, it doesn't.

common/mouse.go
imiric: OK, let's go with `"releasing the mouse button"` and `"moving the mouse pointer"` then. :+1:

inancgumus:
Ivan, I mistakenly deleted one of your comments, sorry.

common/element_handle.go
```diff
@@ -75,7 +75,7 @@ func (h *ElementHandle) checkHitTargetAt(apiCtx context.Context, point Position)
 			return false, err
 		}
 		if box == nil {
-			return false, errors.New("getting bounding box of element")
+			return false, errors.New("missing bounding box of element")
```
ankur22, line 78: Shouldn't this be:

```suggestion
			return false, errors.New("getting bounding box of element")
```

I thought errors describe the action that was occurring when something went wrong?

inancgumus: Most often, that's right. However, this last one is not an action, it's the final result of that action. Similar to the ones that we output errors from CDP/injectedScript actions. It's the final piece in the error chain.

We get the following:

```
tapping element: checking hit target: missing bounding box of element
```

However, the following doesn't make much sense:

```
tapping element: checking hit target: getting bounding box of element
```

It's because we didn't try to get the bounding box, instead, we couldn't find the bounding box. The action is the "checking hit target".

common/browser.go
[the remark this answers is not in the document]
ankur22:
```suggestion
		return fmt.Errorf("internal error while enabling auto-attach to browser pages: %w", err)
```

Just thought it was worth making it match the CDP docs, but happy for this suggestion to be ignored too.

common/browser_context.go
```diff
@@ -214,7 +214,7 @@ func (b *BrowserContext) GrantPermissions(permissions []string, opts goja.Value)
 
 	action := cdpbrowser.GrantPermissions(perms).WithOrigin(origin).WithBrowserContextID(b.id)
 	if err := action.Do(cdp.WithExecutor(b.ctx, b.browser.conn)); err != nil {
-		k6ext.Panic(b.ctx, "override permissions: %w", err)
+		k6ext.Panic(b.ctx, "internal error while overriding browser permissions: %w", err)
```
ankur22, line 217:
```suggestion
		k6ext.Panic(b.ctx, "internal error while granting browser permissions: %w", err)
```

ankur22 (approved):

PR #457, opened 2022-07-14
Disable wrapcheck linter in tests and refactor tests

inancgumus:
We decided to disable the wrapcheck linter for tests. This PR does that and removes `fmt.Errorf("%w", err)` from tests.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,9 +16,10 @@ issues:
   exclude-use-default: false
 
   exclude-rules:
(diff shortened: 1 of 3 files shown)


imiric (approved):
LGTM

BTW, do you think that `TestWaitForFrameNavigationWithinDocument` panic in CI is because of the usual test flakiness or something else? I don't remember seeing `click` panic with `context canceled` before.

PR #462, opened 2022-07-19
Make chromium importable

### LINKED ISSUE #460: Rename launcher to chromium
We might want to rename `launcher` to `chromium` to be semantically more correct, simpler, ready for future browser support, type-safe, and have better compatibility with Playwright.

The following is what we do at the moment to launch a Chromium browser (despite that we only support Chromium):

```js
import launcher from 'k6/x/browser';

const browser = launcher.launch('chromium');
browser.newPage();
```

We panic if something other than `"chromium"` is passed to the `launch()`. We don't need to panic if we directly import `chromium` from `k6/x/browser` as users can catch the problem before running their scripts (as Playwright):

```js

browser.newPage();
```

We can also be ready for future browser support as follows:

```js
import { chromium, firefox, webkit } from 'k6/x/browser';

browser.newPage();
firefox.launch();
 webkit.launch();
```

**Suggestions to make the change**

1. Remove the `browserName` parameter from `main.go#JSModule.Launch()`.
2. Refactor `main.go#JSModule.Launch()` to remove browser name check and panic.
3. Export `"chromium"` in `main.go#Exports()` to import it in a script as `import { chromium } from 'k6/x/browser'`.
4. Update the `README`.
5. Update the `examples/`.
6. Update the documentation (k6-docs).

**In summary (before/after)**

Before:

```js
import launcher from 'k6/x/browser';

const browser = launcher.launch('chromium')
```

After:

```js

```

inancgumus:
This PR closes #460. Please see the issue for details.

### DIFF
--- a/main.go
+++ b/main.go
@@ -62,6 +62,7 @@ func (mi *ModuleInstance) Exports() k6modules.Exports {
 	return k6modules.Exports{
 		Named: map[string]interface{}{
 			"chromium": mi.mod,
+			"devices":  mi.mod.Devices,
 		},
 	}
 }
--- a/main.go
+++ b/main.go
@@ -1,14 +1,11 @@
 package browser
 
 import (
-	"errors"
-
 	"github.com/grafana/xk6-browser/api"
 	"github.com/grafana/xk6-browser/chromium"
 	"github.com/grafana/xk6-browser/common"
 	"github.com/grafana/xk6-browser/k6ext"
 
-	k6common "go.k6.io/k6/js/common"
 	k6modules "go.k6.io/k6/js/modules"
 
 	"github.com/dop251/goja"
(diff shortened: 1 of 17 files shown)


ankur22 (approved):

main.go
```diff
@@ -82,10 +59,16 @@ func (*RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
 // Exports returns the exports of the JS module so that it can be used in test
 // scripts.
 func (mi *ModuleInstance) Exports() k6modules.Exports {
-	return k6modules.Exports{Default: mi.mod}
+	return k6modules.Exports{
+		Named: map[string]interface{}{
+			"chromium": mi.mod,
+			"devices":  mi.mod.Devices,
+		},
```
imiric, line 66: The JS API is better now :+1:, but we'll need to change this once we add support for other browsers. Probably by exporting each browser's `BrowserType` implementation.

PR #467, opened 2022-07-21
Async waitForNavigation

### LINKED ISSUE #440: Async: WaitForNavigation
A usage example can be seen as follows:

```js
import { check } from 'k6';

export default function() {
    headless: true,
  });

  page.goto('https://test.k6.io/', { waitUntil: 'networkidle' });
  const elem = page.$('a[href="/my_messages.php"]');
  elem.click().then(() => {
    page.$('input[name="login"]').type('admin');
    page.$('input[name="password"]').type('123');
  }).then(() =>
    page.$('input[type="submit"]').click()
  ).then(() =>
    page.waitForNavigation()
  ).then(() => {
    check(page, {
      'header': page.$('h2').textContent() == 'Welcome, admin!',
    });
  }).finally(() => {
    page.close();
    browser.close();
  });
}
```

Another example can be using `Promise.all` for `Click` and `WaitForNavigation` as in PW as follows:

```js

export default function() {
    headless: true,
  });

  page.setContent(`
    <html>
      <head>
        <title>Navigation test within the same document</title>
      </head>
      <body>
        <a id="nav-history" href="#">Navigate with History API</a>
        <a id="nav-anchor" href="#anchor">Navigate with anchor link</a>
        <div id="anchor">Some div...</div>
        <script>
          const el = document.querySelector('a#nav-history');
          el.addEventListener('click', function(evt) {
            evt.preventDefault();
            history.pushState({}, 'navigated', '/nav2');
          });
        </script>
      </body>
    </html>  
  `);

  Promise.all([
    page.click('a#nav-anchor'),
    page.waitForNavigation({timeout: 500}),
  ]).then(() => { 
    console.log("ok");
  }).catch(e => {
    console.log("err: "+ e);
  }).finally(() => {
    page.close();
    browser.close();
  })
}
```

inancgumus:
**Please feel free to add additional commits to the PR if you need to and merge it without me.**

Note: I would create another PR to fix the problem in `k6ext.promise`, but since I'll be leaving soon, I wanted to pack it in the same PR.

Closes grafana/xk6-browser#440.

### DIFF
--- a/examples/fillform.js
+++ b/examples/fillform.js
@@ -15,12 +15,13 @@ export default function() {
     // Enter login credentials and login
     page.$('input[name="login"]').type('admin');
     page.$('input[name="password"]').type('123');
-    return page.$('input[type="submit"]').click();
-  }).then(() => {
+  }).then(() =>
+    page.$('input[type="submit"]').click()
+  ).then(() =>
     // We expect the above form submission to trigger a navigation, so wait for it
     // and the page to be loaded.
-    page.waitForNavigation();
-
+    page.waitForNavigation()
+  ).then(() => {
     check(page, {
       'header': page.$('h2').textContent() == 'Welcome, admin!',
     });
--- a/examples/fillform.js
+++ b/examples/fillform.js
@@ -15,13 +15,14 @@ export default function() {
     // Enter login credentials and login
     page.$('input[name="login"]').type('admin');
     page.$('input[name="password"]').type('123');
-  }).then(() =>
-    page.$('input[type="submit"]').click()
-  ).then(() =>
-    // We expect the above form submission to trigger a navigation, so wait for it
-    // and the page to be loaded.
-    page.waitForNavigation()
-  ).then(() => {
+    // We expect the form submission to trigger a navigation, so to prevent a
+    // race condition, setup a waiter concurrently while waiting for the click
+    // to resolve.
+    return Promise.all([
+      page.waitForNavigation(),
+      page.$('input[type="submit"]').click(),
+    ]);
+  }).then(() => {
     check(page, {
       'header': page.$('h2').textContent() == 'Welcome, admin!',
     });
--- a/k6ext/k6test/vu.go
+++ b/k6ext/k6test/vu.go
@@ -30,6 +30,7 @@ func (v *VU) ToGojaValue(i interface{}) goja.Value { return v.Runtime().ToValue(
 
 // RunLoop is a convenience method for running fn in the event loop.
 func (v *VU) RunLoop(fn func() error) error {
+	v.Loop.WaitOnRegistered()
 	return v.Loop.Start(fn)
 }
 
--- a/tests/frame_manager_test.go
+++ b/tests/frame_manager_test.go
@@ -1,6 +1,8 @@
 package tests
 
 import (
+	"fmt"
+	"net/http"
 	"os"
 	"testing"
 	"time"
@@ -66,4 +68,46 @@ func TestWaitForFrameNavigationWithinDocument(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestWaitForFrameNavigation(t *testing.T) {
+	tb := newTestBrowser(t, withHTTPServer())
+	p := tb.NewPage(nil)
+
+	tb.withHandler("/first", func(w http.ResponseWriter, _ *http.Request) {
+		fmt.Fprintf(w, `
+		<html>
+			<head>
+				<title>First page</title>
+			</head>
+			<body>
+				<a href="/second">click me</a>
+			</body>
+		</html>
+		`)
+	})
+	tb.withHandler("/second", func(w http.ResponseWriter, _ *http.Request) {
+		fmt.Fprintf(w, `
+		<html>
+			<head>
+				<title>Second page</title>
+			</head>
+			<body>
+				<a href="/first">click me</a>
+			</body>
+		</html>
+		`)
+	})
+
+	require.NotNil(t, p.Goto(tb.URL("/first"), tb.toGojaValue(&common.FrameGotoOptions{
+		WaitUntil: common.LifecycleEventNetworkIdle,
+		Timeout:   common.DefaultTimeout,
+	})))
+	err := tb.await(func() error {
+		_ = p.Click(`a`, nil)
+		p.WaitForNavigation(nil)
+		return nil
+	})
+	require.NoError(t, err)
+	require.Equal(t, p.Title(), "Second page")
 }
--- a/tests/frame_manager_test.go
+++ b/tests/frame_manager_test.go
@@ -33,35 +33,35 @@ func TestWaitForFrameNavigationWithinDocument(t *testing.T) {
 	for _, tc := range testCases {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
-			tb := newTestBrowser(t, withFileServer())
-			p := tb.NewPage(nil)
+			t.Parallel()
 
-			resp := p.Goto(tb.staticURL("/nav_in_doc.html"), nil)
-			require.NotNil(t, resp)
+			errc := make(chan error, 1)
+			go func() {
+				tb := newTestBrowser(t, withFileServer())
+				p := tb.NewPage(nil)
 
-			// A click right away could possibly trigger navigation before we
-			// had a chance to call WaitForNavigation below, so give it some
-			// time to simulate the JS overhead, waiting for XHR response, etc.
-			time.AfterFunc(timeout*time.Millisecond, func() { //nolint:durationcheck
-				_ = tb.await(func() error {
+				resp := p.Goto(tb.staticURL("/nav_in_doc.html"), nil)
+				require.NotNil(t, resp)
+
+				// A click right away could possibly trigger navigation before we
+				// had a chance to call WaitForNavigation below, so give it some
+				// time to simulate the JS overhead, waiting for XHR
+				// response, etc.
+				<-time.After(timeout * time.Millisecond) //nolint:durationcheck
+
+				// if one of the promises panics, err will contain the error
+				errc <- tb.await(func() error {
 					_ = p.Click(tc.selector, nil)
+					_ = p.WaitForNavigation(tb.toGojaValue(&common.FrameWaitForNavigationOptions{
+						Timeout: 3 * timeout, // interpreted as ms
+					}))
 					return nil
 				})
-			})
-
-			done := make(chan struct{}, 1)
-			go func() {
-				require.NotPanics(t, func() {
-					p.WaitForNavigation(tb.toGojaValue(&common.FrameWaitForNavigationOptions{
-						Timeout: timeout * 3, // interpreted as ms
-					}))
-				})
-				done <- struct{}{}
 			}()
-
 			select {
-			case <-done:
-			case <-time.After(timeout * 5 * time.Millisecond): //nolint:durationcheck
+			case err := <-errc:
+				require.NoError(t, err)
+			case <-time.After(5 * timeout * time.Millisecond): //nolint:durationcheck
 				t.Fatal("Test timed out")
 			}
 		})
--- a/tests/frame_manager_test.go
+++ b/tests/frame_manager_test.go
@@ -112,9 +112,13 @@ func TestWaitForFrameNavigation(t *testing.T) {
 		Timeout:   common.DefaultTimeout,
 	})))
 
+	var timeout time.Duration = 5000 // interpreted as ms
+
 	var wfnPromise, cPromise *goja.Promise
 	err := tb.await(func() error {
-		wfnPromise = p.WaitForNavigation(nil)
+		wfnPromise = p.WaitForNavigation(tb.toGojaValue(&common.FrameWaitForNavigationOptions{
+			Timeout: timeout, // interpreted as ms
+		}))
 		cPromise = p.Click(`a`, nil)
 
 		assert.Equal(t, goja.PromiseStatePending, wfnPromise.State())
--- a/tests/frame_manager_test.go
+++ b/tests/frame_manager_test.go
@@ -39,7 +39,10 @@ func TestWaitForFrameNavigationWithinDocument(t *testing.T) {
 				tb := newTestBrowser(t, withFileServer())
 				p := tb.NewPage(nil)
 
-				resp := p.Goto(tb.staticURL("/nav_in_doc.html"), nil)
+				resp := p.Goto(tb.staticURL("/nav_in_doc.html"), tb.toGojaValue(&common.FrameGotoOptions{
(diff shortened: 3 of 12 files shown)


examples/fillform.js
```diff
@@ -15,12 +15,13 @@ export default function() {
...
-  }).then(() => {
+  }).then(() =>
+    page.$('input[type="submit"]').click()
+  ).then(() =>
     // We expect the above form submission to trigger a navigation, so wait for it
     // and the page to be loaded.
-    page.waitForNavigation();
-
+    page.waitForNavigation()
+  ).then(() => {
```
imiric, line 24: I think we can use `Promise.all()` now to minimize the chances of a `click()`/`waitForNavigation()` race condition, [as per the Playwright docs](https://playwright.dev/docs/api/class-page#page-wait-for-navigation). :tada:

This seemed to work for me:

```javascript
// Goto front page, find login link and click it
page.goto('https://test.k6.io/', { waitUntil: 'networkidle' });
const elem = page.$('a[href="/my_messages.php"]');
elem.click().then(() => {
  // Enter login credentials and login
  page.$('input[name="login"]').type('admin');
  page.$('input[name="password"]').type('123');
  // We expect the form submission to trigger a navigation, so to prevent a
  // race condition, setup a waiter concurrently while waiting for the click
  // to resolve.
  return Promise.all([
    page.waitForNavigation(),
    page.$('input[type="submit"]').click(),
  ]);
}).then(() => {
  check(page, {
    'header': page.$('h2').textContent() == 'Welcome, admin!',
  });
}).finally(() => {
  page.close();
  browser.close();
});
```

tests/frame_manager_test.go
```diff
@@ -33,37 +35,79 @@ func TestWaitForFrameNavigationWithinDocument(t *testing.T) {
...
+
+	require.NotNil(t, p.Goto(tb.URL("/first"), tb.toGojaValue(&common.FrameGotoOptions{
+		WaitUntil: common.LifecycleEventNetworkIdle,
+		Timeout:   common.DefaultTimeout,
+	})))
+	err := tb.await(func() error {
+		_ = p.Click(`a`, nil)
+		p.WaitForNavigation(nil)
+		return nil
+	})
```
imiric, line 110: Same here, I don't see how these Promises are resolved or synchronized, or that `tb.await()` (`vu.RunLoop()`) actually waits for them to run. In fact, this test passes if you remove the `p.WaitForNavigation()` call.

It seems to be relying on timing, as the test consistently takes 500ms on my machine, when it should take much less than that, or at least vary in timing. I saw this in the logs once:

```
time="2022-07-28T12:00:58+02:00" level=debug category="Browser:Close" elapsed="500 ms" goroutine=19
```
... but don't see it consistently, since that elapsed measurement happens by chance from the last logger call.

I don't see where the delay takes place... We set `SlowMo` to 0, and changing [`LifeCycleNetworkIdleTimeout`](https://github.com/grafana/xk6-browser/blob/1997670b94d3252768c2a5da80b645957634be3b/common/consts.go#L35) didn't have an effect. :confused:

So I think we should make sure that this test doesn't rely on timing, and that these Promises are actually resolved and waited on.

tests/frame_manager_test.go
```diff
@@ -33,37 +35,79 @@ func TestWaitForFrameNavigationWithinDocument(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, p.Title(), "Second page")
```
imiric, line 112: Nit: the second argument to `Equal()` (and most of the `testify` methods) is the expected value. Otherwise it will result in confusing Actual/Expected errors.

imiric (changes requested):
LGTM, besides the test and examples issues. The async helpers are a nice touch :+1:

And a meta comment: some of the commits here could be squashed.

tests/frame_manager_test.go
```diff
@@ -33,37 +35,79 @@ func TestWaitForFrameNavigationWithinDocument(t *testing.T) {
...
+				// response, etc.
+				<-time.After(timeout * time.Millisecond) //nolint:durationcheck
+
+				// if one of the promises panics, err will contain the error
+				errc <- tb.await(func() error {
+					_ = p.Click(tc.selector, nil)
+					_ = p.WaitForNavigation(tb.toGojaValue(&common.FrameWaitForNavigationOptions{
+						Timeout: 3 * timeout, // interpreted as ms
 					}))
+					return nil
```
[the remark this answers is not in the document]
ankur22: I think we should be checking all errors to ensure that we're not seeing them -- so i guess we should be ensuring that resolve is called and not reject?

tests/frame_manager_test.go
ankur22: Yeah, same behaviour when i ran it locally. Although the logic seems to be fine, just need to work with `resolve`/`reject`.

We should also test that `reject` is being called when the `javascript` run fails for whatever reason (maybe it times out or doesn't find the element).

examples/fillform.js
ankur22: I've amended this example, other examples that used `click` and `waitForNavigation` and the readme too.

Resolved in: 886eda55893b293ac9ae973e6ee6690cff9adb7d

tests/frame_manager_test.go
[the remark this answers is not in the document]
ankur22: `tb.wait()` is required for the call to `Click` and  `WaitForNavigation` to work on the event loop. The function in `tb.wait()` also needs to return before the event loop either resolves or rejects the result of `Click` and  `WaitForNavigation`. The main change i've done is to add a check to ensure that both `Click` and  `WaitForNavigation` end up in the correct fulfilled state, and this check can only be performed outside of `tb.wait()`.

tests/frame_manager_test.go
ankur22: I've amended this test so now it's checking the state of the promise before ending the test. This should be enough to show that the promises are resolving. What isn't clear is why we don't need `WaitForNavigation` i.e. I can remove it and the navigation occurs and the second page loads before we assert the title. Does `Click` inherently wait for the navigation to complete?

tests/frame_manager_test.go
ankur22: Resolved in: 262eaeed1765e1599b6bdf82d1f2ceabd6eb5cb3

k6ext/k6test/vu.go
```diff
@@ -22,14 +23,19 @@ import (
...
 
 // ToGojaValue is a convenience method for converting any value to a goja value.
 func (v *VU) ToGojaValue(i interface{}) goja.Value { return v.Runtime().ToValue(i) }
 
 // RunLoop is a convenience method for running fn in the event loop.
 func (v *VU) RunLoop(fn func() error) error {
+	v.runLoopMu.Lock()
+	defer v.runLoopMu.Unlock()
+
+	v.Loop.WaitOnRegistered()
```
ankur22, line 38: We're only supposed to have one eventloop per vu, which means we're not supposed to share the eventloop between threads. So each test that requires an eventloop needs its own browser/vu. So this is a bit of hack to allow multiple tests use one browser/vu and one eventloop. How do you feel about that @imiric? Would you rather we followed the eventloop design and had one browser process per test that needs an eventloop?

tests/frame_manager_test.go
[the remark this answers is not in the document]
imiric: OK, that makes sense. I played around with the test, and it looks great, besides the minor suggestions I made just now.

Do you want to leave the functional enhancement (checking for an existing element post-navigation) for later? We can create an issue for it if you prefer not doing it here.

tests/frame_manager_test.go
imiric: > Does `Click` inherently wait for the navigation to complete?

No, but it's likely that the navigation happens very quickly, so that by the time `tb.await()` returns, the promises have already been fulfilled.

You can confirm this by intercepting the click on the element, and delaying the amount of time for the navigation to happen.

For example, change the `/first` body to:

```html
<body>
  <a href="/second">click me</a>
  <script>
    const el = document.querySelector('a');
    el.addEventListener('click', function(evt) {
      evt.preventDefault();
      setTimeout(() => {
        window.location = '/second';
      }, 1000);
    });
  </script>
</body>
```

Then if you remove the `p.WaitForNavigation()` call in `tb.await()`, the test will fail with:

```
Error:          Not equal:
                expected: "Second page"
                actual  : "First page"
```

And if you add it back, it will pass.

BTW, we should probably set a timeout for `WaitForNavigation` like you did in the `WaitForFrameNavigationWithinDocument` test, otherwise if something goes wrong we'll have to wait the default 30s. If we delay the click by 1s, then timing out at 5s for the navigation should be enough, even on CI machines.

k6ext/k6test/vu.go
imiric: I think we generally follow that rule, as otherwise we would run into issues with Goja, which also isn't thread-safe.

We were missing the `WaitOnRegistered()` call, so that's fine, but what happens if you remove the lock? If any tests fail we should rewrite them to not share the browser / eventloop, so it might be better to remove the lock and force us to do that.

imiric (changes requested):
Looks good, thanks for the changes @ankur22. I left just a few more minor suggestions.

k6ext/k6test/vu.go
ankur22: Thanks! Resolved in: 8b231216c784683f90f728dfb2bfa6ccd855d08b

tests/frame_manager_test.go
```diff
@@ -16,7 +20,7 @@ func TestWaitForFrameNavigationWithinDocument(t *testing.T) {
 	}
 	t.Parallel()
 
-	var timeout time.Duration = 200
+	timeout := 2000 * time.Millisecond
```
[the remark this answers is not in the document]
ankur22: Resolved and rebased into: 3054bb37929e37ee2573f51d409e169359216327

tests/frame_manager_test.go
```diff
@@ -16,7 +20,7 @@ func TestWaitForFrameNavigationWithinDocument(t *testing.T) {
...
 	if os.Getenv("CI") == "true" {
 		// Increase the timeout on underprovisioned CI machines to minimize
 		// chances of intermittent failures.
```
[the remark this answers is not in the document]
ankur22: Resolved in: b27f6847aa5a687e2796b447fb52d0874bde9a95

tests/frame_manager_test.go
ankur22: Resolved in: 1e9fda0a58e452bdb611b9a61f5617bdb7e8f60d

NOTE: I didn't add the one second delay. I don't understand why we would need that, and I'm hoping we don't need it 🤞

tests/frame_manager_test.go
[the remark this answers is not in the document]
ankur22: I'm not sure how to test that something has changed when an anchor link is clicked. I was hoping that `isVisible` would do the trick, but that doesn't take into account whether the element is in the viewport or not. Any ideas what I could check for?

Also, the script that is linked to `nav-history`'s `click` event doesn't work, so i think it's best to just remove the `nav-history` test.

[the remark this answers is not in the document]
imiric: > Any ideas what I could check for?

It should be straightforward to determine. E.g. [see here](https://www.javascripttutorial.net/dom/css/check-if-an-element-is-visible-in-the-viewport/). We already have `ElementHandle.boundingBox()`, so we might be able to check that in Go. Having `ElementHandle.isInViewport()` seems like a useful user-facing function to have anyway, even if Playwright doesn't.

But this is a bit out of scope for this PR, so I can create an issue if you agree, and we can work on it later.


> Also, the script that is linked to `nav-history`'s `click` event doesn't work, so i think it's best to just remove the `nav-history` test.

WDYM by "doesn't work"? If you mean that functionally nothing happens on the page, that's expected, since `history.pushState()` simply creates a new history entry in the browser, and it's up to the JS app to interpret that, which is not part of the test.

The only thing we care for is if the `Page.navigatedWithinDocument` CDP event is triggered and if we process it correctly. If you run the test with `XK6_BROWSER_LOG=trace` you can see that happening, so I think the test is still useful.

Though if we want to expand the test to functionally check if an element is within the viewport, the JS will be slightly more complicated.

tests/frame_manager_test.go
imiric: Sure, it's not strictly needed, but it would somewhat ensure that `WaitForNavigation()` is also triggered on fast machines.

But it's not a big deal, let's leave it as is.

tests/frame_manager_test.go
```diff
@@ -33,37 +32,106 @@ func TestWaitForFrameNavigationWithinDocument(t *testing.T) {
...
 			go func() {
-				require.NotPanics(t, func() {
-					p.WaitForNavigation(tb.toGojaValue(&common.FrameWaitForNavigationOptions{
-						Timeout: timeout * 3, // interpreted as ms
+				tb := newTestBrowser(t, withFileServer())
+				p := tb.NewPage(nil)
+
+				resp := p.Goto(tb.staticURL("/nav_in_doc.html"), tb.toGojaValue(&common.FrameGotoOptions{
+					WaitUntil: common.LifecycleEventNetworkIdle,
+					Timeout:   time.Duration(int64(timeout)) * time.Millisecond,
```
imiric, line 44: This should just be:
```suggestion
					Timeout:   timeout,
```

Again, because of the way we parse the value, it's not intuitive using it from Go.

You can confirm this if you add `fmt.Printf("timeout: %s\n", parsedOpts.Timeout)` in `FrameManager.NavigateFrame()`. Currently it prints `got timeout: 1388h53m20s`, when it should be `5s`.

tests/frame_manager_test.go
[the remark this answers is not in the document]
ankur22: Nice, I was thinking about working with the boundingbox values on the go side, but wasn't sure if there was an easier way. Let's create a new issue for this, maybe it could coincide with working on `isInViewport` which would have helped here.

> WDYM by "doesn't work"?

Yeah, it didn't exhibit any visible behaviour. I've double checked it with `XK6_BROWSER_LOG=trace` and the `Page.navigatedWithinDocument` CDP event is triggered 👍

tests/frame_manager_test.go
ankur22: Thanks, that was a bad copy paste. I've reverted the change and rebased it into the last commit

imiric (approved):
Great work @inancgumus and @ankur22! :clap:

tests/frame_manager_test.go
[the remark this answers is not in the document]
ankur22: Created this issue to track the functionality check part: https://github.com/grafana/xk6-browser/issues/492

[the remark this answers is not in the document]
imiric: Ah, thanks. I created #493, but was debating whether there should be a separate issue to change the test, but it makes sense to have one.

Issue #427, opened 2022-06-30
Intermittent test failures and flaky tests


inancgumus:
For a few months now we've noticed an increase in intermittent test failures (e.g. #421), as well very frequent flaky tests and CI failures (#272, #254). This has become a major issue for both users and internal devs working on the extension.

From initial work explored in #281 that aimed to fix #272, it became apparent that these issues stem from the racy nature of how we [process CDP events](https://github.com/grafana/xk6-browser/blob/6baeca02e7515dc5b39c05601078e27688ee0e88/common/connection.go#L279), [which can trigger one](https://github.com/grafana/xk6-browser/blob/6baeca02e7515dc5b39c05601078e27688ee0e88/common/network_manager.go#L368-L382) or [more internal events](https://github.com/grafana/xk6-browser/blob/6baeca02e7515dc5b39c05601078e27688ee0e88/common/frame_manager.go#L162). This is explained in detail in [this discussion](https://github.com/grafana/xk6-browser/discussions/142), along with a proposed architecture to address these issues.

An ancillary problem with the current architecture is that these event flows and race conditions are difficult to trace and understand, and since they're intermittent and difficult to test we're never sure if we're introducing more of them as we make changes.

## Example

To illustrate an example of a race condition that should be addressed, consider the following scenario from [`TestPageGoto`](https://github.com/grafana/xk6-browser/blob/6baeca02e7515dc5b39c05601078e27688ee0e88/tests/page_test.go#L147-L156) (failure example in [CI build #1123](https://github.com/grafana/xk6-browser/runs/7181619757)):

1. xk6-browser sends the `Page.navigate` CDP command:
    ```
    -> {"id":20,"sessionId":"75F7E22B7AD1085646FE46C053DD6729","method":"Page.navigate","params":{"url":"http://127.0.0.1:40721/static/empty.html","frameId":"AF0723A7356FB62639D7BA483EBF615E"}}
    ```
    Done here:
    https://github.com/grafana/xk6-browser/blob/6baeca02e7515dc5b39c05601078e27688ee0e88/common/frame_manager.go#L617

2. The browser responds with the `load` `Page.lifecycleEvent`:
    ```
[...]

PR #478, opened 2022-08-02
WIP Event/CDP refactoring

imiric:
This is a very early WIP draft of some work done for #427.

It's very messy and I don't expect reviews yet, so this PR is more of a discussion starter so that we can decide how we want to proceed. I'll make some specific comments about notable changes, and schedule a meeting to discuss it over a call.

In general terms, it aims to:
- Create a separate `cdp` package where the `cdp.Client` is the only component that sends CDP commands and receives CDP messages.
  It wraps the WS connection to the browser, which is otherwise hidden from the rest of the codebase, and exposes a friendly API separated by CDP domains.

  This is in contrast to the current state where both `common.Connection` and `common.Session` send CDP commands (implement `cdp.Executor`) and receive CDP messages, `Connection` also handles the WS connection, and `Connection` and `Session` are passed around to most of the components, which call the `chromedp/cdproto` API directly.

- Create a notification system only for CDP events, which can be used to wait for events more reliably, in order to get rid of the current flakiness described in #427.

  This is in contrast to the current state where the same event system is used for both CDP and internal events, events can be sent and received from anywhere, and in some cases it's difficult to write non-racy code.

  I'm still debating whether the changes here to decouple the system is an improvement, or whether we should keep the current system with some minor changes, so some discussion is needed there.

- Deprecate our internal event system. This last part is not as straightforward as I previously thought, as we rely on the internal event system in places where it would be unfeasible to do things in any other way, unless we want to completely re-architect the system.

  For example, in `FrameManager.NavigateFrame` and `WaitForNavigation`, we have to wait for a specific `Page.lifecycleEvent`, but also for an internal event where we receive the `Response` of the initial HTTP request. On `main` we currently [emit the `Request` in a roundabout way inside the internal `NavigationEvent`](https://github.com/grafana/xk6-browser/blob/6c507693d4966741e59632d7980253605a87e2d4/common/frame_manager.go#L329) (the actual request is stored in `Frame.pendingDocument` [here](https://github.com/grafana/xk6-browser/blob/6c507693d4966741e59632d7980253605a87e2d4/common/frame_manager.go#L536)), which is all handled under a bunch of locks and racy behavior...

  Refactoring this and deprecating this system is far from trivial, and I did some attempts at it here, but would like to discuss it before making further changes.


imiric:
As discussed with @ankur22, we agreed to proceed as follows:

1. Abandon the event system changes and deprecation of internal events. We'll continue to use `BaseEventEmitter` and the current way of subscribing and listening for events. The reason is that internal events are used throughout the codebase and removing that would involve major architectural changes that could take weeks, if not months, to complete. We might be able to start refactoring to rely less on internal events wherever possible, but this can be done gradually instead of as part of #427.

2. Do the `cdp` package refactoring [as mentioned in point #1 here](https://github.com/grafana/xk6-browser/pull/478#issue-1325933940). We think this is a clear benefit over the current approach and should be relatively painless to complete.

    One thing that's currently not clear is if it would be possible to do this without touching the event system (i.e. without the `cdp.eventWatcher`), since we'd need someway to notify the `BaseEventEmitter` subscribers of incoming CDP events. If this is not straightforward, we might end up with two separate event systems: one for CDP events and one for internal events, which is not ideal. I'll think about this more once I start working on just the `cdp` refactoring.

3. Fix the navigation race conditions by blocking in the waiter goroutine. [This is an approach we currently use for creating Pages in the Browser](https://github.com/grafana/xk6-browser/blob/6c507693d4966741e59632d7980253605a87e2d4/common/browser.go#L365-L388), and I think would work for navigation events as well. It didn't occur to me until I was refactoring this code :facepalm:

    This would only be a specific fix for these navigation race conditions, and wouldn't address the many others we have, but this is the single major reason for test flakiness, and would at least improve things there.

Issue #426, opened 2022-06-30
Data race in ExecutionContext and injectedScript


inancgumus:
Version: 6baeca02e7515d

```bash
WARNING: DATA RACE
Read at 0x00c00432e538 by goroutine 276:
  github.com/grafana/xk6-browser/common.(*ExecutionContext).getInjectedScript()
      /Users/inanc/grafana/xk6-browser/common/execution_context.go:269 +0x188
  github.com/grafana/xk6-browser/common.(*ElementHandle).evalWithScript()
      /Users/inanc/grafana/xk6-browser/common/element_handle.go:1317 +0x80
  github.com/grafana/xk6-browser/common.(*ElementHandle).waitForSelector()
      /Users/inanc/grafana/xk6-browser/common/element_handle.go:705 +0x190
  github.com/grafana/xk6-browser/common.(*Frame).waitForSelector()
      /Users/inanc/grafana/xk6-browser/common/frame.go:645 +0x154
  github.com/grafana/xk6-browser/common.(*Frame).newAction.func1()
      /Users/inanc/grafana/xk6-browser/common/frame.go:1910 +0x11c
  github.com/grafana/xk6-browser/common.callApiWithTimeout.func1()
      /Users/inanc/grafana/xk6-browser/common/helpers.go:121 +0x70

Previous write at 0x00c00432e538 by goroutine 233:
  github.com/grafana/xk6-browser/common.(*ExecutionContext).getInjectedScript()
      /Users/inanc/grafana/xk6-browser/common/execution_context.go:297 +0x370
  github.com/grafana/xk6-browser/common.(*ElementHandle).evalWithScript()
      /Users/inanc/grafana/xk6-browser/common/element_handle.go:1317 +0x80
  github.com/grafana/xk6-browser/common.(*ElementHandle).waitForSelector()
      /Users/inanc/grafana/xk6-browser/common/element_handle.go:705 +0x190
  github.com/grafana/xk6-browser/common.(*Frame).waitForSelector()
      /Users/inanc/grafana/xk6-browser/common/frame.go:645 +0x154
  github.com/grafana/xk6-browser/common.(*Frame).newAction.func1()
      /Users/inanc/grafana/xk6-browser/common/frame.go:1910 +0x11c
  github.com/grafana/xk6-browser/common.callApiWithTimeout.func1()
      /Users/inanc/grafana/xk6-browser/common/helpers.go:121 +0x70

==================
--- FAIL: TestURLSkipRequest (1.42s)
[...]

PR #497, opened 2022-08-12
Fix the data race with injectedScript in ExecutionContext

ankur22:
A data race can occur when two tests perform an action on the same page
but under two different threads. It occurs in ExecutionContext with its
injectedScript parameter. Since we call
ExecutionContext.getInjectedScript() in ElementHandle.evalWithScript(),
it can be called in parallel from different goroutines, so a mutex is
a good fix for this.

Closes: https://github.com/grafana/xk6-browser/issues/426


common/execution_context.go
```diff
@@ -74,6 +75,8 @@ type ExecutionContext struct {
 	injectedScript api.JSHandle
 	vu             k6modules.VU
 
+	isMutex sync.RWMutex
```
imiric, line 78: Let's move this above `injectedScript`, which is usually the convention to indicate what the mutex is protecting.

imiric (changes requested):
:+1: Just a minor suggestion, LGTM otherwise.

Did this fix the issue for you consistently?

common/execution_context.go
ankur22: Thanks! Resolved and rebased into the initial commmit.

imiric (approved):

PR #503, opened 2022-08-17
Add pprof server and protect it with an env var

ankur22:
This will allow us to enable pprof in any environment where we need to
keep an eye on what is going on, especially since we need to track down
some race conditions during long runs.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,6 +1,11 @@
 package browser
 
 import (
+	"log"
+	"net/http"
+	_ "net/http/pprof" // nolint:gosec
+	"os"
+
 	"github.com/grafana/xk6-browser/api"
 	"github.com/grafana/xk6-browser/chromium"
 	"github.com/grafana/xk6-browser/common"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -69,15 +74,11 @@ func (mi *ModuleInstance) Exports() k6modules.Exports {
...
-		if err != nil {
-			log.Fatal(err)
-		}
-		pprof.StartCPUProfile(f)
-		defer pprof.StopCPUProfile()
-		<-ctx.Done()
-	}()*/
+	if _, ok := os.LookupEnv("XK6_BROWSER_PPROF"); ok {
+		go func() {
+			log.Println(http.ListenAndServe("localhost:6060", nil))
```
imiric, line 79: This can't be in `Launch()`, otherwise a new server will try to run on each new VU/iteration.

Maybe make it part of `NewModuleInstance` or `init()`?

Also let's print some debug info to let us know the server was actually started.

imiric (changes requested):

main.go
ankur22: I've moved it to a new `init` function in main,go. I've just used `log.Println` to log when the pprof server starts -- the actual logger gets created when the browser processes is started. Let me know if I've missed something though.

imiric (approved):

PR #508, opened 2022-08-18
Fix concurrent usage of goja.Runtime in waitForNavigation

### LINKED ISSUE #506: Concurrent usage of `goja.Runtime`
On the latest Cloud release (9b08fdb), I ran into this panic during a 5VU/30m test run of [`fillform.js`](https://github.com/grafana/xk6-browser/blob/9b08fdb9238b721a36b69d23210c13448470faf7/examples/fillform.js):

```
fatal error: concurrent map writes

goroutine 2097475 [running]:
runtime.throw({0x15bb280?, 0x89cd54?})
runtime/panic.go:992 +0x71 fp=0xc002b6f960 sp=0xc002b6f930 pc=0x435bb1
runtime.mapassign(0xc001a6b180?, 0x18a0a70?, 0x14329c0?)
runtime/map.go:595 +0x4d6 fp=0xc002b6f9e0 sp=0xc002b6f960 pc=0x40f2b6
github.com/dop251/goja.(*Runtime).typeInfo(0xc001a6b180, {0x18a0a70, 0x14329c0})
github.com/dop251/goja@v0.0.0-20220815083517-0c74f9139fd6/object_goreflect.go:630 +0xd9 fp=0xc002b6fa20 sp=0xc002b6f9e0 pc=0x8b2919
github.com/dop251/goja.(*objectGoReflect).init(0xc0025976c0)
github.com/dop251/goja@v0.0.0-20220815083517-0c74f9139fd6/object_goreflect.go:161 +0x86d fp=0xc002b6faa0 sp=0xc002b6fa20 pc=0x8af08d
github.com/dop251/goja.(*Runtime).reflectValueToValue(0xc001a6b180, {0x13d0580?, 0xc0021cd180?, 0xc000000077?})
github.com/dop251/goja@v0.0.0-20220815083517-0c74f9139fd6/runtime.go:1904 +0x790 fp=0xc002b6fb68 sp=0xc002b6faa0 pc=0x8cc510
github.com/dop251/goja.(*Runtime).ToValue(0xc001a6b180, {0x13d0580?, 0xc0021cd180})
github.com/dop251/goja@v0.0.0-20220815083517-0c74f9139fd6/runtime.go:1823 +0x15cc fp=0xc002b6fc78 sp=0xc002b6fb68 pc=0x8cbb6c
go.k6.io/k6/js/common.Throw(0x1372aa0?, {0x1886240?, 0xc0021cd180?})
go.k6.io/k6@v0.39.1-0.20220817093222-b89a3aaa6bff/js/common/util.go:20 +0x47 fp=0xc002b6fca0 sp=0xc002b6fc78 pc=0x9bf967
github.com/grafana/xk6-browser/k6ext.Panic.func1()
github.com/grafana/xk6-browser@v0.4.1-0.20220817102517-9b08fdb9238b/k6ext/panic.go:34 +0x2e fp=0xc002b6fcc8 sp=0xc002b6fca0 pc=0x9ed84e
github.com/grafana/xk6-browser/k6ext.Panic({0x1894c50, 0xc002a16b70}, {0x15c4784, 0x1a}, {0xc002b94df8, 0x1, 0x1})
github.com/grafana/xk6-browser@v0.4.1-0.20220817102517-9b08fdb9238b/k6ext/panic.go:51 +0x2d7 fp=0xc002b6fd60 sp=0xc002b6fcc8 pc=0x9ed777
github.com/grafana/xk6-browser/common.(*Frame).WaitForNavigation.func3({0x188c060, 0x2504420})
github.com/grafana/xk6-browser@v0.4.1-0.20220817102517-9b08fdb9238b/common/frame.go:1858 +0x11f fp=0xc002b6fe48 sp=0xc002b6fd60 pc=0xbceedf
github.com/grafana/xk6-browser/common.(*Frame).WaitForNavigation.func4()
github.com/grafana/xk6-browser@v0.4.1-0.20220817102517-9b08fdb9238b/common/frame.go:1892 +0x346 fp=0xc002b6ff80 sp=0xc002b6fe48 pc=0xbcecc6
github.com/grafana/xk6-browser/k6ext.promise.func1()
github.com/grafana/xk6-browser@v0.4.1-0.20220817102517-9b08fdb9238b/k6ext/promise.go:41 +0x62 fp=0xc002b6ffe0 sp=0xc002b6ff80 pc=0x9edc22
runtime.goexit()
runtime/asm_amd64.s:1571 +0x1 fp=0xc002b6ffe8 sp=0xc002b6ffe0 pc=0x4688e1
created by github.com/grafana/xk6-browser/k6ext.promise
github.com/grafana/xk6-browser@v0.4.1-0.20220817102517-9b08fdb9238b/k6ext/promise.go:40 +0x105
```

See test run 140728 on staging for details.

It seems to be a regression introduced by #501. k6's `common.Throw()` calls `Runtime.ToValue()` from a different goroutine than the main VU one. Since Goja is not thread safe, we need to find a better way of doing this, or just use a mutex as a quick fix.

This is not the only place where Goja runtimes are shared, but it's the first failure of this type I've seen outside of tests.

imiric:
This is a functional change from before, since the iteration won't be interrupted by a timeout error anymore, and the promise will be rejected instead. This might be an acceptable tradeoff for using a lock, but it would be better if there was a way of interrupting the iteration if the promise is rejected.

Fixes #506

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -1850,7 +1850,7 @@ func (f *Frame) WaitForNavigation(opts goja.Value) *goja.Promise {
 		})
 
 	handleTimeoutError := func(err error) error {
-		if errors.Is(err, context.DeadlineExceeded) {
+		if err != nil {
 			e := &k6ext.UserFriendlyError{
 				Err:     err,
 				Timeout: parsedOpts.Timeout,
--- a/common/frame.go
+++ b/common/frame.go
@@ -1850,6 +1850,8 @@ func (f *Frame) WaitForNavigation(opts goja.Value) *goja.Promise {
 		})
 
 	handleTimeoutError := func(err error) error {
+		f.log.Debugf("Frame:WaitForNavigation",
+			"fid:%v furl:%s timeoutCtx done: %v", f.ID(), f.URL(), err)
 		if err != nil {
 			e := &k6ext.UserFriendlyError{
 				Err:     err,
@@ -1857,9 +1859,6 @@ func (f *Frame) WaitForNavigation(opts goja.Value) *goja.Promise {
 			}
 			return fmt.Errorf("waiting for navigation: %w", e)
 		}
-		f.log.Debugf("Frame:WaitForNavigation",
-			"fid:%v furl:%s timeoutCtx done: %v",
-			f.ID(), f.URL(), err)
 
 		return nil
 	}
--- a/common/frame.go
+++ b/common/frame.go
@@ -1849,17 +1849,19 @@ func (f *Frame) WaitForNavigation(opts goja.Value) *goja.Promise {
 			return false
 		})
 
-	handleTimeoutError := func(err error) {
+	handleTimeoutError := func(err error) error {
 		if errors.Is(err, context.DeadlineExceeded) {
-			err = &k6ext.UserFriendlyError{
+			e := &k6ext.UserFriendlyError{
 				Err:     err,
 				Timeout: parsedOpts.Timeout,
 			}
-			k6ext.Panic(f.ctx, "waiting for navigation: %w", err)
+			return fmt.Errorf("waiting for navigation: %w", e)
 		}
 		f.log.Debugf("Frame:WaitForNavigation",
 			"fid:%v furl:%s timeoutCtx done: %v",
 			f.ID(), f.URL(), err)
+
+		return nil
 	}
 
 	return k6ext.Promise(f.ctx, func() (result interface{}, reason error) {
@@ -1889,8 +1891,7 @@ func (f *Frame) WaitForNavigation(opts goja.Value) *goja.Promise {
 				}
 			}
 		case <-timeoutCtx.Done():
-			handleTimeoutError(timeoutCtx.Err())
-			return nil, nil
+			return nil, handleTimeoutError(timeoutCtx.Err())
 		}
 
 		// A lifecycle event won't be received when navigating within the same
@@ -1900,7 +1901,7 @@ func (f *Frame) WaitForNavigation(opts goja.Value) *goja.Promise {
 			select {
 			case <-lifecycleEvtCh:
 			case <-timeoutCtx.Done():
-				handleTimeoutError(timeoutCtx.Err())
+				return nil, handleTimeoutError(timeoutCtx.Err())
 			}
 		}
 


imiric:
@mstoykov Do you have an idea to improve this?

Previously we called k6's `common.Throw()` to interrupt the iteration, but we can't do that from a promise goroutine, if we want to avoid #506. Rejecting the promise still logs an error, but the iteration will continue.

mstoykov:
>  to interrupt the iteration

I do think the decision on whether the iteration should be interrupted should be of the user writing the test. They *can* do that through just on an exception/failure on this promise - abort the iteration :). 

But IMO there are good reason for this to not be hardcoded like actually maybe retrying or doing other stuff when this fails. Maybe they need to free some resources :shrug: .

So as far as I am concerned this is a double bugfix :P

common/frame.go
```diff
@@ -1889,8 +1891,7 @@ func (f *Frame) WaitForNavigation(opts goja.Value) *goja.Promise {
 				}
 			}
 		case <-timeoutCtx.Done():
-			handleTimeoutError(timeoutCtx.Err())
-			return nil, nil
+			return nil, handleTimeoutError(timeoutCtx.Err())
```
mstoykov, line 1894: Shouldn't it *not* return `nil` and a `nil` error if the timeout isn't DeadlineExceeded? 

no idea what this code does, just it seems .... uh ... strange.

imiric: The first value is a `*Response`, which would be `nil` at this point anyways. If the timeout isn't `DeadlineExceeded`, it probably means the context was cancelled (e.g. iteration done or shutdown), so returning `nil, nil` also makes sense in that case. What do you think it should return instead?

We could return `resp, handleTimeoutError(timeoutCtx.Err())` on line 1904, though. The promise could still be rejected, but at least there's a chance they would have a response to work with. :man_shrugging:

mstoykov: :shrug: 

Arguably you can return the error ... the code should not be executing at this point (*likely* there is still place for a race here). At worst the user will see a bunch of errors that are "context canceled".

But in the other case they could see errors as the response is `nil` but is still used for something. 

So I would argue them seeing the timeout errors will be better. 

In this case I would argue it will be nice to be able to check that k6 is stopping/has stopped and prevent the whole race condition to begin with :shrug:.

So I don't know - maybe just leave it as is and add a comment to be fixed at some point in the future

ankur22: In general it would be better to return an error. In the case it doesn't look like it really matters.

I think the problem is that we could easily forget that we're not returning an error which might be confusing in the future when we look back at this code. I vote to either return the original error if it's not `context.DeadlineExceeded` or log an error and return nil... but it's not idiomatic go to return a `nil` response and `nil` error... i think 🤷‍♂️

imiric: > But in the other case they could see errors as the response is nil but is still used for something.

Yeah, that's a good point. I changed it in 65fbc97. Thanks!

common/frame.go
```diff
@@ -1849,17 +1849,19 @@ func (f *Frame) WaitForNavigation(opts goja.Value) *goja.Promise {
...
 			}
-			k6ext.Panic(f.ctx, "waiting for navigation: %w", err)
+			return fmt.Errorf("waiting for navigation: %w", e)
 		}
 		f.log.Debugf("Frame:WaitForNavigation",
 			"fid:%v furl:%s timeoutCtx done: %v",
 			f.ID(), f.URL(), err)
+
+		return nil
 	}
```
ankur22, line 1865: Might be confusing if the error is not `context.DeadlineExceeded`, although i realise that this lambda func is only used when the context is closed. I would also move the debug log line to just before returning the wrapped error. All minor things though in this case.

ankur22 (approved):

common/frame.go
imiric: Since 65fbc97, this will now always return an error if it's not nil, so this `return` likely won't be reached.

The debug logging was here to avoid a double log of both the timeout err and this debug message, but I guess it doesn't make sense to have it if `err` will be nil here. And double logging is not a big deal, so I moved it in 5751ddc.

mstoykov (approved):

imiric:
I'm not sure if the current failure of #509 is related to these changes or not... :confused:

I've only seen it in tests so far, and we need to fix this issue, so I'll merge this and we can look into that later if it becomes a bigger problem.

Issue #482, opened 2022-08-04
Flaky timeout in `Frame.waitForExecutionContext()`


imiric:
We recently noticed test failures in CI for PRs #326 and #480 (jobs [#1183](https://github.com/grafana/xk6-browser/runs/7636154407?check_suite_focus=true) and [#1197](https://github.com/grafana/xk6-browser/runs/7670466150?check_suite_focus=true) respectively), where different tests timeout after 5m.

The most relevant stack traces seem to be:

From \#1183:
```
goroutine 14087 [select, 4 minutes]:
github.com/grafana/xk6-browser/common.(*Frame).waitForExecutionContext(0xc0009dac40, 0x1b5da7d, 0x4)
   /home/runner/work/xk6-browser/xk6-browser/common/frame.go:543 +0x345
github.com/grafana/xk6-browser/common.(*Frame).Evaluate(0xc0009dac40, 0x1d80388, 0xc0007d7d10, 0x0, 0x0, 0x0, 0xc000799dc8, 0x1)
   /home/runner/work/xk6-browser/xk6-browser/common/frame.go:911 +0x1e5
github.com/grafana/xk6-browser/common.(*Page).Evaluate(0xc00071adc0, 0x1d80388, 0xc0007d7d10, 0x0, 0x0, 0x0, 0x1d80388, 0xc0007d7d10)
   /home/runner/work/xk6-browser/xk6-browser/common/page.go:506 +0x16b
github.com/grafana/xk6-browser/tests.TestPageGotoWaitUntilDOMContentLoaded(0xc0007cf800)
   /home/runner/work/xk6-browser/xk6-browser/tests/page_test.go:194 +0x317
testing.tRunner(0xc0007cf800, 0x1bbc8f0)
   /opt/hostedtoolcache/go/1.16.15/x64/src/testing/testing.go:1203 +0x203
created by testing.(*T).Run
   /opt/hostedtoolcache/go/1.16.15/x64/src/testing/testing.go:1248 +0x5d8
```

And from \#1197:
```
goroutine 17203 [select, 4 minutes]:
github.com/grafana/xk6-browser/common.(*Frame).waitForExecutionContext(0xc004fcf180, {0x18e922f, 0x4})
   /home/runner/work/xk6-browser/xk6-browser/common/frame.go:543 +0x2d7
github.com/grafana/xk6-browser/common.(*Frame).document(0xc004fcf180)
[...]

Issue #496, opened 2022-08-12
Deadlock in the cloud


sniku:
The browser test seems to dead-lock after 3-10min. 
This is likely not related to the cloud, but it's easiest to spot it there because the metrics stop being generated while k6 continues to run. 

- CPU drops to near 0%. 
- The memory stays constant.  (which probably means that the chromium process is running)
- no iterations are being executed
- no browser metrics are being collected

Sample script:
```js
import { check } from 'k6';
import { chromium } from 'k6/x/browser';

export let options = {
  vus: 1,
  duration: '50m'
}

export default function () {
  const browser = chromium.launch('chromium', {
    headless: true,
  });
  const context = browser.newContext();
  const page = context.newPage();

  // Goto front page, find login link and click it
  page.goto('https://test.k6.io/', { waitUntil: 'networkidle' });
  const elem = page.$('a[href="/my_messages.php"]');
  elem.click().then(() => {
    // Enter login credentials and login
    page.$('input[name="login"]').type('admin');
    page.$('input[name="password"]').type('123');
    return page.$('input[type="submit"]').click();
  }).then(() => {
    // We expect the above form submission to trigger navigation, so wait for it
    // and the page to be loaded.
    page.waitForNavigation();

    check(page, {
      'header': page.$('h2').textContent() == 'Welcome, admin!',
    });
  }).finally(() => {
    page.close();
    browser.close();
  });
}
```
![image (7)](https://user-images.githubusercontent.com/442646/184376457-7c52a8aa-761d-442f-ad12-36c03fe4532e.png)
![image](https://user-images.githubusercontent.com/442646/184377814-8202e6ad-c195-40fb-93a5-fb1fd2d004e8.png)

ankur22:
We think the deadlock issue in the cloud ([#496](https://github.com/grafana/xk6-browser/issues/496)) is the same as issue as the test timeout issue in waitForExecutionContext ([#482](https://github.com/grafana/xk6-browser/issues/482)).

[deadlock-cloud-stacktrace.txt](https://github.com/grafana/xk6-browser/files/9375386/deadlock-cloud-stacktrace.txt)

The fix and further details will be raised in [#482](https://github.com/grafana/xk6-browser/issues/482), unless we find something else.

ankur22:
[...]

PR #512, opened 2022-08-22
Add mutex to sync the clearing of the context

ankur22:
Closes: https://github.com/grafana/xk6-browser/issues/482

This tries to synchronize the navigation internal actions by forcing
the system to wait for the context to be cleared before carrying on
with the other actions i.e. handling the response and allowing the
creation of the new contexts.

[Diagram](https://drive.google.com/file/d/17na5TdRjrkZ2aExT-hdb6r-il8uAx8cW/view?usp=sharing).
xk6-browser - Deadlock issue #482   #496 drawio

This has been tested in the cloud and after several tests, xk6-browser
has not deadlocked 🎉 

I have another solution which i prefer as it's closer to the root cause
of the issue -- https://github.com/grafana/xk6-browser/pull/524 (which is still WIP).

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -503,7 +503,7 @@ func (f *Frame) requestByID(reqID network.RequestID) *Request {
 	return frameSession.networkManager.requestFromID(reqID)
 }
 
-func (f *Frame) setContext(world executionWorld, execCtx frameExecutionContext) {
+func (f *Frame) setContext(world executionWorld, execCtx frameExecutionContext) error {
 	f.executionContextMu.Lock()
 	defer f.executionContextMu.Unlock()
 
@@ -518,12 +518,14 @@ func (f *Frame) setContext(world executionWorld, execCtx frameExecutionContext) 
 	if f.executionContexts[world] != nil {
 		f.log.Debugf("Frame:setContext", "fid:%s furl:%q ectxid:%d world:%s, world exists",
 			f.ID(), f.URL(), execCtx.ID(), world)
-		return
+		return errors.New("context already set")
 	}
 
 	f.executionContexts[world] = execCtx
 	f.log.Debugf("Frame:setContext", "fid:%s furl:%q ectxid:%d world:%s, world set",
 		f.ID(), f.URL(), execCtx.ID(), world)
+
+	return nil
 }
 
 func (f *Frame) setID(id cdp.FrameID) {
@@ -537,7 +539,7 @@ func (f *Frame) waitForExecutionContext(world executionWorld) {
 	f.log.Debugf("Frame:waitForExecutionContext", "fid:%s furl:%q world:%s",
 		f.ID(), f.URL(), world)
 
-	t := time.NewTimer(50 * time.Millisecond)
+	t := time.NewTicker(50 * time.Millisecond)
 	defer t.Stop()
 	for {
 		select {
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -28,6 +28,7 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/grafana/xk6-browser/api"
 	"github.com/grafana/xk6-browser/k6ext"
@@ -77,6 +78,9 @@ type FrameSession struct {
 	contextIDToContext   map[cdpruntime.ExecutionContextID]*ExecutionContext
 	isolatedWorlds       map[string]bool
 
+	clearWaitChansMu sync.Mutex
+	clearWaitChans   map[string]chan bool
+
 	eventCh chan Event
 
 	childSessions map[cdp.FrameID]*FrameSession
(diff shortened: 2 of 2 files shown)


common/frame.go
```diff
@@ -537,7 +539,7 @@ func (f *Frame) waitForExecutionContext(world executionWorld) {
 	f.log.Debugf("Frame:waitForExecutionContext", "fid:%s furl:%q world:%s",
 		f.ID(), f.URL(), world)
 
-	t := time.NewTimer(50 * time.Millisecond)
+	t := time.NewTicker(50 * time.Millisecond)
```
ankur22, line 542: `NewTimer` only runs once, but it looks like we want to run this more than once, it's why it's changed to `NewTicker`.

common/frame_session.go
```diff
@@ -604,21 +663,59 @@ func (fs *FrameSession) onExecutionContextDestroyed(execCtxID cdpruntime.Executi
...
+	cPrv, ok := fs.getChannel(frameID)
+
+	if ok {
+		select {
+		case <-cPrv:
+			fmt.Printf("Ankur: addClearWait channel already closed fid:%s\n", frameID)
+		case <-time.After(time.Second):
+			fmt.Printf("Ankur: addClearWait force closing chan fid:%s\n", frameID)
+			close(cPrv)
+		}
```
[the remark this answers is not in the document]
ankur22: Yes, i forgot that. The channel is now removed when `onExecutionContextsCleared` is called and the channel is closed.

common/frame.go
inancgumus: Nice catch! 🏈

common/frame_session.go
```diff
@@ -604,21 +669,69 @@ func (fs *FrameSession) onExecutionContextDestroyed(execCtxID cdpruntime.Executi
...
 func (fs *FrameSession) onExecutionContextsCleared() {
 	fs.logger.Debugf("FrameSession:onExecutionContextsCleared",
 		"sid:%v tid:%v", fs.session.ID(), fs.targetID)
 
 	fs.contextIDToContextMu.Lock()
-	defer fs.contextIDToContextMu.Unlock()
 
+	var fid string
 	for _, context := range fs.contextIDToContext {
+		fid = context.Frame().ID()
```
imiric, line 715: Iterating over a map is unstable, and we'll only get the last iteration's frame ID here.

Couldn't that leave other channels opened, and lead to timeouts in `addClearWait()`?

common/frame.go
```diff
@@ -518,12 +518,14 @@ func (f *Frame) setContext(world executionWorld, execCtx frameExecutionContext)
 	if f.executionContexts[world] != nil {
 		f.log.Debugf("Frame:setContext", "fid:%s furl:%q ectxid:%d world:%s, world exists",
 			f.ID(), f.URL(), execCtx.ID(), world)
-		return
+		return errors.New("context already set")
```
imiric, line 521: Have you considered the suggestion in https://github.com/grafana/xk6-browser/issues/482#issuecomment-1219385591 to remove this check and always overwrite the context with the newly provided one?

Or is this guarding the case when we're processing a second `setContext()` call without having processed the first `onExecutionContextsCleared()`? Wouldn't we "lose" this context in that case? :thinking:

imiric (commented):
I'd like to run some tests with this, as I'm still confused about some of the functionality, as mentioned in the comments.

It would be good to stress test it with more complex scripts, like the Vistaprint one.

common/frame.go
ankur22: Opened a new PR for this fix: https://github.com/grafana/xk6-browser/pull/539

common/frame.go
ankur22: Since the issue is due to out of order events, it didn't feel correct to then change the behaviour in `setContext`, and i also don't know if it's possible to have a `setContext` without first having a `clearContext` after the initial navigation to `about:blank`.

As you've pointed out, this is currently helping to identify when the context has been set, but needs to be cleared (assumed that this must happen first). We shouldn't lose the context as long as the clear does take place before setting it again.

We need a sensible timeout while waiting for the (out of order) clear context event, but there's a small chance that the clear context event is delayed more than 1-5 secs or is missing. In that case:

- should we overwrite the context still, but a late clear context event will clear the context resulting in the same deadlock?
- Or should we leave the behaviour as is, but a late clear context event will clear the context resulting in the same deadlock?
- Or should we panic and exit since we haven't seen the expected clear context event and avoid the deadlock?

imiric: Hhmm right then it would make sense to return an error here, since the situation of overwriting contexts is a bug. I think panicking would be an extreme approach to take, and wouldn't solve the issue, but returning an error is fine in this case. Thanks for explaining.

ankur22:
We believe that the root cause of the issue has been solved in https://github.com/grafana/xk6-browser/pull/555.

PR #515, opened 2022-08-31
Export Chromium `BrowserType` implementation directly

### LINKED ISSUE #485: Export Chromium `BrowserType` implementation directly
In #462 we made a `chromium` object importable in scripts, but the way it's done is a bit convoluted:
https://github.com/grafana/xk6-browser/blob/aadaffec87c8ffba66db89f417bb68e3aacbbe08/main.go#L61-L68

That is, `"chromium"` actually points to the `JSModule` itself, and calling `chromium.launch()` calls the module-level `Launch()` wrapper, which internally creates a Chromium `BrowserType` and calls `BrowserType.Launch()`.

Which means this is possible:

```javascript

export default function() {
  // logs v0.4.0, the xk6-browser version, not the Chromium version
  console.log(chromium.version);
  // logs the device configuration map, which should be browser-agnostic
  console.log(chromium.devices);
}
```

## Suggested solution

As mentioned [here](https://github.com/grafana/xk6-browser/pull/462#discussion_r928747546) and [here](https://github.com/grafana/k6-docs/pull/753#discussion_r941227829), `import { chromium } from 'k6/x/browser';` should actually import the [Chromium `BrowserType` instance](https://github.com/grafana/xk6-browser/blob/aadaffec87c8ffba66db89f417bb68e3aacbbe08/chromium/browser_type.go#L35-L44) directly, so the script can call [all of its public methods](https://github.com/grafana/xk6-browser/blob/aadaffec87c8ffba66db89f417bb68e3aacbbe08/api/browser_type.go#L28-L34). This would allow us to easily support other `BrowserType` implementations in the future.

And if the user wants access to module-level attributes, the script should import the module instead, e.g. `import k6b from 'k6/x/browser;`.

imiric:
Closes #485 

This also exposes the xk6-browser version to JS scripts, which is useful for troubleshooting user scripts until https://github.com/grafana/k6/issues/1741 is implemented.

### DIFF
--- a/main.go
+++ b/main.go
@@ -12,8 +12,6 @@ import (
 	"github.com/grafana/xk6-browser/k6ext"
 
 	k6modules "go.k6.io/k6/js/modules"
-
-	"github.com/dop251/goja"
 )
 
 const version = "v0.4.0"
@@ -23,12 +21,11 @@ type (
 	// instances for each VU.
 	RootModule struct{}
 
-	// JSModule is the entrypoint into the browser JS module.
+	// JSModule exposes the properties available to the JS script.
 	JSModule struct {
-		vu        k6modules.VU
-		k6Metrics *k6ext.CustomMetrics
-		Devices   map[string]common.Device
-		Version   string
+		Chromium api.BrowserType
+		Devices  map[string]common.Device
+		Version  string
 	}
 
 	// ModuleInstance represents an instance of the JS module.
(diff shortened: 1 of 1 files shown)


ankur22 (approved):

imiric:
FYI There was a regression introduced by this change, where `chromium.launch()` fails after the first iteration with a (context) `canceled` error.

I'm looking into it and will push a PR shortly.

PR #517, opened 2022-08-31
Fix context canceled error when running more than 1 iteration

imiric:
This was a regression introduced by #515, caused by the fact [the VU context changes after each iteration](https://github.com/grafana/k6/blob/e09bb87277865d668586429eee97158fbbfc58e5/js/runner.go#L710), so our context initialization needs to happen on each iteration (`launch()` call) as well.

I wanted to add an E2E test to cover this scenario and test multiple iterations, but the process still exits with 0 when this was broken, so we wouldn't have seen it in CI anyway. I'm not sure how to test it otherwise. We should probably dump the summary results and check them in the job, similar to [the way the test-xk6 workflow works in the k6 repo](https://github.com/grafana/k6/blob/master/.github/workflows/xk6.yml#L69-L73). But we can do that later and I didn't want to delay the fix.

### DIFF
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -33,41 +33,34 @@ var _ api.BrowserType = &BrowserType{}
 // BrowserType provides methods to launch a Chrome browser instance or connect to an existing one.
 // It's the entry point for interacting with the browser.
 type BrowserType struct {
-	Ctx             context.Context
-	CancelFn        context.CancelFunc
-	hooks           *common.Hooks
-	fieldNameMapper *common.FieldNameMapper
-	vu              k6modules.VU
-
-	execPath string       // path to the Chromium executable
-	storage  *storage.Dir // stores temporary data for the extension and user
+	// FIXME: This is only exported because testBrowser needs it. Contexts
+	// shouldn't be stored on structs if we can avoid it.
+	Ctx       context.Context
+	vu        k6modules.VU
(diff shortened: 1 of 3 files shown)


chromium/browser_type.go
```diff
@@ -33,41 +33,34 @@ var _ api.BrowserType = &BrowserType{}
...
 	)
 
-	// Create the extension master context.
-	// If this context is cancelled we'll initiate an extension wide cancellation and shutdown.
-	extensionCtx, extensionCancelFn := context.WithCancel(ctx)
-	extensionCtx = common.WithHooks(extensionCtx, hooks)
-
+	// NOTE: vu.InitEnv() *must* be called from the script init scope,
+	// otherwise it will return nil.
+	k6m := k6ext.RegisterCustomMetrics(vu.InitEnv().Registry)
```
imiric, line 56: The downside of doing this here is that it would have to be done in every `BrowserType` implementation, when it really should be done once for the whole extension.

I moved it here since we need to re-initialize the context on each iteration, and since we store these metrics on the context, we have to re-attach it every time.

An alternative would be to keep this in `NewModuleInstance` and pass it to this and every constructor that needs it, but the signature would be messier. We'll have to do it eventually though, once we add a second `BrowserType` implementation, so let's rethink it then.

ankur22 (approved):

Issue #301, opened 2022-04-28
Add MP4 with a sample test execution to the README


sniku:
Similar to the CLI gif we have in https://github.com/grafana/k6 
we should add a sample browser test so users can see it in action.

imiric:
GitHub supports embedding MP4s in READMEs since a few months ago, so this should definitely be a crisp MP4. :slightly_smiling_face:

PR #520, opened 2022-09-07
docs: update README to include an mp4 video

mdcruz:
Fixes https://github.com/grafana/xk6-browser/issues/301

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@
 
 **xk6-browser** is a [k6](https://k6.io/) extension adding support for automation of browsers via the [Chrome Devtools Protocol](https://chromedevtools.github.io/devtools-protocol/) (CDP).
 
-https://user-images.githubusercontent.com/10811379/188867169-c6713f15-4c2d-438a-8217-bb73ecfa6b84.mp4
+https://user-images.githubusercontent.com/10811379/188889687-660ce8f9-264d-4dd0-905c-41a909102be8.mp4
 
 Special acknowledgment to the authors of [Playwright](https://playwright.dev/) and [Puppeteer](https://github.com/puppeteer/puppeteer) for their trailblazing work in this area. This project is heavily influenced and in some regards based on the code of those projects.
 
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@
 
 **xk6-browser** is a [k6](https://k6.io/) extension adding support for automation of browsers via the [Chrome Devtools Protocol](https://chromedevtools.github.io/devtools-protocol/) (CDP).
 
+https://user-images.githubusercontent.com/10811379/188867169-c6713f15-4c2d-438a-8217-bb73ecfa6b84.mp4
+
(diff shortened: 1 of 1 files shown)


imiric:
Thanks for this @mdcruz!

Should we maybe showcase it with `headless: false` and `slowMo: '500ms'`, so that the browser window is shown and the actions can be followed? The demo is more impressive when you actually see the browser. :)

We also might want to show a more complex example, e.g. filling in a form and logging in, similar to [`fillform.js`](https://github.com/grafana/xk6-browser/blob/c5c2cead896d615f06f70f0f4ed24e1cd041b1ed/examples/fillform.js). OTOH, the API might change and we don't want to keep updating the video, but I think what's shown in `fillform.js` on `main` now is relatively stable.

And sorry about the CI failure :disappointed: It's, of course, unrelated, so we can retry the build. And I think you'll need to sign the CLA if you haven't done so already.

mdcruz:
> Thanks for this @mdcruz!
> 
> Should we maybe showcase it with `headless: false` and `slowMo: '500ms'`, so that the browser window is shown and the actions can be followed? The demo is more impressive when you actually see the browser. :)
> 
> We also might want to show a more complex example, e.g. filling in a form and logging in, similar to [`fillform.js`](https://github.com/grafana/xk6-browser/blob/c5c2cead896d615f06f70f0f4ed24e1cd041b1ed/examples/fillform.js). OTOH, the API might change and we don't want to keep updating the video, but I think what's shown in `fillform.js` on `main` now is relatively stable.
> 
> And sorry about the CI failure 😞 It's, of course, unrelated, so we can retry the build. And I think you'll need to sign the CLA if you haven't done so already.
> 
> EDIT: On second thought, the entire `fillform.js` as is would be to long to `cat` it before it's run, but maybe a simplified version of it? Something like:
> 
> ```js
> import { check } from 'k6';
> import { chromium } from 'k6/x/browser';
> 
> export default function () {
>   const browser = chromium.launch({ headless: false, slowMo: '500ms' });
>   const context = browser.newContext();
>   const page = context.newPage();
> 
>   page.goto('https://test.k6.io/my_messages.php', { waitUntil: 'networkidle' });
> 
>   // Enter login credentials and login
>   page.locator('input[name="login"]').type('admin');
>   page.locator('input[name="password"]').type('123');
>   const submitBtn = page.locator('input[type="submit"]');
> 
>   Promise.all([
>     page.waitForNavigation(),
>     submitBtn.click(),
>   ]).then(() => {
>     check(page, {
>       'header': page.locator('h2').textContent() == 'Welcome, admin!',
>     });
>   }).finally(() => {
>     page.close();
>     browser.close();
>   });
> }
> ```
> 
> This is 27 lines. Might be short enough?

Thanks Ivan! This will be short enough. I can use browser.newPage instead to make it shorter

mdcruz:
Updated video: https://user-images.githubusercontent.com/10811379/188889687-660ce8f9-264d-4dd0-905c-41a909102be8.mp4

imiric (approved):
Thanks, looks great! :+1:

ankur22:
This is awesome! Thanks!

ankur22 (approved):

PR #523, opened 2022-09-08
Update k6 dependency to latest stable v0.40.0

imiric:


imiric:
@inancgumus You shouldn't approve this if the tests are failing for a legitimate reason. :sweat_smile:

See https://loadimpact.slack.com/archives/C02H6RF6X1P/p1662643395772169 and https://github.com/grafana/k6/commit/d69cb582741618b93d8321e021009aa375761250.

Looking into it now, but I'll have to leave early today, so if I don't finish maybe someone else can take a look? Otherwise I'll continue tomorrow.

inancgumus:
Hmm, the new k6 version has [some complaints](https://github.com/grafana/xk6-browser/runs/8250141304?check_suite_focus=true).

More info here: https://github.com/grafana/k6/pull/2602

inancgumus:
> @inancgumus You shouldn't approve this if the tests fail for a legitimate reason. 😅

Yeah, my bad; I became aware of the issue after I approved the PR. It was looking harmless :)

imiric:
I pushed a WIP attempt at fixing the test issue in ccf39b4, but some tests fail with a nil panic in `chromium.NewBrowserType()` while looking up the `Registry` at `vu.InitEnv().Registry`. According to @mstoykov the `modulestest` changes better replicate real k6 behavior, but I'm not seeing a way to make it work, and have to leave soon. `Runtime.MoveToVUContext()` should be used instead of setting `Runtime.VU.StateField` directly, but either way it results in a nil panic.

I'll look more into this tomorrow, but would appreciate if someone else could take a look as well. Mihail would be able to help from Monday.

Upgrading to k6 v0.40.0 is not a blocker for releasing xk6-browser v0.5.0, so we could release with k6 v0.39.0 as well. Or we could delay the release by a day, and fix this on Monday.

imiric:
I hackfixed the nil panic and modulestest integrity check, but now the logger is racy... :sob:

Looking into it...

tests/test_browser.go
```diff
@@ -123,11 +124,24 @@ func newTestBrowser(tb testing.TB, opts ...interface{}) *testBrowser {
...
+		Logger:   state.Logger,
+		Registry: registry,
+	}
+	// NewBrowserType must be run from the init context, but since the recent
+	// validation introduced in k6/js/modulestest, InitEnvField and StateField
+	// must not be both set at the same time. So we nil StateField, create
+	// the BrowserType, and revert it before running Launch().
+	// See https://github.com/grafana/k6/pull/2602
+	vu.InitEnvField = initEnv
+	vu.StateField = nil
```
inancgumus, line 137: I have a funny feeling about this fix :flushed: Should we better work on a proper fix in the next cycle? 🤔

imiric:
OK, so it should be fixed with aaae86c. Like I mentioned, it's a _hackfix_ and we should find a cleaner way of doing this.

@mstoykov Could you also review aaae86c and let us know if there's a better way? :pray:

tests/test_browser.go
imiric: Yes, this is temporary and we should find a better way, hopefully with @mstoykov's help.

k6ext/k6test/vu.go
```diff
@@ -10,16 +9,21 @@ import (
...
 	"github.com/stretchr/testify/require"
 	"gopkg.in/guregu/null.v3"
 )
 
 // VU is a k6 VU instance.
+// TODO: Do we still need this VU wrapper?
+// ToGojaValue can be a helper function that takes a goja.Runtime (although it's
+// not much of a helper from calling ToValue(i) directly...), and we can access
+// EventLoop from modulestest.Runtime.EventLoop. I guess we still need the
+// RunLoop() override to call WaitOnRegistered()?
```
inancgumus, line 26: I would argue that k6 APIs (especially the new ones like `modulestest`) are in flux, and we better keep the wrapper and other helpers. Also, the helpers provide easy-to-remember (at least, for me) names for the common and mostly verbose operations.

inancgumus (commented):
🤔

tests/test_browser.go
inancgumus: 🤞

imiric:
OK, merging this then. If someone has an improvement suggestion, we can work on it for the next release.

ankur22 (approved):

tests/test_browser.go
mstoykov: Something like https://github.com/grafana/xk6-browser/pull/525 will is how the API is more appropriate to be used.

k6ext/k6test/vu.go
mstoykov: 1. I would argue you should not have `WaitOnRegistered` *before* the `LoopStart` you might want to have it after to wait on everything to finish in case of errors. But adding it before just masks that a previous test didn't do something properly and you are reusing the same loop.
2. You can still just wrap `modulestest.Runtime` instead of VU ... but maybe you will need to rename the struct ;). And if you embed modulestest.Runtime (or the wrapper) in testBrowser, you should need to touch as many places to rename the access to `VU` due to how embedding fields in structs works.

inancgumus: 1. Yeah, I was suspicious of adding `WaitOnRegistered` there. The problem is in the tests (`slow_mo` tests) that share the same browser and VU.
2. I guess `modulestest.Runtime` is added in k6@v0.4.0 and we can take advantage of it 👍

PR #524, opened 2022-09-09
Remove go and use chan to keep CDP msg order

ankur22:
Closes: https://github.com/grafana/xk6-browser/issues/482

The order of the CDP messages can quite easily be lost after being received and handed to the handlers that are listening in on the events. To alleviate this ordering issue, remove the `goroutine` that is used to place the event on the listening handlers, and instead use a `buffered channel`. The `buffered channel` avoids a deadlock where by an event handler could be busy emitting another event, while the emit method is busy trying to pass an event to the same handler.

[Diagram](https://drive.google.com/file/d/1l8GFLqTe3KiyHQGtOYgVhIKyV1izOEAg/view?usp=sharing).
CDP event loop break ordering drawio

This has been tested in the cloud and after several tests, xk6-browser has not deadlocked 🎉 

I have another solution but I prefer this solution as it's closer to the root cause of the issue -- https://github.com/grafana/xk6-browser/pull/512 (which is still WIP).

### DIFF
--- a/common/event_emitter.go
+++ b/common/event_emitter.go
@@ -22,6 +22,9 @@ package common
 
 import (
 	"context"
+	"time"
+
+	"github.com/grafana/xk6-browser/log"
 )
 
 // Ensure BaseEventEmitter implements the EventEmitter interface.
@@ -76,6 +79,10 @@ const (
 	// Worker
 
 	EventWorkerClose string = "close"
+
+	// Event listener.
+
+	EventListenerDefaultChanBufferSize int64 = 10
 )
 
 // Event as emitted by an EventEmiter.
@@ -103,7 +110,7 @@ type eventHandler struct {
 
 // EventEmitter that all event emitters need to implement.
 type EventEmitter interface {
-	emit(event string, data interface{})
+	emit(logger *log.Logger, event string, data interface{})
 	on(ctx context.Context, events []string, ch chan Event)
 	onAll(ctx context.Context, ch chan Event)
 }
@@ -164,12 +171,22 @@ func (e *BaseEventEmitter) sync(fn func()) {
 	<-done
 }
 
-func (e *BaseEventEmitter) emit(event string, data interface{}) {
+func (e *BaseEventEmitter) emit(logger *log.Logger, event string, data interface{}) {
 	emitEvent := func(eh eventHandler) {
-		select {
-		case eh.ch <- Event{event, data}:
-		case <-eh.ctx.Done():
-			// TODO: handle the error
+		for {
+			select {
+			case eh.ch <- Event{event, data}:
+				return
+			case <-eh.ctx.Done():
+				// TODO: handle the error
+				return
+			case <-time.After(time.Second):
+				// If this is printed then a handler is deadlocked on something
+				// and the channel (which should be buffered) is full. It might
+				// be that by increasing the buffer size on the channel will help
+				// alleviate the problem.
+				logger.Error("emit timed out waiting for handler to read message")
+			}
 		}
 	}
 	emitTo := func(handlers []eventHandler) (updated []eventHandler) {
@@ -180,7 +197,7 @@ func (e *BaseEventEmitter) emit(event string, data interface{}) {
 				handlers = append(handlers[:i], handlers[i+1:]...)
 				continue
 			default:
-				go emitEvent(handler)
+				emitEvent(handler)
 				i++
 			}
 		}
--- a/common/barrier_test.go
+++ b/common/barrier_test.go
@@ -41,7 +41,7 @@ func TestBarrier(t *testing.T) {
 
 	barrier := NewBarrier()
 	barrier.AddFrameNavigation(frame)
-	frame.emit(EventFrameNavigation, "some data")
+	frame.emit(log, EventFrameNavigation, "some data")
 
 	err := barrier.Wait(ctx)
 	require.Nil(t, err)
--- a/common/browser.go
+++ b/common/browser.go
@@ -168,7 +168,7 @@ func (b *Browser) getPages() []*Page {
 func (b *Browser) initEvents() error {
 	var cancelCtx context.Context
 	cancelCtx, b.evCancelFn = context.WithCancel(b.ctx)
-	chHandler := make(chan Event)
+	chHandler := make(chan Event, EventListenerDefaultChanBufferSize)
 
 	b.conn.on(cancelCtx, []string{
 		cdproto.EventTargetAttachedToTarget,
@@ -313,7 +313,7 @@ func (b *Browser) onAttachedToTarget(ev *target.EventAttachedToTarget) {
 		b.sessionIDtoTargetID[ev.SessionID] = evti.TargetID
 		b.sessionIDtoTargetIDMu.Unlock()
 
-		browserCtx.emit(EventBrowserContextPage, p)
+		browserCtx.emit(b.logger, EventBrowserContextPage, p)
 	default:
 		b.logger.Warnf(
 			"Browser:onAttachedToTarget", "sid:%v tid:%v bctxid:%v bctx nil:%t, unknown target type: %q",
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -370,7 +370,7 @@ func (b *BrowserContext) WaitForEvent(event string, optsOrPredicate goja.Value) 
 	}
 
 	evCancelCtx, evCancelFn := context.WithCancel(b.ctx)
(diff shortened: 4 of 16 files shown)


inancgumus:
I once tried using buffered channels across the handlers while working on this problem, which didn't help back then. Another thing that worried me was making the interaction between components/events more unpredictable due to the nature of buffered channels.

ankur22:
> I once tried using buffered channels across the handlers while working on this problem, which didn't help back then.

I've had great success with this approach so far. I've been running stress tests for many minutes and usually I hit upon the following but at the moment nothing:
- nill response in the locator tests,
- deadlock in the locator tests,
- nill error when trying to print the session id.

> Another thing that worried me was making the interaction between components/events more unpredictable due to the nature of buffered channels.

Can you elaborate on what exactly makes channels unpredictable? It's definitely keeping the order of the incoming events now.

inancgumus:
> Can you elaborate on what exactly makes buffered channels unpredictable? It's definitely keeping the order of the incoming events now.

With an unbuffered channel, the communication between parties is synchronized. And _this nature of unbuffered channels helps bringing the deadlocks to the surface_. However, with a buffered channel, there is no synchronization between the concurrent components (no lock-step). So, it is (and will be) harder to figure out the underlying race condition problems.

The race conditions and deadlocks might not be because of the channel capacity. Instead, the problems might be happening because the components' orchestration and logic are incorrect. The deadlocks may seem to go while the work of components is lightweight. But the race condition problems will probably _unpredictably_ come back later.

🤞

ankur22:
> With an unbuffered channel, the communication between parties is synchronized. And _this nature of unbuffered channels helps bring the deadlocks to the surface_. However, with a buffered channel, there is no synchronization between the concurrent components (no lock-step). So, it is (and will be) harder to figure out the underlying race condition problems.

Aha, yes, that make sense. In this case, just maybe, there's a possibility that by removing the goroutine that was placing the events on the handlers, it's preventing the deadlock. It's difficult to tell thought whether it's actually just hidding the probelm like you've suggested 😵‍💫 

> The race conditions and deadlocks might not be because of the channel capacity. Instead, the problems might be happening because the components' orchestration and logic are incorrect. The deadlocks may seem to go while the work of components is lightweight. But the race condition problems will probably _unpredictably_ come back later (Let's hope that won't happen in production!).

I've been running the stress tests locally, and unfortunately I've ran into the "nil response body" and "nil error when trying to print the session id", but still not the deadlock issue that i'm trying to tackle in the first place 🤞 

The next step is testing this in the cloud. Prior to this fix, I was coming across the deadlock issue in the cloud 90% of the time I ran a test. Hopefully it doesn't show up there, which could make this a good candidate. We still have https://github.com/grafana/xk6-browser/pull/512 though which forces the synchronisation of the create and clear contexts, which also needs to be tested in the cloud.

inancgumus:
> In this case, just maybe, there's a possibility that by removing the goroutine that was placing the events on the handlers, it's preventing the deadlock, and it seems that the order of the incoming events is kept.

The thing is each component has its own event emitter. So there can be out-of-order events between different event emitters. The cause of the problem is there is no mechanism to manage this on a higher-level. Maybe we should look into that.

> still not the deadlock issue that i'm trying to tackle in the first place

I'm suspicious that we might be hiding it (deadlock). Even if there isn't a deadlock, it doesn't mean that we don't have race conditions :( I'll try this fix with a complex website like Vistaprint when you finish 🤞

common/event_emitter.go
```diff
@@ -180,7 +197,7 @@ func (e *BaseEventEmitter) emit(event string, data interface{}) {
 				handlers = append(handlers[:i], handlers[i+1:]...)
 				continue
 			default:
-				go emitEvent(handler)
+				emitEvent(handler)
```
ankur22, line 200: Bye bye goroutine -- this helps keep the order of incoming events.

common/event_emitter.go
```diff
@@ -76,6 +79,10 @@ const (
 	// Worker
 
 	EventWorkerClose string = "close"
+
+	// Event listener.
+
+	EventListenerDefaultChanBufferSize int64 = 10
```
ankur22, line 85: I would prefer not have to work with buffered channels, but the issue is with handlers which not only listen and handle incoming events, they can also `emit` events too. Without the buffered channel this would deadlock 😭

inancgumus: Why it is 10?

inancgumus: Why is it `10`?

common/event_emitter.go
inancgumus: 😱

common/event_emitter.go
```diff
@@ -164,12 +171,22 @@ func (e *BaseEventEmitter) sync(fn func()) {
...
+				return
+			case <-eh.ctx.Done():
+				// TODO: handle the error
+				return
+			case <-time.After(time.Second):
+				// If this is printed then a handler is deadlocked on something
+				// and the channel (which should be buffered) is full. It might
+				// be that by increasing the buffer size on the channel will help
+				// alleviate the problem.
+				logger.Error("emit timed out waiting for handler to read message")
```
inancgumus, line 188: Should we then cut this short here (`return` without `for`)? Now it retries, is it expected?

common/event_emitter.go
ankur22: The `10` was a just a random number. I was thinking about this, and actually, it might be better to look at trying to prevent the handler(s) from emitting and receiving event(s), so that we don't need the buffered channels (and leave it as unbuffered). IMO, it doesn't feel right for a handler to be doing these two jobs, and should concentrate on handling the events and not emitting new ones, but I need to look into this a bit more (which could take some time to make sense of).

common/event_emitter.go
ankur22: Yes, i'll add a `return` if we're all happy with this change -- it's currently useful as a tool to debug an issue with pprof 🙂

imiric (commented):
Nice work digging into this :clap:

While I do think this would improve the ordering of emitted events, it wouldn't address the functional concern of dependencies between events. I think the deadlock might not be manifesting in simple scenarios, but would happen on more complex sites, as mentioned in comments above.

For example, consider the `Runtime.executionContextCreated` / `Runtime.executionContextsCleared` situation. Among other things, we need to ensure that:

- `FrameSession.onExecutionContextCreated()` always finishes processing before `FrameSession.onExecutionContextsCleared()` processing starts.
  The `FrameSession.contextIDToContextMu` mutex addresses the concurrency problem, but doesn't enforce the flow of execution. I.e. one could finish before the other, and we could end up with contexts that should've been cleared, or cleared contexts that shouldn't have been.

- That `Frame.waitForExecutionContext()` isn't called, or at least returns early, if the frame is done navigating, and `onExecutionContextsCleared()` was called.
  I.e. if there are no contexts, and no navigation is happening/expected, whatever called `waitForExecutionContext()` is probably anticipating a document to exist, and it points to the same functional race condition we've seen in other places as well.


I'm a bit weary of doing this global change on `EventEmitter`, since it might impact how other events are processed. I.e. given that event emission is no longer async, it might introduce delays for other events, of which we don't know the effects yet.

So I'd prefer a specific fix just for this `waitForExecutionContext()` issue instead. But it's not a strong preference, and if you or Inanc oppose this, I would be OK with merging this and testing it out for some time. We can always revert it later, and we're still early in the cycle.

Are there any major drawbacks of merging the fix from #512? I haven't tested it in depth, and the code wasn't trivial to grok when I first took a look, but it might be preferable to this global change.

<hr>

Another approach I'm still lingering on from #427, is having a separate channel or a small finite-state machine for specific event flows. E.g. we know that for a specific frame `executionContextCreated` should always be processed before `executionContextsCleared`. If that is violated that means that we're receiving CDP messages out of order, and that's a protocol level error out of our hands (and AFAIK we haven't noticed yet). Even though CDP is async, some events are always ordered within the context of a frame.

So I'm picturing a separate frame-specific channel to which these events are sent, and then a separate frame-specific reader would process them _syncronously_ and wait for each event handler to complete, before starting the next one. This would get rid of most locking since we'd be sure things like `FrameSession.contextIDToContext` couldn't be accessed concurrently.

Maybe it's worth exploring that solution for this isolated issue, and we can expand it to other similar event flows later. I tried doing something like this for the #427 refactor, but it was a lot of effort to take on all at once.

Anyway, just an idea. I would be fine with #512 if it improves this issue.

ankur22:
Unsuitable as it's only hiding the deadlock, and with more complex tests this will definitely deadlock.

PR #525, opened 2022-09-12
Use moveToContext only when needed in tests

mstoykov:


imiric (approved):
Thanks, much cleaner :+1:

PR #526, opened 2022-09-12
Parallelize tests

mstoykov:


inancgumus (approved):
LGTM 👍

imiric (approved):
We used to have flakiness issues running some tests in parallel, which is why it wasn't enabled in some.

I'm fine with this change, and hopefully things are more stable now. :crossed_fingers:

It shaves off ~1m in CI and ~20s for me locally, so :+1:

PR #527, opened 2022-09-12
WIP: add more code examples for locator API

mdcruz:
This PR adds more code examples to show the different methods that can be used as part of the Locator API


inancgumus:
Hey @mdcruz, thanks for the examples.

However, I'm not sure whether we need more locator examples like this as the locator API's a forwarder to the inner Frame/ElementHandle methods. We demonstrated the benefit of using locators in [the locator example](https://github.com/grafana/xk6-browser/blob/main/examples/locator.js) already. The benefit is that the locator methods can work across navigations (the main benefit).

For more examples, I believe we can instead add these to the xk6-browser documentation.

As these examples also run in our CI whenever we make a commit, and we don't want to slow down there (although there are some solutions for that.)

mdcruz:
> Hey @mdcruz, thanks for the examples.
> 
> However, I'm not sure whether we need more locator examples like this as the locator API is a forwarder to the inner Frame/ElementHandle methods (exactly mimics their behavior and we have examples for these).
> 
> We demonstrated the benefit of using locators in [the locator example](https://github.com/grafana/xk6-browser/blob/main/examples/locator.js) already. The benefit is that the locator methods can work across navigations (the main benefit).
> 
> For more examples, I believe we can instead add these to the xk6-browser documentation for each locator method (or for the most).
> 
> As these examples also run in our CI whenever we make a commit, and we don't want to slow down there (although there are some solutions for that.)

Hi Inanc, thanks for the feedback! If it's to add it to xk6-browser documentation, I believe @ankur22 has already added them.

I'll close the PR if we feel that the locator.js example is enough already 👍🏼

inancgumus:
> I'll close the PR if we feel that the locator.js example is enough already 👍🏼

I believe it is, but let's also ask the team. @grafana/k6browserdev WDYT?

imiric:
I agree with @inancgumus. There's no need to have standalone Locator examples for each method it wraps, but maybe we could have one additional example to showcase all of them. Though they would definitely make sense on each method's API docs page.

Since the examples in the repo are also used in the E2E pipeline, these could be more complex, showcasing some more advanced workflows, like integration with k6, different scenarios/executors, threshold use, etc.

Also, we should probably adjust the CI workflow to consider JS files from subdirectories as well, since [it currently picks up only a single level](https://github.com/grafana/xk6-browser/blob/38f454007ff2608857f4572d562eaba84b1e1322/.github/workflows/e2e.yml#L68-L70), so it wouldn't have run the ones in the `examples/locator` directory added in this PR.

mdcruz:
Thanks both! I'll close this PR for now.

PR #529, opened 2022-09-12
add example on how to use xk6-browser with k6

mdcruz:
This PR adds a new test to demo how xk6-browser can be used together with k6. This includes:

- creating multiple scenarios where each scenario executes a different function
- different executor used for each scenario
- use of thresholds

### DIFF
--- a/examples/multiple-scenario.js
+++ b/examples/multiple-scenario.js
@@ -1,0 +1,49 @@
+import { chromium } from 'k6/x/browser'
+
+export const options = {
+  scenarios: {
+    messages: {
+      executor: 'constant-vus',
+      exec: 'messages',
+      vus: 2,
+      duration: '2s',
+    },
+    news: {
+      executor: 'per-vu-iterations',
+      exec: 'news',
+      vus: 2,
+      iterations: 4,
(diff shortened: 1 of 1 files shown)


examples/multiple-scenario.js
```diff
@@ -0,0 +1,49 @@
...
+    browser_dom_content_loaded: ['p(90) < 1000'],
+    browser_first_contentful_paint: ['max < 1000']
+  }
+}
+
+export function messages() {
+  const browser = chromium.launch({
+    headless: __ENV.XK6_HEADLESS ? true : false,
+  });
+  const context = browser.newContext();
```
ankur22, line 29: In your WOLT presentation you were suggesting that there was an unnecessary step (possibly `const context = browser.newContext();`). So would it make is easier on new users if this was `const page = browser.newPage();`?

ankur22 (approved):

imiric (approved):
LGTM, thanks!

examples/multiple-scenario.js
mdcruz: You're right! I'll update it.

Issue #469, opened 2022-07-27
Match Dblclick with Playwright's behavior


ankur22:
TODO:

[xk6-browser docs](https://k6.io/docs/javascript-api/xk6-browser/locator/dblclick/) imply that this `dblclick` issue is about locators, however, it's the behavior in our extension in general. 

---

Tested against: https://github.com/grafana/xk6-browser/commit/dbede120c63df43995813a847a25b0e66e289592

The method `dblclick` on `locator` only clicks once.

In the below code it works against `localhost:8080`, which is a test server that can be found here with instructions on how to run it locally: https://github.com/ankur22/testserver.

```javascript
import { sleep } from 'k6';
import { chromium } from 'k6/x/browser';

export default function () {
  const browser = chromium.launch({
    headless: false,
  });
  const context = browser.newContext();
  const page = context.newPage();
  const res = page.goto('https://test.k6.io/browser.php');
  const l = page.locator('#counter-button');
  sleep(1);
  l.dblclick();

  sleep(10);
  browser.close();
}
```

I was expecting the counter on this test page to go up to `2` but it only went up to `1`. When testing with PW i was able to get the expected count of `2`.
```[tasklist]
### Tasks
- [x] FIx issue -- https://github.com/grafana/xk6-browser/pull/1121
- [x] Add to release notes -- https://github.com/grafana/k6/pull/3506
- [x] Docs update -- https://github.com/grafana/k6-docs/pull/1456
```

imiric:
Nice catch!

It doesn't seem related to `Locator`, but to `ElementHandle`:

https://github.com/grafana/xk6-browser/blob/dbede120c63df43995813a847a25b0e66e289592/common/element_handle.go#L242-L244

It should call `Mouse.dblClick()` :smile:

inancgumus:
@ankur22 Nice catch 🏉 @imiric No, it's not because of not calling `dblClick`. Because some locator tests are incorrect 🥲 I'll send a PR.

inancgumus:
@ankur22 Thanks Ankur for helping me find the problems in the locator tests 🙇 

@imiric No, it's not because of not calling `Mouse.dblClick`. Because some locator tests are incorrect 🤦 I'll send a PR.

**Update:**
Ankur, we can double-click. The test server starts the counter from zero. So we see 1 instead of 2 in the k6 logs.

[...]

PR #389, opened 2022-06-10
Refactor locator tests to run in parallel

inancgumus:
Resolves: https://github.com/grafana/xk6-browser/pull/388#discussion_r894552608
Resolves: https://github.com/grafana/xk6-browser/pull/372#discussion_r890936907

Related: #366

PR #532, opened 2022-09-13
Fix parallel locator tests

inancgumus:
The `dblclick` test was wrong. It was listening for the `click` event instead of the `dblclick` event.

And the other tests were wrong too. We missed preventing capturing the loop variables in parallel tests in #389. So we weren't testing the existing locator features so far. The `dblclick` test started to fail when I fix the locator tests (so I fixed it afterward).

Somewhat related: #469


imiric (approved):
:+1:

Issue #510, opened 2022-08-18
Event loop state corruption(?) 


sniku:
I'm struggling to come up with a good name for the issue because the underlying bug is not clear to me.

## Version
Build using the latest main branch [3a1e01b](https://github.com/grafana/xk6-browser/commit/3a1e01b0976cb6c91018e9c1aa01ead47bc3bf43). 

```
➜  ~ xk6-browser4 version
k6 v0.39.0 ((devel), go1.18.5, linux/amd64)
```

## Script 
```js
import { check } from 'k6';
import { Counter } from 'k6/metrics';
import { chromium } from 'k6/x/browser';


const checkpoint1 = new Counter('checkpoint1');
const checkpoint2 = new Counter('checkpoint2');
const checkpoint3 = new Counter('checkpoint3');
const checkpoint4 = new Counter('checkpoint4');
const checkpoint5 = new Counter('checkpoint5');
const checkpoint6 = new Counter('checkpoint6');
const checkpoint7 = new Counter('checkpoint7');

export let options = {
  vus: 1,
  duration: '50m'
}

export default function () {
  const browser = chromium.launch({
    headless: false,
    // executablePath: '/usr/bin/google-chrome-stable',
    executablePath: '/usr/bin/chromium',
  });
  const context = browser.newContext();
  const page = context.newPage();

  checkpoint1.add(1);

  page.goto('https://test.k6.io/', { waitUntil: 'networkidle' });

  checkpoint2.add(1);

  Promise.all([
    page.waitForNavigation(),
    page.locator('a[href="/my_messages.php"]').click(),
  ]).then(() => {
    checkpoint3.add(1)

    page.locator('input[name="login"]').type('admin');
    page.locator('input[name="password"]').type('123');

    checkpoint4.add(1)

    return Promise.all([
      page.waitForNavigation(),
      page.locator('input[type="submit"]').click(),
    ]);
  }).then(() => {
    checkpoint5.add(1)

    check(page, {
      'header': page.locator('h2').textContent() == 'Welcome, admin!',
    });
  }).finally(() => {
    checkpoint6.add(1)
    page.close();
    browser.close();
    checkpoint7.add(1)
  });
}

```

## Result

On a local machine, the test usually runs correctly for a long time. In this case, it was running for about 40 min. In the cloud, the failure happens much quicker - usually within 1-2 min. 

The browser behavior is surprising.

1. The test keeps running. New browser windows are being opened as usual.
[...]

Issue #232, opened 2022-02-09
Error closing the WebSocket connection at end of test


unlikelyzero:
I'm using xk6 and xk6-browser for the first time. When running the example script:

```
import launcher from 'k6/x/browser';

export default function() {
  const browser = launcher.launch('chromium', {
    headless: __ENV.XK6_HEADLESS ? true : false,
  });
  const context = browser.newContext();
  const page = context.newPage();
  page.goto('https://test.k6.io/');
  page.screenshot({ path: 'screenshot.png' });
  // TODO: Assert this somehow. Upload as CI artifact or just an external `ls`?
  // Maybe even do a fuzzy image comparison against a preset known good screenshot?
  page.close();
  browser.close();
}
```

I see the following output:
```
./xk6-browser run ./perf/tests/xk6-browser/example.js

          /\      |‾‾| /‾‾/   /‾‾/
     /\  /  \     |  |/  /   /  /
    /  \/    \    |     (   /   ‾‾\
   /          \   |  |\  \ |  (‾)  |
  / __________ \  |__| \__\ \_____/ .io

  execution: local
     script: ./perf/tests/xk6-browser/example.js
     output: -

  scenarios: (100.00%) 1 scenario, 1 max VUs, 10m30s max duration (incl. graceful stop):
           * default: 1 iterations for each of 1 VUs (maxDuration: 10m0s, gracefulStop: 30s)

WARN[0003] error fetching response body for 'https://test.k6.io/': No data found for resource with given identifier (-32000)  category=cdp elapsed="0 ms" goroutine=115
WARN[0003] error fetching response body for 'https://test.k6.io/static/favicon.ico': context canceled  category=cdp elapsed="0 ms" goroutine=115

running (00m03.1s), 0/1 VUs, 1 complete and 0 interrupted iterations
default ✓ [========================] 1 VUs  00m03.1s/10m0s  1/1 iters, 1 per VU

     browser_dom_content_loaded.......: avg=166.25ms min=113µs    med=166.25ms max=332.39ms p(90)=299.16ms p(95)=315.78ms
     browser_first_contentful_paint...: avg=379.75ms min=379.75ms med=379.75ms max=379.75ms p(90)=379.75ms p(95)=379.75ms
     browser_first_paint..............: avg=379.66ms min=379.66ms med=379.66ms max=379.66ms p(90)=379.66ms p(95)=379.66ms
[...]

PR #534, opened 2022-09-16
Minor BrowserType and Connection refactor

imiric:
These are minor changes I did while working on #510 / #232. See the commits for details.

### DIFF
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -1,11 +1,9 @@
 package chromium
 
 import (
-	"bufio"
 	"context"
 	"errors"
 	"fmt"
-	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


ankur22 (approved):

inancgumus (approved):
**Suggestion:**
If I understand correctly, by looking at the next PR #535, we do this refactoring to get the process-done signal from the command (`exec.CommandContext` in `execute()` to be specific). IMO, since we try to split God package `common`, a better refactoring could be to move `execute()` and `parseDevToolsURL` parsing to another package (other than `chromium`), making there a new type with these functions and listen for the process-done signal from it.

LGTM as it is 👍

imiric:
@inancgumus This refactor is mostly unrelated to #535. That branch is based on this one just because I did the changes together, and it was easier to create like this than have to resolve conflicts.

This is just a minor conceptual cleanup, and I didn't want to get into anything major, like creating a separate package. Besides, what would that package name be? `execute` and `parseDevToolsURL` probably don't belong in an isolated package, since they're closely related to `BrowserProcess`. So maybe `browser` would be a good top-level package candidate? But that opens up a whole can of dependency worms, and like I said, I didn't want to get into it here. :sweat_smile:

Thanks for the review :thumbsup:

inancgumus:
`xk6-browser/process` could be a good package name for dealing with browser process stuff. But, as I said, this PR is OK as it is, and we don't want dependency-worms at this point, fair.

PR #535, opened 2022-09-16
Improve handling of browser process exit and WS connection closing

imiric:
This was initially an effort for #510, but since that became difficult to reproduce even in EC2, it pivoted to fixing #232 and some related issues.

See the commits for details, but in summary, this:
- Ensures we wait for the browser process to exit before ending the iteration.
  Previously, we would cancel the browser context prematurely, which caused `process with PID 282373 unexpectedly ended: signal: killed` errors.
  Now, we wait for the browser process to exit on its own, since the CDP `Browser.close` command does that implicitly, and only forcefully kill it if it exceeds a specific timeout (1s).
- Fixes, or rather _hides_, the `close 1006 (abnormal closure): unexpected EOF` error from #232.
  See, Chrom{e,ium} doesn't close the WS connection cleanly (they don't send a [Close control frame](https://www.rfc-editor.org/rfc/rfc6455#section-1.4)) after we send `Browser.close`, or even the TCP socket for that matter; they just end the process. :facepalm: So in the case this happens as a result of us sending `Browser.close`, we just disregard the error, though still log it as Debug for troubleshooting purposes.

### DIFF
--- a/common/connection.go
+++ b/common/connection.go
@@ -25,7 +25,6 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
-	"net"
 	"net/http"
 	"sync"
 	"sync/atomic"
@@ -57,6 +56,7 @@ type executorEmitter interface {
 type connection interface {
 	executorEmitter
 	Close(...goja.Value)
+	IgnoreIOErrors()
 	getSession(target.SessionID) *Session
 }
 
@@ -130,6 +130,7 @@ type Connection struct {
 	closeCh      chan int
 	errorCh      chan error
 	done         chan struct{}
+	closing      chan struct{}
 	shutdownOnce sync.Once
 	msgID        int64
 
@@ -168,6 +169,7 @@ func NewConnection(ctx context.Context, wsURL string, logger *log.Logger) (*Conn
 		closeCh:          make(chan int),
 		errorCh:          make(chan error),
 		done:             make(chan struct{}),
+		closing:          make(chan struct{}),
 		msgID:            0,
 		sessions:         make(map[target.SessionID]*Session),
 	}
@@ -186,16 +188,10 @@ func (c *Connection) close(code int) error {
 	var err error
 	c.shutdownOnce.Do(func() {
 		defer func() {
-			_ = c.conn.Close()
-
 			// Stop the main control loop
 			close(c.done)
+			_ = c.conn.Close()
 		}()
-
-		err = c.conn.WriteControl(websocket.CloseMessage,
-			websocket.FormatCloseMessage(code, ""),
-			time.Now().Add(10*time.Second),
-		)
 
 		c.sessionsMu.Lock()
 		for _, s := range c.sessions {
@@ -203,6 +199,19 @@ func (c *Connection) close(code int) error {
 			delete(c.sessions, s.id)
 		}
 		c.sessionsMu.Unlock()
+
+		err = c.conn.WriteControl(websocket.CloseMessage,
+			websocket.FormatCloseMessage(code, ""),
+			time.Now().Add(time.Second),
+		)
+
+		// According to the WS RFC[1], we might want to wait for a response
+		// Control frame back from the browser here (possibly for the above
+		// timeout duration), but Chrom{e,ium} never sends one, even when
+		// the browser process exits normally after the Browser.close CDP
+		// command. So we don't bother waiting, since it would just needlessly
+		// delay the k6 iteration.
+		// [1]: https://www.rfc-editor.org/rfc/rfc6455#section-1.4
 
 		c.emit(EventConnectionClose, nil)
 	})
@@ -238,15 +247,21 @@ func (c *Connection) createSession(info *target.Info) (*Session, error) {
 }
 
 func (c *Connection) handleIOError(err error) {
-	c.logger.Errorf("cdp", "communicating with browser: %v", err)
+	// It's either a normal closure initiated by the browser, or we stopped the
+	// connection. In either case, disregard the error.
+	if closing := c.isClosing(); websocket.IsCloseError(
+		err, websocket.CloseNormalClosure, websocket.CloseGoingAway,
+	) || closing {
+		c.logger.Debugf("cdp", "received IO error: %v, connection is closing: %v", err, closing)
+		return
+	}
 
-	if websocket.IsUnexpectedCloseError(err, websocket.CloseNormalClosure, websocket.CloseGoingAway) {
-		// Report an unexpected closure
-		select {
-		case c.errorCh <- err:
-		case <-c.done:
-			return
-		}
+	// Report an unexpected closure
+	c.logger.Errorf("cdp", "communicating with browser: %v", err)
+	select {
+	case c.errorCh <- err:
+	case <-c.done:
+		return
 	}
 	var (
 		cerr *websocket.CloseError
@@ -288,10 +303,7 @@ func (c *Connection) recvLoop() {
 	for {
 		_, buf, err := c.conn.ReadMessage()
 		if err != nil {
-			if !errors.Is(err, net.ErrClosed) {
-				c.logger.Debugf("Connection:recvLoop", "wsURL:%q ioErr:%v", c.wsURL, err)
-				c.handleIOError(err)
-			}
+			c.handleIOError(err)
 			return
 		}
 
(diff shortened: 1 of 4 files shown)


imiric:
I'll look into the CI failure on Monday. It didn't happen locally. :disappointed:

common/connection.go
```diff
@@ -178,31 +179,31 @@ func NewConnection(ctx context.Context, wsURL string, logger *log.Logger) (*Conn
...
-			delete(c.sessions, s.id)
-		}
-		c.sessionsMu.Unlock()
+		// According to the WS RFC[1], we might want to wait for a response
+		// Control frame back from the browser here (possibly for the above
+		// timeout duration), but Chrom{e,ium} never sends one, even when
+		// the browser process exits normally after the Browser.close CDP
+		// command. So we don't bother waiting, since it would just needlessly
+		// delay the k6 iteration.
+		// [1]: https://www.rfc-editor.org/rfc/rfc6455#section-1.4
```
inancgumus, line 206: How are we sure that Chrom{e,ium} never makes a close handshake (sends a close control frame)?

It's because the Chromium code seems like it does so:
* https://chromium.googlesource.com/chromium/src.git/+/master/net/websockets/websocket_channel_test.cc#1100
* https://chromium.googlesource.com/chromium/src.git/+/master/net/websockets/websocket_channel.cc#333

Btw, is it possible that a race condition may cause keep sending messages to the browser after we send the close frame?

mstoykov: From my and @imiric chromium *never* sends close frames when using with CDP. Irregardless of how the closing/shutdown has been initiated.  

The links you ahve provided seem to be related to the websockets functionality used for normal websockets connection initiated by javascript code. 

I have no idea if that code is being used by the CDP connection as well :shrug:.

common/connection.go
```diff
@@ -488,6 +507,9 @@ func (c *Connection) sendLoop() {
 	}
 }
 
+// Close cleanly closes the WebSocket connection.
+// It returns an error if sending the Close control frame fails.
+// TODO: Remove this and close(), since nothing calls them anymore.
 func (c *Connection) Close(args ...goja.Value) {
```
imiric, line 513: BTW, should we do the TODO here? We're not using this anymore, and I doubt that will change, so might as well remove it?

inancgumus: `c.close()` is called from several places. On the other hand, if `Close()` is not being used anymore, why not call `Stop` -> `Close` instead? `connection.Stop` sounds weird.

common/connection.go
inancgumus: > @mstoykov 
> The links you ahve provided seem to be related to the websockets functionality used for normal websockets connection initiated by javascript code.
>
> I have no idea if that code is being used by the CDP connection as well 🤷.

> @imiric 
> I think you're looking at the web implementation, i.e. how the browser itself handles connections initiated from JS. Notice that they mention JS in one of the comments. This might not be the code used by DevTools, though I'm not familiar enough with either C++ or the Chromium code base to know for sure.

DevTools uses the same websocket functionality, see [this](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/devtools_http_handler.cc;l=181?q=%22devtools%20listening%22&ss=chromium).

It's OK if we are sure since we checked the WS packet trace, etc. 👍

common/connection.go
imiric: > `c.close()` is called from several places.

Ah, right, I missed the `closeCh` readers. Fair enough, we shouldn't delete it. And there's not much reason to remove `Close()` either then. It's only used in tests, but we can keep those.

I'll remove the TODO then.


> On the other hand, if `Close()` is not being used anymore, why not call `Stop` -> `Close` instead? `connection.Stop` sounds weird.

The rename would be confusing in that case, since `Stop()` doesn't actually close the connection, just... stops it (or rather, signals the `recvLoop()` error handling to ignore the `unexpected EOF` error). I couldn't think of a clearer name, but since we're not deleting `Close()`, I think we can keep it. Otherwise, I'm open to better name suggestions, just not `Close`.

imiric: I don't think we're bikeshedding yet, just picking a better name. :)

Again, I don't disagree with you about `Stop`. Just that `Close` wouldn't be an improvement.

`IgnoreIOErrors()` is clearer, so I'll rename it to that. Thanks.

imiric: Should be resolved by eb75063. I couldn't think of better names for the internal channel and unexported function to check it, so I settled on `closing` and `isClosing`.

ankur22 (approved):

common/connection.go
```diff
@@ -238,15 +250,25 @@ func (c *Connection) createSession(info *target.Info) (*Session, error) {
...
+	// It's either a normal closure initiated by the browser, or we stopped the
+	// connection. In either case, disregard the error, but close all sessions
+	// and emit a ConnectionClose event, so that the browser event listener can
+	// be stopped.
+	if closing := c.isClosing(); websocket.IsCloseError(
+		err, websocket.CloseNormalClosure, websocket.CloseGoingAway,
+	) || closing {
+		c.logger.Debugf("cdp", "received IO error: %v, connection is closing: %v", err, closing)
+		c.closeAllSessions()
+		c.emit(EventConnectionClose, nil)
```
mstoykov, line 262:
```suggestion
		if closing {
			c.emit(EventConnectionClose, nil)
		}
```

From my local testing the error that we see in the CI as well is because this event gets emitted while the `browser.close` event is still being "done" but after it was already send - so that the actual connection is closed. 

While the transport for that works on to return that `browser.Close` has been finished, the event gets emitted and then that leads to the cancelation of the connection browser context in https://github.com/grafana/xk6-browser/blob/430e82c73115ee34b13721d364dd95c664a2c2e6/common/browser.go#L183-L185
https://github.com/grafana/xk6-browser/blob/430e82c73115ee34b13721d364dd95c664a2c2e6/common/browser.go#L198-L201

mstoykov: Unfortunately as always I am getting other failures - some of which I have a patch for ... the others are timeouts which ... seem bad, but I am not certain if I have somehow fixed them in the last few minutes or they just do not happen at the moment :shrug:

mstoykov: Also you still need to emit that event :facepalm: but after the close has happened. I did return to calling the connection.close method and am currently testing but ... likely we can just emit the event after the browser.Close :thinking: 

On the other hand maybe it should be emitted after we *actually* close the connectiong :shrug:. (given that is the name of the event ;))

ankur22 (commented):
I'm a little lost actually, i need to take another look at this 😅

common/connection.go
mstoykov: https://github.com/grafana/xk6-browser/pull/541/commits/f24516934a417c5ff1511e4308cbfa6c56c5c38b is what I mean

imiric: :+1: Thanks for looking into this, and the fix.

Indeed, emitting `EventConnectionClose` could race and cancel the browser context too early.

I think we can't call `Connection.Close()` right after sending `Browser.close`, since we need to wait for incoming events like `Target.detachedFromTarget` to arrive and be processed. They're not critical in this case since we're shutting down and ending the iteration anyway, but we should handle those cleanly.

Otherwise, you can see this error with `XK6_BROWSER_LOG=trace`:

```
received IO error: read tcp 127.0.0.1:43488->127.0.0.1:40137: use of closed network connection, connection is closing: true
```

There's also some duplication with what's done in `handleIOError()`. E.g. we might emit `EventConnectionClose` twice.

But let's merge your PR and I'll try to clean it up.

imiric: I opted to just move the `Connection.Close()` call after the browser process exits, which is unintuitive, but hopefully the comment in 5e8139e explains the reason. We could always move what's needed into a separate function and call it instead, but a) I couldn't decide what to name that function, and b) it might add some duplication, or change of logic, since we use `sync.Once` there.

I think this is fine, but let me know if you have a better idea.

inancgumus (approved):
LGTM 👏 

I made a lot of tests, and no goroutines left (except pprof, and it's expected) after a test run ends.

Calling `b.conn.emit(EventConnectionClose, nil)` instead of `b.conn.Close()` [here](https://github.com/grafana/xk6-browser/blob/6de9978dfb2df57b7e8b4af43c6aee533143ee78/common/browser.go#L460) got me the same result (all goroutines are cleared). If we do this, we can remove `Close` from the `Connection` interface. Your choice 🤷😄

imiric:
@inancgumus We can't just emit `EventConnectionClose` instead of calling `conn.Close()`, since `conn.Close()` also closes all `Session`s and the `Connection.done` channel, which stops the `Connection.sendLoop()`. This could arguably be refactored, but we also call `conn.Close()` here on `main`, so I wouldn't change it.

imiric:
I'm merging this with only 1 approval, since most of the team is OoO and so we can test it on Cloud today, but let me know if there are any other issues and I'll fix them next week.

PR #540, opened 2022-09-20
Take context Done into considerations in a bunch of places

mstoykov:
This was leaving hanging goroutines all over the place. 

Not certain whether any of those are relevant for real running, but in tests it was both making stack traces harder to read (as you have a bunch of blocked goroutines), but it also likely was taking some CPU/MEM away from tests. The later likely isn't all that much, but still better to not have goroutines that hang around forever ;)

### DIFF
--- a/common/browser_process.go
+++ b/common/browser_process.go
@@ -80,7 +80,11 @@ func NewBrowserProcess(
 	go func() {
 		// If we lose connection to the browser and we're not in-progress with clean
 		// browser-initiated termination then cancel the context to clean up.
-		<-p.lostConnection
+		select {
+		case <-p.lostConnection:
+		case <-ctx.Done():
+		}
+
 		select {
 		case <-p.processIsGracefullyClosing:
 		default:
--- a/common/browser_test.go
+++ b/common/browser_test.go
@@ -22,8 +22,8 @@ func TestBrowserNewPageInContext(t *testing.T) {
 		bc *BrowserContext
 	}
 	newTestCase := func(id cdp.BrowserContextID) *testCase {
-		ctx := context.Background()
-		b := newBrowser(ctx, nil, nil, NewLaunchOptions(), log.NewNullLogger())
+		ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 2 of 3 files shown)


ankur22 (approved):
LGTM, but i'd wait for an approval from someone else first 🙂

imiric (approved):
LGTM, seems safe to do in general.

How did you spot this @mstoykov? Just from the stack traces?

inancgumus (approved):
Ran some tests, and it looks harmless.

> in tests it was both making stack traces harder to read (as you have a bunch of blocked goroutines

However, I couldn't see much difference between the stack traces of `main` and this fix. Out of curiosity, which goroutines you found were hanging out?

Test:

```go
func TestFramePress(t *testing.T) {
	ctx, cancel := context.WithCancel(context.Background()) // added
	tb := newTestBrowser(t, ctx) // added

	p := tb.NewPage(nil)

	p.SetContent(`<input id="text1">`, nil)

	f := p.Frames()[0]

	f.Press("#text1", "Shift+KeyA", nil)
	cancel() // added
	f.Press("#text1", "KeyB", nil)
	f.Press("#text1", "Shift+KeyC", nil)

	require.Equal(t, "AbC", f.InputValue("#text1", nil))
}
```

mstoykov:
> How did you spot this @mstoykov? Just from the stack traces?

yeah - if you run with `-count 10 -race` for example - it usually turns out that something hangs and then in the stacktraces I would have like a page or 20 of just goroutines that are stuck in the places changed here.

This is also is how you can get ... more or less always some kind of failure of the tests.

> However, I couldn't see much difference between the stack traces of main and this fix. Out of curiosity, which goroutines you found were hanging out?

The ones that are running the particular code that is changed - they just sometimes would not be able to send their message as the channel won't be listened to as the context was already done at this point.

You likely will need to run the test multiple times - preferably hitting some other issues in the meantime that dumps all the goroutines. And then what you get is a  lot of those goroutines that are just hanging around, but AFAIK don't actually block anything else. So given that they likely just need to stop as the context was done - I just added the `select`s and now the stack traces are a lot smaller.

PR #541, opened 2022-09-20
Fix/232 ws 1006 abnormal closure dont close durring browser close

mstoykov:


common/browser.go
```diff
@@ -441,6 +441,7 @@ func (b *Browser) Close() {
 		}
 	}
 
+	b.conn.Close()
```
imiric, line 444: Like I mentioned in [this comment](https://github.com/grafana/xk6-browser/pull/535#discussion_r976514043), we should wait for all incoming events to arrive before closing the connection. Since we really have no way of knowing how many and which ones we should expect, the safest way was to wait for the process to exit, and silence the `unexpected EOF` error.

PR #547, opened 2022-09-27
Fix `ignoreDefaultArgs`

### LINKED ISSUE #536: `evaluate.js` example fails in headless mode
On current `main` (96b89a1), Chromium 104.0.5112.79 and Linux, running the [`evaluate.js` example](https://github.com/grafana/xk6-browser/blob/96b89a1d0ad8d2f73d79550545bc61d786a90715/examples/evaluate.js) fails the `width` check, but only when run in headless mode.

Headless:
```
$ XK6_BROWSER_LOG=info XK6_HEADLESS=1 xk6 run -q examples/evaluate.js
INFO[0001] {"deviceScaleFactor":1,"height":720,"width":1280}  source=browser-console-api

     ✗ width
      ↳  0% — ✓ 0 / ✗ 1
     ✓ height
     ✓ scale
```

Headed:
```
$ XK6_BROWSER_LOG=info xk6 run -q examples/evaluate.js
INFO[0001] {"deviceScaleFactor":1,"height":720,"width":1265}  source=browser-console-api

     ✓ width
     ✓ height
     ✓ scale
```

Notice that the returned width is 1280 in headless, but 1265 in headed.

This might be a recent regression, or a change in Chromium(?), since it worked a few months ago.

You can see the failure in CI as well: [job #1147](https://github.com/grafana/xk6-browser/actions/runs/3069371391/jobs/4957938788). The reason it still passes is because failing checks still exit the process with 0. So we need a better way of ensuring we catch these issues in CI...

mstoykov:


chromium/browser_type.go
```diff
@@ -312,13 +312,21 @@ func prepareFlags(lopts *common.LaunchOptions, k6opts *k6lib.Options) map[string
...
 	}
+	ignoreDefaultFlags(f, lopts.IgnoreDefaultArgs)
 
 	setFlagsFromArgs(f, lopts.Args)
 	setFlagsFromK6Options(f, k6opts)
 
 	return f
 }
 
+// ignoreDefaultFlags ignores any flags in teh provide slice
```
ankur22, line 323: nit

```suggestion
// ignoreDefaultFlags ignores any flags in the provided slice
```

chromium/browser_type.go
```diff
@@ -312,13 +312,21 @@ func prepareFlags(lopts *common.LaunchOptions, k6opts *k6lib.Options) map[string
...
+func ignoreDefaultFlags(flags map[string]interface{}, toIgnore []string) {
```
ankur22, line 324: Minor, WDYT of the following?

```suggestion
func ignoreDefaultArgsFlags(flags map[string]interface{}, toIgnore []string) {
```

ankur22 (approved):

chromium/browser_type.go
```diff
@@ -312,13 +312,21 @@ func prepareFlags(lopts *common.LaunchOptions, k6opts *k6lib.Options) map[string
...
+	for _, name := range toIgnore {
+		delete(flags, strings.TrimLeft(name, "--"))
```
inancgumus, line 326: `TrimLeft` is for removing leading characters (cutset), not for removing a prefix string. So, this would work too:

```
delete(flags, strings.TrimLeft(name, "-"))
```

However, I believe it's better to use `strings.TrimPrefix` for removing a string prefix. From the `strings.TrimLeft` documentation:

```go
// ...
// To remove a prefix, use TrimPrefix instead.
func TrimLeft(s, cutset string) string
```

```suggestion
		delete(flags, strings.TrimPrefix(name, "--"))
```

inancgumus (changes requested):
Nice catch, @mstoykov 👏 

I have some suggestions below.

I believe we might want to add a test for this to [BrowserType tests](https://github.com/grafana/xk6-browser/blob/3710294c5186547698af3d1b1ead5bf5ec0ece9b/chromium/browser_type_test.go). We can check whether it removes the given options to be ignored.

inancgumus (approved):
LGTM 👍

mstoykov:
I am going to try to expand the test in the next 30 minutes to handle this and if not will skip them for now :(.

I will also move the evaluate.js changes in a different PR + some CI changes to catch those - but there I have other problems as well :rofl:

common/launch.go
```diff
@@ -102,11 +102,12 @@ func (l *LaunchOptions) Parse(ctx context.Context, opts goja.Value) error {
...
 			case "ignoreDefaultArgs":
 				v := opts.Get(k)
-				switch v.ExportType() {
-				case reflect.TypeOf(goja.Object{}):
-					args := v.Export().([]string)
-					l.IgnoreDefaultArgs = append(l.IgnoreDefaultArgs, args...)
+				var args []string
+				err := rt.ExportTo(v, &args)
+				if err != nil {
+					return fmt.Errorf("ignoreDefaultArgs should be an array of strings: %w", err)
```
imiric, line 108: Do we want to support passing `true` as well, [to be consistent with Playwright](https://playwright.dev/docs/api/class-browsertype#browser-type-launch)?

The use case for that is dubious, and we don't need it to fix #536, but we might as well while we're here.

imiric (approved):
Hey, thanks for this! :+1: Just left a minor suggestion.

We should probably document what all our default browser args are, so that people don't need to go source hunting. We can leave that for later.

inancgumus (approved):

common/launch.go
mstoykov: I will open issue for that as I don't want to get this in this PR anymore and I feel like parsing it as a boolean as well will involve some code I don't particularly want to write :grimacing:

mstoykov: https://github.com/grafana/xk6-browser/issues/549

Issue #553, opened 2022-09-29
The ordering of emitted events is sometimes lost which results in handlers consuming them out of order


ankur22:
The ordering of events that are [emitted](https://github.com/grafana/xk6-browser/blob/ae264530b6fbe0285ac9df9b0f4a907b460bb805/common/event_emitter.go#L167) is sometimes lost due to the use of the goroutine during the [emit to the handlers](https://github.com/grafana/xk6-browser/blob/ae264530b6fbe0285ac9df9b0f4a907b460bb805/common/event_emitter.go#L183).

This was found to eventually cause a [deadlock](https://github.com/grafana/xk6-browser/issues/496) in cloud runs.

To overcome this issue we tried to [remove the goroutine](https://github.com/grafana/xk6-browser/pull/524), but the use of buffered channels didn't help in reducing the possibility of other deadlocks.

We still need to find a robust and maintainable solution whereby the order of the emitted events is somehow kept when the event is emitted to the handlers.

PR #555, opened 2022-09-29
Add a queue per handler to keep events in order

ankur22:
Events are sometimes out of order due to use using a goroutine to publish events to the handler concurrently as there's no guarantee when the goroutines will run (the first goroutine to be created isn't necessarily going to be the first to run). To overcome this, a `queue` has been added per handler to synchronise all publishes of events to that handler as well as a `mutex` so that only one `emitEvent` goroutine can read from the `queue` **and** write to the handler's channel at any one time.

We tried working with a buffered channel instead of a `queue` (which is a slice). Unfortunately it's difficult to determine how large the channel needs to be to work with the test -- the more complex the test, the more events that will be emitted. For now we think it is easier to work with a `queue` that can grow when needed.

After creating the unit tests, we found that although we were receiving events in order we had created a deadlock when a handler was receiving events **and** emitting new events. To overcome this we had to create a "double `queue`" system whereby new emitted events can be stored on one `queue`, while the `emitEvent` goroutines can read off of a separate `queue`. When the read `queue` is depleted, the queues are swapped around.

Closes: https://github.com/grafana/xk6-browser/issues/553

### DIFF
--- a/common/event_emitter.go
+++ b/common/event_emitter.go
@@ -117,8 +117,8 @@ type syncFunc func() (done chan struct{})
 
 // BaseEventEmitter emits events to registered handlers.
 type BaseEventEmitter struct {
-	handlers    map[string][]eventHandler
-	handlersAll []eventHandler
+	handlers    map[string][]*eventHandler
+	handlersAll []*eventHandler
 
 	syncCh chan syncFunc
 	ctx    context.Context
@@ -127,7 +127,7 @@ type BaseEventEmitter struct {
 // NewBaseEventEmitter creates a new instance of a base event emitter.
 func NewBaseEventEmitter(ctx context.Context) BaseEventEmitter {
 	bem := BaseEventEmitter{
-		handlers: make(map[string][]eventHandler),
+		handlers: make(map[string][]*eventHandler),
 		syncCh:   make(chan syncFunc),
 		ctx:      ctx,
 	}
@@ -168,7 +168,7 @@ func (e *BaseEventEmitter) sync(fn func()) {
 }
 
 func (e *BaseEventEmitter) emit(event string, data interface{}) {
-	emitEvent := func(eh eventHandler) {
+	emitEvent := func(eh *eventHandler) {
 		eh.queueMutex.Lock()
 		defer eh.queueMutex.Unlock()
 
@@ -179,7 +179,7 @@ func (e *BaseEventEmitter) emit(event string, data interface{}) {
 			// TODO: handle the error
 		}
 	}
-	emitTo := func(handlers []eventHandler) (updated []eventHandler) {
+	emitTo := func(handlers []*eventHandler) (updated []*eventHandler) {
 		for i := 0; i < len(handlers); {
 			handler := handlers[i]
 			select {
@@ -216,10 +216,10 @@ func (e *BaseEventEmitter) on(ctx context.Context, events []string, ch chan Even
 		for _, event := range events {
 			_, ok := e.handlers[event]
 			if !ok {
-				e.handlers[event] = make([]eventHandler, 0)
+				e.handlers[event] = make([]*eventHandler, 0)
 			}
 			eh := eventHandler{ctx, ch, &sync.Mutex{}, make([]Event, 0)}
-			e.handlers[event] = append(e.handlers[event], eh)
+			e.handlers[event] = append(e.handlers[event], &eh)
 		}
 	})
 }
@@ -227,6 +227,6 @@ func (e *BaseEventEmitter) on(ctx context.Context, events []string, ch chan Even
 // OnAll registers a handler for all events.
 func (e *BaseEventEmitter) onAll(ctx context.Context, ch chan Event) {
 	e.sync(func() {
-		e.handlersAll = append(e.handlersAll, eventHandler{ctx, ch, &sync.Mutex{}, make([]Event, 0)})
+		e.handlersAll = append(e.handlersAll, &eventHandler{ctx, ch, &sync.Mutex{}, make([]Event, 0)})
 	})
 }
--- a/common/event_emitter.go
+++ b/common/event_emitter.go
@@ -181,13 +181,14 @@ func (e *BaseEventEmitter) emit(event string, data interface{}) {
 		eh.queue.readMutex.Lock()
 		defer eh.queue.readMutex.Unlock()
 
-		// We try to read from the current queue (curQueue)
-		// If there isn't anything on curQueue then there must
+		// We try to read from the read queue (queue.read).
+		// If there isn't anything on the read queue, then there must
 		// be something being populated by the synched emitTo
-		// func below. Swap around curQueue with queue. Queue
-		// is now being populated again by emitTo, and all
+		// func below.
+		// Swap around the read queue with the write queue.
+		// Queue is now being populated again by emitTo, and all
 		// emitEvent goroutines can continue to consume from
-		// curQueue until that is again depleted.
+		// the read queue until that is again depleted.
 		if len(eh.queue.read) == 0 {
 			eh.queue.writeMutex.Lock()
 			eh.queue.read, eh.queue.write = eh.queue.write, eh.queue.read
--- a/common/event_emitter.go
+++ b/common/event_emitter.go
@@ -98,10 +98,12 @@ type NavigationEvent struct {
 }
 
 type eventHandler struct {
-	ctx        context.Context
-	ch         chan Event
-	queueMutex *sync.Mutex
-	queue      []Event
+	ctx           context.Context
+	ch            chan Event
+	queueMutex    *sync.Mutex
+	queue         []Event
+	curQueueMutex *sync.Mutex
+	curQueue      []Event
 }
 
 // EventEmitter that all event emitters need to implement.
@@ -169,12 +171,25 @@ func (e *BaseEventEmitter) sync(fn func()) {
 
 func (e *BaseEventEmitter) emit(event string, data interface{}) {
 	emitEvent := func(eh *eventHandler) {
-		eh.queueMutex.Lock()
-		defer eh.queueMutex.Unlock()
+		eh.curQueueMutex.Lock()
+		defer eh.curQueueMutex.Unlock()
+
+		// We try to read from the current queue (curQueue)
+		// If there isn't anything on curQueue then there must
+		// be something being populated by the synched emitTo
+		// func below. Swap around curQueue with queue. Queue
+		// is now being populated again by emitTo, and all
+		// emitEvent goroutines can continue to consume from
+		// curQueue until that is again depleted.
+		if len(eh.curQueue) == 0 {
+			eh.queueMutex.Lock()
+			eh.curQueue, eh.queue = eh.queue, eh.curQueue
+			eh.queueMutex.Unlock()
+		}
 
 		select {
-		case eh.ch <- eh.queue[0]:
-			eh.queue = eh.queue[1:]
+		case eh.ch <- eh.curQueue[0]:
+			eh.curQueue = eh.curQueue[1:]
 		case <-eh.ctx.Done():
 			// TODO: handle the error
 		}
@@ -187,13 +202,6 @@ func (e *BaseEventEmitter) emit(event string, data interface{}) {
 				handlers = append(handlers[:i], handlers[i+1:]...)
 				continue
 			default:
-				// To ensure that goroutines don't break the ordering
-				// of the emitted events, we will need the goroutine to synchronize.
-				// This means that the events need to be stored in a queue per handler. The
-				// goroutine responsible for the the first popped element must complete
-				// publishing the event before the next goroutine can pop and work with
-				// the next event for that one handler. Each handler can process events concurrently still.
-
 				handler.queueMutex.Lock()
 				handler.queue = append(handler.queue, Event{typ: event, data: data})
 				handler.queueMutex.Unlock()
@@ -218,7 +226,14 @@ func (e *BaseEventEmitter) on(ctx context.Context, events []string, ch chan Even
 			if !ok {
 				e.handlers[event] = make([]*eventHandler, 0)
 			}
-			eh := eventHandler{ctx, ch, &sync.Mutex{}, make([]Event, 0)}
+			eh := eventHandler{
+				ctx,
+				ch,
+				&sync.Mutex{},
+				make([]Event, 0),
+				&sync.Mutex{},
+				make([]Event, 0),
+			}
 			e.handlers[event] = append(e.handlers[event], &eh)
 		}
 	})
@@ -227,6 +242,14 @@ func (e *BaseEventEmitter) on(ctx context.Context, events []string, ch chan Even
 // OnAll registers a handler for all events.
 func (e *BaseEventEmitter) onAll(ctx context.Context, ch chan Event) {
 	e.sync(func() {
-		e.handlersAll = append(e.handlersAll, &eventHandler{ctx, ch, &sync.Mutex{}, make([]Event, 0)})
+		e.handlersAll = append(e.handlersAll,
+			&eventHandler{
+				ctx,
+				ch,
+				&sync.Mutex{},
+				make([]Event, 0),
+				&sync.Mutex{},
+				make([]Event, 0),
+			})
 	})
 }
--- a/common/event_emitter.go
+++ b/common/event_emitter.go
@@ -97,13 +97,17 @@ type NavigationEvent struct {
 	err         error
 }
 
+type queue struct {
+	writeMutex sync.Mutex
+	write      []Event
+	readMutex  sync.Mutex
+	read       []Event
+}
+
 type eventHandler struct {
-	ctx           context.Context
-	ch            chan Event
-	queueMutex    sync.Mutex
-	queue         []Event
-	curQueueMutex sync.Mutex
-	curQueue      []Event
+	ctx   context.Context
+	ch    chan Event
+	queue *queue
 }
 
 // EventEmitter that all event emitters need to implement.
@@ -122,6 +126,8 @@ type BaseEventEmitter struct {
 	handlers    map[string][]*eventHandler
 	handlersAll []*eventHandler
 
+	queues map[chan Event]*queue
+
 	syncCh chan syncFunc
 	ctx    context.Context
 }
@@ -132,6 +138,7 @@ func NewBaseEventEmitter(ctx context.Context) BaseEventEmitter {
 		handlers: make(map[string][]*eventHandler),
 		syncCh:   make(chan syncFunc),
 		ctx:      ctx,
+		queues:   make(map[chan Event]*queue),
 	}
 	go bem.syncAll(ctx)
 	return bem
@@ -171,8 +178,8 @@ func (e *BaseEventEmitter) sync(fn func()) {
 
 func (e *BaseEventEmitter) emit(event string, data interface{}) {
 	emitEvent := func(eh *eventHandler) {
-		eh.curQueueMutex.Lock()
-		defer eh.curQueueMutex.Unlock()
+		eh.queue.readMutex.Lock()
+		defer eh.queue.readMutex.Unlock()
 
 		// We try to read from the current queue (curQueue)
 		// If there isn't anything on curQueue then there must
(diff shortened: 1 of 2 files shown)


common/event_emitter.go
```diff
@@ -180,6 +187,17 @@ func (e *BaseEventEmitter) emit(event string, data interface{}) {
...
+				// To ensure that goroutines don't break the ordering
+				// of the emitted events, we will need the goroutine to synchronize.
+				// This means that the events need to be stored in a queue per handler. The
+				// goroutine responsible for the the first popped element must complete
+				// publishing the event before the next goroutine can pop and work with
+				// the next event for that one handler. Each handler can process events concurrently still.
+
+				handler.queueMutex.Lock()
+				handler.queue = append(handler.queue, Event{typ: event, data: data})
+				handler.queueMutex.Unlock()
```
imiric, line 199: A channel would accomplish the same thing, no? It's a FIFO queue after all, and we wouldn't need locking.

ankur22: A channel can't grow as the complexity of the tests increase and thus the number of events increase. So i think the queue is a better approach for this case.

ankur22: I tried that too! 😆 I wish I had saved it as a PR 😞 I'll recreate it and create a PR for you to look at.

imiric (commented):
Hey, nicely done! From my local tests your change seems to work :+1:

Do you think it would be possible to have tests for this? The previous behavior was flaky, so not sure how easily reproducible it is. Worth a try?

inancgumus (commented):
Nice idea! 👏  This ensures that the handlers receive emitted events in order.

One question: How can we be sure that this PR fixes the issue in #482?

By the way, I ran this PR's tests on CI a few times, and there were errors ([1](https://github.com/grafana/xk6-browser/actions/runs/3153039507/jobs/5167295773), [2](https://github.com/grafana/xk6-browser/actions/runs/3153039507/jobs/5167131570), [3](https://github.com/grafana/xk6-browser/actions/runs/3153039507/jobs/5167015995)). Also, locally, I received the following error:

```bash
time="2022-10-03T10:52:42+03:00" level=error msg="process with PID 72688 unexpectedly ended: signal: killed" category=browser elapsed="0 ms" goroutine=99722
--- FAIL: TestLocator (39.38s)
    --- FAIL: TestLocator/Tap (1.63s)
        locator_test.go:188: 
                Error Trace:    /Users/inanc/grafana/xk6-browser/tests/locator_test.go:188
                Error:          Expected value not to be nil.
                Test:           TestLocator/Tap
    --- FAIL: TestLocator/Click (30.62s)
    cleanup panicked with closing the browser: sending a message to browser: websocket: close 1006 (abnormal closure): unexpected EOFpanic: clicking on "#link": timed out after 30s [recovered]
        panic: clicking on "#link": timed out after 30s
```

WDYT about these errors? Do you believe they're related to this PR?

common/event_emitter.go
inancgumus: Maybe something like this?

<details>
  <summary>Channel approach</summary>
  
```diff
diff --git a/common/event_emitter.go b/common/event_emitter.go
index 9c9edaa..43f0b08 100644
--- a/common/event_emitter.go
+++ b/common/event_emitter.go
@@ -22,7 +22,6 @@ package common
 
 import (
 	"context"
-	"sync"
 )
 
 // Ensure BaseEventEmitter implements the EventEmitter interface.
@@ -98,10 +97,9 @@ type NavigationEvent struct {
 }
 
 type eventHandler struct {
-	ctx        context.Context
-	ch         chan Event
-	queueMutex *sync.Mutex
-	queue      []Event
+	ctx   context.Context
+	ch    chan Event
+	queue chan Event
 }
 
 // EventEmitter that all event emitters need to implement.
@@ -169,12 +167,14 @@ func (e *BaseEventEmitter) sync(fn func()) {
 
 func (e *BaseEventEmitter) emit(event string, data interface{}) {
 	emitEvent := func(eh eventHandler) {
-		eh.queueMutex.Lock()
-		defer eh.queueMutex.Unlock()
-
+		var e Event
+		select {
+		case e = <-eh.queue:
+		case <-eh.ctx.Done():
+			return
+		}
 		select {
-		case eh.ch <- eh.queue[0]:
-			eh.queue = eh.queue[1:]
+		case eh.ch <- e:
 		case <-eh.ctx.Done():
 			// TODO: handle the error
 		}
@@ -193,12 +193,14 @@ func (e *BaseEventEmitter) emit(event string, data interface{}) {
 				// goroutine responsible for the the first popped element must complete
 				// publishing the event before the next goroutine can pop and work with
 				// the next event for that one handler. Each handler can process events concurrently still.
-
-				handler.queueMutex.Lock()
-				handler.queue = append(handler.queue, Event{typ: event, data: data})
-				handler.queueMutex.Unlock()
-
 				go emitEvent(handler)
+				select {
+				case handler.queue <- Event{typ: event, data: data}:
+				case <-handler.ctx.Done():
+					handlers = append(handlers[:i], handlers[i+1:]...)
+					continue
+				}
+
 				i++
 			}
 		}
@@ -218,7 +220,7 @@ func (e *BaseEventEmitter) on(ctx context.Context, events []string, ch chan Even
 			if !ok {
 				e.handlers[event] = make([]eventHandler, 0)
 			}
-			eh := eventHandler{ctx, ch, &sync.Mutex{}, make([]Event, 0)}
+			eh := eventHandler{ctx, ch, make(chan Event)}
 			e.handlers[event] = append(e.handlers[event], eh)
 		}
 	})
@@ -227,6 +229,6 @@ func (e *BaseEventEmitter) on(ctx context.Context, events []string, ch chan Even
 // OnAll registers a handler for all events.
 func (e *BaseEventEmitter) onAll(ctx context.Context, ch chan Event) {
 	e.sync(func() {
-		e.handlersAll = append(e.handlersAll, eventHandler{ctx, ch, &sync.Mutex{}, make([]Event, 0)})
+		e.handlersAll = append(e.handlersAll, eventHandler{ctx, ch, make(chan Event)})
 	})
 }
```

</details>

ankur22: @inancgumus thanks for the suggestion. Isn't there a chance that the order of the events received by the handler could still be received in the wrong order when using this channel approach?

Situation:
- Handler is emitting an event.
- Emitter is trying to send events to the Handler.
- Since the Handler is busy trying to emit an event, there's a chance that multiple `emitEvent` goroutines could be waiting to send an event to the Handler.
- Which `emitEvent` writes to the channel first? Second? Third? etc.

@imiric I tried to implement your idea (if i understood it correctly) but I was unable to get very far as it's deadlocking. Here's the single commit: https://github.com/grafana/xk6-browser/commit/328dc888d6e7c6ba12487e6fdce289ff6ad6843b

inancgumus: @ankur22 If so, shouldn’t there be the same problem with the mutex-queue solution? I mean, the channel solution should work like the mutex-queue solution. What am I missing? :)

ankur22: With queue and mutex: https://go.dev/play/p/rUvviX4qvAw
With chan and no mutex: https://go.dev/play/p/_DN7XRMzeQ9

The key here is the mutex lock, not whether it's a `queue` or a `channel`.

IMO The `queue` is still a better solution since there's a chance that the handler could emit an event while the `emitEvent` could be trying to send an event to the same handler. With an unbuffered queue this would deadlock. With a queue that can grow with the complexity of the test code while avoiding deadlocks.

common/event_emitter.go
```diff
@@ -180,6 +187,17 @@ func (e *BaseEventEmitter) emit(event string, data interface{}) {
...
+
 				go emitEvent(handler)
```
ankur22, line 201: Since we have one queue that works on a single event at a time per handler, there's no need to spawn a new goroutine per event. Maybe a goroutine per handler? I'll explore this too.

common/event_emitter.go
```diff
@@ -166,8 +169,12 @@ func (e *BaseEventEmitter) sync(fn func()) {
 
 func (e *BaseEventEmitter) emit(event string, data interface{}) {
 	emitEvent := func(eh eventHandler) {
+		eh.queueMutex.Lock()
+		defer eh.queueMutex.Unlock()
+
 		select {
-		case eh.ch <- Event{event, data}:
+		case eh.ch <- eh.queue[0]:
+			eh.queue = eh.queue[1:]
```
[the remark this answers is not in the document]
ankur22: - Yeah, that should be the case and all events are popped off -- I could try to implement the channel API and ensure it works in the same way though... might be overkill though.
- I need to see this with a fresh pair of eyes, I don't see how that is possible atm.

ankur22:
> Do you think it would be possible to have tests for this? The previous behavior was flaky, so not sure how easily reproducible it is. Worth a try?

Good idea, I'll try.

common/event_emitter.go
inancgumus: Ah, right, that's because of the lock, see: https://go.dev/play/p/H7m6Rr69tS1

common/event_emitter.go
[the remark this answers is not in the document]
ankur22: I took another look at the second point you raised, and I still can't see how we may end up with `index out of range` panic in the `case <-handler.ctx.Done():`. Please let me know if I have missed something though.

common/event_emitter.go
ankur22: Actually, I think it's easier to leave it so that a new goroutine is spawned as it signifies that an event has been received and we need to now send it to the handler. otherwise we might need to implement out own custom channel the encapsulates this behaviour.

ankur22:
> > Do you think it would be possible to have tests for this? The previous behavior was flaky, so not sure how easily reproducible it is. Worth a try?
> 
> Good idea, I'll try.

Unit tests added. I verified that out of the two tests, the test that ensures that events are received in order fails in `main`.

common/event_emitter.go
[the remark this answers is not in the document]
imiric: Right. It wouldn't happen since `emitEvent()` would never be called. Thanks for confirming.

common/event_emitter.go
```diff
@@ -97,8 +98,12 @@ type NavigationEvent struct {
...
 
 type eventHandler struct {
-	ctx context.Context
-	ch  chan Event
+	ctx           context.Context
+	ch            chan Event
+	queueMutex    *sync.Mutex
+	queue         []Event
+	curQueueMutex *sync.Mutex
+	curQueue      []Event
```
inancgumus, line 106: Since `eventHandler` is used as a pointer, we can now drop the sync mutex pointers. I made changes to the relevant parts in the code to be able to use this.

```suggestion
	ctx           context.Context
	ch            chan Event
	queueMutex    sync.Mutex
	queue         []Event
	curQueueMutex sync.Mutex
	curQueue      []Event
```

common/event_emitter.go
```diff
@@ -198,17 +224,32 @@ func (e *BaseEventEmitter) on(ctx context.Context, events []string, ch chan Even
...
+				e.handlers[event] = make([]*eventHandler, 0)
+			}
+			eh := eventHandler{
+				ctx,
+				ch,
+				&sync.Mutex{},
+				make([]Event, 0),
+				&sync.Mutex{},
+				make([]Event, 0),
 			}
```
inancgumus, line 236:
```suggestion
			eh := eventHandler{ctx: ctx, ch: ch}
```

common/event_emitter.go
```diff
@@ -198,17 +224,32 @@ func (e *BaseEventEmitter) on(ctx context.Context, events []string, ch chan Even
...
-		e.handlersAll = append(e.handlersAll, eventHandler{ctx, ch})
+		e.handlersAll = append(e.handlersAll,
+			&eventHandler{
+				ctx,
+				ch,
+				&sync.Mutex{},
+				make([]Event, 0),
+				&sync.Mutex{},
+				make([]Event, 0),
+			})
```
inancgumus, line 253:
```suggestion
		e.handlersAll = append(e.handlersAll, &eventHandler{ctx: ctx, ch: ch})
```

common/event_emitter_test.go
```diff
@@ -125,3 +127,135 @@ func TestEventEmitterAllEvents(t *testing.T) {
...
+			defer wg.Done()
+
+			for i := 0; i < maxInt; i++ {
+				i := i
+				emitter.emit(eventName1, i)
+			}
+		}
+		go emitWorker()
+
+		wg.Wait()
```
imiric, line 259: Nit: this test doesn't assert much, but if we're not expecting a deadlock then it should fail after a specific timeout. Otherwise we run the risk of timing out whenever the global test timeout is reached.

We might want to do the same in the other test, just in case.

imiric (approved):
Excellent work, Ankur! :clap:

It took me a while to get the double queue approach, but it makes sense.

The tests are a great touch, and I also confirmed the order one fails on `main`. :thumbsup:

I ran this with several of our examples, which worked great, and with the Vistaprint script, which needs to be rewritten to use the new async APIs, but at least it didn't regress compared to `main`. :sweat_smile:

Let's create a Cloud release when this is merged, so that it can be battle tested. :shipit:

inancgumus:
@ankur Awesome work, thanks for this! 👏 

I ran the CI tests multiple times and saw the following error in [this CI job](https://github.com/grafana/xk6-browser/actions/runs/3195855218/jobs/5218022064):

```log
=== CONT  TestWaitForFrameNavigationWithinDocument/anchor
    frame_manager_test.go:46: 
        	Error Trace:	/home/runner/work/xk6-browser/xk6-browser/tests/frame_manager_test.go:46
        	            				/home/runner/work/xk6-browser/xk6-browser/tests/asm_amd64.s:1571
        	Error:      	Expected value not to be nil.
        	Test:       	TestWaitForFrameNavigationWithinDocument/anchor
```

I believe this one can be related to #560 or the previous related issues. I think it is not related to this issue. Then again, I wanted to let you know.

I also saw the following error when I ran the tests locally:

```log
--- FAIL: TestLocator (1.00s)
    locator_test.go:271: 
                Error Trace:    /Users/inanc/grafana/xk6-browser/tests/locator_test.go:271
                Error:          Expected value not to be nil.
                Test:           TestLocator
```

This one is related to the previous error I explained above. It's because the origin of the errors seems to be `page.Goto`.

common/event_emitter.go
ankur22: I had to add this in manually

inancgumus (approved):
Thanks for making all of our lives easier 🎉

ankur22:
> @Ankur Awesome work, thanks for this! 👏
> 
> I ran the CI tests multiple times, they worked fabulously, thanks to you.
> 
> On the 3rd run, I saw the following error in [this CI job](https://github.com/grafana/xk6-browser/actions/runs/3195855218/jobs/5218022064):
> 
> ```
> === CONT  TestWaitForFrameNavigationWithinDocument/anchor
>     frame_manager_test.go:46: 
>         	Error Trace:	/home/runner/work/xk6-browser/xk6-browser/tests/frame_manager_test.go:46
>         	            				/home/runner/work/xk6-browser/xk6-browser/tests/asm_amd64.s:1571
>         	Error:      	Expected value not to be nil.
>         	Test:       	TestWaitForFrameNavigationWithinDocument/anchor
> ```
> 
> I believe this one might be related to #560 or the previous related issues. I think it is not related to #553. Then again, I wanted to let you know (in case you may have an idea).
> 
> I also saw the following error when I ran the tests locally:
> 
> ```
> --- FAIL: TestLocator (1.00s)
>     locator_test.go:271: 
>                 Error Trace:    /Users/inanc/grafana/xk6-browser/tests/locator_test.go:271
>                 Error:          Expected value not to be nil.
>                 Test:           TestLocator
> ```
> 
> This one is related to the previous error I explained above. The origin of both errors seems to be `page.Goto`. Although, #560 is about a timeout error, after this change, the underlying problem with `page.Goto` might have reached the surface.

Thanks @inancgumus!

I've taken a look at those errors before actually, and unfortunately they are race conditions unrelated to unordered CDP events. I did start to look into them a couple of months ago before the deadlock cloud issue was prioritised over other issues. I'd be happy to look into these next though as they're a real PITA when stress testing locally, and obviously a pain in CI -- i had to comment out a lot of tests since they kept failing in response being nil when they shouldn't have been 😭

ankur22:
> 🤔 [1](https://github.com/grafana/xk6-browser/actions/runs/3153039507/jobs/5167295773) and [3](https://github.com/grafana/xk6-browser/actions/runs/3153039507/jobs/5167015995) are worrying, and I'll need to investigate those.

These errors occurred prior to adding in the "double `queue`" system for managing the read and write of events from different goroutines. When using a single `queue`, it would deadlock, which is what those test runs showed us. Those same deadlocks shouldn't happen now with the "double `queue`" system (i've done stress tests locally and not had a deadlock so far).

common/event_emitter_test.go
```diff
@@ -125,3 +127,125 @@ func TestEventEmitterAllEvents(t *testing.T) {
...
+		// Success criteria: No deadlock should occur between receiving,
+		//                   emitting, and receiving of events.
+
+		t.Parallel()
+
+		eventName1 := "AtomicIntEvent1"
+		eventName2 := "AtomicIntEvent2"
+		maxInt := 100
+
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*2)
```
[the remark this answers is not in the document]
ankur22: The timeout was based on your earlier suggestion:

> Nit: this test doesn't assert much, but if we're not expecting a deadlock then it should fail after a specific timeout. Otherwise we run the risk of timing out whenever the global test timeout is reached.

I prefer to remove it and revert back to `WithCancel()`. If there's a deadlock it should show up when the global test timeout fires.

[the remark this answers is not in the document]
ankur22: Done 👍

common/event_emitter_test.go
```diff
@@ -125,3 +127,187 @@ func TestEventEmitterAllEvents(t *testing.T) {
...
+				expectedI++
+			}
+
+			close(ch)
+		}
+		go handler()
+
+		emitWorker := func() {
+			for i := 0; i < maxInt; i++ {
+				i := i
```
inancgumus, line 178: There is no need for this.

common/event_emitter_test.go
```diff
@@ -125,3 +127,187 @@ func TestEventEmitterAllEvents(t *testing.T) {
...
+				expectedI++
+			}
+
+			close(ch)
+		}
+		go handler()
+
+		emitWorker := func() {
+			for i := 0; i < maxInt; i += 4 {
+				i := i
```
inancgumus, line 236:
```suggestion
```

common/event_emitter_test.go
```diff
@@ -125,3 +127,187 @@ func TestEventEmitterAllEvents(t *testing.T) {
...
+				}
+			}
+
+			close(ch)
+		}
+		go handler()
+
+		emitWorker := func() {
+			for i := 0; i < maxInt; i++ {
+				i := i
```
inancgumus, line 301:
```suggestion
```

common/event_emitter.go
```diff
@@ -165,21 +177,42 @@ func (e *BaseEventEmitter) sync(fn func()) {
...
+		eh.queue.readMutex.Lock()
+		defer eh.queue.readMutex.Unlock()
+
+		// We try to read from the current queue (curQueue)
+		// If there isn't anything on curQueue then there must
+		// be something being populated by the synched emitTo
+		// func below. Swap around curQueue with queue. Queue
+		// is now being populated again by emitTo, and all
+		// emitEvent goroutines can continue to consume from
+		// curQueue until that is again depleted.
```
inancgumus, line 190:
```suggestion
		// We try to read from the read queue (queue.read).
		// If there isn't anything on the read queue, then there must
		// be something being populated by the synched emitTo
		// func below.
		// Swap around the read queue with the write queue.
		// Queue is now being populated again by emitTo, and all
		// emitEvent goroutines can continue to consume from
		// the read queue until that is again depleted.
```

common/event_emitter.go
```diff
@@ -97,13 +97,17 @@ type NavigationEvent struct {
...
 type eventHandler struct {
-	ctx           context.Context
-	ch            chan Event
-	queueMutex    sync.Mutex
-	queue         []Event
-	curQueueMutex sync.Mutex
-	curQueue      []Event
+	ctx   context.Context
+	ch    chan Event
+	queue *queue
```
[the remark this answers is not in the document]
ankur22: It is safer as a `pointer`. There is a `map` of `chan Event` -> `queue`, and this can be accessed at anytime during the runtime when an element calls `on` in `BaseEventEmitter`; e.g. `frame_session` calls `on` twice. The `queue` in the `map` and the `queue` in the `handler` need to be the same otherwise we could get out of order events and eventually deadlocks.

common/event_emitter.go
```diff
@@ -221,15 +228,27 @@ func (e *BaseEventEmitter) emit(event string, data interface{}) {
 // On registers a handler for a specific event.
 func (e *BaseEventEmitter) on(ctx context.Context, events []string, ch chan Event) {
 	e.sync(func() {
+		q, ok := e.queues[ch]
+		if !ok {
+			q = &queue{}
+			e.queues[ch] = q
+		}
```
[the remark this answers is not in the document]
ankur22: I don't think it needs to be moved -- it feels like an unnecessary abstraction.

common/event_emitter_test.go
ankur22: Yeah, you're correct. No need for var shadowing here.

common/event_emitter.go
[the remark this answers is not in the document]
inancgumus: I suggested it for encapsulating the behavior, but it’s OK 👍😊

PR #556, opened 2022-09-30
Docker example

olegbespalov:
# What?

Add the Docerfile, docker-compose to provide the possibility of running examples inside the docker.

# Why?

Docker is the easiest way to improve the developer's UX.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM golang:1.19-bullseye as builder
+
+RUN go install -trimpath go.k6.io/xk6/cmd/xk6@latest
+
+RUN  xk6 build --output "/tmp/k6" --with github.com/grafana/xk6-browser
+
+FROM debian:bullseye
+
+ARG CHROMIUM_VERSION=106.0.5249.61-1~deb11u1
+
+RUN apt-get update
+RUN apt-get install -y chromium=${CHROMIUM_VERSION}
+
+COPY --from=builder /tmp/k6 /usr/bin/k6
+
+ENV XK6_HEADLESS=true
+
+ENTRYPOINT ["k6"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,8 @@ FROM debian:bullseye
 
 ARG CHROMIUM_VERSION=106.0.5249.61-1~deb11u1
 
-RUN apt-get update
-RUN apt-get install -y chromium=${CHROMIUM_VERSION}
+RUN apt-get update && \
+    apt-get install -y chromium=${CHROMIUM_VERSION}
 
 COPY --from=builder /tmp/k6 /usr/bin/k6
 
--- a/README.md
+++ b/README.md
@@ -391,6 +391,14 @@ export default function () {
 }
 ```
 
+### Running examples in a docker container
+
+To run examples from this directory inside a Docker container you can use `docker-compose`.
+
+```bash
+docker-compose run -T xk6-browser run - <examples/browser_on.js
+```
+
 ## Status
 
 Currently only Chromium is supported, and the [Playwright API](https://playwright.dev/docs/api/class-playwright) coverage is as follows:
--- a/README.md
+++ b/README.md
@@ -393,7 +393,7 @@ export default function () {
 
 ### Running examples in a Docker container
 
-The examples above are also available as standalone files in the [`examples` directory][/examples]. You can run them in a Docker container using Docker Compose with:
+The examples above are also available as standalone files in the [`examples` directory](/examples). You can run them in a Docker container using Docker Compose with:
 
 ```bash
 docker-compose run -T xk6-browser run - <examples/browser_on.js
--- a/README.md
+++ b/README.md
@@ -391,9 +391,9 @@ export default function () {
 }
 ```
 
-### Running examples in a docker container
+### Running examples in a Docker container
 
-To run examples from this directory inside a Docker container you can use `docker-compose`.
+The examples above are also available as standalone files in the [`examples` directory][/examples]. You can run them in a Docker container using Docker Compose with:
 
 ```bash
 docker-compose run -T xk6-browser run - <examples/browser_on.js
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,5 @@
+version: '3.4'
+
+services:
+  xk6-browser:
+    build: .


examples/docker-compose.yaml
```diff
@@ -0,0 +1,5 @@
+version: '3.4'
```
olegbespalov, line 1: Question: maybe move this to the repo root, and then the example can be something like: 

```
docker-compose run xk6-browser run - <examples/browser_on.js
```

WDYT?

ankur22: I don't have a very strong opinion on this. I think leaving it here for now works, and can move it later if we need.

ankur22 (approved):
LGTM, but i'd wait for others to take a look

examples/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM golang:1.19-alpine as builder
+
+RUN apk --no-cache add git
+
+RUN go install -trimpath go.k6.io/xk6/cmd/xk6@latest
+
+RUN  xk6 build --output "/tmp/k6" --with github.com/grafana/xk6-browser
+
+FROM alpine:3.14
```
imiric, line 9: I'm slightly weary of using Chromium on Alpine, because of the musl build. It does seem to work, and `chromium` is an official Alpine package, but if we run into strange issues, it would be difficult to tell whether they're caused by the Alpine build or not. This is why we decided to run on AL2022 in the Cloud instead of Alpine.

So better be safe than sorry, and use official Chromium team builds instead, on a glibc distro? Ubuntu would probably be fine.

examples/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+ENV XK6_HEADLESS=true
+
+RUN apk add chromium
```
imiric, line 13: Maybe we should version pin the `chromium` package as well? Via an `ARG`, so that it could be changed at build time.

imiric (changes requested):
Thanks for this @olegbespalov! :clap:

I don't want to needlessly delay merging this, but think we should discuss some possible improvements first.

In general, I'm not sure if suggesting Docker is a net help for users, or if it only confuses them when it comes to dealing with volumes, ports, etc. We've certainly had many forum questions that were entirely because of issues with Docker. :disappointed:

examples/Dockerfile
olegbespalov: I don't mind changing the base distro, whatever you belive works better :+1:

examples/README.md
```diff
@@ -0,0 +1,9 @@
+## Examples
+
+# Running examples in the docker
+
+To run examples from this folder inside the docker you can use docker-compose.
```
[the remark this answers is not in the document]
olegbespalov: Personally, I believe that publishing the xk6-browser image is a good step. As I said everything that simplifies trying the product from my point of view provides a better developer UX.

examples/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+```bash
+docker-compose run xk6-browser run - <browser_on.js
```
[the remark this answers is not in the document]
olegbespalov: Interesting :thinking:

examples/docker-compose.yaml
inancgumus: How would we run it if we didn't have docker-compose?

olegbespalov: >How would we run a script if we didn't have docker-compose?

Well, by using docker... But are there any issues with the `docker-compose` many other projects use, it's a kind of standard way of quickly trying things :shrug:

examples/Dockerfile
olegbespalov: done :white_check_mark:

examples/docker-compose.yaml
imiric: I think there's much value using Docker Compose when we only have a single service, but if you guys think we should keep it, I'm not strongly against it.

examples/Dockerfile
imiric: You mean it gets stuck with no progress? Strange, I don't see why Ubuntu would affect things.

Do plain k6 scripts work?

olegbespalov: Yep, samples from the original k6 project work:

```
 docker-compose run -T xk6-browser run - <../../k6/samples/http_batch.js
Creating xk6-browser_xk6-browser_run ... done

          /\      |‾‾| /‾‾/   /‾‾/
     /\  /  \     |  |/  /   /  /
    /  \/    \    |     (   /   ‾‾\
   /          \   |  |\  \ |  (‾)  |
  / __________ \  |__| \__\ \_____/ .io

  execution: local
     script: -
     output: -

  scenarios: (100.00%) 1 scenario, 1 max VUs, 10m30s max duration (incl. graceful stop):
           * default: 1 iterations for each of 1 VUs (maxDuration: 10m0s, gracefulStop: 30s)


running (00m00.9s), 0/1 VUs, 1 complete and 0 interrupted iterations
default ✓ [ 100% ] 1 VUs  00m00.9s/10m0s  1/1 iters, 1 per VU

     ✓ main page 200
     ✓ pi page 200
     ✓ pi page has right content

     checks.........................: 100.00% ✓ 3        ✗ 0
     data_received..................: 23 kB   25 kB/s
     data_sent......................: 1.1 kB  1.2 kB/s
     http_req_blocked...............: avg=321.58ms min=239.39ms med=321.59ms max=403.73ms p(90)=403.73ms p(95)=403.73ms
     http_req_connecting............: avg=224.23ms min=203.24ms med=224.23ms max=245.21ms p(90)=245.19ms p(95)=245.2ms
     http_req_duration..............: avg=145.08ms min=143.32ms med=145.08ms max=146.84ms p(90)=146.83ms p(95)=146.84ms
       { expected_response:true }...: avg=145.08ms min=143.32ms med=145.08ms max=146.84ms p(90)=146.83ms p(95)=146.84ms
     http_req_failed................: 0.00%   ✓ 0        ✗ 4
     http_req_receiving.............: avg=176.36µs min=145.04µs med=184.2µs  max=192µs    p(90)=192µs    p(95)=192µs
     http_req_sending...............: avg=139.1µs  min=79.62µs  med=138.39µs max=199.98µs p(90)=199.12µs p(95)=199.55µs
     http_req_tls_handshaking.......: avg=79.2ms   min=0s       med=79.18ms  max=158.44ms p(90)=158.42ms p(95)=158.43ms
     http_req_waiting...............: avg=144.77ms min=143.06ms med=144.78ms max=146.45ms p(90)=146.44ms p(95)=146.44ms
     http_reqs......................: 4       4.278293/s
     iteration_duration.............: avg=933.78ms min=933.78ms med=933.78ms max=933.78ms p(90)=933.78ms p(95)=933.78ms
     iterations.....................: 1       1.069573/s
```

Just in case `Dockerfile`

```
FROM golang:1.19-alpine as builder

RUN apk --no-cache add git

RUN go install -trimpath go.k6.io/xk6/cmd/xk6@latest

RUN  xk6 build --output "/tmp/k6" --with github.com/grafana/xk6-browser

FROM ubuntu:22.04

RUN apt-get update && apt-get install -y \ 
  chromium-browser

COPY --from=builder /tmp/k6 /usr/bin/k6

ENV XK6_HEADLESS=true

ENTRYPOINT ["k6"]
```

imiric: Hhmm if you run with trace logs enabled, like:

```
docker run -i -e XK6_HEADLESS=1 -e XK6_BROWSER_LOG=trace imiric/xk6-browser-test run - <examples/fillform.js
```

This is logged:

```
time="2022-09-30T14:27:56Z" level=debug msg="process with PID 21 unexpectedly ended: exit status 1" category=browser elapsed="0 ms" goroutine=87
```

So for some reason the browser process is crashing. :confused: We're not handling it properly, since the iteration/test should end as well, and this should be logged as an error (after #535 this should've been logged properly, at least, but I tried building from `@main` and the same thing happened).

Initially I thought that building the binary on Alpine and using it in Ubuntu would cause problems, but that works fine, and wouldn't cause the browser to crash. We should probably use the same base image for `builder` and the final stage, just to avoid Docker downloading both Alpine and Ubuntu images.

We'll need some time to look into this, probably enable debug Chromium logs, etc.

@olegbespalov Do you mind creating an issue for this Docker issue? We should fix it even if we don't merge this, and I'm not sure if we should just merge the Alpine version, or wait until we fix the Ubuntu version. WDYT?

imiric: Ah, wait, `chromium-browser` is a Snap package on Ubuntu :facepalm:

Let's use Debian instead then. Base `builder` on `golang:1.19-buster` and the final stage on `debian:buster`. The package name is `chromium`.

examples/docker-compose.yaml
inancgumus: 👍 for the docker-compose. In my opinion, it's simpler to run a single command (`docker-compose`) rather than running it through `docker` (build + run steps).

inancgumus (approved):
Nice, @olegbespalov 👏  I believe this is another user experience improvement for xk6-browser.

examples/Dockerfile
olegbespalov: I've checked the `debian:buster` it has the pretty old version of the Chromium `90.0.4430.212-1~deb10u1` that's why pointed to the `bullseye`.

So, adjust everything to the `bullseye`

examples/docker-compose.yaml
olegbespalov: Yeah, at least till the point when CI gets the builds & pushes them to the docker hub, then probably nothing else apart from the will be needed:

```
docker run -i grafana/xk6-browser run - <examples/fillform.js
```

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN go install -trimpath go.k6.io/xk6/cmd/xk6@latest
+
+RUN  xk6 build --output "/tmp/k6" --with github.com/grafana/xk6-browser
+
+FROM debian:bullseye
+
+ARG CHROMIUM_VERSION=106.0.5249.61-1~deb11u1
+
+RUN apt-get update
+RUN apt-get install -y chromium=${CHROMIUM_VERSION}
```
imiric, line 12: These can be run together, to minimize the image layers.
```suggestion
RUN apt-get update && \
    apt-get install -y chromium=${CHROMIUM_VERSION}
```

README.md
```diff
@@ -391,6 +391,14 @@ export default function () {
 }
 ```
 
+### Running examples in a docker container
+
+To run examples from this directory inside a Docker container you can use `docker-compose`.
```
imiric, line 396: Just to clarify a bit:

```suggestion
### Running examples in a Docker container

The examples above are also available as standalone files in the [`examples` directory][/examples]. You can run them in a Docker container using Docker Compose with:
```

examples/Dockerfile
imiric: Ah, right, sounds good. I'd prefer if we could use an official Chromium repository, to avoid depending on distro packages which are often outdated, but `debian:bullseye` is fine.

imiric (approved):
Thanks Oleg, LGTM! :bow:

Just left the last couple of minor suggestions.

README.md
```diff
@@ -391,6 +391,14 @@ export default function () {
 }
 ```
 
+### Running examples in a Docker container
+
+The examples above are also available as standalone files in the [`examples` directory][/examples]. You can run them in a Docker container using Docker Compose with:
```
imiric, line 396: Wait, sorry, I messed up the Markdown link :facepalm:
```suggestion
The examples above are also available as standalone files in the [`examples` directory](/examples). You can run them in a Docker container using Docker Compose with:
```

PR #563, opened 2022-10-03
Get DevTools URL from data dir file instead of by parsing stdout

### LINKED ISSUE #491: Fails to retrieve the devtools url when running in Windows WSL2
Original community forum post: https://community.k6.io/t/xk6-browser-error-getting-devtools-url-context-deadline-exceeded/4410

xk6-browser version: https://github.com/grafana/xk6-browser/releases/tag/v0.4.0
Platform: Windows 11, WSL2, Ubuntu 20.04

## Issue

The user reported that they were unable to run the example that was posted in [this blog post](https://k6.io/blog/announcing-xk6-browser-testing/). They were able to install and compile a binary using the commands that were in the blog post, but when it came to running the example test, the test times out and fails with:

```
ERRO[0030] launching browser: getting DevTools URL: context deadline exceeded
default at file:/xk6-browser-examples/browser-test.js:4:19(5)
        at native executor=per-vu-iterations scenario=default source=stacktrace
```

I was able to reproduce the issue on my Windows laptop.

imiric:
This is slightly more elegant than parsing stdout, but more importantly, shouldn't cause race conditions where we miss the "DevTools listening on" line, since it doesn't start a goroutine.

A drawback is that it could take slightly longer to return the URL because it's reading from disk and does several read attempts with delays, but in my tests the difference is negligible. In total, it allots 500ms to read the file, which should hopefully be plenty even on IO-starved machines (VMs, CI, etc.). IMO if this takes more time than that then it's likely the environment will not be fit for running a test. But we can always increase it if you prefer.

I didn't add tests for it, since pretty much all our tests check this functionality indirectly, but let me know if I should add some.

Also note that I haven't confirmed this definitely fixes #491, since I wasn't able to reproduce the issue in WSL2 (though did run into some others... :sweat:), and the issue is very rare on Linux/EC2/Cloud, but I'm confident it does, and we can make a Cloud release and test this thoroughly to have more assurance.

### DIFF
--- a/common/browser_process.go
+++ b/common/browser_process.go
@@ -23,11 +23,12 @@ package common
 import (
 	"bufio"
 	"context"
+	"errors"
 	"fmt"
-	"io"
 	"os"
 	"os/exec"
-	"strings"
+	"path/filepath"
+	"time"
 
 	"github.com/grafana/xk6-browser/log"
 	"github.com/grafana/xk6-browser/storage"
@@ -58,12 +59,12 @@ func NewBrowserProcess(
 	ctx context.Context, path string, args, env []string, dataDir *storage.Dir,
 	ctxCancel context.CancelFunc, logger *log.Logger,
 ) (*BrowserProcess, error) {
-	cmd, stdout, procDone, err := execute(ctx, path, args, env, dataDir, logger)
+	cmd, procDone, err := execute(ctx, path, args, env, dataDir, logger)
 	if err != nil {
 		return nil, err
 	}
 
-	wsURL, err := parseDevToolsURL(ctx, stdout)
+	wsURL, err := getDevToolsURL(dataDir.Dir)
 	if err != nil {
 		return nil, fmt.Errorf("getting DevTools URL: %w", err)
 	}
@@ -141,15 +142,9 @@ func (p *BrowserProcess) AttachLogger(logger *log.Logger) {
 func execute(
 	ctx context.Context, path string, args, env []string, dataDir *storage.Dir,
 	logger *log.Logger,
-) (*exec.Cmd, io.Reader, chan struct{}, error) {
+) (*exec.Cmd, chan struct{}, error) {
 	cmd := exec.CommandContext(ctx, path, args...)
 	killAfterParent(cmd)
-
-	stdout, err := cmd.StdoutPipe()
-	if err != nil {
-		return nil, nil, nil, fmt.Errorf("%w", err)
-	}
-	cmd.Stderr = cmd.Stdout
 
 	// Set up environment variable for process
 	if len(env) > 0 {
@@ -158,15 +153,15 @@ func execute(
 
 	// We must start the cmd before calling cmd.Wait, as otherwise the two
 	// can run into a data race.
-	err = cmd.Start()
+	err := cmd.Start()
 	if os.IsNotExist(err) {
-		return nil, nil, nil, fmt.Errorf("file does not exist: %s", path)
+		return nil, nil, fmt.Errorf("file does not exist: %s", path)
 	}
 	if err != nil {
-		return nil, nil, nil, fmt.Errorf("%w", err)
+		return nil, nil, fmt.Errorf("%w", err)
 	}
 	if ctx.Err() != nil {
-		return nil, nil, nil, fmt.Errorf("%w", ctx.Err())
+		return nil, nil, fmt.Errorf("%w", ctx.Err())
 	}
 
 	done := make(chan struct{})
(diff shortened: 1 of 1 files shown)


common/browser_process.go
```diff
@@ -186,37 +181,48 @@ func execute(
...
-				return
-			}
+// getDevToolsURL returns the DevTools WebSocket address by reading the
+// DevToolsActivePort file in the data directory.
+func getDevToolsURL(dataDir string) (wsURL string, rerr error) {
+	var (
+		f                  *os.File
+		fpath              = filepath.Join(dataDir, "DevToolsActivePort")
+		maxReadAttempts    = 10
+		readAttemptDelayMS = 50
```
imiric, line 194: Yeah, `test-tip` failed with `panic: launching browser: getting DevTools URL: unable to read file "/tmp/xk6-browser-data-2308635430/DevToolsActivePort" in 500ms`. :disappointed:

I guess we could just increase it to 5s total to be safe (so 100 attempts)? Even 1 or 2s might be too low for some machines.

mstoykov:
> This is slightly more elegant than parsing stdout, but more importantly, shouldn't cause race conditions where we miss the "DevTools listening on" line, since it doesn't start a goroutine.

From this line I gather that currently it is thought that the reason why xk6-browser can't find the port to connect to is because:
-  the starting of the goroutine takes too long 
- and the line is already printed by the time the goroutine starts to read it?

inancgumus:
@imiric We can have the best of both worlds:

* First, try to get the URL faster via stdout (the previous way)
* Second, if the first one fails, get it from the file (current way)

This way, we can get the URL faster, and only if it fails can we try to get it from the file.

WDYT?

imiric:
@mstoykov That's the theory, but it wasn't confirmed since this issue is so difficult to reproduce.

It could also be that it just takes more than 30s to start the browser process, which is very excessive, but not out of the question on VMs with bursty/unreliable IO like in CI or EC2. In that case, this won't help and we should increase the number of attempts/timeout, but I would still argue this approach is cleaner than parsing stdout.

mstoykov:
> That's the theory, but it wasn't confirmed since this issue is so difficult to reproduce.

Is there something else reading the output? Because if it isn't - there is no way that we can miss it. If anything the current code should continue to read the output as otherwise the buffer used to save it (in the OS) might overfill. This likely isn't a problem with the current usage where (from my observations) the browser does not output more then 3-4 lines at the very beginning of the execution.

So it seems to me that 
> It could also be that it just takes more than 30s to start the browser process
Is the more likely culprit?

I meant to open an issue ot log the actual output of the browser :thinking: Maybe that will help with diagnosing this?

imiric:
@inancgumus I'd rather have one, preferably simpler, way of doing this.

Besides, I'm not certain whether the file is written before or after that line is printed to stdout, so reading the file might not make that big of a difference. At least it doesn't from my local tests. I'll try to confirm with `strace` or _* shudders *_ reading the Chromium source.

common/browser_process.go
```diff
@@ -186,37 +181,48 @@ func execute(
...
+	)
+
+	// The browser might not have created the file yet, so try reading it
+	// multiple times after a slight delay.
+	for readAttempts := 0; readAttempts < maxReadAttempts; readAttempts++ {
+		var err error
+		f, err = os.Open(fpath) //nolint:gosec  // false positive, https://github.com/securego/gosec/issues/439
+		if errors.Is(err, os.ErrNotExist) {
+			time.Sleep(time.Duration(readAttemptDelayMS) * time.Millisecond)
```
ankur22, line 203: Might we need to use a select statement here with `time.Timer` and also watch for the `context` being cancelled?

common/browser_process.go
```diff
@@ -58,12 +59,12 @@ func NewBrowserProcess(
...
 	ctxCancel context.CancelFunc, logger *log.Logger,
 ) (*BrowserProcess, error) {
-	cmd, stdout, procDone, err := execute(ctx, path, args, env, dataDir, logger)
+	cmd, procDone, err := execute(ctx, path, args, env, dataDir, logger)
 	if err != nil {
 		return nil, err
 	}
 
-	wsURL, err := parseDevToolsURL(ctx, stdout)
+	wsURL, err := getDevToolsURL(dataDir.Dir)
```
ankur22, line 67: If the `getDevToolsURL` was called prior to calling `execute` (which will require some goroutine orchestration so that execute is only called once `getDevToolsURL` has started waiting for the file)  then we should avoid any possible race conditions, no?

imiric: True, that could be a solution as well. Though a) we're not 100% sure this is caused by a race, see the discussion above, and b) starting a goroutine with proper orchestration and parsing stdout is arguably messier than just reading a file.

imiric:
According to `strace`, the file is written to immediately after `DevTools listening on ...` is written to stderr:

```
write(2, "\nDevTools listening on ws://127."..., 98) = 98
creat("/tmp/xk6-browser-data-3945058657/DevToolsActivePort", 0666) = 48
write(48, "40575\n/devtools/browser/fc99a07c"..., 60) = 60
```

Still, this approach could have a slight delay (~50ms), depending on when we succeed reading the file.

Judging by the feedback, you all prefer the stdout parsing approach, and we should drop this? Since a race is unlikely as stdout would be buffered, as Mihail says, then the only explanation is that it truly does take >30s to start the browser in some cases. That's out of our hands to fix here, but maybe we can run on more stable and better performing EC2 instances to minimize the chances of this happening.

ankur22:
> then the only explanation is that it truly does take >30s to start the browser in some cases. That's out of our hands to fix here, but maybe we can run on more stable and better performing EC2 instances to minimize the chances of this happening.

I think it would be good to find a way to replicate this issue. Let me know if you need me to test any changes on my Windows laptop.

Issue #533, opened 2022-09-14
Wrong metric sample timestamp pushed to Cloud


imiric:
On current `main` (cf033cb), running the following script:

<details>
  <summary>script.js</summary>

```javascript
import { check } from 'k6';
import { chromium } from 'k6/x/browser';

export const options = {
  vus: 1,
  duration: '60s',
}

export default function () {
  const browser = chromium.launch({
    headless: true,
  });
  const context = browser.newContext();
  const page = context.newPage();

  // Goto front page, find login link and click it
  page.goto('https://test.k6.io/', { waitUntil: 'networkidle' });
  Promise.all([
    page.waitForNavigation(),
    page.locator('a[href="/my_messages.php"]').click(),
  ]).then(() => {
    // Enter login credentials and login
    page.locator('input[name="login"]').type('admin');
    page.locator('input[name="password"]').type('123');
    // We expect the form submission to trigger a navigation, so to prevent a
    // race condition, setup a waiter concurrently while waiting for the click
    // to resolve.
    return Promise.all([
      page.waitForNavigation(),
      page.locator('input[type="submit"]').click(),
    ]);
  }).then(() => {
    check(page, {
      'header': page.locator('h2').textContent() == 'Welcome, admin!',
    });
  }).finally(() => {
    page.close();
    browser.close();
  });
}
```

</details>

... and sending the metrics to the Cloud with:

```
xk6-browser run --output cloud script.js
```

Results in the test run ending with a `Timed out` status, not showing any metric results (the metrics do appear while the test is running, but disappear when it finishes):

![2022-09-14-154703_1556x909_scrot](https://user-images.githubusercontent.com/1009277/190171633-cf8b940a-fabc-43ba-a2ff-a4143c6426f0.png)

And a strange "Started" timestamp that is ~11 days in the past in this example:

![2022-09-14-144430_551x313_scrot](https://user-images.githubusercontent.com/1009277/190171206-3c02b54a-d6ee-432e-bf10-9aa22ada192f.png)

See test run 140998 on staging for details.

After further troubleshooting it seems that _some_ metric samples are being sent with this wrong timestamp, which confuses the backend. I added some debug print statements to `k6.output.cloud.MetricsClient.PushMetric()`:

<details>
  <summary>patch</summary>

```diff
diff --git a/output/cloud/metrics_client.go b/output/cloud/metrics_client.go
index 6aef2076a..4faa66c4b 100644
--- a/output/cloud/metrics_client.go
[...]

PR #564, opened 2022-10-05
Refactor to NewRequestParams

inancgumus:
Using a relatively long list of primitive types makes it hard to initialize `Request` and can bring uninvited bugs. I noticed this when testing `Request`/`Response` metric emissions for Issue #533. I needed to pass `nil` and false values for unwanted fields:

```go
req, err := NewRequest(vu.Context(), evt, nil, nil, "intercept", false)
```

Now:

```go
req, err := NewRequest(vu.Context(), NewRequestParams{
	event:          evt,
	interceptionID: "intercept",
})
```

I'm using this approach in #551 and will keep doing so if this PR is accepted.

### DIFF
--- a/common/network_manager.go
+++ b/common/network_manager.go
@@ -460,7 +460,13 @@ func (m *NetworkManager) onRequest(event *network.EventRequestWillBeSent, interc
 			event.Request.URL, event.Request.Method, event.Initiator.Type, event.FrameID)
 	}
 
-	req, err := NewRequest(m.ctx, event, frame, redirectChain, interceptionID, m.userReqInterceptionEnabled)
+	req, err := NewRequest(m.ctx, NewRequestParams{
+		event:             event,
+		frame:             frame,
+		redirectChain:     redirectChain,
+		interceptionID:    interceptionID,
+		allowInterception: m.userReqInterceptionEnabled,
+	})
 	if err != nil {
 		m.logger.Errorf("NetworkManager", "creating request: %s", err)
 		return
--- a/common/request.go
+++ b/common/request.go
@@ -67,56 +67,60 @@ type Request struct {
 	vu                  k6modules.VU
 }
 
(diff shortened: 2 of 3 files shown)


mstoykov (approved):

ankur22 (approved):
Seems like a sensible change, although generally a long list of parameters might be a sign of something else. Consolidating params into a `XXXParams` struct feels like we're just hiding the problem when maybe we should step back and think about how it's being used and how we can refactor it. I can't point to any specific example though which would help explain what I'm getting at.

imiric (approved):
I'm not a big fan of `Params` types, as usually your IDE should let you know the name of the parameter you're passing, so I haven't experienced the confusion personally, but don't mind this if you prefer it.

inancgumus:
@ankur22 There is definitely an underlying issue, and the long list of parameters reflects that. However, the issue won't be solved in a day. The `Request` type has a lot of dependencies as much as the other types in the `common` package. This issue makes them hard to test, and having a long parameter list does not help either.

I know that this refactoring is not the best one, but I believe it makes the usage of the `Request` type more readable and opens the door for further refactoring.

@ivan The issue is not that my IDE cannot show me the name of the parameters I'm passing. The problem is we're unnecessarily passing `nil` and `false` parameters whenever we want to use the type. I believe this causes _unnecessary_ duplication and makes it less readable. Since this is a subjective issue, I'd like not to discuss it further :)

andrewslotin:
@imiric, I believe there are two benefits to use parameter types in such cases:
1. As @inancgumus said, it allows one to omit specifying zero values explicitly.
2. Probably minor in this case, since this project is not a library, but using a struct does not break backward-compatibility whenever you need to add another parameter, which is often the case with methods that initialize and configure instances. A new parameter almost always can be added to the configuration object type in a way that its zero value does not change the original behavior.

Issue #456, opened 2022-07-13
`setDefaultTimeout` on `BrowserContext` isn't converting the time correctly


ankur22:
Tested against: https://github.com/grafana/xk6-browser/commit/dbede120c63df43995813a847a25b0e66e289592

I was expecting `setDefaultTimeout` to timeout but the timeout didn't occur when I ran the following script:

In the below code it works against `localhost:8080`, which is a test server that can be found here with instructions on how to run it locally: https://github.com/ankur22/testserver.

```javascript
import { sleep } from 'k6';
import launcher from 'k6/x/browser';

export default function () {
  const browser = launcher.launch('chromium', {
      headless: false,
  });
  const context = browser.newContext();
  const page = context.newPage();
  context.setDefaultTimeout(1);
  const res = page.goto('http://localhost:8080/slow');

  sleep(10);
  browser.close();
}
```

In the documentation I found this comment which might give us a clue as to why:

```javascript
const context = browser.newContext();
// FIXME: This is interpreted as 16m40s. Wrong JS->Go conversion.
context.setDefaultTimeout(1000); // 1s
const page = context.newPage();
page.click('h2'); // times out
```

mstoykov:
As I mention in https://github.com/grafana/xk6-browser/pull/574#discussion_r992035380 

we can use the k6's https://github.com/grafana/k6/blob/fcbbcd54acb2ee58cf16d6a15c730809a541afe9/lib/types/types.go#L190-L211 but it will add support for "1s" , "5m" which is something playwright doesn't support. 

Argaubly that is better for us ... but I would expect that if somebody writes "5m" in playwright they will get `5` milliseconds so it might be a thing we don't want to do :shrug:

ankur22:
Hmm, i think being aligned with k6 would be the best, especially since the browser will one day be merged into k6.

inancgumus:
Should we both support Playwright-like integers and k6-like human-readable ones to keep the compatibility promise? We can do that by wrapping the k6 one. I'm also OK with directly using the k6 one.

mstoykov:
@inancgumus the k6 one already supports integers(and floats) and interpret them as milliseconds, which seems to be the de facto standard across the js world :shrug:

ankur22:
[...]

PR #574, opened 2022-10-10
Add more promiseHelpers and awaitWithTimeout

mstoykov:
- promiseThen and promiseAll are callign the underlying js methods making it easier to write tests with promises.
- awaitWithTimeout is the same with `await` but waits for the provided timeout and returns an error if the timeout was reached.

### DIFF
--- a/tests/test_browser.go
+++ b/tests/test_browser.go
@@ -283,7 +283,8 @@ func (b *testBrowser) awaitWithTimeout(timeout time.Duration, fn func() error) e
 func (b *testBrowser) promiseThen(promise *goja.Promise, onFulfilled, onRejected func(goja.Value)) *goja.Promise {
 	b.t.Helper()
 	rt := b.runtime()
-	val, err := rt.RunString(`(function(promise, onFulfilled, onRejected) { return promise.then(onFulfilled, onRejected) })`)
+	val, err := rt.RunString(
+		`(function(promise, onFulfilled, onRejected) { return promise.then(onFulfilled, onRejected) })`)
 	require.NoError(b.t, err)
 	cal, ok := goja.AssertFunction(val)
 	require.True(b.t, ok)
@@ -295,6 +296,7 @@ func (b *testBrowser) promiseThen(promise *goja.Promise, onFulfilled, onRejected
 	require.NoError(b.t, err)
 	newPromise, ok := val.Export().(*goja.Promise)
 	require.True(b.t, ok)
+
 	return newPromise
 }
 
@@ -313,6 +315,7 @@ func (b *testBrowser) promiseAll(promises ...*goja.Promise) *goja.Promise {
 	require.NoError(b.t, err)
 	newPromise, ok := val.Export().(*goja.Promise)
 	require.True(b.t, ok)
+
 	return newPromise
 }
 


tests/frame_manager_test.go
```diff
@@ -20,7 +20,7 @@ func TestWaitForFrameNavigationWithinDocument(t *testing.T) {
 	}
 	t.Parallel()
 
-	var timeout time.Duration = 5000 // interpreted as ms
+	timeout := 5 * time.Second
```
imiric, line 23: This shouldn't be changed until the wrong timeout conversion is fixed (related to #456).

The reason is that `WaitForNavigation()` expects an integer, which is then converted to ms:
https://github.com/grafana/xk6-browser/blob/508bcb861edcdad7615e53cbf64c3829b4dfd0b3/common/frame_options.go#L676

If you print `o.Timeout` here, you'll see that `5 * time.Second` sets it to `1388h53m20s`.

Last time I took a look at this, it involved changing how the default timeout is set, which affected a bunch of other places, and I never fixed it. Worth taking a look now and seeing if it can be fixed only for `WaitForNavigation()`?

tests/test_browser.go
```diff
@@ -259,6 +260,26 @@ func (b *testBrowser) await(fn func() error) error {
...
+
+	// use timer instead of time.After to not leak time.After for the duration of the timeout
+	t := time.NewTimer(timeout)
+	defer t.Stop()
+
+	select {
+	case err := <-errC:
+		return err
+	case <-t.C:
+		return fmt.Errorf("test timeouted after %s", timeout)
```
imiric, line 279:
```suggestion
		return fmt.Errorf("test timed out after %s", timeout)
```

imiric (changes requested):
Makes sense :+1:, but we should either revert the timeout value change, or fix the parsing.

tests/frame_manager_test.go
mstoykov: I decided against going down another rabbit hole especially as from the playwright documentation this should be taking integers only. 

But if you want I can try to make a fix using https://github.com/grafana/k6/blob/fcbbcd54acb2ee58cf16d6a15c730809a541afe9/lib/types/types.go#L190-L211 which is what k6 does.

tests/frame_manager_test.go
```diff
@@ -41,32 +41,18 @@ func TestWaitForFrameNavigationWithinDocument(t *testing.T) {
...
-					}
-
-					assert.Equal(t, goja.PromiseStateFulfilled, wfnPromise.State())
-					assert.Equal(t, goja.PromiseStateFulfilled, cPromise.State())
+					wfnPromise := p.WaitForNavigation(tb.toGojaValue(&common.FrameWaitForNavigationOptions{
+						Timeout: time.Duration(timeout.Milliseconds()), // interpreted as ms
+					}))
+					cPromise := p.Click(tc.selector, nil)
+					tb.promiseThen(tb.promiseAll(wfnPromise, cPromise), func(_ goja.Value) {
+						// this is a bit pointless :shrug:
```
inancgumus, line 52: Why is it pointless?

mstoykov: Because we are basically checkking that they are fulfilled in the callback that gets called when they get fulfilled - in practice not checking anything.

tests/frame_manager_test.go
```diff
@@ -41,32 +41,18 @@ func TestWaitForFrameNavigationWithinDocument(t *testing.T) {
...
-						wfnPromise = p.WaitForNavigation(tb.toGojaValue(&common.FrameWaitForNavigationOptions{
-							Timeout: timeout, // interpreted as ms
-						}))
-						cPromise = p.Click(tc.selector, nil)
-
-						assert.Equal(t, goja.PromiseStatePending, wfnPromise.State())
-						assert.Equal(t, goja.PromiseStatePending, cPromise.State())
-
-						return nil
-					})
```
inancgumus: I wonder what is the difference between:
* Initiating two promises in the `await` closure, vs,
* Putting them in a `promiseAll`?

Is it about making it as much closer to the Javascript implementation as possible?

tests/frame_manager_test.go
inancgumus: Hmm, should we remove the checks in closure then?

tests/frame_manager_test.go
mstoykov: > Initiating two promises in the await closure 

`await` is misleadingly named (and I will likely make a PR once I figure a better name). `await` here is literally "start the event loop" which will wait until it doesn't have any more work or error occured. 

Calling `await` *within* `await` starts the event loop within the event loop, which is not supported - and doesn't even have a defined behaviour. It worked in this particular case, because the outer await is basically doing nothing, but in general *should* not be done.

Given that I am actually moving `goto` to be async (already done, just figuring out how to make the test less ... painful to write) I am in practice building up the codebase to make writing tests with async parts easier. 

> much closer to the Javascript implementation as possible

This is another argument and arguably a lot of the tests should just run `js` code instead. But I am not going to rewrite that ;). Also at least some need (or at least use) access to things that aren't available from JavaScript and should not be available. So from that perspective things such as PromiseThen and promiseAll will be needed (or will need to be written on case by case basis ;).

tests/frame_manager_test.go
mstoykov: :shrug: then it will be empty and a bit ... strange. I guess I can make certain that it gets called by setting a boolean :shrug:

tests/frame_manager_test.go
inancgumus: Thanks for explaining 👍

tests/frame_manager_test.go
imiric: Nah, let's just leave the timeout as is for now, but it makes sense to use `GetDurationValue()` once we fix this.

I think we should break compatibility with Playwright in this case, since duration strings are much friendlier than unitless integers, but we can discuss that in #456.

tests/frame_manager_test.go
imiric: This test is a bit pointless to begin with. The only reason it was initially like this is because `WaitForNavigation()` panicked on same-document navigations, and this tested that it didn't panic anymore.

But to make it useful, we should make it more functional, and assert that e.g. a certain element is visible after navigating on the same page.

Doesn't have to be done here, just FYI.

imiric (approved):
LGTM

I like the `promiseThen` and `promiseAll` helpers as well :+1: Maybe update the PR title and description to match?

ankur22 (approved):

tests/test_browser.go
```diff
@@ -259,6 +260,65 @@ func (b *testBrowser) await(fn func() error) error {
 	return b.vu.Loop.Start(fn)
 }
 
+// awaitWithTimeout is the same as await but takes a timeout and times out the function after the time runs out.
+func (b *testBrowser) awaitWithTimeout(timeout time.Duration, fn func() error) error {
```
imiric, line 264: @mstoykov One thing I just noticed is that if a `testify/assert` or `require` fails somewhere in a Promise created in `fn`, then we would get the test failure, but would still have to wait for the timeout to expire, and would then get the timeout failure.

Is there a way of exposing the first error here, and aborting the timeout in that case?

mstoykov: https://github.com/grafana/xk6-browser/pull/585

PR #576, opened 2022-10-10
Check session for nilness before making a page from it

### LINKED ISSUE #509: Nil panic calling `Session.ID()`
This seems to be a new flaky nil panic :disappointed: Spotted 2 times in #508 (job [#1237](https://github.com/grafana/xk6-browser/runs/7894489703?check_suite_focus=true) and [#1240](https://github.com/grafana/xk6-browser/runs/7895924239?check_suite_focus=true)):

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x112daaa]

goroutine 1155 [running]:
github.com/grafana/xk6-browser/common.(*Session).ID(0x0)
	/home/runner/work/xk6-browser/xk6-browser/common/session.go:77 +0x2a
github.com/grafana/xk6-browser/common.(*Page).sessionID(...)
	/home/runner/work/xk6-browser/xk6-browser/common/page.go:922
github.com/grafana/xk6-browser/common.NewPage({0x1cace60?, 0xc0006548d0}, {0x1cb1d10?, 0x0}, 0xc000706b40, {0xc0000486c0, 0x20}, 0x0, 0x1, 0xc000705e30)
	/home/runner/work/xk6-browser/xk6-browser/common/page.go:129 +0x7f1
github.com/grafana/xk6-browser/common.(*Browser).onAttachedToTarget(0xc000447e00, 0xc00065c0a0)
	/home/runner/work/xk6-browser/xk6-browser/common/browser.go:287 +0xacf
github.com/grafana/xk6-browser/common.(*Browser).initEvents.func1()
	/home/runner/work/xk6-browser/xk6-browser/common/browser.go:194 +0x425
created by github.com/grafana/xk6-browser/common.(*Browser).initEvents
	/home/runner/work/xk6-browser/xk6-browser/common/browser.go:179 +0x35e
FAIL	github.com/grafana/xk6-browser/tests	5.593s
?   	github.com/grafana/xk6-browser/tests/ws	[no test files]
```

It happened in different tests, `TestLocator/DispatchEvent` and `TestElementHandleBoundingBoxInvisibleElement`.

We do a nil check before calling `session.ID()`:
https://github.com/grafana/xk6-browser/blob/67499f7b2608ada93d467ff8fc1c4890c09d559e/common/page.go#L920-L923

But I suppose it's possible for that to race and `session` to be nil after it? This is only called for logging, though `session` shouldn't be nil at that point... :thinking:

mstoykov:
closes #509

### DIFF
--- a/common/browser.go
+++ b/common/browser.go
@@ -243,8 +243,9 @@ func (b *Browser) onAttachedToTarget(ev *target.EventAttachedToTarget) {
 
 	session := b.conn.getSession(ev.SessionID)
 	if session == nil {
-		b.logger.Debugf("Browser:onAttachedToTarget seesion got closed before attachToTarget was handled",
-			"sid:%v tid:%v", ev.SessionID, evti.TargetID)
+		b.logger.Warnf("Browser:onAttachedToTarget",
+			"session closed before attachToTarget is handled. sid:%v tid:%v",
+			ev.SessionID, evti.TargetID)
 		return // ignore
 	}
 
--- a/common/browser.go
+++ b/common/browser.go
@@ -242,6 +242,11 @@ func (b *Browser) onAttachedToTarget(ev *target.EventAttachedToTarget) {
 	}
 
 	session := b.conn.getSession(ev.SessionID)
+	if session == nil {
+		b.logger.Debugf("Browser:onAttachedToTarget seesion got closed before attachToTarget was handled",
+			"sid:%v tid:%v", ev.SessionID, evti.TargetID)
+		return // ignore
+	}
 
 	switch evti.Type {
 	case "background_page":


common/browser.go
```diff
@@ -242,6 +242,11 @@ func (b *Browser) onAttachedToTarget(ev *target.EventAttachedToTarget) {
 	}
 
 	session := b.conn.getSession(ev.SessionID)
+	if session == nil {
+		b.logger.Debugf("Browser:onAttachedToTarget seesion got closed before attachToTarget was handled",
+			"sid:%v tid:%v", ev.SessionID, evti.TargetID)
```
inancgumus, line 247: IMO, this should produce a warning for us to catch it later on when it happens.

```suggestion
		b.logger.Warnf("Browser:onAttachedToTarget",
			"session closed before attachToTarget is handled. sid:%v tid:%v",
			ev.SessionID, evti.TargetID)
```

imiric (approved):
On the one hand, this check makes sense. We don't want to create a page on a nil session.

But on the other, it feels like a bandaid on the root issue of lacking synchronization between handlers. I.e. it should be safe for `Browser.onAttachedToTarget()` to finish, before a `DetachedFromTarget` event is processed. After all, the session could be closed after this nil check, in which case it wouldn't panic on access, but any CDP calls that use it would fail.

So this is fine for fixing #509, but we might want to think about how event handlers can be synchronized, which wouldn't be a trivial change.

PR #585, opened 2022-10-12
Don't wait in awaitWithTimeout if tests fails

mstoykov:


ankur22 (approved):

tests/test_browser.go
```diff
@@ -265,6 +265,7 @@ func (b *testBrowser) awaitWithTimeout(timeout time.Duration, fn func() error) e
 	b.t.Helper()
 	errC := make(chan error)
 	go func() {
+		defer close(errC)
```
ankur22, line 268: Actually, I don't get this change. Although it is good practice to close a channel, why do we need this defer? If we an error is published to the channel, the read side will read and return, so the behaviour is the same, no?

PR #587, opened 2022-10-13
Refactor testPromises

inancgumus:
Mihail did a great job adding promise test helpers to the test browser test suite 👏  I refactored some parts of the test promises to make it slightly more readable (hopefully). But I'm open to suggestions.

### DIFF
--- a/tests/browser_context_options_test.go
+++ b/tests/browser_context_options_test.go
@@ -97,17 +97,15 @@ func TestBrowserContextOptionsExtraHTTPHeaders(t *testing.T) {
 	p := bctx.NewPage()
 
 	err := tb.awaitWithTimeout(time.Second*5, func() error {
-		tb.promiseThen(p.Goto(tb.URL("/get"), nil),
-			func(resp api.Response) {
-				require.NotNil(t, resp)
-				var body struct{ Headers map[string][]string }
-				require.NoError(t, json.Unmarshal(resp.Body().Bytes(), &body))
-				h := body.Headers["Some-Header"]
-				require.NotEmpty(t, h)
-				assert.Equal(t, "Some-Value", h[0])
-			})
+		tb.promise(p.Goto(tb.URL("/get"), nil)).then(func(resp api.Response) {
+			require.NotNil(t, resp)
+			var body struct{ Headers map[string][]string }
+			require.NoError(t, json.Unmarshal(resp.Body().Bytes(), &body))
+			h := body.Headers["Some-Header"]
+			require.NotEmpty(t, h)
+			assert.Equal(t, "Some-Value", h[0])
+		})
 		return nil
 	})
-
 	require.NoError(t, err)
 }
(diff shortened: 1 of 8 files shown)


tests/browser_context_options_test.go
```diff
@@ -97,17 +97,16 @@ func TestBrowserContextOptionsExtraHTTPHeaders(t *testing.T) {
...
-		tb.promiseThen(p.Goto(tb.URL("/get"), nil),
-			func(resp api.Response) {
-				require.NotNil(t, resp)
-				var body struct{ Headers map[string][]string }
-				require.NoError(t, json.Unmarshal(resp.Body().Bytes(), &body))
-				h := body.Headers["Some-Header"]
-				require.NotEmpty(t, h)
-				assert.Equal(t, "Some-Value", h[0])
-			})
+		gotoPage := tb.promise(p.Goto(tb.URL("/get"), nil))
```
inancgumus, line 100: WDYT?

"test browser promise":

```suggestion
		gotoPage := tb.promise(p.Goto(tb.URL("/get"), nil))
```

vs. "test browser wrap promise":


```suggestion
		gotoPage := tb.wrapPromise(p.Goto(tb.URL("/get"), nil))
```

vs. "test browser test promise"

```suggestion
		gotoPage := tb.testPromise(p.Goto(tb.URL("/get"), nil))
```

tests/test_browser.go
```diff
@@ -281,62 +281,75 @@ func (b *testBrowser) awaitWithTimeout(timeout time.Duration, fn func() error) e
...
+		val, err = t.call(goja.Undefined(), rt.ToValue(t.Promise), rt.ToValue(resolve))
+	case 1:
+		val, err = t.call(goja.Undefined(), rt.ToValue(t.Promise), rt.ToValue(resolve), rt.ToValue(reject[0]))
+	default:
+		panic("too many arguments to promiseThen")
+	}
+	require.NoError(t.tb.t, err)
+
+	var ok bool
+	t.Promise, ok = val.Export().(*goja.Promise)
```
mstoykov, line 349: This changes the original testPromise, IMO this shouldn't be the case you should make a new `testPromise`.

mstoykov: That likely can reuse the `call` I think :thinking:

mstoykov (changes requested):

tests/test_browser.go
inancgumus: Yeah, since there is a single promise function created in `promise`, it reuses it. Only changes the current `testPromise.Promise` so that the subsequent calls can use the new promise. Do you see any problem with that?

inancgumus: Or, something like this:

```go
// then(...) {
//      ...
	p, ok := val.Export().(*goja.Promise)
	require.True(t.tb.t, ok)

	return t.tb.promise(p)
```

inancgumus: Thanks, I've updated the test browser as we discussed.

mstoykov (approved):

ankur22 (approved):

PR #588, opened 2022-10-13
Remove copyright disclaimer

inancgumus:
I ran the script that I was keeping to remove the noisy copyright disclaimer from the files.


ankur22 (approved):
Seems like a good idea. Out of curiosity, why was this here in the first place? Could this have superseded the license file in the root dir?

inancgumus:
Yes. We discussed it before that the license file is enough.

Issue #542, opened 2022-09-21
The "complex" test fails on the very first step (`page.goto`)


ankur22:
Commit tested against: 8b52a028840dbaa5ae61937766d473fcde9fca10 (pre-v0.6.0).
Test script: "[complex](https://github.com/grafana/xk6-browser-tests/blob/main/complex/vistaprint.js)" test.

The "complex" test does not work against the version of xk6-browser mentioned above. It navigates to the site, but seems to fail to get passed the `waitUntil: 'networkidle'` stage of the `page.goto` action.

In a randomly chosen tagged version of xk6-browser (`v0.4.0`) it is able to consistently get through most of the "complex" test, so well passed the `page.goto` initial step.

inancgumus:
There is regression somewhere, and `git bisect` can help even though the complex test script is somewhat flaky. Weirdly, we cannot even pass the `page.goto` stage with the script. I vote to make this a high priority after dealing with the current high-priority issues.

mstoykov:
removing the `debug: false` fixed it for me

mstoykov:
> removing the debug: false fixed it for me

This is no longer true with the latest version - I first hit what I fixed in #579 and then I get
```
go.k6.io/k6/js/common.Throw(0x1394a00?, {0x18b3ca0?, 0xc0012c0120?})
        go.k6.io/k6@v0.40.1-0.20221004153024-912b2790958d/js/common/util.go:20 +0x55
github.com/grafana/xk6-browser/k6ext.Panic({0x18c2b68, 0xc002384e40}, {0x1600224, 0x27}, {0xc003491df8, 0x2, 0x2})
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/k6ext/panic.go:51 +0x2d7
github.com/grafana/xk6-browser/common.(*FrameSession).onFrameNavigated(0xc001973c30, 0xc000af66c0, 0x20?)
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame_session.go:650 +0x2e9
github.com/grafana/xk6-browser/common.(*FrameSession).initEvents.func1()
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame_session.go:257 +0x694
created by github.com/grafana/xk6-browser/common.(*FrameSession).initEvents
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/frame_session.go:228 +0x1ab
```

[...]

PR #589, opened 2022-10-13
Fix races around touching state.Group outside of the event loop

mstoykov:
There is no locking around the `state.Group` and it is modified by calls to `group()` while on the event loop. So for the time being to modify it - it is needed to be on the event loop.

Also the adding of the tag isn't needed as that is done automatically by `group()`

### DIFF
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -515,9 +515,11 @@ func (fs *FrameSession) onConsoleAPICalled(event *cdpruntime.EventConsoleAPICall
 		WithTime(event.Timestamp.Time()).
 		WithField("source", "browser-console-api")
 
-	if s := fs.vu.State(); s.Group.Path != "" {
-		l = l.WithField("group", s.Group.Path)
-	}
+		/* accessing the state Group while not on the eventloop is racy
+		if s := fs.vu.State(); s.Group.Path != "" {
+			l = l.WithField("group", s.Group.Path)
+		}
+		*/
 
 	var parsedObjects []interface{}
 	for _, robj := range event.Args {
@@ -688,9 +690,11 @@ func (fs *FrameSession) onLogEntryAdded(event *cdplog.EventEntryAdded) {
 		WithField("url", event.Entry.URL).
 		WithField("browser_source", event.Entry.Source.String()).
 		WithField("line_number", event.Entry.LineNumber)
-	if s := fs.vu.State(); s.Group.Path != "" {
-		l = l.WithField("group", s.Group.Path)
-	}
+		/* accessing the state Group while not on the eventloop is racy
+		if s := fs.vu.State(); s.Group.Path != "" {
+			l = l.WithField("group", s.Group.Path)
+		}
+		*/
 	switch event.Entry.Level {
 	case "info":
 		l.Info(event.Entry.Text)
--- a/common/network_manager.go
+++ b/common/network_manager.go
@@ -166,9 +166,6 @@ func (m *NetworkManager) emitRequestMetrics(req *Request) {
 	state := m.vu.State()
 
 	tags := state.Tags.GetCurrentValues()
-	if state.Options.SystemTags.Has(k6metrics.TagGroup) {
-		tags = tags.With("group", state.Group.Path)
-	}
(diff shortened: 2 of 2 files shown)


mstoykov:
found while testing for #542

common/frame_session.go
```diff
@@ -515,9 +515,11 @@ func (fs *FrameSession) onConsoleAPICalled(event *cdpruntime.EventConsoleAPICall
...
 		WithField("source", "browser-console-api")
 
-	if s := fs.vu.State(); s.Group.Path != "" {
-		l = l.WithField("group", s.Group.Path)
-	}
+		/* accessing the state Group while not on the eventloop is racy
+		if s := fs.vu.State(); s.Group.Path != "" {
+			l = l.WithField("group", s.Group.Path)
+		}
+		*/
```
ankur22, line 522: I'm assuming this wouldn't be an issue if whatever action that was used was running on the event loop i.e. was async?

inancgumus, line 522: Should we remove the comment?

```suggestion
```

common/frame_session.go
```diff
@@ -688,9 +690,11 @@ func (fs *FrameSession) onLogEntryAdded(event *cdplog.EventEntryAdded) {
...
 		WithField("browser_source", event.Entry.Source.String()).
 		WithField("line_number", event.Entry.LineNumber)
-	if s := fs.vu.State(); s.Group.Path != "" {
-		l = l.WithField("group", s.Group.Path)
-	}
+		/* accessing the state Group while not on the eventloop is racy
+		if s := fs.vu.State(); s.Group.Path != "" {
+			l = l.WithField("group", s.Group.Path)
+		}
+		*/
```
inancgumus, line 697: Should we remove the comment?

```suggestion
```

common/frame_session.go
mstoykov: yes if it was *on* the event loop it would've been fine as a `group()` call couldn't have changed the value ... as it also needs to run on the event loop.

> was async? 

async is way too generic term ... in practice it *is* async - it doesn't run syncrhonisly ;) It also doesn't run on the event loop - which is in general the idea of async ... anything. To free the event loop to do other stuff and as they don't run on the event loop (and consequently on/by the JS VM) you can do multiple stuff in parallel (or at least concurrently). The problem is that IMO "async" is used too generically. 

From what I could gather, this is just something that *always* ran asynchronously. And maybe the addition of *async* APIs made it more likely to race ... or maybe it always was possible to race - I haven't tested with a really old version of xk6-browser and `-race` :shrug: 

But yes - if we can make this code ot run on the event loop - that will make it possible to keep the lines, which is why I just commented them. Unfortunately that (IMO) will require that most if not all of the APIs are async as otherwise this code will not run if the JS VM is currently blocked on doing an *synchronous* call.

I also am not certain that this (or some of the surrounding) code does not *unblock* a certain syncrhnous call that will otherwise not continue. 

Not to speak that making this run on the event loop will be *way* bigger change, which is why I opted for dropping some info so I get rid of this race condition instead of doing something way more involved.

ankur22: I'm happy with this approach for now.

ankur22: I'd prefer to have them left commented out and maybe even link to a issue so that we know why it has been commented out and what it will take to uncomment them.

inancgumus: > if we can make this code ot run on the event loop - that will make it possible to keep the lines, which is why I just commented them

👍

ankur22 (approved):

inancgumus (approved):
LGTM, but I believe we need a new issue to add `Group` support back in the xk6-browser.

mstoykov:
>  a new issue to add Group support back in the xk6-browser.

what do you mean?


the tag `group` is still added - the code that is removed here didn't need to exist. 

Do you want an issue to make the logging of `group` to be brought back for the two cases where it was removed?

inancgumus:
> Do you want an issue to make the logging of `group` to be brought back for the two cases where it was removed?

Never mind, I'm confused. It's good as it is 👍

PR #594, opened 2022-10-18
Add frame lifecycle event listener

inancgumus:
Totally experimental, probably a super dumb idea, and just for POC to concretely show what I mean [here](https://github.com/grafana/xk6-browser/pull/578#issuecomment-1282335614). I'm not expecting this to be merged.

We also need to handle shutting down the goroutine when a `Frame` is detached, etc.

We could probably do it without a goroutine, but I didn't explore that.


inancgumus:
Since the architecture is based on a pubsub event system, IMO tracking the lifecycle events is a proper solution for the current architecture. However, the current architecture and the solution in this PR is complicated, it would be better to write tests first and then refactor the architecture with a simpler one.

Issue #566, opened 2022-10-05
Error when initializing networking: network.EnableParams


inancgumus:
[This CI job](https://github.com/grafana/xk6-browser/actions/runs/3189218790/jobs/5202778121) failed with the following error:

```
time="2022-10-05T10:59:35Z" level=error msg="communicating with browser: websocket: close 1006 (abnormal closure): unexpected EOF" category=cdp elapsed="0 ms" goroutine=25
panic: creating a new page: initializing networking *network.EnableParams: websocket: close 1006 (abnormal closure): unexpected EOF

goroutine 28 [running]:
go.k6.io/k6/js/common.Throw(0x13c8720?, {0x187c860?, 0xc0011b1c80?})
	go.k6.io/k6@v0.40.1-0.20220927101031-b5bbf7e96050/js/common/util.go:20 +0x55
github.com/grafana/xk6-browser/k6ext.Panic({0x188b3f8, 0xc0018c7fb0}, {0x1618379, 0x17}, {0xc002605c70?, 0x1, 0x1})
	github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/k6ext/panic.go:51 +0x2c6
github.com/grafana/xk6-browser/common.(*Browser).onAttachedToTarget(0xc0005ea300, 0xc001187f60)
	github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/browser.go:303 +0x6af
github.com/grafana/xk6-browser/common.(*Browser).initEvents.func1()
	github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/browser.go:195 +0x337
created by github.com/grafana/xk6-browser/common.(*Browser).initEvents
	github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/common/browser.go:180 +0x19e
```

Update: This started happening a lot. See: [1](https://github.com/grafana/xk6-browser/actions/runs/3190315932/jobs/5205251760).

Maybe related: #558, #557, #545, #510 — evaluation needed.

imiric:
Strange, I haven't seen that one before.

It seems to only happen in that PR, though. Are we sure that it's not caused by the Go or dependency updates? cc: @mstoykov

Have you guys seen it on `main`? Let's not be quick about creating issues if they only happen in a PR...

inancgumus:
Seems like a Github issue, closing (for now).

inancgumus:
[...]

PR #598, opened 2022-10-19
Throw first browser error from stderr

### LINKED ISSUE #363: Allocate browser: parse websocket url: file already closed
It seems we have a brand new bug:

"panic: cannot allocate browser: cannot parse websocket url: scanner err: read |0: file already closed [recovered]"

IDK whether we can fix this problem as it seems to happen when we can't launch a browser. It might be possible that Github has a glitch.

---

[From the log](https://github.com/grafana/xk6-browser/runs/6705682654?check_suite_focus=true):

```
=== RUN   TestPageInputSpecialCharacters
--- FAIL: TestPageInputSpecialCharacters (0.09s)
panic: cannot allocate browser: cannot parse websocket url: scanner err: read |0: file already closed [recovered]

goroutine 14162 [running]:
...
github.com/grafana/xk6-browser/chromium.(*BrowserType).Launch(0xc0001efdc0, 0x1d2f[698](https://github.com/grafana/xk6-browser/runs/6705682654?check_suite_focus=true#step:8:699), 0xc0007e0a50, 0x0, 0x0)
	/home/runner/work/xk6-browser/xk6-browser/chromium/browser_type.go:143 +0xc4e
github.com/grafana/xk6-browser/tests.newTestBrowser(0x1d30918, 0xc000b07080, 0x0, 0x0, 0x0, 0xf)
	/home/runner/work/xk6-browser/xk6-browser/tests/test_browser.go:127 +0x66f
github.com/grafana/xk6-browser/tests.TestPageInputSpecialCharacters(0xc000b07080)
	/home/runner/work/xk6-browser/xk6-browser/tests/page_test.go:324 +0x73
testing.tRunner(0xc000b07080, 0x1b72b48)
...
```

imiric:
This ensures we don't log a cryptic `file already closed` error if the browser process exits prematurely, and instead try to read stderr and return the first error message found. For example, when running a headed test on Linux without a display server, we will now log `launching browser: Missing X server or $DISPLAY`.

I did an alternative fix in [`fix/363-file-already-closed-v2`](https://github.com/grafana/xk6-browser/commits/fix/363-file-already-closed-v2) (c439edb), which intercepts writes to stderr and returns either a WS URL or error over a channel. I think this is cleaner as it avoids the separate `parseDevToolsURL()` and goroutine, and [the `select` makes it very clear what we're expecting](https://github.com/grafana/xk6-browser/commit/c439edb0d18d18f3fcaba85f570b9bdd44d4d4ef#diff-1d503ce53003467b3e0ae0b563e116b4b9c1789696ee04a2e7ee1979ebc6abd0R55-R62), but this version has some performance issues as the test suite takes 4x longer on my machine, so there seems to be some blocking I didn't look into. I can dig into it further if someone else prefers this approach. Otherwise we can just merge this, which would be safer.

Closes #363


mstoykov:
I would really prefer if we can actually output the whole stdout and stderr from the browser to the user. Maybe behind an option, but that will be useful and from my expriments it is pretty silent unless you tell it to be verbose. 

But this is enough for now, just IMO we should expand it.

tests/browser_test.go
```diff
@@ -141,3 +142,23 @@ func TestBrowserUserAgent(t *testing.T) {
...
+		gobj, _ := robj.(*goja.Object)
+		err, ok := gobj.Export().(error)
+		require.Truef(t, ok, "recovered object wasn't an error, but %T", err)
+		assert.Equal(t, "launching browser: Invalid devtools server port", err.Error())
+	}()
+
+	lopts := defaultLaunchOpts()
+	lopts.Args = []any{"remote-debugging-port=99999"}
+
+	newTestBrowser(t, lopts, withLogCache())
```
imiric, line 161: `withLogCache` is not needed; leftover from some experiments.

```suggestion
	newTestBrowser(t, lopts)
```

imiric:
Agree @mstoykov, that would be good. From my tests I didn't see anything written to stdout, but it would be good to see the full output of both. It may happen that just getting a single line from stderr won't capture the full message, though I haven't run into that yet.

Do you want to create an issue for this?

common/browser_process.go
```diff
@@ -166,28 +178,43 @@ func execute(
...
+			// workaround. We need to use regex, unfortunately, to extract the
+			// error message properly. It would be awkward to do otherwise, and
+			// this isn't a hot path, so it shouldn't impact performance.
+			for _, submatches := range errRx.FindAllSubmatchIndex([]byte(line), -1) {
+				errM = errRx.Expand(errM, errTmpl, []byte(line), submatches)
+			}
+			if len(errM) > 0 {
+				c <- result{"", errors.New(string(errM))}
+				return
+			}
```
inancgumus, line 217: Since the Chromium errors have [defined patterns](https://source.chromium.org/chromium/chromium/src/+/main:base/logging.h;drc=4be70d0a54a18a36221e3733fd1dea827fc5f7ad;l=395) and never use `:ERROR:` in the rest of the message, could we use `strings` to extract the error?

```suggestion
		const urlPrefix = "DevTools listening on "
		scanner := bufio.NewScanner(cmd.stderr)

		for scanner.Scan() {
			line := scanner.Text()
			if strings.HasPrefix(line, urlPrefix) {
				c <- result{
					strings.TrimPrefix(strings.TrimSpace(line), urlPrefix),
					nil,
				}
				return
			}
			if strings.Contains(line, ":ERROR:") {
				if i := strings.Index(line, "] "); i > 0 {
					c <- result{"", errors.New(line[i+2:])}
					return
				}
			}
		}
```

inancgumus (approved):
I think this is much better 👏  Some suggestions.

tests/browser_test.go
```diff
@@ -141,3 +142,23 @@ func TestBrowserUserAgent(t *testing.T) {
...
+	t.Parallel()
+
+	defer func() {
+		robj := recover()
+		require.IsType(t, &goja.Object{}, robj)
+		gobj, _ := robj.(*goja.Object)
+		err, ok := gobj.Export().(error)
+		require.Truef(t, ok, "recovered object wasn't an error, but %T", err)
+		assert.Equal(t, "launching browser: Invalid devtools server port", err.Error())
+	}()
```
inancgumus, line 156: We can use Mihail's recent assert func here:

```suggestion
	defer func() {
		assertPanicErrorContains(t, recover(), "launching browser: Invalid devtools server port")
	}()
```

ankur22 (approved):
Looks good. Inanc raised some good points though.

common/browser_process.go
```diff
@@ -38,23 +39,26 @@ func NewBrowserProcess(
...
 	if err != nil {
-		return nil, fmt.Errorf("getting DevTools URL: %w", err)
+		procCtxCancel()
+		return nil, err
 	}
 
 	p := BrowserProcess{
-		ctx:                        ctx,
+		ctx:                        procCtx,
 		cancel:                     ctxCancel,
```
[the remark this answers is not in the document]
imiric: I didn't think about it much TBH, but looking at it now, we want `BrowserProcess.ctx` to be a new context associated with the process, in case the caller depends on this specific context being done. This doesn't seem to be the case now, but it should be a separate context only associated to the browser process.

And we're setting the parent context's cancel function, because, well, that cancels the parent context, which interrupts the iteration, etc. If this were `procCtxCancel`, then only the process itself would be cancelled.

common/browser_process.go
imiric: We could, but I still find the regex safer for parsing this. If this was a performance concern, I would agree, and would probably done it that way to begin with, but I think it's more important that we capture the relevant information correctly, than be a few microseconds faster.

Or do you have another reason to change this?

inancgumus: My reasons were (IMHO) string matching is simpler, easy to read, probably slightly faster, and has less code.

> We could, but I still find the regex safer for parsing this.

So I'm not sure if the regex is safer 😒 For example, let's say the line contains the following:

```
"launching browser: Invalid devtools server [1020/141559.181070:ERROR:headless_shell.cc(771)] port"
```

* The regex captures the error as follows: `"launching browser: port"`
* However, the string match above captures it as this: `"launching browser: Invalid devtools server [1020/141559.181070:ERROR:headless_shell.cc(771)] port"`.

common/browser_process.go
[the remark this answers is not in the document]
inancgumus: I guess that makes sense 👍

common/browser_process.go
inancgumus: It's totally up to you 😄

tests/browser_test.go
imiric: Resolved by 9e4850b.

common/browser_process.go
imiric: Resolved by 64e2810.

tests/browser_test.go
imiric: Resolved by f1a6a6d.

imiric:
I'm starting to think this introduced a regression, since `main` builds have failed 3 times now. Some with #566, and the latest one with:

```
Error:      	"canceled" does not contain "launching browser: Invalid devtools server port"
103
Test:       	TestBrowserCrashErr
```

I'll try reverting just e35e88c9, since the other changes shouldn't be relevant.

imiric:
I'm seeing another possible regression from this.

Since a couple of days ago, possibly after a system update (Linux), I'm not able to launch headed tests. They fail with:

```
ERRO[0000] GoError: launching browser: Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
```

This is probably some `dbus` misconfiguration on my machine that I don't care about fixing, but the interesting thing is that this is _not_ a fatal error. I.e. if I ignore this and continue to parse `stderr`, Chromium will output the DevTools URL and start correctly, allowing me to run a headed test without issues.

Considering there might be other non-fatal errors, this introduces a needless abort of the test when it could actually continue.

This is my full `stderr`:

```
(chromium:23400): Gtk-WARNING **: 11:54:55.838: Theme parsing error: gtk.css:3611:22: 'none' is not a valid color name
[23400:23418:1028/115455.877614:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[23400:23418:1028/115455.877661:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[23400:23418:1028/115455.898207:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[23400:23418:1028/115455.898231:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")

DevTools listening on ws://127.0.0.1:41315/devtools/browser/d1d3f8eb-b362-4f12-9370-bd25778d0da7
```

So I think we need to be smarter about this, and possibly introduce a timeout for getting the DevTools URL, and only if that is expired, then return the first error found on `stderr`.

@grafana/k6browserdev WDYT?

PR #599, opened 2022-10-20
Revert "Use context cancellation for signaling that browser process is done"

imiric:
This reverts commit e35e88c949757d56b707d88ad0f4e749f44352d8.

It seems the previous change might've been causing some flakiness issues. See https://github.com/grafana/xk6-browser/issues/566#issuecomment-1285672600 and https://github.com/grafana/xk6-browser/pull/598#issuecomment-1285750715 .

Using a context instead of a channel was cleaner, and while on the surface shouldn't have any functional differences, the context cancellation might be time sensitive, so a channel should be safer.

### DIFF
--- a/common/browser.go
+++ b/common/browser.go
@@ -429,7 +429,7 @@ func (b *Browser) Close() {
 	// forcefully after the timeout.
 	timeout := time.Second
 	select {
-	case <-b.browserProc.ctx.Done():
+	case <-b.browserProc.processDone:
 	case <-time.After(timeout):
 		b.logger.Debugf("Browser:Close", "killing browser process with PID %d after %s", b.browserProc.Pid(), timeout)
 		b.browserProc.Terminate()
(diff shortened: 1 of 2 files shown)


ankur22 (approved):

imiric:
#566 happened again :disappointed: Maybe the whole thing is racy? I really don't see how anything else in #598 could be causing this. Unless it's GitHub being flaky again. :man_shrugging:

Any suggestions?

ankur22:
> Any suggestions?

Can you run the CI job several times, and compare those results with running the CI against `main`? Which ever one encounters the least errors we go with 🙂 

I think we should reopen the issue, and dedicate some time in the next cycle to it -- key thing being trying to reproduce it locally.

inancgumus:
Hey @imiric, I haven't reviewed this yet because it seems one of the E2E tests is failing.

What's the plan for this PR?

imiric:
The plan was to see if reverting e35e88c would improve the flakiness situation, but judging by the CI failure, I'm not sure it did. Plus I got sidetracked looking into #566 instead.

Don't worry about this for now, we can decide in a few days what to do with it.

imiric:
I needed this for #617, and I think it's overall safer than the previous context cancellation approach, so @inancgumus you can review it now. Although it's just a reversion; it shouldn't have any effect on the existing flakiness.

inancgumus (approved):

PR #606, opened 2022-10-25
Add iteration ID field to log events

imiric:
This adds an `iteration_id` field to all log events that uniquely identifies a k6 iteration, and thus a single browser start/stop cycle. I'm finding it relatively useful when testing issues like #566 that only happen when stress testing the entire `tests` package, and not a single test in isolation, since I can filter log events for a specific iteration.

It's only added to events when `XK6_BROWSER_LOG` is set to `debug` or `trace`.

Related to #422

### DIFF
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -118,7 +118,7 @@ func (b *BrowserType) initContext() context.Context {
 	ctx := k6ext.WithVU(b.vu.Context(), b.vu)
 	ctx = k6ext.WithCustomMetrics(ctx, b.k6Metrics)
 	ctx = common.WithHooks(ctx, b.hooks)
-	ctx = common.WithTraceID(ctx, fmt.Sprintf("%x", b.randSrc.Uint64()))
+	ctx = common.WithIterationID(ctx, fmt.Sprintf("%x", b.randSrc.Uint64()))
 	return ctx
 }
 
@@ -375,7 +375,7 @@ func makeLogger(ctx context.Context, launchOpts *common.LaunchOptions) (*log.Log
 	var (
 		k6Logger            = k6ext.GetVU(ctx).State().Logger
 		reCategoryFilter, _ = regexp.Compile(launchOpts.LogCategoryFilter)
-		logger              = log.New(k6Logger, common.GetTraceID(ctx), launchOpts.Debug, reCategoryFilter)
+		logger              = log.New(k6Logger, common.GetIterationID(ctx), launchOpts.Debug, reCategoryFilter)
 	)
 
 	// set the log level from the launch options (usually from a script's options).
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -375,7 +375,7 @@ func makeLogger(ctx context.Context, launchOpts *common.LaunchOptions) (*log.Log
 	var (
 		k6Logger            = k6ext.GetVU(ctx).State().Logger
 		reCategoryFilter, _ = regexp.Compile(launchOpts.LogCategoryFilter)
-		logger              = log.New(k6Logger, launchOpts.Debug, reCategoryFilter)
+		logger              = log.New(k6Logger, common.GetTraceID(ctx), launchOpts.Debug, reCategoryFilter)
 	)
 
 	// set the log level from the launch options (usually from a script's options).
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -11,6 +12,7 @@ import (
 	"runtime"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/grafana/xk6-browser/api"
 	"github.com/grafana/xk6-browser/common"
@@ -39,6 +41,7 @@ type BrowserType struct {
 	k6Metrics *k6ext.CustomMetrics
 	execPath  string       // path to the Chromium executable
 	storage   *storage.Dir // stores temporary data for the extension and user
+	randSrc   *rand.Rand
 }
 
 // NewBrowserType registers our custom k6 metrics, creates method mappings on
@@ -57,6 +60,7 @@ func NewBrowserType(vu k6modules.VU) api.BrowserType {
 		hooks:     hooks,
 		k6Metrics: k6m,
 		storage:   &storage.Dir{},
+		randSrc:   rand.New(rand.NewSource(time.Now().UnixNano())), //nolint: gosec
 	}
 	rt.SetFieldNameMapper(common.NewFieldNameMapper())
 
@@ -114,6 +118,7 @@ func (b *BrowserType) initContext() context.Context {
 	ctx := k6ext.WithVU(b.vu.Context(), b.vu)
 	ctx = k6ext.WithCustomMetrics(ctx, b.k6Metrics)
 	ctx = common.WithHooks(ctx, b.hooks)
+	ctx = common.WithTraceID(ctx, fmt.Sprintf("%x", b.randSrc.Uint64()))
 	return ctx
 }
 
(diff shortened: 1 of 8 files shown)


common/context.go
```diff
@@ -2,15 +2,21 @@ package common
...
 
 type ctxKey int
 
 const (
 	ctxKeyLaunchOptions ctxKey = iota
 	ctxKeyHooks
+	ctxKeyTraceID
 )
 
+var rnd = rand.New(rand.NewSource(time.Now().UnixNano())) //nolint
```
imiric, line 18: I'm not happy with this global, but I don't think we need to pass it somewhere from `BrowserType`, and it would be slightly awkward, so I don't mind it much.

Though I'm currently running into an issue on this branch after a few minutes of stress testing that ends with:

```
panic: GoError: launching browser: inotify_init() failed: Too many open files (24)
```

... which is strange, since a) my `ulimit -n` is set to `50000`, and b) I don't see how the changes here could cause it, unless this line does it somehow. Initially I had this inside `WithTraceID()`, and thought that might be the issue, but I moved it to a global, and still see the error. I haven't run into it on `main`, so I'm skeptical it's something here.

Any thoughts?

imiric: OK, I moved it to `NewBrowserType()` in f01e97e.

Still, please (stress) test this and let me know if you run into any issues.

common/context.go
```diff
@@ -23,6 +24,21 @@ func GetHooks(ctx context.Context) *Hooks {
 	return v.(*Hooks)
 }
 
+// WithTraceID adds a random unique hexadecimal trace ID to the context.
```
imiric, line 27: I guess this no longer makes sense... :face_exhaling:

inancgumus:
Cool! I like this idea. I'll review it tomorrow.

By the way, IMO, `trace_id` does not convey the meaning that each trace is for an iteration. Should we call it `iteration` or something more descriptive? 🤔

Asking out of curiosity, what is the reason we don't just increment a counter for each iteration like `1`, `2`, N?

imiric:
> By the way, IMO, trace_id does not convey the meaning that each trace is for an iteration. Should we call it iteration or something more descriptive?

Good point. I suppose we could, though I wanted to leave this open for other use cases besides tracking iterations. Though if we were to add another tracing ID then it should definitely be called something more unique. So, sure, I'll rename it to `iteration_id` instead.

> Asking out of curiosity, what is the reason we don't just increment a counter for each iteration like 1, 2, N?

I briefly considered exposing the [`scenario.iterationInTest`](https://k6.io/docs/javascript-api/k6-execution/#scenario) value, but besides it being problematic to access (I'm not sure if extensions are able to, without some goja hacks... @mstoykov ?), it's not unique across scenarios, nor across distributed test runs. So while sequential IDs could possibly work for local execution, random hex IDs are safer against collision for distributed test runs.

tests/browser_test.go
```diff
@@ -153,3 +153,34 @@ func TestBrowserCrashErr(t *testing.T) {
...
+	tb.logCache.mu.RLock()
+	defer tb.logCache.mu.RUnlock()
+	for _, evt := range tb.logCache.entries {
+		for k, v := range evt.Data {
+			if k == "trace_id" {
+				require.IsType(t, "", v)
+				require.NotEmpty(t, v)
+				tracedEvts++
+				if traceID == "" {
+					traceID, _ = v.(string)
```
ankur22: Is there no way to get the `traceID` from the context, and then compare that against the `traceID` from the log lines?

ankur22: If there isn't a way to retrieve the `traceID` from the context, then should we ensure that the `v.(string)` itself isn't an empty string?

imiric: Yeah, sure, that seems slightly better :+1:

ankur22 (approved):
Nice! This would've been very useful when debugging the issues i've worked on recently 🙂 

Maybe worth renaming it to `iterationID`, since `traceID` is generally the term used in microservices to trace a request through a system... actually, i guess it's being used in the same way, so `traceID` might be ok. I wonder if there's a way to feed the logs through Tempo 🤔 😄

tests/browser_test.go
imiric: Resolved by 7642c45.

imiric:
@ankur22 Yeah, I renamed it to `iteration_id` after Inanc mentioned it, though "trace ID" is a fairly generic term and depends on the context of what you want to trace. :) Essentially, any log field that's unique and can be used to aggregate a bunch of events can be thought of as a trace ID.

I do think we'll want to expand this if we want to fully implement #422, so that we can trace unique components and actions. That's why I was somewhat reluctant to rename everything here to `iteration_id`, since we probably don't want to add N args to `log.New()` for each traced field, but maybe pass it a struct or context with a few of these trace IDs. In any case, that's something to think about later. :)

common/context.go
imiric: Resolved by 95271e1.

common/context.go
```diff
@@ -9,6 +9,7 @@ type ctxKey int
 const (
 	ctxKeyLaunchOptions ctxKey = iota
 	ctxKeyHooks
+	ctxKeyTraceID
```
inancgumus, line 12: Should we be consistent with the logger field name?

```suggestion
	ctxKeyIterationID
```

common/context.go
```diff
@@ -23,6 +24,21 @@ func GetHooks(ctx context.Context) *Hooks {
...
+}
+
+// GetIterationID returns the iteration identifier attached to the context.
+func GetIterationID(ctx context.Context) string {
+	v := ctx.Value(ctxKeyTraceID)
+	val, ok := v.(string)
+	if v == nil || !ok {
+		return ""
+	}
+	return val
```
inancgumus, line 39: The following code will do the same thing:

```suggestion
	s, _ := ctx.Value(ctxKeyIterationID).(string)
	return s // it will return "" on error
```

log/logger.go
```diff
@@ -95,11 +97,15 @@ func (l *Logger) Logf(level logrus.Level, category string, msg string, args ...a
...
-	entry := l.WithFields(logrus.Fields{
+	fields := logrus.Fields{
 		"category":  category,
 		"elapsed":   fmt.Sprintf("%d ms", elapsed),
 		"goroutine": goRoutineID(),
-	})
+	}
+	if l.GetLevel() > logrus.InfoLevel {
+		fields["iteration_id"] = l.iterID
+	}
```
inancgumus, line 107: This will print `iteration_id=` if `l.iterID` is empty. Not a big deal, but should we do the following?

```suggestion
	if l.iterID != "" && l.GetLevel() > logrus.InfoLevel {
		fields["iteration_id"] = l.iterID
	}
```

So we can skip seeing the `iteration_id` in the logs if it's empty.

inancgumus (approved):
LGTM. I have some non-blocking suggestions. Feel free to apply them or not 👍

common/context.go
imiric: Ah, thought I renamed them all. Sure :+1:

chromium/browser_type.go
```diff
@@ -57,6 +60,7 @@ func NewBrowserType(vu k6modules.VU) api.BrowserType {
 		hooks:     hooks,
 		k6Metrics: k6m,
 		storage:   &storage.Dir{},
+		randSrc:   rand.New(rand.NewSource(time.Now().UnixNano())), //nolint: gosec
```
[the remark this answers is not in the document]
imiric: Hhmm but why would it matter that iteration IDs are deterministic? It's not like issues would only happen on a specific iteration, so how would that be useful? We would also get collisions in distributed tests runs, so I don't think this is a good idea.

common/context.go
imiric: Resolved by a65d830.

common/context.go
imiric: Thanks! We should probably update all the other getters as well then.

Resolved by fd3e409.

log/logger.go
imiric: Resolved by a7e79ce.

PR #615, opened 2022-10-28
Optimize TestPageScreenshotFullpage

inancgumus:
This PR is a demonstration of a simple optimization and its effects.

`TestPageScreenshotFullpage` is the slowest test we have (see #612). So it's a low-hanging fruit to optimize. A quick fix sped up the test by 35% on my machine.

```
Before: 671259438 ns/op 43410268 B/op 8771 allocs/op
After : 363529042 ns/op  5970741 B/op 8533 allocs/op
```

Updates #612

### DIFF
--- a/tests/page_test.go
+++ b/tests/page_test.go
@@ -399,7 +399,7 @@ func TestPageScreenshotFullpage(t *testing.T) {
 
 		const div = document.createElement('div');
 		div.style.width = '1280px';
-		div.style.height = '8000px';
+		div.style.height = '800px';
 		div.style.background = 'linear-gradient(red, blue)';
 
 		document.body.appendChild(div);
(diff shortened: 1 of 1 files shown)


tests/page_test.go
```diff
@@ -399,7 +399,7 @@ func TestPageScreenshotFullpage(t *testing.T) {
 
 		const div = document.createElement('div');
 		div.style.width = '1280px';
-		div.style.height = '8000px';
+		div.style.height = '800px';
```
ankur22, line 402: Can we not lower these resolutions even further? what's stopping us from using 800x600? Or even 1x1?

ankur22 (approved):

tests/page_test.go
ankur22: I get a error on the last assert when i change it 800x600 🤷

imiric (approved):

PR #616, opened 2022-10-28
Browser should not override defaults with null values

### LINKED ISSUE #407: Browser should not override defaults with null values
If the executablePath is set to null, then chromium will not be found. My expectation would be that null should 
imply default behavior.

```
export default function() {
        executablePath: null,       // Override search for browser executable in favor of specified absolute path
    });
```

inancgumus:
Fixes #407 by ignoring **any options** with `null` or `undefined` values given by a test script.

For example:
* Setting `executablePath` to `null` will be ignored, and its default value will be used.
* Or, setting `headless` to `undefined` will be ignored, and `headless` will stay as `true` ([its default value](https://github.com/grafana/xk6-browser/blob/4a47b35e132d6d188e80c7543686c3665f4c942b/common/browser_options.go#L46)).
* Setting `args` to `null` will be ignored, and `args` will stay `nil` (its default).

---

WDYT about returning an error when an option is set to `null` or `undefined`?

**Current approach.**
* Pros: Allows defaults to be used when one of the options is set to `null`. This can also be a double-edged sword (cons-es below). Slightly faster parsing (since it skips the `null`/unset properties).
* Cons: It can be confusing for users to get default behavior even if they deliberately want to set an option to `null`. IDK when this can happen, though.

**Returning an error on null**
* Pros: Strict behavior and lets users know an option should not be set to `null`. One way of doing things.
* Cons: Users can't set an option to `null`. Might there edge cases exist where users want to intentionally set an option to `null` (not only specific to `BrowserType.Launch`)?

### DIFF
--- a/common/browser_options.go
+++ b/common/browser_options.go
@@ -60,10 +60,11 @@ func (l *LaunchOptions) Parse(ctx context.Context, opts goja.Value) error { //no
 		o  = opts.ToObject(rt)
 	)
 	for _, k := range o.Keys() {
-		var (
-			err error
-			v   = o.Get(k)
-		)
+		v := o.Get(k)
+		if v.Export() == nil {
+			continue // don't override the defaults on `null``
+		}
+		var err error
 		switch k {
 		case "args":
 			err = exportOpt(rt, k, v, &l.Args)
--- a/common/browser_options_test.go
+++ b/common/browser_options_test.go
@@ -30,6 +30,23 @@ func TestBrowserLaunchOptionsParse(t *testing.T) {
 				}, lo)
 			},
 		},
+		"nulls": { // don't override the defaults on `null`
+			opts: map[string]any{
+				"env":               nil,
+				"headless":          nil,
+				"logCategoryFilter": nil,
+				"timeout":           nil,
+			},
+			assert: func(tb testing.TB, lo *LaunchOptions) {
+				tb.Helper()
+				assert.Equal(tb, &LaunchOptions{
+					Env:               make(map[string]string),
+					Headless:          true,
+					LogCategoryFilter: ".*",
+					Timeout:           DefaultTimeout,
+				}, lo)
+			},
+		},
 		"args": {
 			opts: map[string]any{
 				"args": []any{"browser-arg1='value1", "browser-arg2=value2", "browser-flag"},


ankur22 (commented):
My vote would be to be more explicit when a user has passed in an invalid value, so rather than assuming default values, return an error.

Worth also looking at what k6 does with its options.

inancgumus:
I looked into how k6 handles options.

When you give an `undefined` or a `null` value to an option, it uses the defaults (mostly!).

IMO, we also should not override the defaults. WDYT? @grafana/k6browserdev 

**Example script:**

```js
import http from 'k6/http';

export const options = {
  vus: -1,           // defaults to 1
  // vus: undefined, // defaults to 1
  // vus: 0,         // err: the number of VUs must be more than 0
  // vus: 'A',       // err: invalid syntax
  // iterations: 0,  // err: can't be less than the number of VUs

  teardownTimeout: undefined,    // defaults to 1m
  // teardownTimeout: null,      // defaults to 1m
  // teardownTimeout: '-10s',    // -10s: overrides the default!

  userAgent: null,         // defaults to k6/0.40.0
  // userAgent: undefined, // default to k6/0.40.0
  // userAgent: 1,         // err: cannot unmarshall
  // userAgent: 'MyK6UserAgentString/1.0',

  summaryTrendStats: undefined, // defaults to avg,min,med,max,p(90),p(95)
  summaryTrendStats: null,      // defaults to avg,min,med,max,p(90),p(95)

  maxRedirects: 0,   // 0: overrides the default (10)
  maxRedirects: -10, // 0: overrides the default (10)

  httpDebug: 'full', // for testing purposes
};

export default function () {
  http.get('https://httpbingo.org/absolute-redirect/10');
}
```

common/browser_options.go
```diff
@@ -60,10 +60,11 @@ func (l *LaunchOptions) Parse(ctx context.Context, opts goja.Value) error { //no
 		o  = opts.ToObject(rt)
 	)
 	for _, k := range o.Keys() {
-		var (
-			err error
-			v   = o.Get(k)
-		)
+		v := o.Get(k)
+		if v.Export() == nil {
+			continue // don't override the defaults on `null``
```
imiric, line 65: Let's log a warning that we're falling back to a default value, and, if possible, what that value is.

imiric (changes requested):
To be fair, k6 option handling is a mess and internally inconsistent (see https://github.com/grafana/k6/issues/883 :sweat:), so we won't be able to derive consistent behavior that we can follow... The best we can do is to cherry pick some good practices and do the same in xk6-browser.

Since we're not using nullable types yet, we can't determine whether `null`/`undefined` was specified by the user, and if so, whether it makes sense as a valid value for some options and how to handle it.

The main question here is about throwing a validation error or falling back to the default value.

So while we're unable to determine that, I think we should keep this fallback to default, just make sure that we log a warning for it. And let's create an issue for using nullable types, so that we can properly validate it.

inancgumus:
> The best we can do is to cherry pick some good practices and do the same in xk6-browser.

Makes sense. I wonder what those good practices are. Skip this if it's in the k6 issue you shared. I haven't checked it out yet :-)

> Let's create an issue for using nullable types, so that we can properly validate it.

Since you're more knowledgeable about it, would you mind creating the issue? 🙇

common/browser_options.go
inancgumus: Welcome to the chain of PRs: #608 -> #616 -> Logger PR. These two are much better than the current `main`. I don't see any issues merging them. I'll send the next PR (logging) tomorrow.

imiric:
> I'd like to learn more about those good practices

The issue mentions the problems we have, but you can pickup any best practices from browsing the code and the documentation; I certainly don't have a specific list in mind :slightly_smiling_face:. Nullable types is one, as well as having a consistent [order of precendence](https://k6.io/docs/using-k6/k6-options/how-to#order-of-precedence), which is a bit limiting for us, since we can't define new CLI args, but do rely on JS options and env vars.

Sure, I'll create the issue.

common/browser_options.go
inancgumus: Please see #620.

imiric (approved):

ankur22 (approved):

PR #617, opened 2022-10-28
Wait for browser process to exit before returning first error from stderr

### LINKED ISSUE #614: Failure to run headed test after non-fatal Chromium error
### Brief summary

After #598, and since a couple of days ago, possibly after a system update (Linux), I'm not able to launch headed tests on `main` (47bdfb0). They fail with:

```
ERRO[0000] GoError: launching browser: Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
```

This is probably some `dbus` misconfiguration on my machine that I don't care about fixing, but the interesting thing is that this is not a fatal error. I.e. if I ignore this and continue to parse stderr, Chromium will output the DevTools URL and start correctly, allowing me to run a headed test without issues.

Considering there might be other non-fatal errors, this introduces a needless abort of the test when it could actually continue.

This is my full stderr:

```
(chromium:23400): Gtk-WARNING **: 11:54:55.838: Theme parsing error: gtk.css:3611:22: 'none' is not a valid color name
[23400:23418:1028/115455.877614:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[23400:23418:1028/115455.877661:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[23400:23418:1028/115455.898207:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[23400:23418:1028/115455.898231:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")

DevTools listening on ws://127.0.0.1:41315/devtools/browser/d1d3f8eb-b362-4f12-9370-bd25778d0da7
```

## Suggested solution

We need to have a safer way of parsing stderr. Possibly introduce a timeout for getting the DevTools URL, and only if that is expired, then return the first error found on stderr.

### xk6-browser version

47bdfb07269afbe05daeefb743b9e7d95feaaf40

### OS

Linux

### Chrome version

106.0.5249.103

N/A

Start any script with `headless: false` on a machine that launches Chromium with some non-fatal error (e.g. try stopping the `dbus` service).

### Expected behaviour

The test run should proceed without issues.

### Actual behaviour

The test run is aborted.

imiric:
This ensures we don't return on the first error we see, since they could be non-fatal, but actually wait for the browser process to exit. I confirmed this fixed the issue locally for me with the `dbus` errors mentioned in #614, and I added some unit tests for `parseDevToolsURL()`.

It's based on #599, since it needed to also check if the process ended unexpectedly, and that was safer with the channel approach. So let's merge that first.

Fixes #614

### DIFF
--- a/common/browser_process.go
+++ b/common/browser_process.go
@@ -181,49 +181,69 @@ func execute(
 
 // parseDevToolsURL grabs the WebSocket address from Chrome's output and returns
 // it. If the process ends abruptly, it will return the first error from stderr.
-func parseDevToolsURL(ctx context.Context, cmd command) (string, error) { //nolint: cyclop
-	var (
-		lines = make(chan string)
-		errCh = make(chan error)
-	)
+func parseDevToolsURL(ctx context.Context, cmd command) (_ string, err error) {
+	parser := &devToolsURLParser{
+		sc: bufio.NewScanner(cmd.stderr),
+	}
+	done := make(chan struct{})
 	go func() {
-		scanner := bufio.NewScanner(cmd.stderr)
-		for scanner.Scan() {
-			lines <- scanner.Text()
+		for parser.scan() {
 		}
-		if err := scanner.Err(); err != nil {
-			errCh <- err
-		}
+		close(done)
 	}()
+	for err == nil {
+		select {
+		case <-done:
+			err = parser.err()
+		case <-ctx.Done():
+			err = ctx.Err()
+		case <-cmd.done:
+			err = errors.New("browser process ended unexpectedly")
+		}
+	}
+	if parser.url != "" {
+		err = nil
+	}
+
+	return parser.url, err
+}
+
+type devToolsURLParser struct {
+	sc *bufio.Scanner
+
+	errs []error
+	url  string
+}
+
+func (p *devToolsURLParser) scan() bool {
+	if !p.sc.Scan() {
+		return false
+	}
 
 	const urlPrefix = "DevTools listening on "
-	var stdErr error
 
-	preferStdErr := func(e error) error {
-		if stdErr != nil {
-			return stdErr
-		}
-		return e
+	line := p.sc.Text()
+	if strings.HasPrefix(line, urlPrefix) {
+		p.url = strings.TrimPrefix(strings.TrimSpace(line), urlPrefix)
 	}
-
-	for {
-		select {
-		case line := <-lines:
-			if strings.HasPrefix(line, urlPrefix) {
-				return strings.TrimPrefix(strings.TrimSpace(line), urlPrefix), nil
-			}
-			// TODO: Capture more than one error?
-			if stdErr == nil && strings.Contains(line, ":ERROR:") {
-				if i := strings.Index(line, "] "); i > 0 {
-					stdErr = errors.New(line[i+2:])
-				}
-			}
-		case err := <-errCh:
-			return "", preferStdErr(err)
-		case <-cmd.done:
-			return "", preferStdErr(errors.New("browser process ended unexpectedly"))
-		case <-ctx.Done():
-			return "", preferStdErr(fmt.Errorf("%w", ctx.Err()))
+	if strings.Contains(line, ":ERROR:") {
+		if i := strings.Index(line, "] "); i > 0 {
+			p.errs = append(p.errs, errors.New(line[i+2:]))
 		}
 	}
+
+	return p.url == ""
+}
+
+func (p *devToolsURLParser) err() error {
+	if p.url != "" {
+		return io.EOF
+	}
+	if len(p.errs) > 0 {
+		return p.errs[0]
+	}
+	if err := p.sc.Err(); err != nil {
+		return fmt.Errorf("%w", err)
+	}
+	return nil
 }
--- a/common/browser_process.go
+++ b/common/browser_process.go
@@ -179,41 +179,51 @@ func execute(
 	return command{cmd, done, stdout, stderr}, nil
 }
 
-// parseDevToolsURL grabs the websocket address from chrome's output and returns it.
-func parseDevToolsURL(ctx context.Context, cmd command) (wsURL string, _ error) {
-	type result struct {
-		devToolsURL string
-		err         error
-	}
-	c := make(chan result, 1)
+// parseDevToolsURL grabs the WebSocket address from Chrome's output and returns
+// it. If the process ends abruptly, it will return the first error from stderr.
+func parseDevToolsURL(ctx context.Context, cmd command) (string, error) { //nolint: cyclop
+	var (
+		lines = make(chan string)
+		errCh = make(chan error)
+	)
 	go func() {
-		const urlPrefix = "DevTools listening on "
 		scanner := bufio.NewScanner(cmd.stderr)
-
 		for scanner.Scan() {
-			line := scanner.Text()
-			if strings.HasPrefix(line, urlPrefix) {
-				c <- result{
-					strings.TrimPrefix(strings.TrimSpace(line), urlPrefix),
-					nil,
-				}
-				return
-			}
-			if strings.Contains(line, ":ERROR:") {
-				if i := strings.Index(line, "] "); i > 0 {
-					c <- result{"", errors.New(line[i+2:])}
-					return
-				}
-			}
+			lines <- scanner.Text()
 		}
 		if err := scanner.Err(); err != nil {
-			c <- result{"", err}
+			errCh <- err
 		}
 	}()
-	select {
-	case r := <-c:
-		return r.devToolsURL, r.err
-	case <-ctx.Done():
-		return "", fmt.Errorf("%w", ctx.Err())
+
+	const urlPrefix = "DevTools listening on "
+	var stdErr error
+
+	preferStdErr := func(e error) error {
+		if stdErr != nil {
+			return stdErr
+		}
+		return e
+	}
+
+	for {
+		select {
+		case line := <-lines:
+			if strings.HasPrefix(line, urlPrefix) {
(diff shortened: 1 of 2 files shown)


common/browser_process.go
```diff
@@ -179,41 +179,51 @@ func execute(
...
-// parseDevToolsURL grabs the websocket address from chrome's output and returns it.
-func parseDevToolsURL(ctx context.Context, cmd command) (wsURL string, _ error) {
-	type result struct {
-		devToolsURL string
-		err         error
-	}
-	c := make(chan result, 1)
+// parseDevToolsURL grabs the WebSocket address from Chrome's output and returns
+// it. If the process ends abruptly, it will return the first error from stderr.
+func parseDevToolsURL(ctx context.Context, cmd command) (string, error) { //nolint: cyclop
```
imiric, line 184: I made the goroutine lambda smaller, in order to react on `cmd.done` in the same `select`, but it caused a slight increase in cyclomatic complexity, and found it simpler to disable the linter here.

I actually think the current version is more readable, but let me know if you prefer refactoring it in some way.

ankur22 (approved):
LGTM 👍 I think it's clearer now.

imiric:
I added some unit tests for `parseDevToolsURL()`. @ankur22 Please take another look. :pray:

common/browser_process.go
```diff
@@ -179,41 +179,51 @@ func execute(
...
+	var (
+		lines = make(chan string)
+		errCh = make(chan error)
```
inancgumus, line 187: The scanner would never return when we find the devToolsURL. To be on the safe side, let's buffer each one.

```suggestion
		lines = make(chan string, 1)
		errCh = make(chan error, 1)
```

common/browser_process.go
```diff
@@ -179,41 +179,51 @@ func execute(
...
+	}()
+
+	const urlPrefix = "DevTools listening on "
+	var stdErr error
+
+	preferStdErr := func(e error) error {
+		if stdErr != nil {
+			return stdErr
+		}
+		return e
```
inancgumus, line 206: You could use `defer` instead of calling `preferStdErr` on each case below.

For example:

```go
func parseDevToolsURL(ctx context.Context, cmd command) (devToolsURL string, err error) { //nolint:cyclop
	var (
		lines = make(chan string, 1)
		errCh = make(chan error, 1)
	)
	go func() {
		scanner := bufio.NewScanner(cmd.stderr)
		for scanner.Scan() {
			lines <- scanner.Text()
		}
		if err := scanner.Err(); err != nil {
			errCh <- err
		}
	}()
	var stdErr error
	defer func() {
		if stdErr != nil && devToolsURL == "" {
			err = stdErr
		}
	}()
	for err == nil && devToolsURL == "" {
		select {
		case line := <-lines:
			const urlPrefix = "DevTools listening on "

			if strings.HasPrefix(line, urlPrefix) {
				devToolsURL = strings.TrimPrefix(strings.TrimSpace(line), urlPrefix)
			}
			// TODO: Capture more than one error?
			if stdErr == nil && strings.Contains(line, ":ERROR:") {
				if i := strings.Index(line, "] "); i > 0 {
					stdErr = errors.New(line[i+2:])
				}
			}
		case err = <-errCh:
		case <-cmd.done:
			err = errors.New("browser process ended unexpectedly")
		case <-ctx.Done():
			err = fmt.Errorf("%w", ctx.Err())
		}
	}

	return devToolsURL, err
}
```

common/browser_process.go
inancgumus, line 184: > a slight increase in cyclomatic complexity
> the current version is more readable

TBH, I had a hard time trying to understand it. Can you simplify the function further?

I'm also not comfortable adding another `nolint` (#58) while there is a chance for further refactoring.

common/browser_process.go
```diff
@@ -179,41 +179,51 @@ func execute(
...
+		lines = make(chan string)
+		errCh = make(chan error)
+	)
 	go func() {
-		const urlPrefix = "DevTools listening on "
 		scanner := bufio.NewScanner(cmd.stderr)
-
 		for scanner.Scan() {
-			line := scanner.Text()
+			lines <- scanner.Text()
```
inancgumus, line 192: Can we stop scanning after grabbing the devTools URL?

inancgumus (dismissed):
It's nice that we have brilliant tests for this issue now 👏

common/browser_process_test.go
```diff
@@ -0,0 +1,151 @@
...
+	lines []string
+	err   error
+}
+
+func (r mockReader) Read(p []byte) (n int, err error) {
+	for _, l := range r.lines {
+		n += copy(p, []byte(l+"\n"))
+	}
+	return n, r.err
+}
```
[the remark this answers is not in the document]
imiric: Thanks, you're right :+1: I didn't check what `Scanner.Text()` returned, and just assumed that `Scan()` would tokenize whatever `Read()` returned in a single call. :facepalm:

I pushed the change in 024381d, with some minor tweaks. Mainly, returning `io.EOF` to avoid the `multiple Read calls return no data or error` error.

But I'm not sure what you're suggesting with that previous code. Let's discuss the refactor in the other thread.

common/browser_process_test.go
```diff
@@ -0,0 +1,151 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			mr := mockReader{lines: tc.stderr, err: tc.readErr}
+			cmdDone := make(chan struct{})
+			cmd := command{done: cmdDone, stderr: mr}
```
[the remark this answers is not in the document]
imiric: Done in 024381d.

common/browser_process_test.go
```diff
@@ -0,0 +1,151 @@
...
+				err   error
+			)
+
+			go func() {
+				wsURL, err = parseDevToolsURL(ctx, cmd)
+				close(done)
+			}()
+
+			if tc.prematureCmdDone {
+				time.Sleep(200 * time.Millisecond)
```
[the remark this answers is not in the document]
imiric: They simulate the scenario where we get to read from stderr before the process ends prematurely. Since we have no way of synchronizing the `Scanner` goroutine with the test, a slight delay accomplishes that, albeit flakily.

Actually, these tests were incomplete, since they're not testing that we actually return the error from stderr. So there are two scenarios we should simulate:
- we read an error from stderr, and then the process ends prematurely.
- the process ends prematurely before we read an error from stderr (or there's no error written).

Also, I think there's a better way of doing this. See 833d1d0. This should be more reliable ensuring we do get to read from stderr, and then trigger the process ending some time after that. It's slightly more hacky, and it avoids `time.Sleep()`, but still uses `time.AfterFunc()` to trigger the premature process end.

Let me know if you have a better suggestion.

common/browser_process.go
imiric: > TBH, I had a hard time trying to understand it.

I sincerely doubt that :slightly_smiling_face: It's a trivial `Scanner` goroutine, and `select` below it. The refactor just simplifies the goroutine, and moves the line parsing to the `select`.

I struggled trying to simplify it, as the cyclomatic complexity comes from the `for { select {} }` loop, and the `preferStdErr()` function. The cyclomatic complexity measures the number of possible code paths; not necessarily how difficult the code is to understand. This is subjective, of course, so if you find this version more difficult to understand, and you strongly object `nolint`ing it, then I'll try to refactor it further, with some of your other suggestions.

Otherwise we can revert the refactor and keep the previous version.

common/browser_process_test.go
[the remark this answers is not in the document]
inancgumus: > I'm not sure what you're suggesting with that previous code

The previous code I shared is the one on the current `main`. I just added `<-cmd.done` to it to make it work with the test in this PR. It wasn't about refactoring.

common/browser_process.go
inancgumus: > I struggled trying to simplify it

I experimented with this and separated the parser code in #619. You can cherry-pick it if you want. This one doesn't have cyclomatic complexity.

common/browser_process.go
inancgumus: #619 fixes this issue.

imiric:
@inancgumus I tried to take some of your suggestions into account, but considering you completely rewrote this fix in #619, I'm leaving it up to you and @ankur22 what to do with this PR.

common/browser_process.go
```diff
@@ -181,49 +181,69 @@ func execute(
...
-		for scanner.Scan() {
-			lines <- scanner.Text()
-		}
-		if err := scanner.Err(); err != nil {
-			errCh <- err
+		for parser.scan() {
 		}
+		close(done)
 	}()
+	for err == nil {
```
ankur22, line 194: Why is this a for loop?

inancgumus: Hmm 🤔 Sometimes, the browser process can only return an empty string. And the scanner will return `false` (_no more lines to read_). So, the `err` from the `case <-done: err = parser.err()` will be `nil`. The loop waits until the command (`<-cmd.done`) finishes and returns with an error (or the context (<- `ctx.done`).

We're waiting for the command to exit so that we can run this:

https://github.com/grafana/xk6-browser/blob/6c3449f1d5ac697b21a66d3c7215e862ea51e919/common/browser_process.go#L162-L177

The `done` channel there is closed when we successfully delete the user data directory.

ankur22 (approved):

PR #619, opened 2022-11-01
Refactor parseDevToolsURL (#617)

inancgumus:
Separates the parsing from concurrently waiting.

Advantages:
+ Separation of concerns.
+ Parsing logic or error handling can be changed (maintained) and tested separately.
+ Simplified channel handling.
+ Scanner returns when it grabs the devToolsURL.
+ Scanner URL parsing/error handling is managed where it belongs (in the parser code).
+ Supports multiple errors.

Disadvantages:
+ Verbose.

PR #620, opened 2022-11-01
Log when a launch option is set to null

inancgumus:
It logs when one of the default browser launch options is set to either `null` or `undefined` and overridden by its default option. I needed to move `makeLogger` from `launch` to `Launch` so we could log something in `BrowserType.Launch`.

Currently, `Parse` doesn't mind the `logCategoryFilter` and `debug` options. There was a chicken-egg situation: `Parse` needs a logger, but we don't know about the `logCategoryFilter` and `debug` options until `Parse` finishes. And, `Parse` needs to log!

This should [allow](https://github.com/grafana/xk6-browser/pull/608#discussion_r1008050172) merging #608 and #619.

PS: I can squash this after the review if you want.

### DIFF
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -133,7 +133,7 @@ func (b *BrowserType) Launch(opts goja.Value) api.Browser {
 	}
 
 	launchOpts := common.NewLaunchOptions()
-	if err := launchOpts.Parse(ctx, opts); err != nil {
+	if err := launchOpts.Parse(ctx, opts, b.logger); err != nil {
 		k6ext.Panic(ctx, "parsing launch options: %w", err)
 	}
 	ctx = common.WithLaunchOptions(ctx, launchOpts)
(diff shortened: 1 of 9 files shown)


examples/locator.js
```diff
@@ -9,6 +9,7 @@ export const options = {
 export default function() {
   const browser = chromium.launch({
     headless: __ENV.XK6_HEADLESS ? true : false,
+    debug: true,
```
inancgumus: TODO: Remove this from the commit.

examples/hosts.js
```diff
@@ -5,7 +5,7 @@ export const options = {
   hosts: { 'test.k6.io': '127.0.0.254' },
   thresholds: {
     checks: ["rate==1.0"]
-  }
+  },
```
inancgumus, line 8: TODO: Remove this from the commit.

imiric (approved):

ankur22 (approved):

Issue #622, opened 2022-11-02
Fix lifecycle event waitUntil for Reload


ankur22:
The `waitUntil` options for `Reload` doesn't work when we set `waitUntil` to `networkIdle`. The unit test below currently fails on `main`:

```go
func TestLifecycleReloadNetworkIdle(t *testing.T) {
	t.Parallel()

	tb := newTestBrowser(t, withFileServer())
	p := tb.NewPage(nil)
	tb.withHandler("/home", func(w http.ResponseWriter, r *http.Request) {
		http.Redirect(w, r, tb.staticURL("reload_lifecycle.html"), http.StatusMovedPermanently)
	})

	var counter int64
	var counterMu sync.Mutex
	tb.withHandler("/ping", func(w http.ResponseWriter, _ *http.Request) {
		counterMu.Lock()
		defer counterMu.Unlock()

		time.Sleep(time.Millisecond * 30)

		counter++
		fmt.Fprintf(w, "pong %d", counter)
	})

	tb.withHandler("/ping.js", func(w http.ResponseWriter, _ *http.Request) {
		fmt.Fprintf(w, `
				var pingJSTextOutput = document.getElementById("pingJSText");
				pingJSTextOutput.innerText = "ping.js loaded from server";
			`)
	})

	waitUntil := common.LifecycleEventNetworkIdle
	assertHome(t, tb, p, waitUntil, func() {
		result := p.TextContent("#pingRequestText", nil)
		assert.EqualValues(t, "Waiting... pong 5 - for loop complete", result)

		result = p.TextContent("#pingJSText", nil)
		assert.EqualValues(t, "ping.js loaded from server", result)

		opts := tb.toGojaValue(common.PageReloadOptions{
			WaitUntil: waitUntil,
			Timeout:   30 * time.Second,
		})
		p.Reload(opts)

		result = p.TextContent("#pingRequestText", nil)
		assert.EqualValues(t, "Waiting... pong 10 - for loop complete", result)

		result = p.TextContent("#pingJSText", nil)
		assert.EqualValues(t, "ping.js loaded from server", result)
	})
}
```

reload_lifecycle.html:
```html
<html>

<head></head>

<body>
    <div id="pingRequestText">Waiting...</div>
    <div id="pingJSText">Waiting...</div>

    <script>
        var pingRequestTextOutput = document.getElementById("pingRequestText");

        var p = requestPings();
        p.then(() => {
            pingRequestTextOutput.innerText += ' - for loop complete';
[...]

PR #623, opened 2022-11-02
Fix waituntil for page.reload

ankur22:
Closes: https://github.com/grafana/xk6-browser/issues/622

The logic to `waitUntil` has been copied from `FrameManager.NavigateFrame`. The issue was that `waitUntil` when used with `networkIdle` wasn't working as expected -- the wait would end too early.

I've also refactored the previous lifecycle event tests into a new file. There will be a lot more lifecycle event tests coming in future PRs, so it's best to keep them all in one place.

### DIFF
--- a/common/page.go
+++ b/common/page.go
@@ -687,32 +687,52 @@ func (p *Page) Reload(opts goja.Value) api.Response {
 		k6ext.Panic(p.ctx, "parsing reload options: %w", err)
 	}
 
-	ch, evCancelFn := createWaitForEventHandler(p.ctx,
-		p.frameManager.MainFrame(),
-		[]string{EventFrameNavigation}, func(data any) bool {
+	timeoutCtx, timeoutCancelFn := context.WithTimeout(p.ctx, parsedOpts.Timeout)
+	defer timeoutCancelFn()
+
+	ch, evCancelFn := createWaitForEventHandler(
+		timeoutCtx, p.frameManager.MainFrame(), []string{EventFrameNavigation},
+		func(data any) bool {
 			return true // Both successful and failed navigations are considered
 		},
 	)
 	defer evCancelFn() // Remove event handler
+
+	lifecycleEvtCh, lifecycleEvtCancel := createWaitForEventPredicateHandler(
+		timeoutCtx, p.frameManager.MainFrame(), []string{EventFrameAddLifecycle},
+		func(data any) bool {
+			if le, ok := data.(LifecycleEvent); ok {
+				return le == parsedOpts.WaitUntil
+			}
+			return false
+		})
+	defer lifecycleEvtCancel()
 
 	action := cdppage.Reload()
 	if err := action.Do(cdp.WithExecutor(p.ctx, p.session)); err != nil {
 		k6ext.Panic(p.ctx, "reloading page: %w", err)
 	}
 
+	wrapTimeoutError := func(err error) error {
+		if errors.Is(err, context.DeadlineExceeded) {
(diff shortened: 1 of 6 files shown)


common/page.go
```diff
@@ -687,34 +687,54 @@ func (p *Page) Reload(opts goja.Value) api.Response {
...
+
+	ch, evCancelFn := createWaitForEventHandler(
+		timeoutCtx, p.frameManager.MainFrame(), []string{EventFrameNavigation},
+		func(data any) bool {
 			return true // Both successful and failed navigations are considered
 		},
 	)
 	defer evCancelFn() // Remove event handler
 
+	lifecycleEvtCh, lifecycleEvtCancel := createWaitForEventPredicateHandler(
```
ankur22, line 701: This is what helps us `waitUntil` the correct lifecycle event.

common/page.go
```diff
@@ -687,34 +687,54 @@ func (p *Page) Reload(opts goja.Value) api.Response {
...
+			return false
+		})
+	defer lifecycleEvtCancel()
+
 	action := cdppage.Reload()
 	if err := action.Do(cdp.WithExecutor(p.ctx, p.session)); err != nil {
 		k6ext.Panic(p.ctx, "reloading page: %w", err)
 	}
 
+	wrapTimeoutError := func(err error) error {
```
ankur22, line 716: Copied, but maybe a good candidate to be it's own function?

common/page.go
```diff
@@ -725,6 +745,12 @@ func (p *Page) Reload(opts goja.Value) api.Response {
 
 	applySlowMo(p.ctx)
 
+	select {
+	case <-lifecycleEvtCh:
```
ankur22, line 749: We're waiting for the lifecycle event.

common/page.go
```diff
@@ -723,6 +743,12 @@ func (p *Page) Reload(opts goja.Value) api.Response {
 		req.responseMu.RUnlock()
 	}
 
+	select {
+	case <-lifecycleEvtCh:
+	case <-timeoutCtx.Done():
+		k6ext.Panic(p.ctx, "%w", wrapTimeoutError(timeoutCtx.Err()))
+	}
+
 	applySlowMo(p.ctx)
```
ankur22, line 752: Do we need this slow mo here?

inancgumus: Yes, so we can slow down the page reload process.

inancgumus (approved):
Thanks!

common/page.go
ankur22: There is quite a bit of refactoring to be done with all these lifecycle events. I will do them in a new PR once tall these PRs are in (including moving `wrapTimeoutError` into its own function). I think that's the best strategy for now, but I will be mindful in the future to do them earlier when possible.

inancgumus: No worries 🤝

tests/lifecycle_wait_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+		assertHome(t, tb, p, common.LifecycleEventNetworkIdle, func() {
+			result := p.TextContent("#pingRequestText", nil)
+			assert.EqualValues(t, "Waiting... pong 10 - for loop complete", result)
+		})
+	})
+}
+
+func assertHome(
+	t *testing.T,
```
ankur22, line 270: Note to self: it's idiomatic go to group similar args on the same line e.g.

```
func assertHome(
	t *testing.T, tb *testBrowser,
        p api.Page, waitUntil common.LifecycleEvent,
```

Issue #593, opened 2022-10-17
Simplify lifecycle events for frames


ankur22:
This issue was opened after some discussions in https://github.com/grafana/xk6-browser/pull/578.

Basically, the issue was that a test would time out when xk6-browser navigates to google.com and waits for `networkIdle`. The issue was that we weren't correctly handling the CDP event from the browser.

While working on the fix, it became clear that the lifecycle event code was too complex for what we needed and it might be worth removing.

https://github.com/grafana/xk6-browser/issues/593#tasklist-block-fa274e2a-04dd-4247-b3f4-83b797ca776a

inancgumus:
> The issue was that we weren't correctly handling the CDP event from the browser.
> 
> While working on the fix, it became clear that the lifecycle event code was too complex for what we needed and it might be worth removing.

I believe the issue should not be about removing stuff. Instead, [finding ways to understand](https://github.com/grafana/xk6-browser/pull/578#issuecomment-1282056312) why the existing constructors exist in the first place. For example, in #578, [it turned out](https://github.com/grafana/xk6-browser/pull/578#issuecomment-1282335614) that the actual problem was about _not handling the removal events_.

ankur22:
> Research: How do Playwright and Puppeteer do it. Maybe add a Mermaid diagram.

I don't think looking at puppeteer and playwright should be how we determine what's best for xk6-browser. They're a good source of information on why things have been done a certain way in xk6-browser though.

> I believe the issue should not be about removing stuff. Instead, https://github.com/grafana/xk6-browser/pull/578#issuecomment-1282056312 why the existing constructs exist in the first place. For example, in https://github.com/grafana/xk6-browser/pull/578, https://github.com/grafana/xk6-browser/pull/578#issuecomment-1282335614 that the actual problem was about not handling the removal events.

I agree that a suite of tests is what we need now, and then we can refactor the code, whatever the solution maybe as long as the tests pass.

inancgumus:
> I don't think looking at puppeteer and playwright should be how we determine what's best for xk6-browser. They're a good source of information on why things have been done a certain way in xk6-browser though.

They can give us a better overall picture. I'm not saying that we should copycat them :)

inancgumus:
I've removed the PR from the project board and linked PR #647 to this issue. This way, we can reduce the number of issues and can easily track their implementation and tests from a single place. IMO, the project items should mostly be about public-facing issues rather than PRs (not always possible).

PR #625, opened 2022-11-02
Add tests to lock in the waitUntil lifecycle event behaviour

ankur22:
Before we attempt to cleanup the lifecycle event code (#593) we need to lock in the current behaviour so that any refactoring does not affect the behaviour.

The tests are:
- test that `waitUntil` on `load` and `DOMContentLoaded` work as expected when used in `goto`.
- test that `waitUntil` on `load`, `DOMContentLoaded` and `networkidle` work as expected when we navigate to a page with a main and sub frame when used in `goto`.
- test that `waitUntil` on `load`, `DOMContentLoaded` and `networkidle` work as expected when used as an option on `WaitForNavigation`.

### DIFF
--- a/tests/lifecycle_wait_test.go
+++ b/tests/lifecycle_wait_test.go
@@ -15,6 +15,17 @@ import (
 )
 
 func TestLifecycleWaitForNavigationLoad(t *testing.T) {
+	// Test description
+	//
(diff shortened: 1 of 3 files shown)


ankur22:
I've tidied the commits and moved these changes into [PR #643](https://github.com/grafana/xk6-browser/pull/643)

PR #627, opened 2022-11-03
Force browser opts in cloud

### LINKED ISSUE #414: Force browser options in Cloud test runs
In order to ensure users don't run into confusing issues when running xk6-browser tests in the Cloud, we should override certain [options passed to `BrowserType.Launch`](https://github.com/grafana/xk6-browser/blob/eb136726e37282fffad0c642062a04a6680a6adc/README.md?plain=1#L106-L115) and ultimately to the browser process.

For example:
- `headless` in the Cloud must currently be set to `true`, as we don't have the capability to run with a browser UI, and arguably don't need to.
- `executablePath` shouldn't be configurable in the Cloud.
- `devtools` should be ignored.

We should ensure that whatever these options are configured as, they are passed a fixed value in Cloud test runs.

## Suggested solution

We have to detect some way that we're running in the Cloud, maybe with an existing or additional env var(?). Consult with the k6 or Backend team about the best approach.

And then the change to override the options should be done around here:
https://github.com/grafana/xk6-browser/blob/eb136726e37282fffad0c642062a04a6680a6adc/chromium/browser_type.go#L208-L213

## Update

We can detect whether we're running in the cloud by checking the environment variable `K6_CLOUDRUN_INSTANCE_ID`. See [this page](https://k6.io/docs/cloud/creating-and-running-a-test/cloud-tests-from-the-cli/#cloud-environment-variables) for more details.

https://github.com/grafana/xk6-browser/issues/414#tasklist-block-ca1a6fa1-4189-43f1-8bb9-d1229b670ebe

inancgumus:
This PR ignores cloud-only options and logs a message when xk6-browser is running in the cloud.

The cloud-only options are: `devtools`, `executablePath`, and `headless`.

* The `k6ext.OnCloud` detects the cloud run and returns `true` if the [`K6_CLOUDRUN_INSTANCE_ID`](https://k6.io/docs/cloud/creating-and-running-a-test/cloud-tests-from-the-cli/#cloud-environment-variables) environment variable exists.
* `LaunchOptions.Parse` ignores the cloud-only options if we're running in the cloud.

Closes #414.

### DIFF
--- a/common/browser_options_test.go
+++ b/common/browser_options_test.go
@@ -212,7 +212,7 @@ func TestBrowserLaunchOptionsParse(t *testing.T) {
 			t.Parallel()
 			var (
 				vu = k6test.NewVU(t)
-				lo = NewLaunchOptions()
+				lo = NewLaunchOptions(false)
 			)
 			err := lo.Parse(vu.Context(), vu.ToGojaValue(tt.opts), log.NewNullLogger())
 			if tt.err != "" {
--- a/common/browser_options_test.go
+++ b/common/browser_options_test.go
@@ -15,9 +15,10 @@ func TestBrowserLaunchOptionsParse(t *testing.T) {
 	t.Parallel()
 
 	for name, tt := range map[string]struct {
-		opts   map[string]any
-		assert func(testing.TB, *LaunchOptions)
-		err    string
+		opts    map[string]any
+		assert  func(testing.TB, *LaunchOptions)
+		err     string
+		onCloud bool
 	}{
 		"defaults": {
 			opts: map[string]any{},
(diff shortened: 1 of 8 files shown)


common/browser_options_test.go
```diff
@@ -31,6 +32,24 @@ func TestBrowserLaunchOptionsParse(t *testing.T) {
...
+				"executablePath": "something else",
+			},
+			assert: func(tb testing.TB, lo *LaunchOptions) {
+				tb.Helper()
+				assert.Equal(t, &LaunchOptions{
+					Env:               make(map[string]string),
+					Headless:          true,
+					LogCategoryFilter: ".*",
+					Timeout:           DefaultTimeout,
+					onCloud:           true,
```
ankur22, line 49: Do we need to assert only on `headless`, `devtools` and `executablePath`?

ankur22 (changes requested):
Looks good, just need to double check the assert on the test.

common/browser_options_test.go
inancgumus: Only those fields are forced, and we check whether users can overwrite them in the cloud. The other fields are taken care of already by the other test cases.

inancgumus: 👍 Resolved in 78c3f7a.

ankur22 (approved):
LGTM. Thanks for the changes in the test, they're very useful comments.

Issue #635, opened 2022-11-07
Deadlock in TestPageWaitForFunction tests


ankur22:
[logs_4015.zip](https://github.com/grafana/xk6-browser/files/9951185/logs_4015.zip)
### Brief summary

A CI pipeline failed with the following log, and it looks to be a deadlock around the `WaitForFunction`, and specifically `waitForExecutionContext`. 

### xk6-browser version

CI

### OS

Ubuntu   20.04.5   LTS

### Chrome version

Unknown

### Docker version and image (if applicable)

NA

### Steps to reproduce the problem

A CI pipeline failed with the following log (only the goroutine logs that are of interest have been added), and it looks to be a deadlock around the `WaitForFunction`, and specifically `waitForExecutionContext`. We had a similar issue with [deadlock in the cloud](https://github.com/grafana/xk6-browser/issues/496), which was due to out of order CDP events which was fixed in https://github.com/grafana/xk6-browser/pull/555. Could be something holding on `executionContextMu` in `frame`.

NOTE: The log is just the starting place to try and recreate the issue, but the root cause might be elsewhere in the system.

CI action log: https://github.com/grafana/xk6-browser/actions/runs/3409862907/jobs/5672145446

```bash
goroutine 17250 [sync.RWMutex.Lock, 4 minutes]:
sync.runtime_SemacquireRWMutex(0xc000581400?, 0x77?, 0x116f62e?)
	/home/runner/sdk/gotip/src/runtime/sema.go:87 +0x26
sync.(*RWMutex).Lock(0xc000581400)
	/home/runner/sdk/gotip/src/sync/rwmutex.go:152 +0x8b
github.com/grafana/xk6-browser/common.(*Frame).setContext(0xc0005812c0, {0x1afba9d, 0x7}, {0x1d37578?, 0xc001398d10})
	/home/runner/work/xk6-browser/xk6-browser/common/frame.go:518 +0x87
github.com/grafana/xk6-browser/common.(*FrameSession).onExecutionContextCreated(0xc001341080, 0xc00058bac8)
	/home/runner/work/xk6-browser/xk6-browser/common/frame_session.go:565 +0xa4e
github.com/grafana/xk6-browser/common.(*FrameSession).initEvents.func1()
	/home/runner/work/xk6-browser/xk6-browser/common/frame_session.go:253 +0x645
created by github.com/grafana/xk6-browser/common.(*FrameSession).initEvents
[...]

PR #636, opened 2022-11-07
Fix deadlock with executionContextMu

ankur22:
executionContextMu isn't needed to be locked and unlocked in this (WaitForFunction) method as it doesn't r/w to executionContext or documentHandle. This function later calls an unexported method with the same name, which does require access to executionContext.

A deadlock was occurring later in waitForExecutionContext since this method was holding on to the lock.

Closes: https://github.com/grafana/xk6-browser/issues/635


inancgumus (approved):
Nice! However, there is still a devil lurking in those waters. The following deadlocks after a while:

```go
go test ./tests -run='TestPageWaitForFunction' -count=100 -timeout=10s
```

We can look at this in-depth later.

PR #628, opened 2022-11-03
Fix WaitForLoadState and add tests

ankur22:
This PR was opened to help in part with issue https://github.com/grafana/xk6-browser/issues/593.

This PR's aim is to do two things:
1. Add missing tests for `WaitForLoadState` when using `waitUntil` options.
2. Fix an issue where `WaitForLoadState` needs to wait for new events from the browser.

https://github.com/grafana/xk6-browser/pull/623 needs to be merged into `main` first before this PR.

The tests all follow a similar setup to what is already present in `main` and in https://github.com/grafana/xk6-browser/pull/623.

The fix involved copying the `waitUntil` logic used in [FrameManager.NavigateFrame](https://github.com/grafana/xk6-browser/blob/main/common/frame_manager.go#L578).

PR #644, opened 2022-11-08
Refactor the frame's lifecycle event handling code

ankur22:
Closes: https://github.com/grafana/xk6-browser/issues/593

These were the prerequisite PRs to fix and add tests around lifecycle events:
- https://github.com/grafana/xk6-browser/pull/578
- https://github.com/grafana/xk6-browser/pull/623
- https://github.com/grafana/xk6-browser/pull/628
- https://github.com/grafana/xk6-browser/pull/647 (https://github.com/grafana/xk6-browser/pull/643)

Now that these are in place, I feel confident that the refactoring in this PR doesn't affect the `waitUntil` lifecycle behaviour.

Each commit explains why the change was made and eventually we end up in a situation where we no longer need the networkIdle timer, `recalculateLifecycle` and `subtreeLifecycleEvents`.

I believe the reasons why lifecycle event was so complicated in the first place was down to unordered CDP events, which has  since been resolved in PR #555. Now we're seeing all CDP events come in order, starting with events coming in for the sub frame and ending on the root/main frame, from `DOMContentLoaded`, to `load`, and ending on `networkIdle`. NOTE: there are other lifecycle events that the browser sends through, but we're not interested in those yet.


inancgumus (commented):
I'm glad we got rid of all these 🥳 Thanks! Let's wait for #643 before merging this one.

inancgumus (approved):
🚀

PR #647, opened 2022-11-10
Refactor lifecycle tests and add any missing tests

ankur22:
Before we attempt to cleanup the lifecycle event code (#593) we need to lock in the current behaviour so that any refactoring does not affect the behaviour.

The tests are:
- test that `waitUntil` on `load` and `DOMContentLoaded` work as expected when used in `goto`.
- test that `waitUntil` on `load`, `DOMContentLoaded` and `networkidle` work as expected when we navigate to a page with a main and sub frame when used in `goto`.
- test that `waitUntil` on `load`, `DOMContentLoaded` and `networkidle` work as expected when used as an option on `WaitForNavigation`.
- test that `waitUntil` on `networkidle` timesout if we set it too late after the navigation has completed.

Linked issue: https://github.com/grafana/xk6-browser/issues/593


tests/lifecycle_wait_test.go
```diff
@@ -14,366 +14,546 @@ import (
...
+		return nil
 	})
+	require.NoError(t, err)
+
+	assert.False(t, resolved)
+	assert.True(t, rejected)
 }
 
-func TestLifecycleWaitForLoadStateDOMContentLoadedThenNetworkIdle(t *testing.T) {
+func TestLifecycleWaitForLoadState(t *testing.T) {
```
inancgumus, line 197: Can we put these tests into `TestLifecycleWaitForNavigation`? Optional: To do that, maybe we can create a `home` struct and have some methods. I believe fewer things to read and less code are mostly better regarding comprehensibility and maintenance. I think `assertHome` deserves a better home.

tests/lifecycle_wait_test.go
```diff
@@ -14,366 +14,546 @@ import (
...
-		assert.EqualValues(t, "ping.js loaded from server", result)
-	})
+				return testPromise{}
+			}, nil)
+		})
+	}
 }
 
-func TestLifecycleReloadLoad(t *testing.T) {
+func TestLifecycleReload(t *testing.T) {
```
inancgumus, line 313: As far as I can see, there is little difference between this one and `TestLifecycleWaitForNavigation`. Maybe we can join them together too.

inancgumus (approved):
Excellent work 👏  Thanks for considering my previous feedback and refactoring this 🙇

**Commits:**
IMHO, we don't need to make separate commits for minor changes, such as the following. Please consider squashing them
- Add general lifecycle event guidelines
- Rename the goto network lifecycle event tests
- Move general guideline comment to top
- Refactor ... test commits

It is now clear that the same code exercises the test cases of the table test. And each test case is small enough. A commit for each top-level test is enough. IMO, no need to make a commit for each table test. It is better to squash them into one commit.

**The following `static/*.html` files are nearly identical.**
- prolonged_network_idle.html
- prolonged_network_idle_10.html
- reload_lifecycle.html
- wait_for_nav_lifecycle.html

We can parameterize them for each test case instead of duplicating these files. You can use `page.Eval` to do that. I found that only the following parts change:

Changing parts:
```html
+ <div id="pingJSText">Waiting...</div>
+ Loop count
+ <script src="/ping.js" async></script>
+ <a href="/home" id="homeLink">Home</a>
```

**Summary** :)
As the PR is big, I couldn't validate its functionality. I'll check it out once again when I'm back.

I'm approving this PR as it is, and I have some suggestions I want you to consider. All are optional. I believe there are some areas in which we can further decrease duplication and make it more straightforward and maintainable.

tests/lifecycle_wait_test.go
```diff
@@ -14,366 +14,546 @@ import (
...
-		p.WaitForLoadState(waitUntil.String(), nil)
-	})
+	tests := []struct {
+		name                  string
+		pingSlowness          time.Duration
+		pingJSSlow            bool
+		waitUntil             common.LifecycleEvent
+		pingRequestTextAssert func(result string, pingCount int)
+		pingJSTextAssert      func(result string)
+		assertFunc            func(p api.Page)
```
inancgumus, line 57: It's good to see that you could find a way to simplify my previous example 👏 Clever!

ankur22:
Thanks for the suggestions! I'll take a look at them 🙂

tests/lifecycle_wait_test.go
```diff
@@ -36,17 +36,7 @@ func TestLifecycleWaitForLoadStateLoad(t *testing.T) {
...
-	tb.withHandler("/ping", func(w http.ResponseWriter, _ *http.Request) {
-		counterMu.Lock()
-		defer counterMu.Unlock()
-
-		time.Sleep(time.Millisecond * 100)
-
-		counter++
-		fmt.Fprintf(w, "pong %d", counter)
-	})
+	withPingHandler(t, tb, time.Millisecond*100, nil)
```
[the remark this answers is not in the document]
ankur22: Done in 6d6b2ed5dec69486fca453218e02a4c2251f1472

tests/lifecycle_wait_test.go
ankur22: I can take a look at making `assertHome` more generic.

All these separate tests functions test different methods (`goto`, `waitForNavigation`, `waitForLoadState`, `reload`), and although each of them use `waitUntil` in a similar way, they are still different methods. If we accidentally change one of these, then only the related test function will break, which should be easier to understand.

`goto` is tested in a few different ways, one is with `subFrames`, another is without `subFrames` and finally we test `networkIdle` in different scenarios where the website is different enough whereby the lifecycle events are emitted by the browser (subtly) differently -- i.e. they are edge cases. I think they should also live in their own functions.

tests/lifecycle_wait_test.go
ankur22: One tests `reload` with `waitUntil` option and the other tests `WaitForNavigation` with `waitUntil` option. So i think it's better to keep them separated.

tests/lifecycle_wait_test.go
```diff
@@ -14,366 +14,546 @@ import (
...
+			},
+		)
+		prm.then(
+			func() {
+				resolved = true
+			},
+			func() {
+				rejected = true
+			},
+		)
```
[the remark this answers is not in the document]
ankur22: Resolved in 0dd3708f4f0b10482c67194bddd50677d0216977

PR #649, opened 2022-11-11
Add/641 naked browser launch opts

### LINKED ISSUE #641: Support naked browser launch options
The following script code results in `GoError: parsing launch options: LaunchOptions does not exist in the runtime`. 

```js
```

So we have to provide an empty launch options object as follows:

```js
const browser = chromium.launch({});
```

A better way is to support passing the options as _naked_ in the first example above.

inancgumus:
Fixes #641.

### DIFF
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -131,9 +131,8 @@ func (b *BrowserType) Launch(opts goja.Value) api.Browser {
 	if b.logger, err = makeLogger(ctx); err != nil {
 		k6ext.Panic(ctx, "setting up logger: %w", err)
 	}
-
 	launchOpts := common.NewLaunchOptions(k6ext.OnCloud())
-	if err := launchOpts.Parse(ctx, opts, b.logger); err != nil {
+	if err := launchOpts.Parse(ctx, b.logger, opts); err != nil {
 		k6ext.Panic(ctx, "parsing launch options: %w", err)
 	}
 	ctx = common.WithLaunchOptions(ctx, launchOpts)
--- a/common/browser_options_test.go
+++ b/common/browser_options_test.go
@@ -14,6 +14,13 @@ import (
 func TestBrowserLaunchOptionsParse(t *testing.T) {
 	t.Parallel()
 
+	defaultOptions := &LaunchOptions{
+		Env:               make(map[string]string),
+		Headless:          true,
+		LogCategoryFilter: ".*",
+		Timeout:           DefaultTimeout,
+	}
+
 	for name, tt := range map[string]struct {
 		opts    map[string]any
 		assert  func(testing.TB, *LaunchOptions)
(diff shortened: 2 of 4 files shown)


chromium/browser_type.go
```diff
@@ -124,16 +124,15 @@ func (b *BrowserType) initContext() context.Context {
 
 // Launch allocates a new Chrome browser process and returns a new api.Browser value,
 // which can be used for controlling the Chrome browser.
-func (b *BrowserType) Launch(opts goja.Value) api.Browser {
+func (b *BrowserType) Launch(opts ...goja.Value) api.Browser {
```
ankur22, line 127: Why does it need to be done in this way?

For example `browser.newContext()` can be used without supplying any options and it only takes a single `goga.Value` argument:

```go
func (b *Browser) NewContext(opts goja.Value) api.BrowserContext
```

examples/launch_naked.js
```diff
@@ -0,0 +1,15 @@
+import exec from 'k6/execution';
```
ankur22, line 1: Nice, this is good 👍

common/browser_options_test.go
```diff
@@ -262,4 +263,41 @@ func TestBrowserLaunchOptionsParse(t *testing.T) {
 			}
 		})
 	}
+
```
ankur22, line 266: Thanks for the test too 👍

common/browser_options_test.go
```diff
@@ -262,4 +263,41 @@ func TestBrowserLaunchOptionsParse(t *testing.T) {
...
+	// will be passed to NewLaunchOptions to set onCloud to false.
+	const onCloud = false
```
ankur22, line 268: Very minor, so you can ignore this suggestion.

```suggestion
	const notOnCloud = false
```

I think this reads better when it's used:

```go
wantOpts: NewLaunchOptions(notOnCloud),
```

vs

```go
wantOpts: NewLaunchOptions(onCloud),
```

ankur22: Or maybe set `onCloud` to `true`, and then:

```go
wantOpts: NewLaunchOptions(!onCloud),
```

chromium/browser_type.go
ankur22: The browser context parse returns `ni`l if `opts` is `nil`:

```go
if opts != nil && !goja.IsUndefined(opts) && !goja.IsNull(opts) {
   ...
}
return nil
```

Couldn't we do something similar?

ankur22 (changes requested):
I'm a little confused as to why we need to use a variadic function.

The tests are a great addition 👍

chromium/browser_type.go
inancgumus: That's a very good point 👍  I'll check it out, thanks.

inancgumus:
You're right. IDK how I missed that 🤦 Thanks, @ankur22 🙇 I've changed the signature to the original and updated the tests. I only left-shifted the logger as I wanted to clarify that `context` and `logger` are peripherals, and we're interested in the options.

ankur22 (approved):
Looks good, thanks for this!

PR #652, opened 2022-11-17
Subscribe to CDP performance metrics

ankur22:
This is a test to see what is returned from the browser via CDP. Does it return us the valuable Web Vital metrics we need to show the end user?

### DIFF
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -25,6 +25,7 @@ import (
 	cdplog "github.com/chromedp/cdproto/log"
 	"github.com/chromedp/cdproto/network"
 	cdppage "github.com/chromedp/cdproto/page"
+	"github.com/chromedp/cdproto/performancetimeline"
 	cdpruntime "github.com/chromedp/cdproto/runtime"
 	"github.com/chromedp/cdproto/security"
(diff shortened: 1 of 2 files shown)


inancgumus:
> Although we're not getting back the `startTime` we can just use [initTime](https://github.com/grafana/xk6-browser/blob/7a45ec758172a42d663d80ab6f913c8f28e8540d/common/frame.go#L352) like we're doing with our other metric measurements.
> 
> * I tried this and it doesn't work as `initTime` is relative to the lifecycle events, and for some reason it's out by more than a day.

This might be because of the walltime vs timestamp difference.

Here we set it using a timestamp:

https://github.com/grafana/xk6-browser/blob/7a45ec758172a42d663d80ab6f913c8f28e8540d/common/frame_session.go#L702

We might want to do a similar thing like we did here:

https://github.com/grafana/xk6-browser/blob/ae5c339128b06bf3771bde6152773fd7ea34f05d/common/request.go#L44-L57

inancgumus:
> I don't think retrieving the Web Vitals from the browser over CDP is a viable solution in its current state.

Why do you think so? Could you elaborate more?

ankur22:
This was a POC, so closing it.

PR #663, opened 2022-11-29
dismiss dialog box automatically

ankur22:
Partly closes: https://github.com/grafana/xk6-browser/issues/655

We have had a feature request recently as well as seen dialog boxes being used elsewhere. There currently isn't a work around to removing dialog boxes in the path of a test. This change will automatically dismiss the dialog boxes (and accept it for `beforeunload` type dialogs).

### DIFF
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -272,6 +272,11 @@ func (fs *FrameSession) onEventJavascriptDialogOpening(event *cdppage.EventJavas
 		"sid:%v tid:%v url:%v dialogType:%s",
 		fs.session.ID(), fs.targetID, event.URL, event.Type)
 
+	// Dialog type of beforeunload needs to accept the
+	// dialog, instead of dismissing it. We're unable to
+	// dismiss beforeunload dialog boxes at the moment as
+	// it seems to pause the exec of any other action on
+	// the page. I believe this is an issue in Chromium.
 	action := cdppage.HandleJavaScriptDialog(false)
 	if event.Type == cdppage.DialogTypeBeforeunload {
 		action = cdppage.HandleJavaScriptDialog(true)
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -273,6 +273,10 @@ func (fs *FrameSession) onEventJavascriptDialogOpening(event *cdppage.EventJavas
 		fs.session.ID(), fs.targetID, event.URL, event.Type)
 
 	action := cdppage.HandleJavaScriptDialog(false)
+	if event.Type == cdppage.DialogTypeBeforeunload {
+		action = cdppage.HandleJavaScriptDialog(true)
+	}
+
 	if err := action.Do(cdp.WithExecutor(fs.ctx, fs.session)); err != nil {
 		fs.logger.Errorf("FrameSession:onEventJavascriptDialogOpening", "failed to dismiss dialog box: %v", err)
 	}
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -259,10 +259,23 @@ func (fs *FrameSession) initEvents() {
 					fs.onAttachedToTarget(ev)
 				case *target.EventDetachedFromTarget:
 					fs.onDetachedFromTarget(ev)
+				case *cdppage.EventJavascriptDialogOpening:
+					fs.onEventJavascriptDialogOpening(ev)
 				}
 			}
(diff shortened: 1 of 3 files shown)


tests/frame_test.go
```diff
@@ -21,3 +23,58 @@ func TestFramePress(t *testing.T) {
...
+		{
+			name: "confirm",
+		},
+		{
+			name: "prompt",
+		},
+		{
+			name: "beforeunload",
+		},
+	}
```
inancgumus, line 45: Since all we need is a string, to make it simpler, we can use `[]string` here instead of a `struct`. We can add a struct back if needed, but, currently it's not needed.

```suggestion
	tests := []string{"alert", "confirm", "prompt", "beforeunload"}
```

tests/frame_test.go
```diff
@@ -21,3 +23,58 @@ func TestFramePress(t *testing.T) {
...
+
+			p := b.NewPage(nil)
+
+			err := b.await(func() error {
+				opts := b.toGojaValue(struct {
+					WaitUntil string `js:"waitUntil"`
+				}{
+					WaitUntil: "networkidle",
+				})
+				b.promise(p.Goto(b.staticURL("dialog.html?dialogType="+tt.name), opts)).then(func() *goja.Promise {
```
inancgumus, line 61: Consider (optional) to make it slightly more readable and comprehensible.

```suggestion
				opts := b.toGojaValue(struct {
					WaitUntil string `js:"waitUntil"`
				}{
					WaitUntil: "networkidle",
				})
				pageGoto := p.Goto(
					b.staticURL("dialog.html?dialogType="+tt.name),
					opts,
				)
				b.promise(pageGoto).then(func() *goja.Promise {
```

common/frame_session.go
```diff
@@ -259,12 +259,29 @@ func (fs *FrameSession) initEvents() {
...
 
+func (fs *FrameSession) onEventJavascriptDialogOpening(event *cdppage.EventJavascriptDialogOpening) {
+	fs.logger.Debugf("FrameSession:onEventJavascriptDialogOpening",
+		"sid:%v tid:%v url:%v dialogType:%s",
+		fs.session.ID(), fs.targetID, event.URL, event.Type)
+
+	action := cdppage.HandleJavaScriptDialog(false)
+	if event.Type == cdppage.DialogTypeBeforeunload {
+		action = cdppage.HandleJavaScriptDialog(true)
+	}
```
inancgumus, line 278: Consider adding a comment here about why we're doing this `false` -> `true` change for `beforeUnload`.

inancgumus (approved):
Nice work 👏

tests/frame_test.go
ankur22: I prefer keeping it as structs. If someone takes a quick look at it, it's clear that this is a table test.

common/frame_session.go
ankur22: I think this is an issue in Chromium. When we try to dismiss the dialog box, nothing happens. It's the same issue that Playwright runs into with Chromium -- https://github.com/microsoft/playwright/issues/14431

ankur22: Resolved in: f598949abe8156e2d7eabbb5f5837ad1b1415c49

ankur22:
> Nice work 👏

Thanks @inancgumus 🙂 I've actioned on all but one comment. Thanks for the review!

tests/frame_test.go
inancgumus: Table testing is not related whether you're using a struct or not.

inancgumus: Table testing is not related to whether you're using a struct or not 😅

Examples: [1](https://github.com/golang/go/blob/656d3f4401e1fdb628e24027cf91cc803c7a0dac/src/cmd/go/internal/work/exec_test.go#L37) and [2](https://github.com/golang/go/blob/50b8f490e1701773254fdbfe0e3fea00ff1f2a68/src/net/mail/message_test.go#L1029).

ankur22: Oh nice! Thanks for the references. I'll change it.

inancgumus: Sure thing :)

ankur22: Resolved in: 40b4a87aefa1bdd6783b4d13368ee49e596ef092

Issue #664, opened 2022-12-01
Restrict browser runs with environment variables


ankur22:
We need an env var to disable the browser test runs in restricted environments.

PR #665, opened 2022-12-01
Add a env var to disable test runs

ankur22:
This flag can stop tests runs that shouldn't occur, along with another messaging env var that override the default messaging.

Closes: https://github.com/grafana/xk6-browser/issues/664


inancgumus (changes requested):
LGTM. Let's squash the commits into a single one first 🙇 

**Some notes about our workflow:**

A PR should change one part of the code once, so we can easily review it. Changing the same code part in subsequent commits makes it harder to see what is going on (and forces me to review the same feature multiple times) + it produces noise for the commit history. So we should squash commits when the initial solution changes.

Refactoring commits are valuable when you're refactoring an existing part of the codebase instead of refactoring your previous commits 😄 So this is another candidate for squashing (for the same reasons above).

Atomic commits are good, but super atomic commits for every step we take are not 😄 🙇  We can discuss this more if you like; please let me know. [A great blog post from Github](https://github.blog/2022-06-30-write-better-commits-build-better-projects/) can shed more light on this.

ankur22:
@inancgumus thanks for the review and the explanation. This sounds good to me 👍 I've squashed the changes.

PR #667, opened 2022-12-05
xk6-browser should be importable as a k6 module

### LINKED ISSUE #662: xk6-browser should be importable as a k6 module
There is currently no way to import the xk6-browser and assign it a custom module name without also setting `k6/x/browser`. 

To do that the [RootModule](https://github.com/grafana/xk6-browser/blob/main/main.go#L52) will need to be in package different from the "main" one and the main one to continue importing it.

You can see a very simplified example in https://github.com/grafana/xk6-timers repo for example.

inancgumus:
Closes #662


ankur22 (approved):
LGTM. Thanks for making it consistent with the other experimental modules 👍 

I think we should get approval from the k6 team too.

browser/module_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/grafana/xk6-browser/chromium"
+
+	k6common "go.k6.io/k6/js/common"
+	k6modulestest "go.k6.io/k6/js/modulestest"
+	k6metrics "go.k6.io/k6/metrics"
+)
+
+// TestModuleNew tests registering the module.
+// It doesn't test the module's remaining functionality as it is
+// already tested in the tests/ integration tests.
```
mstoykov, line 18: Shall the `tests` then just import `browser` and test it ?

register.go
```diff
@@ -0,0 +1,28 @@
...
+	_ "net/http/pprof" //nolint:gosec
+	"os"
+
+	"github.com/grafana/xk6-browser/browser"
+
+	k6modules "go.k6.io/k6/js/modules"
+)
+
+func init() {
+	if _, ok := os.LookupEnv("K6_BROWSER_PPROF"); ok {
```
mstoykov, line 16: I guess we should remember that this won't work with core k6.

browser/module_test.go
mstoykov: preferably not in this PR but later ;)

mstoykov (approved):
LGTM!

I have left some comments that are likely mostly for openning new issues or taking notes.

register.go
ankur22: That's fine tbh.

Out of curiosity, why wouldn't the `init` function run when `k6` imports the module? I thought that all `init` functions had to to run (which is an unwanted side affect sometimes).

mstoykov: you import a *package*. 
And we are not going to import *this* package as that will also register the extension as `k6/x/browser` - the init below this one. 

Which is like the whole point of this PR ;)

ankur22: Aha, yep, of course 🙂

browser/module_test.go
inancgumus: That makes sense, thanks :) However, I liked keeping this simple test next to the module creation. So that it would be easier to make changes to both code and tests when needed (instead of jumping to the other parts of the code base `tests/`.

register.go
inancgumus: We use this code in our own instances, so that won't be a problem. Thanks for letting us know, though 👍

PR #671, opened 2022-12-08
Fix k6 Hosts option incompatibility

### LINKED ISSUE #670: Fix k6 Hosts incompatibility
k6-core v0.42 will introduce [Support wildcards in options.Hosts](https://github.com/grafana/k6/issues/2724). Doing so will break our current code. We need to make changes to be compatible with this new breaking change.

The following is no longer possible as `Hosts` is no longer a map:

https://github.com/grafana/xk6-browser/blob/54b570e54c3be6fe9e82afe97dfc46e124721b2f/chromium/browser_type.go#L369-L371

inancgumus:
Fixes #670.


chromium/browser_type.go
```diff
@@ -356,24 +360,44 @@ func setFlagsFromArgs(flags map[string]any, args []string) {
...
+	//
+	// This is done by marshaling the k6 hosts option to JSON and then
+	// unmarshaling it to a map[string]string. This is done because the
+	// k6 v0.42 changed Hosts from a map to types.NullHosts and doesn't
+	// expose the map anymore.
+	//
+	// TODO: A better way to do this would be to handle the resolver
+	// rules by communicating with Chromium (and then using Hosts's
+	// Match method) instead of passing the rules via the command line
+	// to Chromium.
```
ankur22, line 383: How would we communicate with the chromium? CDP? I don't know why that would be better than passing the rules via the command line. Maybe we need another issue for that? 🙂

ankur22 (approved):
LGTM 👍 

I'm not sure about the details in the comment, and it might be good to add those details either in the same comment or in a new issue.

chromium/browser_type.go
inancgumus: Right. I'll create another issue next week. Reminder set 👍😄

PR #673, opened 2022-12-13
Map k6 browser API to Goja

### LINKED ISSUE #661: Localize query method mappings
xk6-browser should not call `rt.SetFieldNameMapper`.

[SetFieldNameMapper](https://pkg.go.dev/github.com/dop251/goja#Runtime.SetFieldNameMapper) changes how goja maps properties from go types to js objects. 

Given that this is global for the whole goja runtime, doing this in an extension is not a good idea and k6 already [does it](https://github.com/grafana/k6/blob/5fa71b761185200d29c87eff16faa5e4a6c372b7/js/bundle.go#L307) Internally.

 This likely will need changes to xk6-browser as I have not looked into what exactly xk6-browser currently differs from k6

inancgumus:
As discussed in issue grafana/xk6-browser#661, we're now "manually" mapping the browser module methods to Goja. The global Goja field name mappers are removed from the code. This can be automated in the future as the mapping can be error-prone.

The current mappers are just forwarders to the inner methods, and we add `$` and `$$` for the `Query` and `QueryAll` methods.

Let me know if I missed something while mapping the methods; and whether the way we solved it is good enough.

Closes grafana/xk6-browser#661.


chromium/browser_type.go
```diff
@@ -48,22 +48,16 @@ type BrowserType struct {
...
 	k6m := k6ext.RegisterCustomMetrics(vu.InitEnv().Registry)
 	b := BrowserType{
 		vu:        vu,
-		hooks:     hooks,
+		hooks:     common.NewHooks(),
 		k6Metrics: k6m,
 		storage:   &storage.Dir{},
 		randSrc:   rand.New(rand.NewSource(time.Now().UnixNano())), //nolint: gosec
 	}
-	rt.SetFieldNameMapper(common.NewFieldNameMapper())
```
mstoykov, line 66: nit; from a very quick look it seems like you do not delete the FIeldNameMapper code inside the xk6-browser/common package, was there a reason to not delete it ?

inancgumus: Ah, right. Thanks. Resolved in [21ad06b](https://github.com/grafana/xk6-browser/pull/673/commits/21ad06b0e0e7ac271c125a4759c9e31f971328c9).

browser/module.go
```diff
@@ -69,3 +76,459 @@ func (*RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
...
+			return rt.ToValue(mr).ToObject(rt)
+		},
+		"resourceType": r.ResourceType,
+		"response": func() *goja.Object {
+			mr := mapResponse(rt, r.Response())
+			return rt.ToValue(mr).ToObject(rt)
+		},
+		"size":   r.Size,
+		"timing": r.Timing,
+		"uRL":    r.URL,
```
mstoykov, line 134: is this really the name of the property `uRL`?

browser/module.go
```diff
@@ -69,3 +76,459 @@ func (*RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
...
+		"request": func() *goja.Object {
+			mr := mapRequest(rt, r.Request())
+			return rt.ToValue(mr).ToObject(rt)
+		},
+		"securityDetails": r.SecurityDetails,
+		"serverAddr":      r.ServerAddr,
+		"size":            r.Size,
+		"status":          r.Status,
+		"statusText":      r.StatusText,
+		"uRL":             r.URL,
```
mstoykov, line 164: uRL ?

browser/module.go
```diff
@@ -69,3 +76,459 @@ func (*RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
...
+		"press":             f.Press,
+		"selectOption":      f.SelectOption,
+		"setContent":        f.SetContent,
+		"setInputFiles":     f.SetInputFiles,
+		"tap":               f.Tap,
+		"textContent":       f.TextContent,
+		"title":             f.Title,
+		"type":              f.Type,
+		"uncheck":           f.Uncheck,
+		"uRL":               f.URL,
```
mstoykov, line 360: `uRL` again

browser/module.go
```diff
@@ -69,3 +76,459 @@ func (*RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
...
+		"textContent":            eh.TextContent,
+		"type":                   eh.Type,
+		"uncheck":                eh.Uncheck,
+		"waitForElementState":    eh.WaitForElementState,
+		"waitForSelector": func(selector string, opts goja.Value) *goja.Object {
+			eh := eh.WaitForSelector(selector, opts)
+			ehm := mapElementHandle(rt, eh)
+			return rt.ToValue(ehm).ToObject(rt)
+		},
+		"query": func(selector string) *goja.Object {
```
mstoykov, line 433: Do you want to have those as well?

mstoykov (changes requested):
I have left some comments, but looks good in general. 

I would not be able to guess if that is enough to not break anything, but I would argue (again) that you should have more JS code tests that use the API as users will do. 


This will also help with making certain that this mapping are what you want - as it will exercise them ;)

browser/module.go
```diff
@@ -54,13 +57,17 @@ func (*RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
...
-			Chromium: chromium.NewBrowserType(vu),
-			Devices:  common.GetDevices(),
-			Version:  version,
-		},
+		mod: m.NewJSModule(vu),
+	}
+}
+
+// NewJSModule returns a new JSModule instance.
+func (*RootModule) NewJSModule(vu k6modules.VU) *JSModule {
```
ankur22, line 66: Nit -- I don't think this refactor was needed.

browser/module.go
```diff
@@ -69,3 +76,459 @@ func (*RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
...
+		"waitForEvent":                p.WaitForEvent,
+		"waitForFunction":             p.WaitForFunction,
+		"waitForLoadState":            p.WaitForLoadState,
+		"waitForNavigation":           p.WaitForNavigation,
+		"waitForRequest":              p.WaitForRequest,
+		"waitForResponse":             p.WaitForResponse,
+		"waitForSelector":             p.WaitForSelector,
+		"waitForTimeout":              p.WaitForTimeout,
+		"workers":                     p.Workers,
+		"query": func(selector string) *goja.Object {
```
ankur22, line 259: We don't need `query` and `queryAll`, no?

browser/module.go
```diff
@@ -69,3 +76,459 @@ func (*RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
...
+		err := obj.Set(k, rt.ToValue(v))
+		if err != nil {
+			k6common.Throw(rt, fmt.Errorf("mapping: %w", err))
+		}
+	}
+
+	return obj
+}
+
+func mapRequest(rt *goja.Runtime, r api.Request) mapping {
```
ankur22, line 103: Are we having to `mapRequest` and `mapResponse` because it can return a `frame`, which in turn can return `pages`? Seems like this is going to catch people out as we will forget to add new methods to the mappings. I preferred the approach where we looped through all the methods and just set them to the object, like above in `mapBrowserToGoja`.

ankur22 (changes requested):
It feels like there are lot of things that could go wrong (i'm not saying they have) i.e. the mappings could be incorrect. Is there anyway we can test the objects that are now manually being created and mapped?

browser/module.go
inancgumus: Good catch. Resolved in https://github.com/grafana/xk6-browser/pull/673/commits/6cf7bcdb6792cd830a5ee2534d334c9a1e58357e.

browser/module.go
inancgumus: No, nice catch again, thanks, removed in https://github.com/grafana/xk6-browser/pull/673/commits/0c388d50abb43a43f1ab0c6c88e6f274839b47f6.

browser/module.go
inancgumus: No, we don't :)

browser/module.go
inancgumus: Hi there!

I agree that the solution we have in place is not future-proof, but it's good enough for now. In the future, we should consider automating or generating this mapping code.

I've added some tests to verify the behavior, which you can find in the following commits:

* https://github.com/grafana/xk6-browser/pull/673/commits/d062048411cf225ca00da0b9f206b148478e9a8f
* https://github.com/grafana/xk6-browser/pull/673/commits/0c388d50abb43a43f1ab0c6c88e6f274839b47f6

On the bright side, if something goes wrong and a method can't be found, users will receive an error message. We can fix any issues that arise, and with the tests in place, the likelihood of that happening is low.

Additionally, these manual mappings will allow us to move the goja layer from our domain logic code, making it more type-safe and easier to work with. I'll create an issue about that soon.

Let me know if you have any questions or concerns.

---

Related:

[The PR description](https://github.com/grafana/xk6-browser/pull/673#issue-1494144809):

> As discussed in the issue https://github.com/grafana/xk6-browser/issues/661, we're now "manually" mapping the browser module methods to Goja. The global Goja field name mappers are removed from the code. This can be automated in the future as the mapping can be error-prone.

Also [here](https://github.com/grafana/xk6-browser/issues/661):

> However, not sure the current solution will be future-proof. Since we have a relatively large API surface area; it can be error-prone to manually do this kind of mappings. I believe we might want to automate this in the future (maybe by writing a source-code generator). That way the resulting mapping can be in the code itself and people can easily track them without magic.

inancgumus:
@ankur22 

> It feels like there are lot of things that could go wrong (i'm not saying they have) i.e. the mappings could be incorrect. Is there anyway we can test the objects that are now manually being created and mapped?
>
> Are we having to mapRequest and mapResponse because it can return a frame, which in turn can return pages? Seems like this is going to catch people out as we will forget to add new methods to the mappings. I preferred the approach where we looped through all the methods and just set them to the object, like above in mapBrowserToGoja.

`mapBrowserToGoja` is the root mapping, that's why it has a loop. The other mappings will automatically added, so they don't need a loop.

I agree that the solution we have in place is not future-proof, but it's good enough for now. In the future, we should consider automating or generating this mapping code.

I've added some tests to verify the behavior, which you can find in the following commits:

* https://github.com/grafana/xk6-browser/pull/673/commits/d062048411cf225ca00da0b9f206b148478e9a8f
* https://github.com/grafana/xk6-browser/pull/673/commits/0c388d50abb43a43f1ab0c6c88e6f274839b47f6

On the bright side, if something goes wrong and a method can't be found, users will receive an error message. We can fix any issues that arise, and with the tests in place, the likelihood of that happening is low. Since the tests test whether all API methods (`api/`) is in the mappings or not.

Additionally, these manual mappings will allow us to move the goja layer from our domain logic code, making it more type-safe and easier to work with. I'll create an issue about that soon.

Let me know if you have any questions or concerns.

---

Related:

[The PR description](https://github.com/grafana/xk6-browser/pull/673#issue-1494144809):

> As discussed in the issue https://github.com/grafana/xk6-browser/issues/661, we're now "manually" mapping the browser module methods to Goja. The global Goja field name mappers are removed from the code. This can be automated in the future as the mapping can be error-prone.

Also [here](https://github.com/grafana/xk6-browser/issues/661):

> However, not sure the current solution will be future-proof. Since we have a relatively large API surface area; it can be error-prone to manually do this kind of mappings. I believe we might want to automate this in the future (maybe by writing a source-code generator). That way the resulting mapping can be in the code itself and people can easily track them without magic.

inancgumus:
> I have left some comments, but looks good in general.
> 

Thank you for your review 🙇 

> I would not be able to guess if that is enough to not break anything, but I would argue (again) that you should have more JS code tests that use the API as users will do.
> This will also help with making certain that this mapping are what you want - as it will exercise them ;)

We now have tests to check whether we map the entire API surface that have the wildcard methods. I agree we should have more js tests, preferably in the `examples/` folder. We should work on this next.

I'm also thinking of changing the directory name to `jstests` or something like that. Since they are not examples anymore and we already have examples in the k6-browser docs.

inancgumus:
> I have left some comments, but looks good in general.
> 

@mstoykov, thank you for your review 🙇 

> I would not be able to guess if that is enough to not break anything, but I would argue (again) that you should have more JS code tests that use the API as users will do.
> This will also help with making certain that this mapping are what you want - as it will exercise them ;)

We now have tests to check whether we map the entire API surface that have the wildcard methods. I agree we should have more js tests, preferably in the `examples/` folder. We should work on this next.

I'm also thinking of changing the directory name to `jstests` or something like that. Since they are not examples anymore and we already have examples in the k6-browser docs.

browser/module.go
inancgumus: Replied [here](https://github.com/grafana/xk6-browser/pull/673#issuecomment-1350594447).

browser/module.go
```diff
@@ -74,6 +74,27 @@ func (mi *ModuleInstance) Exports() k6modules.Exports {
 
 type mapping map[string]any
 
+// wildcards is a list of extra mappings for our API (api/).
+func wildcards() map[string]string {
```
ankur22, line 81: Looks like this and `isWidlCard` are only used in the tests, maybe they could be moved to the test file for now?

browser/module.go
```diff
@@ -74,6 +74,27 @@ func (mi *ModuleInstance) Exports() k6modules.Exports {
...
+		"Frame.queryAll":         "$$",
+		"ElementHandle.query":    "$",
+		"ElementHandle.queryAll": "$$",
+	}
+}
+
+// isWildcard returns true if the method is a wildcard method and
+// returns the name of the method to be called instead of the original
+// method.
+func isWildcard(exceptions map[string]string, typ, method string) (string, bool) {
```
ankur22, line 95: Instead of passing in the `wildcards`, can it not just call `wildcards` from `isWildCard`? A simpler approach could be to define the `wildscards` as a global in the package.

```go
var wildcards = map[string]string{
  "Page.query":             "$",
  "Page.queryAll":          "$$",
  "Frame.query":            "$",
  "Frame.queryAll":         "$$",
  "ElementHandle.query":    "$",
  "ElementHandle.queryAll": "$$",
}

func isWildcard(typ, method string) (string, bool) {
	name := typ + "." + method
	s, ok := wildcards[name]
	return s, ok
}
```

ankur22 (approved):
LGTM. I've left some minor comments.

Having looked at the unit tests you've added in this, i'm not sure adding JS tests would add any more benefit to this, I could be wrong though.

browser/module_test.go
```diff
@@ -45,3 +48,113 @@ func TestModuleNewDisabled(t *testing.T) {
...
+				require.NotNil(t, method)
+
+				m := toFirstLetterLower(method.Name)
+				// change the method name if it is mapped to a wildcard
+				// method. these wildcard methods are not exist on our
+				// API. so we need to use the mapped method instead.
+				if wm, ok := isWildcard(wildcards, typ.Name(), m); ok {
+					m = wm
+				}
+				if _, ok := mapped[m]; !ok {
```
ankur22, line 143: Should we also ensure that `query` and `queryAll` aren't being mapped in the `goja.Object`?

mstoykov (approved):

mstoykov:
> It feels like there are lot of things that could go wrong (i'm not saying they have) i.e. the mappings could be incorrect. Is there anyway we can test the objects that are now manually being created and mapped?

Arguably having them "manually" written (and hopefully tested manually soon as well) makes it less likely that something will just break because the "magic" translation layer changes or someone unknowingly renames a method which was used.

Additioanlly apart from `$` and `$$` I would argue that having fields such as [page.request](https://playwright.dev/docs/api/class-page#page-request) is way easier and more explicit with having a goja.Object created and defined "manually" then it being "magically"  created by goja.

browser/module.go
inancgumus: That makes sense to keep it simple. However, we'd need to add a `nolint` directive (which we don't want #58). Also, the function lets us make the `wildcards` map read-only.

browser/module_test.go
inancgumus: Good idea. Resolved in https://github.com/grafana/xk6-browser/pull/673/commits/220be028aae76e9a1e98489ed1da2cc6ccaa3b7b.

browser/module.go
inancgumus: Resolved in https://github.com/grafana/xk6-browser/pull/673/commits/b298d140f163bcc7594eb303588c631a8bcc40fc.

inancgumus:
> LGTM. I've left some minor comments.

Thank you for the re-review @ankur22.

> Having looked at the unit tests you've added in this, i'm not sure adding JS tests would add any more benefit to this, I could be wrong though.

JS tests can ensure that we are mapping the methods correctly to the ones we mapped. For example, what would happen if we map `request.frame` to something else? Since `request.frame` now returns a `*goja.Object`, we won't detect whether we correctly mapped to a `Frame`. Also, JStests are necessary for the rest of the extension too. So we can exercise them from the point of view of a user. These tests can also allow us to refactor the extension more freely instead of changing the Go code.

PR #681, opened 2023-01-02
Use external Chrome process whenever its DevTools URL is provided via an env var

andrewslotin:
This is a proof-of-concept to make `xk6-browser` use an external Chrome process to run tests.

Whenever the DevTools URL is provided via the `CHROME_DEVTOOLS_URL=` env var, `chromium.launch()` skips launching the browser process altogether and uses provided URL instead. `browser.close()` is a noop in this case, since we don't own the process and it can be reused by other test runs. It is possible though to terminate the browser process via CDP.

This code by no means should be considered production-ready and provided only to prove the idea.

### DIFF
--- a/common/browser.go
+++ b/common/browser.go
@@ -392,6 +392,11 @@ func (b *Browser) newPageInContext(id cdp.BrowserContextID) (*Page, error) {
 
 // Close shuts down the browser.
 func (b *Browser) Close() {
+	if b.browserProc.Pid() < 0 { // we don't own browser process
+		b.logger.Debugf("Ignoring Browser:Close as it has not been started by xk6-browser", "")
+		return
+	}
+
 	defer func() {
 		if err := b.browserProc.userDataDir.Cleanup(); err != nil {
 			b.logger.Errorf("Browser:Close", "cleaning up the user data directory: %v", err)
--- a/common/browser_process.go
+++ b/common/browser_process.go
@@ -39,6 +39,14 @@ func NewBrowserProcess(
 	ctx context.Context, path string, args, env []string, dataDir *storage.Dir,
 	ctxCancel context.CancelFunc, logger *log.Logger,
 ) (*BrowserProcess, error) {
+	if wsURL, ok := os.LookupEnv("CHROME_DEVTOOLS_URL"); ok {
+		return &BrowserProcess{
+			ctx:    ctx,
+			cancel: ctxCancel,
+			wsURL:  wsURL,
+		}, nil
+	}
+
 	cmd, err := execute(ctx, ctxCancel, path, args, env, dataDir, logger)
 	if err != nil {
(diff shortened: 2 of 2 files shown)


inancgumus:
PS: I added #682 for tracking the general work about this.

inancgumus (commented):
Thanks, @andrewslotin!

I've tried this with the example JavaScripts (`examples/`) and they seem to work 🥳 

For now, some Go tests don't pass (see below). But, we can figure it out later on. As a POC, it looks OK. I believe we still need to do some more tests to ensure 😄 

```bash
$ chromium --remote-debugging-port=8079 &
...
$ CHROME_DEVTOOLS_URL=ws://127.0.0.1:8079/devtools/browser/5e49211e-d8f3-4f79-8646-7a2a8b90027d go test ./tests -race -v
...
=== RUN   TestBrowserContextOptionsExtraHTTPHeaders
    browser_context_options_test.go:90: 
                Error Trace:    /Users/inanc/grafana/k6browser/main/tests/browser_context_options_test.go:90
                Error:          Received unexpected error:
                                test timed out after 5s
                Test:           TestBrowserContextOptionsExtraHTTPHeaders
--- FAIL: TestBrowserContextOptionsExtraHTTPHeaders (5.12s)
...
--- PASS: TestBrowserNewPage (0.21s)
=== RUN   TestTmpDirCleanup
    browser_test.go:59: 
                Error Trace:    /Users/inanc/grafana/k6browser/main/tests/browser_test.go:59
                Error:          Should be empty, but was [xk6-browser-data-3329750862]
                Test:           TestTmpDirCleanup
                Messages:       a dir shouldn't exist which matches the pattern `xk6-browser-data-*
...
=== RUN   TestBrowserUserAgent
    browser_test.go:145: 
                Error Trace:    /Users/inanc/grafana/k6browser/main/tests/browser_test.go:145
                Error:          "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4753.0 Safari/537.36" does not contain "Headless"
                Test:           TestBrowserUserAgent
...
=== RUN   TestElementHandleClickConcealedLink
DEADLOCK HERE
...
and so on.
```

ankur22:
This is a great POC! Simple to understand and it works 💪 

It confirms that we can indeed quite easily hook into an existing Chrome browser process ✅ 

I guess the next steps are to identify the best strategy based on how we can retrieve the URL of the running browser process from the container or browser farm. Would it be best to think about the next strategy based on the simplest and most secure approach, which is the browser running in a container? I think we need to have a discussion with the k6 and backend teams to identify what's the best approach of retrieving the URL from a running browser process.

inancgumus:
Closing as it is solved by the having `Connect` API.

Issue #659, opened 2022-11-24
Use latest master/main version of k6 in CI


ankur22:
Create a CI job which can build and run all the tests against the `master`/`main` version of `k6`. This should run alongside the current build job where the latest `release` version of `k6` is used.

This will help us spot breaking changes early so that we can tackle them well in advance of the release when we usually upgrade to the latest version of `k6`.

PR #694, opened 2023-01-09
Add a CI job to test against the latest k6

ankur22:
This job will test xk6-browser against the latest version of k6 from master. We should be able to catch breaking changes earlier during the cycle so that we're not rushing to fix them just before the release.

closes: https://github.com/grafana/xk6-browser/issues/659


inancgumus (approved):
LGTM

PR #695, opened 2023-01-09
Update to the latest version of k6

ankur22:
This version contains the much awaited await/async keyword support.

### DIFF
--- a/common/frame_manager.go
+++ b/common/frame_manager.go
@@ -14,7 +14,6 @@ import (
 	k6modules "go.k6.io/k6/js/modules"
 
 	"github.com/chromedp/cdproto/cdp"
-	"github.com/chromedp/cdproto/network"
 	cdppage "github.com/chromedp/cdproto/page"
 )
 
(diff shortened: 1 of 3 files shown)


inancgumus:
E2E tests are failing (`examples/`). I looked at this a little, but it didn't help.

I was suspicious of the mapping. Then, I disabled the mapping and switched back to [the old way](https://github.com/grafana/xk6-browser/blob/e8af2abe5c4bd1f20a33bc34c425f1c7535703f1/browser/module.go), and the same error in this PR happened again. So it should not be related to the mapping.

I suggest asking this issue to the k6-core team :-)

inancgumus:
E2E tests are failing (`examples/`). I looked at this a little, but it didn't help.

I was suspicious of the mapping. Then, I disabled the mapping and switched back to [the old way](https://github.com/grafana/xk6-browser/blob/e8af2abe5c4bd1f20a33bc34c425f1c7535703f1/browser/module.go), and the same E2E CI error in this PR happened again. So it should not be related to the mapping.

```
// for example, browser.close() in examples/getattribute.js
// causes the following:
ERRO[0000] TypeError: Not a function: [object Object]
```

Somehow the new k6 async/await update made the goja not see our functions.

I suggest asking about this issue to the k6-core team :-)

@ankur22

Issue #683, opened 2023-01-03
Pre-merge: k6 browser Async API to Goja mapping


inancgumus:
This issue concerns an architectural redesign for the upcoming Async API and improving the k6 browser Async API to Goja mapping.

### The problems this issue solves

* Provide an easy switch to [the k6 browser Async API](https://github.com/grafana/xk6-browser/issues/428).
* Increase developer productivity by clearly separating the k6 browser code and Goja (updates [#271](https://github.com/grafana/xk6-browser/issues/271)).
* Unlock merging of k6 browser to k6 core: Mapping Async APIs.

### POC

Here's [a POC](https://github.com/grafana/xk6-browser/pull/704) for this issue.

### Sub issues

- [x] Planning and experimentation
- [x] #702 
- [x] #704
- [x] #705
- [x] #707
- [x] #706
- [x] #708
- [x] #709
- [x] #723

---

### Mapping Async APIs

The current problem is that when a function returns a value in a promise, we can't map the returned value in [the mapping layer](https://github.com/grafana/xk6-browser/blob/e62e7c12372f210914ea8d9ba70ddb392dbdc93e/browser/mapping.go).

So the missing piece of the puzzle is:
* Mapping the returned value
* Putting the value in a promise
* Returning it to Goja in the mapping layer.
 
These all should happen in the mapping layer to correctly map the Async methods.

### Example

* [`page.goto(...)`](https://pkg.go.dev/github.com/grafana/xk6-browser@v0.7.0/common#Page.Goto) returns a promise.
* And the promise resolves to a`response` object.
* The `response` object [has a `frame()` method](https://pkg.go.dev/github.com/grafana/xk6-browser@v0.7.0/common#Response.Frame).
* Frame [has a `query` method](https://pkg.go.dev/github.com/grafana/xk6-browser@v0.7.0/common#Frame.Query) that we want to map to the `$` wildcard.

But we can't map the returned `response.frame()` method in the mapping layer because the `response` object is returned in a promise. We can't get the `response` object in the mapping layer until the promise is resolved (happening on the Goja side).

**To understand it better, let's take a look at the following.**

For example, the following code works: We use the `query` method to get the element.

```javascript
page.goto('https://test.k6.io/', { waitUntil: 'networkidle' }).then(response => {
  ...
  const f = response.frame();
  const el = f.query('a[href="/my_messages.php"]')
  console.log(el.textContent());
  ...
});
```

However, the following code does not work:

```javascript
[...]

PR #702, opened 2023-01-11
Expose VU to the mapping layer

inancgumus:
Exposing VU is preliminary for using promises in the mapping layer.

### Context

Normally [this code](https://github.com/grafana/xk6-browser/blob/update/697-js-examples-to-async/chromium/browser_type.go#L112) loads the VU so that the code in the `common` package can call [k6ext/promise](https://github.com/grafana/xk6-browser/blob/a11233766d5580b5ba8cf1ed5b7b9bb4dcfd8f66/k6ext/promise.go#L34-L39).

### Problem

Since we're moving the promise handling to the mapping layer (#683), the mapping layer needs a VU from the context to pass it to the `k6ext.Promise()`.

`k6ext.Promise()` gets the current VU from the context:

https://github.com/grafana/xk6-browser/blob/a11233766d5580b5ba8cf1ed5b7b9bb4dcfd8f66/k6ext/promise.go#L34-L39

### Solution

So, in the mapping layer, we need to inject VU into the context:

```go
func (*RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
    	...
    	ctx := k6ext.WithVU(vu.Context(), vu)
    	...
    	return &ModuleInstance{
		mod: &JSModule{
			Chromium: mapBrowserToGoja(ctx, vu),
			...
		},
	}
}
```

So that we can use `k6ext.Promise` in the mapping layer as follows:

```go
func mapPage(vu k6modules.VU, p api.Page) mapping {
    ...
    "goto": func(url string, opts goja.Value) *goja.Promise {
      // get the VU from the context
      return k6ext.Promise(ctx, func() (any, error) {
        resp, err := p.Goto(url, opts)
        if err != nil {
          return nil, err
        }
 
        return mapResponse(vu, resp), nil
      })
    },
    ...
```

PS: We can directly pass the VU to the `k6ext.Promise` after we finish our work in #683 (_after moving every promise code to the mapping layer_). But that would be another task than the mapping task as I don't know the implications of it yet—and we don't have enough time for it right now. Plus, I'm trying to make a gradual switch and keep the PRs small and focused.

Updates: #683

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -39,13 +39,12 @@ func wildcards() map[string]string {
 // methods.
 // See issue #661 for more details.
 func mapBrowserToGoja(vu k6modules.VU) *goja.Object {
-	rt := vu.Runtime()
-
 	var (
+		rt          = vu.Runtime()
 		obj         = rt.NewObject()
 		browserType = chromium.NewBrowserType(vu)
 	)
-	for k, v := range mapBrowserType(rt, browserType) {
+	for k, v := range mapBrowserType(vu, browserType) {
 		err := obj.Set(k, rt.ToValue(v))
 		if err != nil {
 			k6common.Throw(rt, fmt.Errorf("mapping: %w", err))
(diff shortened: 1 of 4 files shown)


inancgumus:
Blocked until the next PR that uses promises in the mapping layer.

inancgumus:
@ankur Please take another look at this PR.

inancgumus:
@ankur22 Please take another look at this PR.

ankur22 (approved):
LGTM. Thanks for taking the time to make the POC and taking the comments onboard 🙏

Issue #697, opened 2023-01-10
Convert promisified (.then()) examples to async/await


inancgumus:
This issue aims to convert non-await examples (`examples/`) to async/await counterparts.

~It seems like [the latest k6 version can't run our examples](https://github.com/grafana/xk6-browser/actions/runs/3875450875/jobs/6619503338) (the reason is not yet clear).~

We would also need to update the documentation to the new await/async API (@mdcruz 👀 🙇)

inancgumus:
Note: Also needed for https://github.com/grafana/k6-blog/pull/277.

PR #703, opened 2023-01-11
Convert promisified (.then()) examples to async/await

inancgumus:
Note that, IMO, `examples/browser_on.js` does not need to be converted to async/await. I experimented with it, but it didn't look good. Be my guest if you have any ideas :)

Closes: #697

### DIFF
--- a/examples/fillform.js
+++ b/examples/fillform.js
@@ -7,36 +7,35 @@ export const options = {
   }
 }
 
-export default function() {
+export default async function() {
   const browser = chromium.launch({
     headless: __ENV.XK6_HEADLESS ? true : false,
   });
   const context = browser.newContext();
   const page = context.newPage();
 
-  // Goto front page, find login link and click it
-  page.goto('https://test.k6.io/', { waitUntil: 'networkidle' }).then(() => {
-    return Promise.all([
+  try {
+    // Goto front page, find login link and click it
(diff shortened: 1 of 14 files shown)


ankur22 (approved):
LGTM.

Happy to leave the `browser_on.js` script as is for now since those promises are being handled like pub/sub async style workflows, whereas the other tests require a more synchronous style of workflow.

examples/fillform.js
```diff
@@ -7,36 +7,35 @@ export const options = {
...
   const context = browser.newContext();
   const page = context.newPage();
 
-  // Goto front page, find login link and click it
-  page.goto('https://test.k6.io/', { waitUntil: 'networkidle' }).then(() => {
-    return Promise.all([
+  try {
+    // Goto front page, find login link and click it
+    await page.goto('https://test.k6.io/', { waitUntil: 'networkidle' });
+    await Promise.all([
```
mdcruz, line 20: @inancgumus Just for my own understanding, can we also do?
```
await page.waitForNavigation()
await page.locator('a[href="/my_messages.php"]').click()
```

mdcruz (approved):
LGTM!

examples/fillform.js
inancgumus: Yeah, it should work, but `goto` and `waitForNavigation` have different implementations beneath. Btw, `locator.click` is not yet async.

PR #704, opened 2023-01-12
683 POC for unblocking 702

inancgumus:
@ankur22 check this out. I hope we can now unblock #702. Most tests would fail since this is about delivering the idea of #683. I didn't want to spend time fixing those for now.

For more context, please see the explanations I wrote before in the relevant issues:

- https://github.com/grafana/xk6-browser/pull/702
- https://github.com/grafana/xk6-browser/issues/683


common/frame.go
```diff
@@ -946,20 +946,25 @@ func (f *Frame) getAttribute(selector, name string, opts *FrameBaseOptions) (goj
...
-		applySlowMo(f.ctx)
-		return resp, err
-	})
+	resp, err := f.manager.NavigateFrame(f, url, parsedOpts)
+	if err != nil {
+		return nil, fmt.Errorf("navigating frame to %q: %w", url, err)
+	}
+	applySlowMo(f.ctx)
+
+	return resp, nil
```
inancgumus, line 967: Removing the promise handling from here and make it a native method that returns `Response` and `error` instead of a boxed `*goja.Promise`.

common/frame.go
```diff
@@ -946,20 +946,25 @@ func (f *Frame) getAttribute(selector, name string, opts *FrameBaseOptions) (goj
 }
 
 // Goto will navigate the frame to the specified URL and return a HTTP response object.
-func (f *Frame) Goto(url string, opts goja.Value) *goja.Promise {
-	netMgr := f.manager.page.mainFrameSession.getNetworkManager()
-	defaultReferer := netMgr.extraHTTPHeaders["referer"]
-	parsedOpts := NewFrameGotoOptions(
-		defaultReferer, time.Duration(f.manager.timeoutSettings.navigationTimeout())*time.Second)
+func (f *Frame) Goto(url string, opts goja.Value) (api.Response, error) {
```
inancgumus, line 949: `*goja.Promise` -> `(api.Response, error)`

tests/browser_context_options_test.go
```diff
@@ -76,15 +75,17 @@ func TestBrowserContextOptionsExtraHTTPHeaders(t *testing.T) {
...
+	require.Nil(t, err)
+	require.NotNil(t, resp)
+	err = tb.awaitWithTimeout(time.Second*5, func() error {
+		var body struct{ Headers map[string][]string }
+		require.NoError(t, json.Unmarshal(resp.Body().Bytes(), &body))
+
+		h := body.Headers["Some-Header"]
+		require.NotEmpty(t, h)
+		assert.Equal(t, "Some-Value", h[0])
+
```
inancgumus, line 88: The test becomes simple and clearer 🥳

browser/mapping.go
```diff
@@ -239,22 +243,56 @@ func mapFrame(vu k6modules.VU, f api.Frame) mapping {
...
+		//
+		"goto": func(url string, opts goja.Value) *goja.Promise {
+			return k6ext.Promise(ctx, func() (any, error) {
+				resp, err := f.Goto(url, opts)
+				if err != nil {
+					return nil, err //nolint:wrapcheck
+				}
+
+				return mapResponse(vu, resp), nil
+			})
```
inancgumus, line 277: Handling the promise in the mapping layer. `ctx` delivers `VU` to the `k6ext.Promise`.

common/frame_manager.go
```diff
@@ -663,6 +663,11 @@ func (m *FrameManager) NavigateFrame(frame *Frame, url string, parsedOpts *Frame
 		return nil, wrapTimeoutError(timeoutCtx.Err())
 	}
 
+	// fix the bug here in this code for the POC to pass
+	if resp == nil {
+		return nil, nil
+	}
+
```
inancgumus, line 670: Making things explicit surfaced a bug here. Isn't it cool?

common/frame.go
ankur22: This is great. I like that the promise is now part of the goja mapping instead of buried in the business logic 👍

common/frame.go
ankur22: Nice 👍

tests/browser_context_options_test.go
ankur22: Yep, this is a lot clearer 👏

common/frame.go
inancgumus: That's exactly the idea! 💯

browser/mapping.go
```diff
@@ -239,22 +246,56 @@ func mapFrame(vu k6modules.VU, f api.Frame) mapping {
...
+		// layer.
+		//
+		// Benefits:
+		// 1- This allows us to map the Response.
+		// 2- It's clearer, more manageable, and less error-prone
+		//    to have all the promise handling in one place
+		//    (in the mapping layer).
+		// 3- Tests are getting clearer (free from the promise
+		//    handling). See TestBrowserContextOptionsExtraHTTPHeaders
+		//    in browser_context_options_test.go.
```
ankur22, line 269: To be clear, i never had a problem with moving the promise to the mapping layer, this is an excellent idea 👏

inancgumus: Thanks, good to hear that we have been on the same page 🙇

browser/mapping.go
```diff
@@ -211,7 +215,10 @@ func mapElementHandle(vu k6modules.VU, eh api.ElementHandle) mapping {
 //
 //nolint:funlen
 func mapFrame(vu k6modules.VU, f api.Frame) mapping {
```
ankur22, line 217: In an attempt to make xk6-browser follow go good practices, could we not pass in `ctx` as the first param here (and all other mappings methods)?

```
func mapFrame(ctx, context.Context, vu k6modules.VU, f api.Frame) mapping {
```

That way we also don't need to wrap the `k6modules.VU` in a `moduleVU`. Please let me know if i've missed something here.

browser/mapping.go
```diff
@@ -211,7 +215,10 @@ func mapElementHandle(vu k6modules.VU, eh api.ElementHandle) mapping {
...
-	rt := vu.Runtime()
+	var (
+		rt  = vu.Runtime()
+		ctx = vu.Context()
```
ankur22, line 220: Do we need to create a new context in each mapping method?

ankur22 (commented):
@inancgumus thanks for this POC, this is awesome!

My only thought is around how we work with `ctx`, and whether we can try to pass in `ctx` as the first param to the mapping methods, so that it's a little more idiomatic go. WDYT?

browser/mapping.go
inancgumus: This doesn't create a new context but gets the existing wrapped context from`vu.Context()`. So it's the same context we set [here](https://github.com/grafana/xk6-browser/blob/3a4bee6af98006b544ae0a5af1267d23b1c472da/browser/module.go#L46-L50) only once. Were you asking this, or maybe I didn't get what you meant? 🤔

browser/mapping.go
inancgumus: I've done it like so as it's kind of a k6 standard, and it's less code :) Passing ctx as the first parameter is a dogma in the Go world. You can even see core Go developers use the embedded context pattern everywhere in their code. Then again, I'll think about this more and get back to you. Thanks for your review! 🙇

browser/mapping.go
inancgumus: @ankur22 please check [this](https://github.com/grafana/xk6-browser/pull/704/commits/089689d5e8a8b4422418b7abec109b55bd5399f4) out and tell me WDYT.

browser/module.go
```diff
@@ -71,9 +71,17 @@ func (*RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
 		k6common.Throw(vu.Runtime(), errors.New(msg))
 	}
 
+	// using our custom VU so that we can be ready for the future
+	// changes to the k6-core VU code. And we can have a fine-grained
+	// control over it, now and in the future.
+	//
+	// this puts the VU object in the context so that it can be
+	// used by inner objects such as k6ext.Promise.
+	vu = moduleVU{vu}
```
inancgumus, line 80: @ankur22 This is my rationale for using our own `moduleVU` instead of a plain VU.

browser/mapping.go
ankur22: I agree that passing around `ctx` is tedious, but it's less confusing than when we end up with multiple `ctx`s. We should only ever need to work with one `ctx`. 🙂

browser/mapping.go
ankur22: @inancgumus yes, that's what i think we should be doing. WDYT?

inancgumus: Ah, I got you now. You're right. I'll fix it.

browser/module.go
ankur22: It's stepping into the "over engineering" realm. Have we needed this before?

I'd be happy with it if this is all we ever needed to interact with and when we interacted with `vu` it would abstract away any of the `k6modules.VU` internals.

browser/mapping.go
inancgumus: Check [this](https://github.com/grafana/xk6-browser/pull/704/commits/5822656b1d4f06f92eae38eb8334122b54d39afe) out.

browser/module.go
inancgumus: Fair enough. See: https://github.com/grafana/xk6-browser/pull/704/commits/5822656b1d4f06f92eae38eb8334122b54d39afe

browser/mapping.go
ankur22: Yeah, i think that makes more sense, and is easier to reason with. 👍

inancgumus:
> @inancgumus thanks for this POC, this is awesome!
> 
> My only thought is around how we work with `ctx`, and whether we can try to pass in `ctx` as the first param to the mapping methods, so that it's a little more idiomatic go. WDYT?

Thanks for your review @ankur22. I missed the context-copying stuff, thanks for that as well 🙇  Since we got rid of the `moduleVU`, and passing `Context`; I'm moving the relevant code (_except the `page.goto` mapping part_) to #702.

ankur22 (approved):

inancgumus:
Closing the POC PR.

PR #712, opened 2023-01-13
Fix lifecycle tests parallelism bug

inancgumus:
So far, we haven't been testing the lifecycle tests. The test variables were captured because `t.Parallel`. This fixes it.

Related: #647, #644, #643


ka3de (approved):
LGTM.

PR #714, opened 2023-01-13
Fix frame manager navigate frame bug

inancgumus:
Returning an interface makes the `response` never `nil`. This creates problems when we want to check whether the `response` is `nil` or not.

Here, it never becomes `nil` even if it happens to be `nil`:

https://github.com/grafana/xk6-browser/blob/f2e001837a6b426fba95c873d4f97b3378da5d94/common/frame_manager.go#L644-L666

So we can't check whether it's `nil` or not here:

https://github.com/grafana/xk6-browser/blob/d2785768849aa7cc976eba27cb983cf2dd1475af/browser/mapping.go#L99-L102


ka3de (approved):
LGTM.

PR #715, opened 2023-01-17
Map [Page|ElementHandle|Frame|Locator].Click promises and fix tests

### LINKED ISSUE #707: Map [Page|ElementHandle|Frame|Locator].Click promises and fix tests
This issue is about mapping `Page.Click`, `Frame.Click`, etc. methods in the mapping layer.

See issue #683 for more details.

inancgumus:
Closes #707.

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -152,7 +152,7 @@ func mapElementHandle(ctx context.Context, vu k6modules.VU, eh api.ElementHandle
 		"click": func(opts goja.Value) *goja.Promise {
 			return k6ext.Promise(ctx, func() (any, error) {
 				err := eh.Click(opts)
-				return nil, fmt.Errorf("%w", err)
+				return nil, err //nolint:wrapcheck
 			})
 		},
 		"contentFrame": func() *goja.Object {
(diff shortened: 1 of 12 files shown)


browser/mapping.go
```diff
@@ -149,7 +149,12 @@ func mapElementHandle(ctx context.Context, vu k6modules.VU, eh api.ElementHandle
 		"objectID":       eh.ObjectID,
 		"boundingBox":    eh.BoundingBox,
 		"check":          eh.Check,
-		"click":          eh.Click,
+		"click": func(opts goja.Value) *goja.Promise {
+			return k6ext.Promise(ctx, func() (any, error) {
+				err := eh.Click(opts)
+				return nil, fmt.Errorf("%w", err)
```
inancgumus, line 155: Forgot: `//nolint:wrapcheck` here. Harmless, but, you know.

tests/frame_test.go
```diff
@@ -57,29 +54,11 @@ func TestFrameDismissDialogBox(t *testing.T) {
...
-
-					return nil
-				}).then(func() {
-					result := p.TextContent("#textField", nil)
-					assert.EqualValues(t, tt+" dismissed", result)
-				})
-
-				return nil
-			})
+			err = p.Click("#clickHere", nil)
```
inancgumus, line 57: @ankur22 This looks wrong to me, please double check 🤦

browser/mapping.go
ankur22: Maybe make it consistent. I prefer `fmt.Errorf("%w", err)` over `//nolint:wrapcheck`. I leave it you though 👍

ankur22 (approved):
LGTM

browser/mapping.go
inancgumus: We decided against that in sync back in July and disabled the wrapcheck linter. I'm confused about what changed your mind in that time frame 😄 I prefer the second, the first one is unnecessarily verbose and `%w` should not be preferred unless you want to unwrap the error. I'll go with the wrapcheck.

ankur22: I can't remember that far back :) Happy with the disabling of the wrapcheck linter. Why do we need the nolint still if the wrapcheck is disabled?

Issue #706, opened 2023-01-12
Map [Page|Frame].WaitForNavigation promises and fix tests


inancgumus:
This issue is about mapping `Page.WaitForNavigation` and `Frame.WaitForNavigation` methods in the mapping layer.

See issue #683 for more details.

PR #717, opened 2023-01-18
Map [Page|Frame].WaitForNavigation and fix tests

inancgumus:
This moves the promise handling for [Page|Frame].WaitForNavigation to the mapping layer. The lifecycle tests are still using promises since its logic is complex to figure out for now. We should refactor them when we finish—I've added a task on #683.

Related: #683 
Closes: #706


ankur22 (approved):
LGTM

PR #718, opened 2023-01-18
Refactor the Launch code to work with errors

ankur22:
This is a POC and proves that this does work in the same way as if we panicked.


inancgumus:
@ankur22, can we close this one if it was only an experiment? You can still refer to it in another comment you wrote (I don't remember where it was).

ankur22:
Closing as similar examples are now in `main`.

PR #719, opened 2023-01-19
Add support for BrowserContext.addInitScript

ka3de:
Closes #8.

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -482,8 +482,13 @@ func mapPage(ctx context.Context, vu k6modules.VU, p api.Page) mapping {
 func mapBrowserContext(ctx context.Context, vu k6modules.VU, bc api.BrowserContext) mapping {
 	rt := vu.Runtime()
 	return mapping{
-		"addCookies":                  bc.AddCookies,
-		"addInitScript":               bc.AddInitScript,
+		"addCookies": bc.AddCookies,
+		"addInitScript": func(script goja.Value, arg goja.Value) *goja.Promise {
+			return k6ext.Promise(ctx, func() (any, error) {
+				err := bc.AddInitScript(script, arg)
+				return nil, err //nolint:wrapcheck
+			})
+		},
 		"browser":                     bc.Browser,
 		"clearCookies":                bc.ClearCookies,
 		"clearPermissions":            bc.ClearPermissions,
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -98,7 +98,7 @@ func (b *BrowserContext) AddInitScript(script goja.Value, arg goja.Value) error 
 	case reflect.Func:
 		source, err = b.parseFuncInitScript(script, arg, isCallable, hasArgs)
 	default:
-		return errors.New("parsing init script: invalid script type")
+		err = errors.New("parsing init script: invalid script type")
 	}
 
 	if err != nil {
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -78,7 +78,7 @@ func (b *BrowserContext) AddInitScript(script goja.Value, arg goja.Value) error 
 	b.logger.Debugf("BrowserContext:AddInitScript", "bctxid:%v", b.id)
 
 	if !gojaValueExists(script) {
-		return errors.New("parsing init script: invalid script")
+		return errors.New("parsing init script: script is undefined or null")
 	}
 
 	_, isCallable := goja.AssertFunction(script)
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -149,10 +149,10 @@ func (b *BrowserContext) parseFuncInitScript(script goja.Value, arg goja.Value,
 		if err != nil {
 			return "", fmt.Errorf("marshaling arg: %v", err)
 		}
-		return fmt.Sprintf("(%s)(%s);", script.ToString().String(), string(input)), nil
+		return fmt.Sprintf("(%s)(%s);", script.ToString(), string(input)), nil
 	}
 
-	return fmt.Sprintf("(%s)();", script.ToString().String()), nil
+	return fmt.Sprintf("(%s)();", script.ToString()), nil
 }
 
 // Browser returns the browser instance that this browser context belongs to.
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -131,7 +131,7 @@ func (b *BrowserContext) parseObjectInitScript(script goja.Value) (string, error
 			path := obj.Get(k).String()
 			source, err := readFile(path)
 			if err != nil {
-				return "", fmt.Errorf("reading file: %v", err)
+				return "", fmt.Errorf("reading file: %v", err) //nolint:errorlint
 			}
 			// See sourcemaps (https://developer.chrome.com/blog/sourcemaps/).
 			return fmt.Sprint(source, "\n//# sourceURL=", path), nil
@@ -147,7 +147,7 @@ func (b *BrowserContext) parseFuncInitScript(script goja.Value, arg goja.Value,
 	if isCallable && hasArgs {
 		input, err := json.Marshal(arg.Export())
 		if err != nil {
-			return "", fmt.Errorf("marshaling arg: %v", err)
+			return "", fmt.Errorf("marshaling arg: %v", err) //nolint:errorlint
 		}
 		return fmt.Sprintf("(%s)(%s);", script.ToString(), string(input)), nil
 	}
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -84,8 +84,11 @@ func (b *BrowserContext) AddInitScript(script goja.Value, arg goja.Value) error 
 	_, isCallable := goja.AssertFunction(script)
 	hasArgs := gojaValueExists(arg)
 
-	var err error
-	var source string
+	var (
+		err    error
+		source string
+	)
+
 	//nolint:exhaustive
 	switch script.ExportType().Kind() {
 	case reflect.String:
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -147,7 +147,7 @@ func (b *BrowserContext) parseFuncInitScript(script goja.Value, arg goja.Value,
 	if isCallable && hasArgs {
 		input, err := json.Marshal(arg.Export())
 		if err != nil {
-			return "", fmt.Errorf("can't marshal arg: %w", err)
+			return "", fmt.Errorf("marshaling arg: %v", err)
 		}
 		return fmt.Sprintf("(%s)(%s);", script.ToString().String(), string(input)), nil
 	}
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -20,8 +20,10 @@ import (
 )
 
 // Ensure BrowserContext implements the EventEmitter and api.BrowserContext interfaces.
-var _ EventEmitter = &BrowserContext{}
-var _ api.BrowserContext = &BrowserContext{}
+var (
+	_ EventEmitter       = &BrowserContext{}
+	_ api.BrowserContext = &BrowserContext{}
+)
 
 // BrowserContext stores context information for a single independent browser session.
 // A newly launched browser instance contains a default browser context.
@@ -70,7 +72,7 @@ func (b *BrowserContext) AddCookies(cookies goja.Value) *goja.Promise {
 }
 
 // AddInitScript adds a script that will be initialized on all new pages.
-func (b *BrowserContext) AddInitScript(script goja.Value, arg goja.Value) {
+func (b *BrowserContext) AddInitScript(script goja.Value, arg goja.Value) error {
 	b.logger.Debugf("BrowserContext:AddInitScript", "bctxid:%v", b.id)
 
 	rt := b.vu.Runtime()
@@ -101,8 +103,12 @@ func (b *BrowserContext) AddInitScript(script goja.Value, arg goja.Value) {
 	b.evaluateOnNewDocumentSources = append(b.evaluateOnNewDocumentSources, source)
 
 	for _, p := range b.browser.getPages() {
-		p.evaluateOnNewDocument(source)
+		if err := p.evaluateOnNewDocument(source); err != nil {
+			return fmt.Errorf("adding init script: %w", err)
+		}
 	}
+
+	return nil
 }
 
 // Browser returns the browser instance that this browser context belongs to.
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -83,13 +83,20 @@ func (b *BrowserContext) AddInitScript(script goja.Value, arg goja.Value) error 
 		return errors.New("parsing init script: arguments are not supported")
 	}
 
+	var err error
 	var source string
 	//nolint:exhaustive
 	switch script.ExportType().Kind() {
 	case reflect.String:
 		source = script.String()
+	case reflect.Map, reflect.Struct:
+		source, err = b.parseObjectInitScript(script)
 	default:
 		return errors.New("parsing init script: invalid script type")
+	}
+
+	if err != nil {
+		return fmt.Errorf("parsing init script: %w", err)
 	}
 
 	b.evaluateOnNewDocumentSources = append(b.evaluateOnNewDocumentSources, source)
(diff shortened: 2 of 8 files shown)


common/browser_context.go
```diff
@@ -73,38 +74,74 @@ func (b *BrowserContext) AddCookies(cookies goja.Value) *goja.Promise {
 func (b *BrowserContext) AddInitScript(script goja.Value, arg goja.Value) {
 	b.logger.Debugf("BrowserContext:AddInitScript", "bctxid:%v", b.id)
 
-	rt := b.vu.Runtime()
+	if script == nil || goja.IsUndefined(script) || goja.IsNull(script) {
+		k6ext.Panic(b.ctx, "parsing init script: invalid script")
+	}
```
inancgumus, line 79: Here, you should return an error and use the helper `gojaValueExists`. Please see the PRs of the sub-issues of #683 for more details (for returning errors and mapping them in the mapping layer).

common/browser_context.go
```diff
@@ -21,8 +21,10 @@ import (
 )
 
 // Ensure BrowserContext implements the EventEmitter and api.BrowserContext interfaces.
-var _ EventEmitter = &BrowserContext{}
-var _ api.BrowserContext = &BrowserContext{}
+var (
+	_ EventEmitter       = &BrowserContext{}
+	_ api.BrowserContext = &BrowserContext{}
+)
```
inancgumus, line 27: 👍 for this grouping.

common/browser_context.go
```diff
@@ -82,6 +84,7 @@ func (b *BrowserContext) AddInitScript(script goja.Value, arg goja.Value) {
 	hasArgs := arg != nil && !goja.IsUndefined(arg) && !goja.IsNull(arg)
 
 	var source string
+	//nolint:exhaustive
```
inancgumus, line 87: Keep this in mind while adding nolint: https://github.com/grafana/xk6-browser/issues/58 :)

ka3de: Got it! Does that mean that we should try to apply the lint suggestions? In this case there are quite a few missing cases and for me it was "polluting" the code, especially considering that we have a default case.

common/browser_context.go
ka3de: I actually wanted to maintain the current format to comply with the rest of the code, but `gofumpt` was applying this change automatically, so I guess we should progressively apply this "rule" for the rest :)

common/browser_context.go
ka3de: So if I understand correctly, that implies making the `addInitScript` method return a promise in the JS API right?

ka3de: Implemented in [a3c7e31](https://github.com/grafana/xk6-browser/pull/719/commits/a3c7e31431db1363bb590441170488e1ddb2b4ee).

common/browser_context.go
```diff
@@ -70,39 +74,82 @@ func (b *BrowserContext) AddCookies(cookies goja.Value) *goja.Promise {
...
 
 // AddInitScript adds a script that will be initialized on all new pages.
-func (b *BrowserContext) AddInitScript(script goja.Value, arg goja.Value) {
+func (b *BrowserContext) AddInitScript(script goja.Value, arg goja.Value) error {
 	b.logger.Debugf("BrowserContext:AddInitScript", "bctxid:%v", b.id)
 
-	rt := b.vu.Runtime()
+	if !gojaValueExists(script) {
+		return errors.New("parsing init script: invalid script")
+	}
```
inancgumus, line 82: Suggestion: We better put this in a previous refactoring commit first. All kinds of refactoring should be in a refactoring commit (or multiple ones) to make the new feature addition ready—as long as it makes sense. The reasoning is it makes it harder to see the actual changes (the goal of the PR)—the difference between the refactorings and the core changes.

_I know you did this because of my previous comment: I'm trying to educate you on how we work._

common/browser_context.go
```diff
@@ -70,39 +74,82 @@ func (b *BrowserContext) AddCookies(cookies goja.Value) *goja.Promise {
...
-				source = fmt.Sprintf("(%s);", script.ToString().String())
-			} else {
-				source = fmt.Sprintf("(%s)(...args);", script.ToString().String())
+	_, isCallable := goja.AssertFunction(script)
+	hasArgs := gojaValueExists(arg)
+
+	var err error
+	var source string
+	//nolint:exhaustive
+	switch script.ExportType().Kind() {
```
inancgumus, line 90: This could go to a refactoring commit first.

`script.ExportType()` -> `...Kind()` and the cases.

Also, the `parseObjectInitScript` function, could go to a refactoring one.

common/browser_context.go
```diff
@@ -70,39 +74,82 @@ func (b *BrowserContext) AddCookies(cookies goja.Value) *goja.Promise {
...
+	case reflect.String:
+		source = script.String()
+	case reflect.Map, reflect.Struct:
```
inancgumus, line 93: What is the reasoning behind adding `Map` here? Do `Map` and `Struct` cover the previous `reflect.TypeOf(goja.Object{})`? Should we add more types or are `Map` and `Struct` enough?

inancgumus (commented):
By the way, I was reviewing the changes, then everything changed—I guess you force-pushed more things. I'll do another review if you have finished 😅 

My notes:
Refactorings or betterments such as changing the current logic or adding more functions could go to specific refactoring commits. Better yet, we do no refactoring at this stage, add the core feature first (what the issue aims), and then do the refactoring in subsequent commits or PR if it makes sense. If doing them in subsequent commits/PR won't make sense, then we better do these refactoring first with the explanations in the commit why we've done so. Otherwise, it makes it harder to see what's changed and what's to be reviewed. We can easily make mistakes while reviewing the code.

common/browser_context.go
```diff
@@ -70,39 +74,82 @@ func (b *BrowserContext) AddCookies(cookies goja.Value) *goja.Promise {
...
+	return nil
+}
+
+func (b *BrowserContext) parseObjectInitScript(script goja.Value) (string, error) {
+	obj := script.ToObject(b.vu.Runtime())
+	for _, k := range obj.Keys() {
+		switch k {
+		case "content":
+			return obj.Get(k).String(), nil
+		case "path": //nolint:goconst
```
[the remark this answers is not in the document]
ka3de: I personally don't see the benefit of defining a constant for an element that appears only once in the code. I think the linter here is being confused by the presence of `path` literal in the error messages, but defining a constant and using it there I think would make it even worse. What do you think?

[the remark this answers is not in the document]
inancgumus: Yeah, not that critical here.

common/browser_context.go
```diff
@@ -102,6 +109,26 @@ func (b *BrowserContext) AddInitScript(script goja.Value, arg goja.Value) error
...
+			path := obj.Get(k).String()
+			source, err := readFile(path)
+			if err != nil {
+				return "", fmt.Errorf("can't read script from path '%s': %w", path, err)
```
inancgumus, line 131: The error stutters: `parsing init script: can't read script from path 'static/dinit_script.js': open /Users/inanc/grafana/k6browser/main/tests/static/dinit_script.js: no such file or directory`.

We can make it concise and informative at the same time:

```suggestion
				return "", fmt.Errorf("reading file: %v", err)
```

This will result in the following:

`parsing init script: reading file: open /Users/inanc/grafana/k6browser/main/tests/static/dinit_script.js: no such file or directory`

tests/browser_context_test.go
```diff
@@ -0,0 +1,302 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBrowserContextAddInitScript(t *testing.T) {
+	t.Parallel()
+
+	t.Run("string_script_on_new_page", func(t *testing.T) {
```
inancgumus, line 15: Consider putting similar tests in a table test—they have more or less the same body. Let them accept differing parts from the fields. You can add more table tests if things become hard to read.

tests/browser_context_test.go
```diff
@@ -0,0 +1,302 @@
...
+			t.Run(tt.name, func(t *testing.T) {
+				t.Parallel()
+
+				tb := newTestBrowser(t, withFileServer())
+				bctx := tb.NewContext(nil)
+				t.Cleanup(bctx.Close)
+
+				rt := tb.vu.Runtime()
+				_, err := rt.RunString(tt.argScript)
+				require.NoError(t, err)
```
inancgumus, line 153: You can use `tb.runJavaScript()` here and in the other places.

common/browser_context.go
```diff
@@ -75,33 +76,23 @@ func (b *BrowserContext) AddCookies(cookies goja.Value) *goja.Promise {
...
-				source = fmt.Sprintf("(%s)(...args);", script.ToString().String())
-			}
-		}
+	var source string
+	//nolint:exhaustive
+	switch script.ExportType().Kind() {
+	case reflect.String:
+		source = script.String()
+	default:
+		return errors.New("parsing init script: invalid script type")
```
inancgumus, line 98: Removing the switch cases here first and then adding them (mixed in new funcs) in the subsequent commit makes it hard to follow. Let's leave them here as they were, make minimal changes to the code to pass, and then add new code (a new commit to refactoring functions). While working on this, you have all the context, but the reviewer has not :)

common/page.go
```diff
@@ -166,8 +168,20 @@ func (p *Page) didCrash() {
 	p.emit(EventPageCrash, p)
 }
 
-func (p *Page) evaluateOnNewDocument(source string) {
-	// TODO: implement
+func (p *Page) evaluateOnNewDocument(source string) error {
```
inancgumus, line 179: This change could go to a single commit (maybe the first commit as it's the actual goal of the PR).

common/browser_context.go
```diff
@@ -129,6 +138,20 @@ func (b *BrowserContext) parseObjectInitScript(script goja.Value) (string, error
 	return "", errors.New("content or path properties must be set for script as object")
 }
 
+func (b *BrowserContext) parseFuncInitScript(script goja.Value, arg goja.Value,
+	isCallable, hasArgs bool,
+) (string, error) {
+	if isCallable && hasArgs {
+		input, err := json.Marshal(arg.Export())
+		if err != nil {
+			return "", fmt.Errorf("can't marshal arg: %w", err)
```
inancgumus, line 147: Let's use `-ing` as it's a convention we use. Do we need to wrap the error?

```suggestion
			return "", fmt.Errorf("marshaling arg: %v", err)
```

common/browser_context.go
```diff
@@ -129,6 +138,20 @@ func (b *BrowserContext) parseObjectInitScript(script goja.Value) (string, error
...
+		}
+		return fmt.Sprintf("(%s)(%s);", script.ToString().String(), string(input)), nil
```
inancgumus, line 149: `ToString()` is enough here:

```suggestion
		return fmt.Sprintf("(%s)(%s);", script.ToString(), string(input)), nil
```

common/browser_context.go
```diff
@@ -129,6 +138,20 @@ func (b *BrowserContext) parseObjectInitScript(script goja.Value) (string, error
...
+	}
+
+	return fmt.Sprintf("(%s)();", script.ToString().String()), nil
```
inancgumus, line 152: Also here:

```suggestion
	return fmt.Sprintf("(%s)();", script.ToString()), nil
```

common/browser_context.go
```diff
@@ -481,3 +481,9 @@ func (b *BrowserContext) runWaitForEventHandler(
 func (b *BrowserContext) getSession(id target.SessionID) *Session {
 	return b.browser.conn.getSession(id)
 }
+
+func (b *BrowserContext) getEvaluateOnNewDocumentSources() []string {
+	sources := make([]string, len(b.evaluateOnNewDocumentSources))
+	copy(sources, b.evaluateOnNewDocumentSources)
+	return sources
+}
```
inancgumus, line 489: Could you explain the reasoning for the copying here? I suggest adding a comment to explain.

By the way, this commit also makes sense to be in this [commit](4d75cfd3f5a9cc486d1334fc9eb232a94b1d592c) as they both deal with the same behavior: `evaluateOnNewDocument`.

inancgumus (commented):
Thanks, Daniel for the new changes :) We still need to make some more adjustments to the commits. I'm keeping this as through as possible, so the next PRs will be smooth.

common/browser_context.go
```diff
@@ -84,8 +84,11 @@ func (b *BrowserContext) AddInitScript(script goja.Value, arg goja.Value) error
 	_, isCallable := goja.AssertFunction(script)
 	hasArgs := gojaValueExists(arg)
 
-	var err error
-	var source string
+	var (
+		err    error
+		source string
+	)
```
inancgumus, line 90: @ka3de You could do this refactoring in [this commit](https://github.com/grafana/xk6-browser/pull/719/commits/a91153d75c12579d785881bbf359464264e9c799) as it's a minor change. Refactoring/change commits are mostly needed when things can be confusing (for example, removing the switch cases and readding them in functions in a subsequent commit—as I explained in another comment).

common/browser_context.go
```diff
@@ -147,7 +147,7 @@ func (b *BrowserContext) parseFuncInitScript(script goja.Value, arg goja.Value,
 	if isCallable && hasArgs {
 		input, err := json.Marshal(arg.Export())
 		if err != nil {
-			return "", fmt.Errorf("can't marshal arg: %w", err)
+			return "", fmt.Errorf("marshaling arg: %v", err)
```
inancgumus, line 150: Linter is confused. Deserves a `nolint:wrapcheck` here.

ka3de:
Hey @inancgumus I addressed some of the comments. It would be good if you can take a look when you have some time so we can clean the commits and open discussions before force pushing to fix the other issues. Thx!

browser/mapping.go
```diff
@@ -482,8 +482,13 @@ func mapPage(ctx context.Context, vu k6modules.VU, p api.Page) mapping {
 func mapBrowserContext(ctx context.Context, vu k6modules.VU, bc api.BrowserContext) mapping {
 	rt := vu.Runtime()
 	return mapping{
-		"addCookies":                  bc.AddCookies,
-		"addInitScript":               bc.AddInitScript,
+		"addCookies": bc.AddCookies,
+		"addInitScript": func(script goja.Value, arg goja.Value) *goja.Promise {
+			return k6ext.Promise(ctx, func() (any, error) {
+				err := bc.AddInitScript(script, arg)
+				return nil, err //nolint:wrapcheck
```
ankur22, line 489: Can we just return an error?

common/helpers.go
```diff
@@ -240,3 +242,22 @@ func asGojaValue(ctx context.Context, v any) goja.Value {
 func gojaValueToString(ctx context.Context, v any) string {
 	return asGojaValue(ctx, v).String()
 }
+
+// readFile reads a file from FS. If path is relative, it's
+// converted to absolute path based on current working dir.
+func readFile(path string) (string, error) {
```
ankur22, line 248: Actually, i think it might be best to drop the readFile functionality. It will be difficult to work with in the cloud. We can add it back in if customer's ask for it. It's also a security risk since it looks like we can read any (permissions granted) file from the filesystem.

examples/browsercontext_addinitscript.js
```diff
@@ -0,0 +1,32 @@
...
+
+export default async function() {
+  const browser = chromium.launch({
+    headless: __ENV.XK6_HEADLESS ? true : false,
+  });
+  const context = browser.newContext();
+  await context.addInitScript(
+    `(function () {
+        document.open();
+        document.write("k6Test");
```
ankur22, line 18: I think we're going to have to restrict where we read/write files from/to. Even though the browser might be dockerised, it's worth mitigating issues by narrowing the scope of where customer's can read/write.

ankur22 (changes requested):
I think in general it looks good 🙂👍 

Unfortunately we will need to reduce the scope of what init scripts can do so that we help mitigate some issues. I also think we should remove the read from file functionality since we don't support (i might be wrong) a way of uploading or creating other files along with the test script. Sorry about that.

inancgumus:
Agreed with the team to close this PR.

Issue #709, opened 2023-01-12
Map Browser.On promises and fix tests


inancgumus:
This issue is about mapping `Browser.On` method in the mapping layer.

See issue #683 for more details.

Issue #675, opened 2022-12-15
Add JS tests


inancgumus:
TBA

I will fill this out when I'm back. Just a reminder for now.

Related: #661

inancgumus:
@grafana/k6-browser I did some experiments with this in #778. However, since we deprioritized it, I removed my assignment for now.

PR #722, opened 2023-01-20
Map Browser.On and fix test

ankur22:
This moves the promise handling of `Browser.on` to the mapping layer.

Ideally, a test should be written by the JavaScript side. See issue #675 for details. The reason is, they are completely concerned about the JS-side of things and the Go tests become futile, complicated and hard to understand.

Closes https://github.com/grafana/xk6-browser/issues/709. Updates https://github.com/grafana/xk6-browser/issues/683.

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -533,9 +533,17 @@ func mapBrowser(ctx context.Context, vu k6modules.VU, b api.Browser) mapping {
 		"close":       b.Close,
 		"contexts":    b.Contexts,
 		"isConnected": b.IsConnected,
-		"on":          b.On,
-		"userAgent":   b.UserAgent,
-		"version":     b.Version,
+		"on": func(event string) *goja.Promise {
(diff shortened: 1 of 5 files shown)


browser/mapping.go
```diff
@@ -533,9 +533,17 @@ func mapBrowser(ctx context.Context, vu k6modules.VU, b api.Browser) mapping {
...
-		"on":          b.On,
-		"userAgent":   b.UserAgent,
-		"version":     b.Version,
+		"on": func(event string) *goja.Promise {
+			return k6ext.Promise(ctx, func() (result any, reason error) {
+				res, err := b.On(event)
+				if err != nil {
+					return nil, err //nolint:wrapcheck
+				}
+				return res, nil
```
inancgumus, line 542: The following is the same (less code is less things to maintain):

```suggestion
return b.On(event) //nolint:wrapcheck
```

inancgumus (approved):
Thanks @ankur22. Only a small suggestion.

PR #723, opened 2023-01-20
Remove await method from testBrowser

ankur22:
Now that promises have been moved to a level above the business logic, we can remove the need to run things in promises and assume (in most cases) that things will run synchronously. Where there's still a need to run things in parallel we can use the run method on testBrowser.

Updates https://github.com/grafana/xk6-browser/issues/683.

### DIFF
--- a/tests/lifecycle_wait_test.go
+++ b/tests/lifecycle_wait_test.go
@@ -13,7 +13,6 @@ import (
 
 	"github.com/grafana/xk6-browser/api"
 	"github.com/grafana/xk6-browser/common"
-	"github.com/grafana/xk6-browser/k6ext"
 )
 
 // TODO
@@ -71,7 +70,7 @@ func TestLifecycleWaitForNavigation(t *testing.T) {
 		waitUntil             common.LifecycleEvent
 		pingRequestTextAssert func(result string, pingCount int)
 		pingJSTextAssert      func(result string)
-		assertFunc            func(tb *testBrowser, p api.Page) testPromise
+		assertFunc            func(tb *testBrowser, p api.Page) error
 		wantError             string
 	}{
 		{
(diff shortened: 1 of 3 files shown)


tests/lifecycle_wait_test.go
```diff
@@ -126,21 +125,24 @@ func TestLifecycleWaitForNavigation(t *testing.T) {
...
+					opts := tb.toGojaValue(&common.FrameWaitForNavigationOptions{
+						Timeout:   1000,
+						WaitUntil: common.LifecycleEventNetworkIdle,
+					})
+					_, err := p.WaitForNavigation(opts)
+					return err
+				}
 
-				return tb.promise(waitForNav)
+				ctx, cancel := context.WithTimeout(tb.ctx, 5*time.Second)
```
inancgumus: Should the following be like:

```suggestion
				ctx, cancel := context.WithTimeout(tb.ctx, time.Second)
```

As the error message on line 145 says: _"waiting for navigation: timed out after **1s**"._

tests/lifecycle_wait_test.go
```diff
@@ -181,13 +183,10 @@ func TestLifecycleWaitForNavigation(t *testing.T) {
...
 					return p.Click(`a`, nil)
 				}
-				runAll := k6ext.Promise(tb.ctx, func() (result any, reason error) {
-					ctx, cancel := context.WithTimeout(tb.ctx, 5*time.Second)
-					defer cancel()
-					return nil, tb.run(ctx, waitForNav, click)
-				})
 
-				return tb.promise(runAll)
+				ctx, cancel := context.WithTimeout(tb.ctx, 5*time.Second)
```
inancgumus, line 182: 5 seconds or 1 sec as the previous suggestion?

inancgumus (approved):
Nice that we got rid of this method as well. One minor suggestion there. Fix the linter errors, squash, and 🚀

tests/lifecycle_wait_test.go
ankur22: The `timeout` test runs an `asserFunc` instead of running the main body of the test, so it completes before reaching the 5 second timeout step.

tests/lifecycle_wait_test.go
ankur22: The timeout test runs an asserFunc instead of running the main body of the test, so it completes before reaching the 5 second timeout step.

tests/lifecycle_wait_test.go
inancgumus: Oh, it's good then 👍

PR #726, opened 2023-01-24
Map forgotten types

### LINKED ISSUE #725: Map forgotten mappings: JSHandle and Worker
Map forgotten mappings: [JSHandle](https://github.com/grafana/xk6-browser/blob/e433cb765803dffcb77f1032ac7bd369c9e415da/api/js_handle.go) and Worker. Otherwise, we'll be delivering a half-baked feature.

We should have caught this while reviewing the early PRs, but we missed it. It happens, I understand.

`Route` is not implemented yet, and it's complicated to tackle right now. We can do it afterward (see [this](https://github.com/grafana/xk6-browser/issues/10)).

Related: #683 #661

inancgumus:
Closes #725

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -130,25 +130,50 @@ func mapResponse(ctx context.Context, vu k6modules.VU, r api.Response) mapping {
 	return maps
 }
 
+// mapJSHandle to the JS module.
+func mapJSHandle(ctx context.Context, vu k6modules.VU, jsh api.JSHandle) mapping {
+	rt := vu.Runtime()
+	return mapping{
+		"asElement": func() *goja.Object {
+			m := mapElementHandle(ctx, vu, jsh.AsElement())
+			return rt.ToValue(m).ToObject(rt)
+		},
+		"dispose":  jsh.Dispose,
+		"evaluate": jsh.Evaluate,
+		"evaluateHandle": func(pageFunc goja.Value, args ...goja.Value) *goja.Object {
+			var (
+				h = jsh.EvaluateHandle(pageFunc, args...)
+				m = mapJSHandle(ctx, vu, h)
+			)
+			return rt.ToValue(m).ToObject(rt)
+		},
+		"getProperties": func() *goja.Object {
+			dst := make(map[string]any)
+			for k, v := range jsh.GetProperties() {
+				dst[k] = mapJSHandle(ctx, vu, v)
+			}
+			return rt.ToValue(dst).ToObject(rt)
+		},
+		"getProperty": func(propertyName string) *goja.Object {
+			var (
+				h = jsh.GetProperty(propertyName)
+				m = mapJSHandle(ctx, vu, h)
+			)
+			return rt.ToValue(m).ToObject(rt)
+		},
+		"jsonValue": jsh.JSONValue,
+		"objectID":  jsh.ObjectID,
+	}
+}
+
 // mapElementHandle to the JS module.
 //
 //nolint:funlen
 func mapElementHandle(ctx context.Context, vu k6modules.VU, eh api.ElementHandle) mapping {
 	rt := vu.Runtime()
 	maps := mapping{
-		"asElement": func() *goja.Object {
-			m := mapElementHandle(ctx, vu, eh.AsElement())
-			return rt.ToValue(m).ToObject(rt)
-		},
-		"dispose":        eh.Dispose,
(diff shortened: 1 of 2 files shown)


browser/mapping.go
```diff
@@ -130,25 +130,50 @@ func mapResponse(ctx context.Context, vu k6modules.VU, r api.Response) mapping {
...
+			return rt.ToValue(dst).ToObject(rt)
+		},
+		"getProperty": func(propertyName string) *goja.Object {
+			var (
+				h = eh.GetProperty(propertyName)
+				m = mapJSHandle(ctx, vu, h)
+			)
+			return rt.ToValue(m).ToObject(rt)
+		},
+		"jSONValue": eh.JSONValue,
```
ankur22, line 164: This should be `jsonValue`, no?

Reference: https://playwright.dev/docs/api/class-jshandle#js-handle-json-value

browser/mapping.go
```diff
@@ -130,25 +130,50 @@ func mapResponse(ctx context.Context, vu k6modules.VU, r api.Response) mapping {
 	return maps
 }
 
+// mapJSHandle to the JS module.
+func mapJSHandle(ctx context.Context, vu k6modules.VU, eh api.JSHandle) mapping {
```
ankur22, line 134: Minor: rename from `eh` to `jsh`?

browser/mapping.go
```diff
@@ -214,6 +239,11 @@ func mapElementHandle(ctx context.Context, vu k6modules.VU, eh api.ElementHandle
 		return rt.ToValue(mehs).ToObject(rt)
 	}
 
+	jsHandleMap := mapJSHandle(ctx, vu, eh)
```
ankur22, line 242: Why do we implement `JSHandle` in `ElementHandle`? Playwright doesn't do this -- https://playwright.dev/docs/api/class-elementhandle

ankur22 (changes requested):
Thanks for this.

A couple of minor comments, and one comment that i think we need to think about and we've missed when doing the original mapping -- why does `ElementHandle` implement `JSHandle`?

browser/mapping.go
inancgumus: If remove mapping of the `JSHandle` from the `ElementHandle`, our API would change. I want to keep the current code intact (the following). We can deal with this later if it becomes necessary.

https://github.com/grafana/xk6-browser/blob/e433cb765803dffcb77f1032ac7bd369c9e415da/api/element_handle.go#L6-L9

Also, `mapJSHandle` maps the `JSHandle` part of the `ElementHandle`. So we don't need to do it twice.

browser/mapping.go
inancgumus: Resolved [here](https://github.com/grafana/xk6-browser/blob/d50f4ba1a7934371c122116fdcf2c8653618a08d/browser/mapping.go#L134).

browser/mapping.go
inancgumus: Resolved [here](https://github.com/grafana/xk6-browser/blob/d50f4ba1a7934371c122116fdcf2c8653618a08d/browser/mapping.go#L164).

inancgumus:
> i think we need to think about and we've missed when doing the original mapping -- why does `ElementHandle` implement `JSHandle`?

We didn't miss it: It's what the current API is. The mapping is about mapping the current API. It's not about changing behavior. More details are [here](https://github.com/grafana/xk6-browser/pull/726#discussion_r1085093796).

browser/mapping.go
ankur22: Hmm, my preference would be to break the API so that we match Playwright. I had a look at the comments from past commits and I can't see any reason as to why we would need it. These APIs aren't documented anywhere, and only users who checked the source code would likely find them. The worst case scenario is that some users would complain that the API has broken their tests, but we can at that point put it back in -- this would prove that it is in fact useful.

I feel that this is the best time to remove this, since we're going to only add more xk6-browser users after the merge, which will make breaking things a little more challenging.

WDYT?

inancgumus: I'm on the side of not complicating things right now. It's not a high priority—but the mapping is. We shouldn't block a thing that is not related to this PR and mapping in general.

We preserve the current behavior. If you want to change the API, be my guest: I've created a new issue: #728.

ankur22: Ok, np 🙂

ankur22 (approved):

PR #727, opened 2023-01-24
Run test and e2e CI actions daily

### LINKED ISSUE #701: Auto-run CI tasks when k6-core updates
Auto-running our CI tasks when k6 makes an update will allow us to detect the breaking changes as soon as possible. We might also think about doing this daily instead of per push by the k6-core.

ka3de:
This PR modifies the `e2e` and `test` GH actions in order to trigger their execution on a daily basis. The motivation for this is to increase the feedback loop for possible braking changes due to modifications done in k6 core.

Closes #701.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -6,6 +6,9 @@ on:
     branches:
       - main
   pull_request:
+  schedule:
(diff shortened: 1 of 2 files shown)


ankur22 (approved):
LGTM 👍

PR #734, opened 2023-01-27
Reduce dependencies

### LINKED ISSUE #729: Reduce dependency usage
Some dependencies currently required by the project are marginally used and could potentially be removed. This acquires more relevance with the upcoming merge with k6 core due to the importance that is placed in this project to minimize the dependencies usage and limit its heterogeneity in order to prevent malfunctions between extensions code.

Some examples for dependencies that could potentially be reevaluated are:
- [github.com/hashicorp/go-multierror](https://github.com/grafana/xk6-browser/blob/5bd4fa1d7d9c9345eeee8107a97c021b7fee08b2/common/remote_object.go#L16)
- [github.com/fatih/color](https://github.com/grafana/xk6-browser/blob/5bd4fa1d7d9c9345eeee8107a97c021b7fee08b2/log/logger.go)
- ~~[github.com/mccutchen/go-httpbin/httpbin](https://github.com/grafana/xk6-browser/blob/5bd4fa1d7d9c9345eeee8107a97c021b7fee08b2/tests/ws/server.go#L22): Although in this case this one has 0 transitive dependencies~~ -> Not included in binary as is only required in `tests` package.
- [github.com/oxtoacart/bpool](https://github.com/grafana/xk6-browser/blob/5bd4fa1d7d9c9345eeee8107a97c021b7fee08b2/k6ext/k6test/vu.go#L15): Only 1 transitive dependency for testify, which is nevertheless used in xk6-browser anyway.

Other dependencies such as:
- golang.org/x/exp
- golang.org/x/net
- ~~golang.org/x/sync~~ -> Only used in `tests` pkg
- gopkg.in/guregu/null.v3
Are marginally used but are required as transitive dependencies by other modules.

Additionally this is in interest of reducing the xk6-browser / k6 binary size.

ka3de:
This PR tries to get rid of some dependencies currently required by the main module that are marginally used. More specifically:
- Removes the usage of `github.com/fatih/color`
- Removes the requirement for `golang.org/x/exp` from main module (still required as transitive)
  This produces the direct inclusion of `github.com/google/go-cmp` but that is only used inside `tests` pkg
- Replaces the usage of `github.com/hashicorp/go-multierror` with a custom implementation that fulfills our current use case

Closes #729.

### DIFF
--- a/common/remote_object.go
+++ b/common/remote_object.go
@@ -13,7 +13,6 @@ import (
 
 	cdpruntime "github.com/chromedp/cdproto/runtime"
 	"github.com/dop251/goja"
-	"github.com/hashicorp/go-multierror"
 	"github.com/sirupsen/logrus"
 )
 
@@ -40,17 +39,62 @@ func (pe *objectPropertyParseError) Unwrap() error {
 	return pe.error
 }
 
+type multiError struct {
+	Errors []error
+}
+
+func (me *multiError) append(err error) {
+	me.Errors = append(me.Errors, err)
+}
+
+func (me multiError) Error() string {
+	if len(me.Errors) == 0 {
+		return ""
+	}
+	if len(me.Errors) == 1 {
+		return me.Errors[0].Error()
+	}
+
+	var buf strings.Builder
+	fmt.Fprintf(&buf, "%d errors occurred:\n", len(me.Errors))
+	for _, e := range me.Errors {
+		fmt.Fprintf(&buf, "\t* %s\n", e)
+	}
+
+	return buf.String()
+}
+
+func multierror(err error, errs ...error) error {
+	me := &multiError{}
+	// We can't use errors.As(), as we want to know if err is of type
+	// multiError, not any error in the chain. If err contains a wrapped
+	// multierror, start a new multiError that will contain err.
+	e, ok := err.(*multiError) //nolint:errorlint
+
+	if ok {
+		me = e
+	} else if err != nil {
+		me.append(err)
+	}
+
+	for _, e := range errs {
+		me.append(e)
+	}
+
+	return me
+}
+
 func parseRemoteObjectPreview(op *cdpruntime.ObjectPreview) (map[string]any, error) {
 	obj := make(map[string]any)
 	var result error
 	if op.Overflow {
-		result = multierror.Append(result, &objectOverflowError{})
+		result = multierror(result, &objectOverflowError{})
 	}
 
 	for _, p := range op.Properties {
 		val, err := parseRemoteObjectValue(p.Type, p.Value, p.ValuePreview)
 		if err != nil {
-			result = multierror.Append(result, &objectPropertyParseError{err, p.Name})
+			result = multierror(result, &objectPropertyParseError{err, p.Name})
 			continue
 		}
 		obj[p.Name] = val
@@ -145,8 +189,8 @@ func handleParseRemoteObjectErr(ctx context.Context, err error, logger *logrus.E
 		ooe *objectOverflowError
 		ope *objectPropertyParseError
 	)
-	merr, ok := err.(*multierror.Error)
-	if !ok {
+	var merr *multiError
+	if !errors.As(err, &merr) {
 		// If this panics it's a bug :)
 		k6ext.Panic(ctx, "parsing remote object value: %w", err)
 	}
--- a/common/remote_object_test.go
+++ b/common/remote_object_test.go
@@ -2,6 +2,8 @@ package common
 
 import (
 	"encoding/json"
+	"errors"
+	"fmt"
 	"math"
 	"testing"
 
@@ -231,6 +233,77 @@ func TestParseRemoteObject(t *testing.T) {
 			} else {
 				assert.NoError(t, err)
 			}
+		})
+	}
+}
+
+func TestMultierror(t *testing.T) {
+	t.Parallel()
+
+	var (
+		mockErr1 = errors.New("mockErr1")
+		mockErr2 = errors.New("mockErr2")
+		mockErr3 = errors.New("mockErr3")
+
+		mockMultiErr = &multiError{
+			Errors: []error{mockErr1},
+		}
+		mockWrappedMultiErr = fmt.Errorf("%w", mockMultiErr)
+	)
+
+	tests := []struct {
+		name    string
+		initial error
+		errs    []error
+		exp     []error
+		expStr  string
+	}{
+		{
+			name:    "initial error is nil",
+			initial: nil,
+			errs:    []error{mockErr1},
+			exp:     []error{mockErr1},
+			expStr:  mockErr1.Error(),
+		},
+		{
+			name:    "initial error is std error",
+			initial: mockErr1,
+			errs:    []error{mockErr2, mockErr3},
+			exp:     []error{mockErr1, mockErr2, mockErr3},
+			expStr: fmt.Sprintf(
+				"3 errors occurred:\n\t* %s\n\t* %s\n\t* %s\n",
+				mockErr1, mockErr2, mockErr3),
+		},
+		{
+			name:    "initial error is multiError",
+			initial: mockMultiErr,
+			errs:    []error{mockErr3},
+			exp:     []error{mockErr1, mockErr3},
+			expStr: fmt.Sprintf(
+				"2 errors occurred:\n\t* %s\n\t* %s\n",
+				mockErr1, mockErr3),
+		},
+		{
+			name:    "initial error is wrapped multiError",
+			initial: mockWrappedMultiErr,
+			errs:    []error{mockErr3, mockErr2},
+			exp:     []error{mockWrappedMultiErr, mockErr3, mockErr2},
+			expStr: fmt.Sprintf(
+				"3 errors occurred:\n\t* %s\n\t* %s\n\t* %s\n",
+				mockWrappedMultiErr, mockErr3, mockErr2),
+		},
+	}
+
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			err := multierror(tc.initial, tc.errs...)
+			var me *multiError
+			assert.True(t, errors.As(err, &me))
+			assert.Equal(t, tc.exp, me.Errors)
+			assert.Equal(t, tc.expStr, me.Error())
 		})
 	}
 }
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,6 @@ require (
 	github.com/sirupsen/logrus v1.9.0
 	github.com/stretchr/testify v1.8.0
 	go.k6.io/k6 v0.42.1-0.20230109152238-8418147fa630
-	golang.org/x/exp v0.0.0-20221106115401-f9659909a136
 	golang.org/x/net v0.1.0
 	golang.org/x/sync v0.1.0
 	gopkg.in/guregu/null.v3 v3.5.0
@@ -31,6 +30,7 @@ require (
 	github.com/fatih/color v1.13.0 // indirect
 	github.com/go-sourcemap/sourcemap v2.1.4-0.20211119122758-180fcef48034+incompatible // indirect
 	github.com/golang/protobuf v1.5.2 // indirect
+	github.com/google/go-cmp v0.5.8 // indirect
 	github.com/hashicorp/errwrap v1.1.0 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/klauspost/compress v1.15.11 // indirect
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,6 @@ go 1.19
 require (
 	github.com/chromedp/cdproto v0.0.0-20221023212508-67ada9507fb2
 	github.com/dop251/goja v0.0.0-20221229151140-b95230a9dbad
-	github.com/fatih/color v1.13.0
 	github.com/gorilla/websocket v1.5.0
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/mailru/easyjson v0.7.7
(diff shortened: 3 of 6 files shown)


ankur22 (changes requested):
Thanks for doing this! Most of it looks good. I've left some suggestions that I think are worth considering, and let me know in the comments if my suggestions don't work 🙂 

EDIT: Could you also double check the CI pipeline issues and resolve those? There are some tests that fail often in the CI pipeline (we have raised issues for those so you should be able to find them, and then rerun the failed tests runs).

common/remote_object.go
```diff
@@ -40,17 +39,59 @@ func (pe *objectPropertyParseError) Unwrap() error {
...
+type multiError struct {
+	Errors []error
+}
+
+func (me *multiError) append(err error) {
+	me.Errors = append(me.Errors, err)
+}
+
+func (me multiError) Error() string {
+	if len(me.Errors) > 1 {
```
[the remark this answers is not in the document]
ka3de: Makes sense, fixed in [b4f3aad](https://github.com/grafana/xk6-browser/pull/734/commits/b4f3aad5ab0efc980a1fa40affe0dc0a32aafdce). Looks better now!

go.mod
```diff
@@ -5,16 +5,13 @@ go 1.19
...
 	github.com/mailru/easyjson v0.7.7
 	github.com/mccutchen/go-httpbin v1.1.2-0.20190116014521-c5cb2f4802fa
 	github.com/oxtoacart/bpool v0.0.0-20190530202638-03653db5a59c
 	github.com/sirupsen/logrus v1.9.0
 	github.com/stretchr/testify v1.8.0
 	go.k6.io/k6 v0.42.1-0.20230109152238-8418147fa630
-	golang.org/x/exp v0.0.0-20221106115401-f9659909a136
 	golang.org/x/net v0.1.0
 	golang.org/x/sync v0.1.0
 	gopkg.in/guregu/null.v3 v3.5.0
```
ankur22, line 17: Let's match the versions with k6 deps. For example, k6 is using `gopkg.in/guregu/null.v3 v3.3.0`

common/remote_object.go
```diff
@@ -145,7 +186,7 @@ func handleParseRemoteObjectErr(ctx context.Context, err error, logger *logrus.E
 		ooe *objectOverflowError
 		ope *objectPropertyParseError
 	)
-	merr, ok := err.(*multierror.Error)
+	merr, ok := err.(*multiError)
```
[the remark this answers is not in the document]
ka3de: The problem with `As()` AFAIK is that it will return the first error in the chain that matches the given type, if found any. That means that you might skip some "wrapped errors" in the chain, but considering the current implementation I think it should be safe here, changed [here](https://github.com/grafana/xk6-browser/blob/d76ff69916458824d567f80c8849a648b4efde8a/common/remote_object.go#L193).

common/remote_object.go
```diff
@@ -40,17 +39,59 @@ func (pe *objectPropertyParseError) Unwrap() error {
 	return pe.error
 }
 
+type multiError struct {
```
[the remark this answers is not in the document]
ka3de: Absolutely. Done [here](https://github.com/grafana/xk6-browser/blob/d76ff69916458824d567f80c8849a648b4efde8a/common/remote_object_test.go#L240).

common/remote_object.go
```diff
@@ -40,17 +39,59 @@ func (pe *objectPropertyParseError) Unwrap() error {
...
+
+	if len(me.Errors) == 1 {
+		return me.Errors[0].Error()
+	}
+
+	return ""
+}
+
+func multierror(err error, errs ...error) error {
+	if err == nil {
```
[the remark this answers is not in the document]
ka3de: I have refactored this function, take a look [here](https://github.com/grafana/xk6-browser/blob/d76ff69916458824d567f80c8849a648b4efde8a/common/remote_object.go#L67).

go.mod
ka3de: Thanks for taking a look at that! Changed in [364f771](https://github.com/grafana/xk6-browser/pull/734/commits/364f7716220ecf3c36ac4a25be65eaf6fb0e8aa3).

log/logger.go
```diff
@@ -89,8 +88,7 @@ func (l *Logger) Logf(level logrus.Level, category string, msg string, args ...a
 		return
 	}
 	if l.Logger == nil {
```
ankur22, line 90: Let's remove this completely -- `l.Logger` cannot be `nil` at this point, since the actions before it would have failed will `nil` pointer errors.

common/remote_object.go
```diff
@@ -40,17 +39,62 @@ func (pe *objectPropertyParseError) Unwrap() error {
...
 
+type multiError struct {
+	Errors []error
+}
+
+func (me *multiError) append(err error) {
+	me.Errors = append(me.Errors, err)
+}
+
+func (me multiError) Error() string {
```
ankur22, line 50: Can you also add a test for this `Error` function?

common/remote_object.go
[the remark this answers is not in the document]
ankur22: yep, makes sense. maybe add a `nolint` directive here and explain why.

log/logger.go
ka3de: I think the proper action here would be to, in the constructor [here](https://github.com/grafana/xk6-browser/blob/364f7716220ecf3c36ac4a25be65eaf6fb0e8aa3/log/logger.go#L35), provide a "default" logrus Logger in case the input one is `nil`. And then we can remove this part of the code. What do you think?

ka3de: Done in: [61609ee](https://github.com/grafana/xk6-browser/pull/734/commits/61609ee2ccd55f921b99b83ca59228cdadd64d81).

common/remote_object.go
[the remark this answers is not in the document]
ka3de: As mentioned, for our current implementation, using `As()` should be safe, changed.

common/remote_object.go
ka3de: Sure, done in [c6e7e9a](https://github.com/grafana/xk6-browser/pull/734/commits/c6e7e9ac8cb3e3b610ffbb2905d8b88c4c444494).

log/logger.go
inancgumus: @ka3de doing so makes more sense 👍

common/remote_object_test.go
```diff
@@ -256,30 +256,41 @@ func TestMultierror(t *testing.T) {
...
+			expStr:  mockErr1.Error(),
 		},
 		{
 			name:    "initial error is std error",
 			initial: mockErr1,
 			errs:    []error{mockErr2, mockErr3},
 			exp:     []error{mockErr1, mockErr2, mockErr3},
+			expStr: fmt.Sprintf(
+				"3 errors occurred:\n\t* %s\n\t* %s\n\t* %s\n",
+				mockErr1.Error(), mockErr2.Error(), mockErr3.Error()),
```
inancgumus, line 275: No need to say `.Error` as `Sprintf` will automatically expand an error to `string`.

```suggestion
			expStr: fmt.Sprintf(
				"3 errors occurred:\n\t* %s\n\t* %s\n\t* %s\n",
				mockErr1, mockErr2, mockErr3,
			),
```

common/remote_object_test.go
```diff
@@ -256,30 +256,41 @@ func TestMultierror(t *testing.T) {
...
 		},
 		{
 			name:    "initial error is multiError",
 			initial: mockMultiErr,
 			errs:    []error{mockErr3},
 			exp:     []error{mockErr1, mockErr3},
+			expStr: fmt.Sprintf(
+				"2 errors occurred:\n\t* %s\n\t* %s\n",
+				mockErr1.Error(), mockErr3.Error()),
```
inancgumus, line 284: Same here.

```suggestion
			expStr: fmt.Sprintf(
				"2 errors occurred:\n\t* %s\n\t* %s\n",
				mockErr1, mockErr3,
			),
```

common/remote_object_test.go
```diff
@@ -256,30 +256,41 @@ func TestMultierror(t *testing.T) {
...
 		},
 		{
 			name:    "initial error is wrapped multiError",
 			initial: mockWrappedMultiErr,
 			errs:    []error{mockErr3, mockErr2},
 			exp:     []error{mockWrappedMultiErr, mockErr3, mockErr2},
+			expStr: fmt.Sprintf(
+				"3 errors occurred:\n\t* %s\n\t* %s\n\t* %s\n",
+				mockWrappedMultiErr.Error(), mockErr3.Error(), mockErr2.Error()),
```
inancgumus, line 293: Here too:

```suggestion
			expStr: fmt.Sprintf(
				"3 errors occurred:\n\t* %s\n\t* %s\n\t* %s\n",
				mockWrappedMultiErr, mockErr3, mockErr2,
			),
```

inancgumus (changes requested):
Didn't look into this thoroughly yet, here is what I saw after skimming it.

It's great that we're getting rid of that excess calories 👏

common/remote_object_test.go
ka3de: Fixed in [805bf89](https://github.com/grafana/xk6-browser/pull/734/commits/805bf899fd3e5cc0dac0221731416bf375273812).

common/remote_object_test.go
ka3de: Fixed in [805bf89](https://github.com/grafana/xk6-browser/pull/734/commits/805bf899fd3e5cc0dac0221731416bf375273812).

log/logger.go
ankur22: Could we log a warning too? just so that we're aware that we've forgotten to do something?

ankur22 (approved):
LGTM. Thanks for taking on the feedback 🙂

common/remote_object.go
```diff
@@ -40,17 +39,62 @@ func (pe *objectPropertyParseError) Unwrap() error {
 	return pe.error
 }
 
+type multiError struct {
```
inancgumus, line 42: Would it make more sense to move this to `multierror.go` file? Instead of overcrowding this file or the already crowded `helpers.go` file. Also, IMO, having a test for this in `remote_object_test.go` file does not make sense.

log/logger.go
ka3de: I think that makes sense! although here I see 2 options:
Comply with our logging API to produce the log:
```
// New creates a new logger.
func New(logger *logrus.Logger, iterID string) *Logger {
	var defLogger bool
	if logger == nil {
		defLogger = true
		logger = logrus.New()
	}
	l := &Logger{
		Logger: logger,
		iterID: iterID,
	}
	if defLogger {
		l.Warnf("Logger", "no logger supplied, using default")
	}

	return l
}
```
Or log directly with default logrus logger:
```
// New creates a new logger.
func New(logger *logrus.Logger, iterID string) *Logger {
	if logger == nil {
		logger = logrus.New()
		logger.Warn("no logger supplied, using default")
	}
	return &Logger{
		Logger: logger,
		iterID: iterID,
	}
}
```

What do you think?

ankur22: My vote would be 1, but i'd be happy with either.

common/remote_object.go
```diff
@@ -40,17 +39,62 @@ func (pe *objectPropertyParseError) Unwrap() error {
...
+	}
+	if len(me.Errors) == 1 {
+		return me.Errors[0].Error()
+	}
+
+	var buf strings.Builder
+	buf.WriteString(fmt.Sprintf("%d errors occurred:\n", len(me.Errors)))
+	for _, e := range me.Errors {
+		buf.WriteString(fmt.Sprintf("\t* %s\n", e.Error()))
+	}
```
inancgumus, line 62: You could use `fmt.Fprintf` here instead:

```suggestion
	fmt.Fprintf(&buf, "%d errors occurred:\n", len(me.Errors))
	for _, e := range me.Errors {
		fmt.Fprintf(&buf, "\t* %s\n", e)
	}
```

inancgumus (changes requested):
LGTM with some suggestions.

log/logger.go
inancgumus: My vote is 2 since it's simple—1 is complicated.

> Debugging is twice as hard as writing the code in the first place. Therefore, if you write the code as cleverly as possible, you are, by definition, not smart enough to debug it." - Brian W. Kernighan

common/remote_object.go
ankur22: I was thinking the same, but then I realised that `multierror` is only used in one file, so it makes sense to leave the type close to where it's used.

common/remote_object.go
ankur22: Out of curiosity, why would you chose one over the other?

common/remote_object.go
inancgumus: Yes, I thought that too, but as a helper, it still makes more sense to me in a new file. Also, in Go 1.20+, we will remove this file as there are multiple errors support is on the way. Having it in a new file will also make us spot it more easily.

log/logger.go
ka3de: Ops, just pushed option 1 as I thought it was better to comply with our log output, but we can discuss!
The log message itself in this case I don't think it's a big deal, but at the same time I don't think the implementation for the first option is so cumbersome.

common/remote_object.go
ka3de: Exactly the multierror implementation is only used in this file, and I would keep its definition inside this file only and try to not widespread its usage. Actually after the merge is done I would create an issue in order to get rid of this implementation or replace it by something simpler.

log/logger.go
inancgumus: Less code is better than more code. But, 🤷 , I leave it to you. I said what I wanted to say :)

common/remote_object.go
inancgumus: Yes, it's more performant and less code (=easier to read and maintain). More function calls often result in a less performant code.

common/remote_object.go
ka3de: My reason to moving it to a new file would be if we think that there is a valid use case for "multierrors" and therefore want to use them in more places. In that case I think it's good to move it as a helper and improve its implementation in future releases. Otherwise I would keep it in this file and try to remove it for something simpler.

What do you think?

inancgumus: @ka3de I would rather create that issue soon without waiting for the merge.

ka3de: Done! See #738 .

inancgumus: Thanks!

common/remote_object.go
ka3de: You are right! Done.

inancgumus (approved):
🚀

Issue #648, opened 2022-11-11
Fix launching a browser results in closed file already closed error


inancgumus:
### Brief summary

`TestBrowserCrashErr` [ends with the following error](https://github.com/grafana/xk6-browser/actions/runs/3438417446/jobs/5734442790) when trying to launch a browser:

```
--- FAIL: TestBrowserCrashErr (0.03s)
    browser_test.go:151: 
        	Error Trace:	/home/runner/work/xk6-browser/xk6-browser/tests/page_test.go:705
        	            				/home/runner/work/xk6-browser/xk6-browser/tests/browser_test.go:151
        	Error:      	Error "GoError: launching browser: read |0: file already closed" does not contain "launching browser: Invalid devtools server port"
        	Test:       	TestBrowserCrashErr
```

Although we only observe the error on CI for now. It can be problematic in some instances when users launch a browser too. We need to fix it before it happens again.

### xk6-browser version

79b7f087

### OS

Github CI

### Chrome version

107.0.5304.110

### Steps to reproduce the problem

Last detected in: `TestBrowserCrashErr `.

https://github.com/grafana/xk6-browser/actions/runs/3438417446/jobs/5734442790

### Expected behaviour

No error.

### Actual behaviour

Error.

### Related issues

#415, #363.

ankur22:
I was able to reproduce this issue with the help of https://stackoverflow.com/questions/70422836/error-in-reading-the-byte-read-0-file-already-closed-in-go. I wasn't able to reproduce this just by stress testing the test unfortunately. It had to be done in a manual way:

1. Checkout 79b7f08.
2. Use the following diff:
    ```go
      diff --git a/common/browser_process.go b/common/browser_process.go
      index 78de55d..8fc5a7c 100644
      --- a/common/browser_process.go
      +++ b/common/browser_process.go
      @@ -159,6 +159,10 @@ func execute(
       		return command{}, fmt.Errorf("%w", ctx.Err())
       	}
       
      +	if cmd.Process != nil && cmd.Process.Pid != 0 {
      +		fmt.Println(cmd.Process.Pid)
      +	}
      +
       	done := make(chan struct{})
       	go func() {
       		// TODO: How to handle these errors?
      diff --git a/tests/test_browser.go b/tests/test_browser.go
      index 26bfa7e..710eee7 100644
      --- a/tests/test_browser.go
      +++ b/tests/test_browser.go
[...]

PR #735, opened 2023-01-27
Update the message when browser failed to load

ankur22:
When the browser fails to load prior to xk6-browser connecting to it via CDP, then we need to make it clear what's happened.

Partially Resolves: https://github.com/grafana/xk6-browser/issues/648

### DIFF
--- a/common/browser_process.go
+++ b/common/browser_process.go
@@ -243,11 +243,14 @@ func (p *devToolsURLParser) err() error {
 	if len(p.errs) > 0 {
 		return p.errs[0]
 	}
-	if err := p.sc.Err(); err != nil {
-		if errors.Is(err, fs.ErrClosed) {
-			return fmt.Errorf("browser process shutdown unexpectedly before a connection could be made to it: %w", err)
-		}
-		return fmt.Errorf("%w", err)
+
+	err := p.sc.Err()
+	if errors.Is(err, fs.ErrClosed) {
+		return fmt.Errorf("browser process shutdown unexpectedly before establishing a connection: %w", err)
 	}
+	if err != nil {
+		return err //nolint:wrapcheck
+	}
+
 	return nil
 }
--- a/common/browser_process.go
+++ b/common/browser_process.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


common/browser_process.go
```diff
@@ -243,6 +244,9 @@ func (p *devToolsURLParser) err() error {
 		return p.errs[0]
 	}
 	if err := p.sc.Err(); err != nil {
+		if errors.Is(err, fs.ErrClosed) {
+			return fmt.Errorf("browser process shutdown unexpectedly before a connection could be made to it: %w", err)
+		}
```
inancgumus, line 249: Please make the inner `if` unintended for a better line-of-sight ([see](https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html)):

```go
	err := p.sc.Err()
	if errors.Is(err, fs.ErrClosed) {
		return fmt.Errorf("browser process shutdown unexpectedly before a connection could be made to it: %w", err)
	}
	if err != nil {
		return fmt.Errorf("%w", err)
	}
```

The error message can be shorter without losing the intent:

`"browser process shutdown unexpectedly before establishing a connection: %w"`

Lastly, you don't need to wrap the error with `%w` ([see](https://go.dev/blog/go1.13-errors)). It might deserve a `nolint` directive when you don't wrap as the linter is not that smart.

inancgumus (changes requested):
LGTM with some minor details.

common/browser_process.go
ankur22: Yep, good suggestion. Resolved in 4d95e2ee63224b948bdd628064b02908e3a1497c

inancgumus (approved):

inancgumus:
@ankur, let's squash that last commit into the previous one before merging. It doesn't make sense to have that small linter commits.

inancgumus:
@ankur22 , let's squash that last commit into the previous one before merging. It doesn't make sense to have that small linter

ankur22:
> @ankur22 , let's squash that last commit into the previous one before merging. It doesn't make sense to have that small linter commit to be have in the git history.

Yep, i know, it's only there to show what's going on and for reviewers to reference :)

PR #741, opened 2023-01-31
Revert env var that disables test runs

ankur22:
We have decided that after the merge k6 will take control of restricting test runs, allowing current browser test users to not be affected (for now).


ka3de (approved):
LGTM.

inancgumus (approved):

PR #743, opened 2023-01-31
Revert the update of some of the dependency

ankur22:
As part of the merge into k6, we need to ensure that we're not updating the version of dependencies that k6 is already depending on. If there's a good reason for them to be at the version they were set at then please comment on them explaining why.

Related comment: https://github.com/grafana/k6/pull/2884/files#r1091880936


inancgumus:
> If there's a good reason for them to be at the version they were set at then please comment on them explaining why.

It's almost impossible to remember the reasoning while upgrading them before :( If the tests pass and examples work when we downgrade them, then it should be OK.

inancgumus (approved):
> If there's a good reason for them to be at the version they were set at then please comment on them explaining why.

It's almost impossible to remember the reasoning while upgrading them before :( If the tests pass and examples work when we downgrade them, then it should be OK.

ankur22:
> It's almost impossible to remember the reasoning while upgrading them before :( If the tests pass and examples work when we downgrade them, then it should be OK.

Thanks @inancgumus, that was my thinking too, if the tests pass and all seems fine then let's go with it for now.

mstoykov (approved):
LGTM!

I can confirm that this no longer bumps version when `go get`-ed in k6

PR #744, opened 2023-02-01
Return an error from BrowserContext.AddCookies and handle panic in mapping layer

ankur22:
At the moment the codebase uses panic too liberally and we don't have control over when we actually want to abort all test runs vs when we only want to fail the current test iteration.

This is a POC (which if we're all on agreement with could become the first PR for this way of working with errors and panics) of what was suggested in a [comment](https://github.com/grafana/xk6-browser/issues/688#issuecomment-1408543524) on #688. The idea is that we will bubble up errors all the way to the mapping layer, where it can decide on whether to panic or to return an error to goja.

This would mean that we have more control over when to panic, and to be more aware of what we want to panic on, as well as using good GO error handling practices throughout the codebase.

### DIFF
--- a/api/browser_context.go
+++ b/api/browser_context.go
@@ -1,11 +1,14 @@
 package api
 
 import (
+	"context"
+
 	"github.com/dop251/goja"
 )
 
 // BrowserContext is the public interface of a CDP browser context.
 type BrowserContext interface {
+	// Externally available methods
 	AddCookies(cookies goja.Value) error
 	AddInitScript(script goja.Value, arg goja.Value)
 	Browser() Browser
@@ -35,4 +38,7 @@ type BrowserContext interface {
 	StorageState(opts goja.Value)
 	Unroute(url goja.Value, handler goja.Callable)
 	WaitForEvent(event string, optsOrPredicate goja.Value) any
+
+	// Internally available methods
+	Ctx() context.Context
 }
--- a/api/browser_context.go
+++ b/api/browser_context.go
@@ -6,7 +6,7 @@ import (
 
 // BrowserContext is the public interface of a CDP browser context.
 type BrowserContext interface {
-	AddCookies(cookies goja.Value)
+	AddCookies(cookies goja.Value) error
 	AddInitScript(script goja.Value, arg goja.Value)
 	Browser() Browser
 	ClearCookies()
--- a/k6error/internal.go
+++ b/k6error/internal.go
@@ -1,0 +1,11 @@
+// Package k6error contains ErrInternal.
+package k6error
+
+import (
+	"errors"
+)
+
+// ErrInternal should be wrapped into an error
+// to signal to the mapping layer that the error
+// is an internal error and we should abort the test run.
+var ErrInternal = errors.New("internal error")
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -539,7 +539,7 @@ func mapBrowserContext(ctx context.Context, vu k6modules.VU, bc api.BrowserConte
 		"addCookies": func(cookies goja.Value) *goja.Promise {
 			return k6ext.Promise(ctx, func() (result any, reason error) {
 				err := bc.AddCookies(cookies)
-				panicIfInternalError(ctx, err)
+				panicIfInternalError(bc, err)
 				return nil, err //nolint:wrapcheck
 			})
 		},
@@ -628,8 +628,12 @@ func mapBrowserType(ctx context.Context, vu k6modules.VU, bt api.BrowserType) ma
 	}
 }
 
-func panicIfInternalError(ctx context.Context, err error) {
+type withCtx interface {
+	Ctx() context.Context
+}
+
+func panicIfInternalError(w withCtx, err error) {
 	if errors.Is(err, k6error.ErrInternal) {
-		k6ext.Panic(ctx, err.Error())
+		k6ext.Panic(w.Ctx(), err.Error())
 	}
 }
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -2,12 +2,14 @@ package browser
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/dop251/goja"
 
 	"github.com/grafana/xk6-browser/api"
 	"github.com/grafana/xk6-browser/chromium"
+	"github.com/grafana/xk6-browser/k6error"
 	"github.com/grafana/xk6-browser/k6ext"
 
 	k6common "go.k6.io/k6/js/common"
@@ -534,7 +536,13 @@ func mapWorker(ctx context.Context, vu k6modules.VU, w api.Worker) mapping {
 func mapBrowserContext(ctx context.Context, vu k6modules.VU, bc api.BrowserContext) mapping {
 	rt := vu.Runtime()
 	return mapping{
-		"addCookies":                  bc.AddCookies,
+		"addCookies": func(cookies goja.Value) *goja.Promise {
+			return k6ext.Promise(ctx, func() (result any, reason error) {
+				err := bc.AddCookies(cookies)
+				panicIfInternalError(ctx, err)
+				return nil, err //nolint:wrapcheck
+			})
+		},
 		"addInitScript":               bc.AddInitScript,
 		"browser":                     bc.Browser,
 		"clearCookies":                bc.ClearCookies,
@@ -617,5 +625,11 @@ func mapBrowserType(ctx context.Context, vu k6modules.VU, bt api.BrowserType) ma
 			m := mapBrowser(ctx, vu, bt.Launch(opts))
 			return rt.ToValue(m).ToObject(rt)
 		},
+	}
+}
+
+func panicIfInternalError(ctx context.Context, err error) {
+	if errors.Is(err, k6error.ErrInternal) {
+		k6ext.Panic(ctx, err.Error())
 	}
 }
--- a/browser/mapping_test.go
+++ b/browser/mapping_test.go
@@ -37,6 +37,14 @@ func customMappings() map[string]string {
 	}
 }
 
+// ignoreUnmapped is used to ignore the unmapped
+// exported methods.
+func ignoreUnmapped() map[string]interface{} {
+	return map[string]interface{}{
+		"ctx": nil,
+	}
+}
+
 // TestMappings tests that all the methods of the API (api/) are
 // to the module. This is to ensure that we don't forget to map
 // a new method to the module.
@@ -56,6 +64,7 @@ func TestMappings(t *testing.T) {
 			},
 		}
 		customMappings = customMappings()
+		ignoreUnmapped = ignoreUnmapped()
 	)
 
 	// testMapping tests that all the methods of an API are mapped
@@ -87,7 +96,9 @@ func TestMappings(t *testing.T) {
 			if cmok {
 				m = cm
 			}
-			if _, ok := mapped[m]; !ok {
+			_, iOk := ignoreUnmapped[m]
+			_, mOk := mapped[m]
+			if !mOk && !iOk {
 				t.Errorf("method %s not found", m)
 			}
 		}
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/grafana/xk6-browser/api"
+	"github.com/grafana/xk6-browser/k6error"
 	"github.com/grafana/xk6-browser/k6ext"
 	"github.com/grafana/xk6-browser/log"
 
@@ -64,8 +65,8 @@ func NewBrowserContext(
 }
 
 // AddCookies is not implemented.
-func (b *BrowserContext) AddCookies(cookies goja.Value) {
-	k6ext.Panic(b.ctx, "BrowserContext.addCookies(cookies) has not been implemented yet")
+func (b *BrowserContext) AddCookies(cookies goja.Value) error {
+	return fmt.Errorf("BrowserContext.addCookies(cookies) has not been implemented yet: %w", k6error.ErrInternal)
 }
 
 // AddInitScript adds a script that will be initialized on all new pages.
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -362,6 +362,14 @@ func (b *BrowserContext) WaitForEvent(event string, optsOrPredicate goja.Value) 
 	return b.waitForEvent(event, predicateFn, timeout)
 }
 
+// Ctx returns the go context that is associated
+// with BrowserContext. It's main uses is to help
+// shutdown the browser process if a panic occurs
+// in the mapping layer.
+func (b *BrowserContext) Ctx() context.Context {
+	return b.ctx
+}
+
 func (b *BrowserContext) waitForEvent(event string, predicateFn goja.Callable, timeout time.Duration) any {
 	evCancelCtx, evCancelFn := context.WithCancel(b.ctx)
 	chEvHandler := make(chan Event)


inancgumus:
Thanks, @ankur22, for following up on the discussion we made earlier and putting it into reality in this PR.

One question: Do you think we need a separate error type like that (`InternalError`) and all the machinery around it? Can we have a single error variable, wrap it using `Errorf: %w`, and detect it with `errors.Is`?

For example:

```
// k6error package
var ErrInternal = errors.New("internal error")

// common package
func (b *BrowserContext) AddCookies(cookies goja.Value) error {
  return fmt.Errorf("BrowserContext.addCookies(cookies) has not been implemented yet: %w", k6error.ErrInternal)
}

// mapping layer
...
"addCookies": func(cookies goja.Value) *goja.Promise {
  return k6ext.Promise(ctx, func() (result any, reason error) {
    err := bc.AddCookies(cookies)
    panicIfInternalError(ctx, err)
}

func panicIfInternalError(ctx context.Context, err error) {
  if errors.Is(err, k6error.ErrInternal) {
    k6ext.Panic(ctx, err.Error())
  }
}
```

k6error/internal.go
```diff
@@ -0,0 +1,42 @@
...
+func AsInternalError(msg string) *Internal {
+	return &Internal{err: fmt.Errorf("internal: %w", errors.New(msg))}
+}
+
+// Error will return the msg of the wrapped error.
+func (i *Internal) Error() string { return i.err.Error() }
+
+// Internal will return true to signal that it
+// is an internal error.
+func (i *Internal) Internal() bool { return true }
```
ka3de: Do we need this? Can't we work with std's `errors.Is()` ? Also in order to call this method wouldn't we need to cast an `error` into our `Internal` (maybe through `errors.As()`) ?

k6error/internal.go
```diff
@@ -0,0 +1,42 @@
...
+
+type internal interface {
+	Internal() bool
+}
+
+// IsInternal returns true if err is of type
+// Internal.
+func IsInternal(err error) bool {
```
ka3de, line 43: Same consideration as in https://github.com/grafana/xk6-browser/pull/744/commits/e4aba5550f1fcad42ba0b0773a893313b150339e#r1093114677.

k6error/internal.go
inancgumus: I have the same question [here](https://github.com/grafana/xk6-browser/pull/744#issuecomment-1411925677). Take a look if you haven't seen it :)

inancgumus (commented):
Thanks, @ankur22, for following up on the discussion we made earlier and putting it into reality in this PR.

One question: Do you think we need a separate error type like that (`InternalError`) and all the machinery around it? Can we have a single error variable, wrap it using `Errorf: %w`, and detect it with `errors.Is`?

For example:

```go
// k6error package
var ErrInternal = errors.New("internal error")

// common package
func (b *BrowserContext) AddCookies(cookies goja.Value) error {
  return fmt.Errorf("BrowserContext.addCookies(cookies) has not been implemented yet: %w", k6error.ErrInternal)
}

// mapping layer
...
"addCookies": func(cookies goja.Value) *goja.Promise {
  return k6ext.Promise(ctx, func() (result any, reason error) {
    err := bc.AddCookies(cookies)
    panicIfInternalError(ctx, err)
}

func panicIfInternalError(ctx context.Context, err error) {
  if err == nil || !errors.Is(err, k6error.ErrInternal) {
    return
  }
  k6ext.Panic(ctx, err.Error())
}
```

k6error/internal.go
ankur22: Yeah good point, we don't need this 🙂

I've found it to be useful when working with `temporary` errors (mainly for backend services) since there are other error types within go that implement `Temporary()`.

ankur22:
> One question: Do you think we need a separate error type like that (`InternalError`) and all the machinery around it? Can we have a single error variable, wrap it using `Errorf: %w`, and detect it with `errors.Is`?

Nice! I do like the simplicity. Let me try that and see.

ankur22:
One question for you @inancgumus and @ka3de:

If you're scanning through the code, which one looks clearer:

```go
fmt.Errorf("BrowserContext.addCookies(cookies) has not been implemented yet: %w", k6error.ErrInternal)
```

or 

```go
k6error.AsInternalError("BrowserContext.addCookies(cookies) has not been implemented yet")
```

inancgumus:
> If you're scanning through the code, which one looks clearer:

That's the idiomatic way of doings things after Go 1.13. So it depends on the Go authors, not us :)

However, you can still make it cleaner if you like as follows:

```go
// k6error
func Internal(err error) error {
  return fmt.Errorf("%s: %w", err, ErrInternal)
}

// common
return k6error.Internal("BrowserContext.addCookies(cookies) has not been implemented yet")
```

You can go fancier if needed, like copying `Errorf`, something like:

```go
// k6error
func Internal(format string, a ...any) error {
  a = append(a, ErrInternal)
  return fmt.Errorf(format+": %w", a...)
}


```

ankur22:
> That's the idiomatic way of doings things after Go 1.13. So it depends on the Go authors, not us :)

Updated to the simpler idiomatic way: 5abd5e4dae486b7af96a8e5e9bcaa7f00e5db207

ka3de (approved):
LGTM.

inancgumus:
The code is OK as it is 👍 We will get a huge stack trace because of an unimplemented method. The stack trace—hence the panic—would benefit if an "unexpected" error occurs, and we can track the bug through stack trace. In this case, it will harm the user experience. Let's save the `panic` for really unexpected situations.

Instead of [doing it like this](https://github.com/grafana/xk6-browser/blob/32d494441ab49588ecd30d70fe05e5811f854d6a/browser/mapping.go#L539-L545), we can return the error as follows:

```go
...
"addCookies": func(cookies goja.Value) *goja.Promise {
	return k6ext.Promise(ctx, func() (result any, reason error) {
		err := bc.AddCookies(cookies)
	})
},
```

For example, let's try it with the following script:

```js

export const options = {
  thresholds: {
    checks: ["rate==1.0"]
  }
}

  const browser = chromium.launch({ headless: true });

  try {
    await context.addCookies("...");
    console.log("should not reach here");
  } finally {
    browser.close();
  }
}
```

Just returning the error in the mapping layer will result in the following error output in the console:

```shell
ERRO[0000] Uncaught (in promise) BrowserContext.addCookies(cookies) has not been implemented yet  executor=per-vu-iterations scenario=default
```

I believe it's clear enough for the user.

We can make it better by using the [`AbortingPromise`](https://github.com/grafana/xk6-browser/blob/32d494441ab49588ecd30d70fe05e5811f854d6a/k6ext/promise.go#L29-L32) function instead—which I wrote for this sort of case:

```go
...
"addCookies": func(cookies goja.Value) *goja.Promise {
	return k6ext.AbortingPromise(ctx, func() (result any, reason error) {
		err := bc.AddCookies(cookies)
	})
},
```

Which will result in the following error output:

```shell
ERRO[0000] BrowserContext.addCookies(cookies) has not been implemented yet  executor=per-vu-iterations scenario=default
```

@grafana/k6-browser WDYT?

ka3de:
> @grafana/k6-browser WDYT?

But in this cases the test would continue after the error is returned right? I thought the idea of the definition of `ErrInternal` was a way for us to understand in the mapping layer when a `panic` must occur so the test is ended because it's a "non-recoverable error", therefore I understand that the panic was the proper action. Is this the behavior that we are trying to define? Or did I misunderstood it?

inancgumus:
> But in this cases the test would continue after the error is returned right?

You're right: The test run will continue if there are multiple VUs if we don't panic. Then we can compromise and have a long stack trace—even in this case. Maybe we can use `go.k6.io/k6/execution.AbortTestRun()` instead of panicking 🤔 But I haven't looked into it yet. I'm happy to keep the PR as it is without complicating things further 👍 

By the way, we get the following error when we run the script I shared above:

```shell
panic: no browser process ID in context
        panic: GoError: BrowserContext.addCookies(cookies) has not been implemented yet: internal error
```

The interesting line is this:

```
panic: no browser process ID in context
```

The `k6ext.Panic` can't get the browser process ID and can't clear the browser process. We better check what's going on there.

inancgumus:
> But in this cases the test would continue after the error is returned right?

@ka3de, you're right: The test run will continue if there are multiple VUs if we don't panic. Then we can compromise and have a long stack trace—even in this case. Maybe we can use `go.k6.io/k6/execution.AbortTestRun()` instead of panicking 🤔 But I haven't looked into it yet. I'm happy to keep the PR as it is without complicating things further 👍 

---

@ankur22, by the way, we get the following error when we run the script I shared above:

```shell
panic: no browser process ID in context
        panic: GoError: BrowserContext.addCookies(cookies) has not been implemented yet: internal error
```

The interesting line is this:

```
panic: no browser process ID in context
```

The `k6ext.Panic` can't get the browser process ID and can't clear the browser process. We better check what's going on there.

ka3de:
> The interesting line is this: 
> ```
> panic: no browser process ID in context
> ```
> The `k6ext.Panic` can't get the browser process ID and can't clear the browser process. We might want to check what's going on there.

You are right @inancgumus that does not look very friendly. Maybe we should pause this until we have analyzed better the different options and establish a consensus on the behavior from now one.

For me personally I see various different ways to return an error for which I don't fully understand the differences. E.g.:
- Return plain error to Goja.
- Std `panic` 
- [k6ext.Panic](https://github.com/grafana/xk6-browser/blob/0fdcdff67cca53b82be4f6c3509de474999e5b2a/k6ext/panic.go#L18)
- [k6common.Throw](https://github.com/grafana/k6/blob/b750e206859a495b8a1fe12aa016dc8ed0e1c9db/js/common/util.go#L16)
- [k6execution.AbortTestRun](https://github.com/grafana/k6/blob/b750e206859a495b8a1fe12aa016dc8ed0e1c9db/execution/abort.go#L62-L72)

Or maybe this should be tackled in a different discussion, but I think we need a better analysis for this.
What do you think @ankur22 ?

ankur22:
@grafana/k6-browser Thanks for the reviews and suggestions.

My feeling is that we're looking for the perfect solution instead of one that will get us moving in the right direction.

There is confusion between all the ways of panicking and erroring (as i've mentioned in https://github.com/grafana/xk6-browser/issues/688).

We need to move towards a solution where we:
- Stop the whole test run when we find a genuine reason to do so (e.g. panic when a method is not implemented or an unexpected error such as when parsing string to json, etc.).

That is the key goal for issue [#688](https://github.com/grafana/xk6-browser/issues/688), but at the moment panics are placed throughout the code and it's difficult to reason where we should be erroring vs panicking.

This PR is an attempt to:
- Force us to think about errors and bubble them up to the mapping layer.
- At the mapping layer, it can easily discern an internal error and panic, otherwise return an error.

Panics should rarely occur. When they do, we will see a messy stack trace. The stack trace is a concern, but one that could be solved in another PR which is unrelated to this PR, which is trying to solve a different problem.

inancgumus:
@ankur22, when we panic, we clean up the browser process. That is the current behavior until v0.8.

However, this PR will introduce a regression I explained in [my earlier comment](https://github.com/grafana/xk6-browser/pull/744#issuecomment-1418901227) 😒 I'd be happy to merge this PR if we can solve that issue 👍 

We can tackle finding the perfect solution later—I agree 🚀

ankur22:
> However, this PR will introduce a regression I explained in [my earlier comment](https://github.com/grafana/xk6-browser/pull/744#issuecomment-1418901227) 😒 I'd be happy to merge this PR if we can solve that issue 👍

Apologies, this is a valid blocker. I've attempted to solve it with 42871c3153193209f0e19fc05221ba422cbd624f.

> We can tackle finding the perfect solution later (#753)—I agree 🚀

💪

api/browser_context.go
```diff
@@ -35,4 +38,7 @@ type BrowserContext interface {
 	StorageState(opts goja.Value)
 	Unroute(url goja.Value, handler goja.Callable)
 	WaitForEvent(event string, optsOrPredicate goja.Value) any
+
+	// Internally available methods
+	Ctx() context.Context
```
ankur22, line 43: Another solution could be to have a `BrowserProcessRegister` which would keep track of all the running browser process (i.e. store the pid). When the `panic` occurs, it would just be a case of running something like `bpr.Shutdown()`.

This way we wouldn't dirty the APIs with these "internal" exported methods.

Another reason to work with a `BrowserProcessRegister` would be to clean up the context a bit by removing the the pid from the context.

ka3de: I have to understand better how this would be done, but on a first sight this seems like a better option to me. Otherwise we'll have to expose a method in order to retrieve the `ctx` from every mapping function (as a panic could be called from methods on each one of them), right?

ankur22: I've created https://github.com/grafana/xk6-browser/pull/755 which uses a register.

inancgumus: Nice! I'd prefer merging #755 first and then applying those changes here then 👍

inancgumus:
I was wondering whether we can solve [this issue](https://github.com/grafana/xk6-browser/pull/744#issuecomment-1418986802) more simply without doing medium-sized changes (#755 and #762).

The idea is this. Since `BrowserType.Launch` is the root method that launches a browser and gets the process ID, we can make the process ID available to the mapping layer.

1. Let `Launch` return the process ID.
2. Put the `processID` in `moduleVU` after we get it from `Launch`.
3. Return a new context with the process ID in `moduleVU.Context()`.

I know it's not the cleanest solution, and we can perfect it later, as we [discussed above](https://github.com/grafana/xk6-browser/pull/744#issuecomment-1418966132).

```go
func (vu moduleVU) Context() context.Context {
	...
	ctx := k6ext.WithProcessID(vu.VU.Context(), vu.browserProcessID)
	return k6ext.WithVU(ctx, vu.VU)
}
...
func mapBrowserType(vu moduleVU, bt api.BrowserType) mapping {
	...
	return mapping{
		...
		"launch": func(opts goja.Value) *goja.Object {
			var b api.Browser
			b, vu.browserProcessID = bt.Launch(opts)
			m := mapBrowser(vu, b)
			return rt.ToValue(m).ToObject(rt)
		},
	}
}
```

@grafana/k6-browser

ankur22:
> I was wondering whether we can solve [this issue](https://github.com/grafana/xk6-browser/pull/744#issuecomment-1418986802) more simply without doing medium-sized changes (#755 and #762).

This is indeed a more elegant solution.

I still prefer the #755 solution since it handles the closing of all browser processes that a test might start up. I'd be happy to remove the ugly `runID` code if we were in agreement to using the hacky workaround of using `Kill` in #762.

I think #762 is still a valid issue that we should merge in since calling `Release` on the process will prevent the `Kill` from functioning as we expect, which can lead to zombie browser processes.

inancgumus:
Thanks, @ankur22! I believe we can start simpler with this solution and work towards the others next. Since this solution is a low-hanging fruit.

> I still prefer the #755 solution since it handles the closing of all browser processes that a test might start up.

Currently (v0.8), we should also be closing all test browser processes since each test gets its own context, and we should be already cleaning them up.

ankur22:
> Currently (v0.8), we should also be closing all test browser processes since each test gets its own context, and we should already be cleaning them up. Then we can include #762 for sure if you think that's critical.

That's true about the context. What i meant was when a panic occurs, in [k6ext.Panic](https://github.com/grafana/xk6-browser/blob/main/k6ext/panic.go#L36), it can only retrieve the single PID, but if the browser test spawns multiple browsers then the panic will only try to kill the last opened browser. With the `register` in #755, we can register multiple pids, and it's encapsulated in the `k6os` package.

inancgumus:
@ankur22, this should be unblocked now 🥳 Thanks for your cooperation and understanding 🙇

ankur22:
Closing and changes moved to https://github.com/grafana/xk6-browser/pull/796 as reworking this PR would be too much work.

PR #746, opened 2023-02-02
Remove Async handling from the core methods

### LINKED ISSUE #742: Remove Async handling from the core methods
Since the mapping layer  (#683) is now responsible for handling of the Goja promise, we don't need to return promises from the methods in the `common` package (also called _core_, I guess). This issue is about removing `*goja.Promise` from all methods that return it in the `common` package.

inancgumus:
This is a continuation of the efforts set in #683. Handling the Async APIs in the mapping layer. Please take a look at issue #742 for more details. This can also help the work in #688—where we need to remove the promises and return errors 🤔?

Closes #742

### DIFF
--- a/api/browser_context.go
+++ b/api/browser_context.go
@@ -6,23 +6,23 @@ import (
 
 // BrowserContext is the public interface of a CDP browser context.
 type BrowserContext interface {
-	AddCookies(cookies goja.Value) *goja.Promise
+	AddCookies(cookies goja.Value)
 	AddInitScript(script goja.Value, arg goja.Value)
 	Browser() Browser
 	ClearCookies()
 	ClearPermissions()
 	Close()
-	Cookies() *goja.Promise
-	ExposeBinding(name string, callback goja.Callable, opts goja.Value) *goja.Promise
-	ExposeFunction(name string, callback goja.Callable) *goja.Promise
+	Cookies() []any // TODO: make it []Cookie later on
+	ExposeBinding(name string, callback goja.Callable, opts goja.Value)
+	ExposeFunction(name string, callback goja.Callable)
 	GrantPermissions(permissions []string, opts goja.Value)
-	NewCDPSession() *goja.Promise
+	NewCDPSession() CDPSession
 	NewPage() Page
 	Pages() []Page
-	Route(url goja.Value, handler goja.Callable) *goja.Promise
+	Route(url goja.Value, handler goja.Callable)
 	SetDefaultNavigationTimeout(timeout int64)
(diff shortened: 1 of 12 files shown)


ka3de (approved):
LGTM.

ankur22 (approved):
LGTM 👍

inancgumus:
Guys, @grafana/k6browserdev, I forgot to return values from some functions and force-pushed now. Take another look if you want. No behavior change—just a reminder for us when we implement these methods in the future.

ka3de:
> Guys, @grafana/k6-browser, I forgot to return values from some functions and force-pushed now. Take another look if you want. No behavior change (_since the methods are not implemented_)—returned values are reminders for us when we want to implement these methods in the future.

I guess adding the return types for these methods do not harm, but I wonder if it's necessary or the best to do that in this PR considering that we are now implementing them and they will panic in any case?

Also, I personally lack a little bit of context to know if all the defined types are correct, although they look correct! But for example, in this method the returned type does not seem to match with the behavior specified in the Godoc, no?
https://github.com/grafana/xk6-browser/blob/a2bb6a28576a6af8c9567c3d91e93d9211130411/common/response.go#L191-L196

inancgumus:
> Could you have done a fixup instead of a force push?

@ankur22, yeah, TBH, I didn't want to deal with the flaky tests and put the work on you guys instead. Sorry about that :-(

inancgumus:
> I guess adding the return types for these methods do not harm, but I wonder if it's necessary or the best to do that in this PR considering that we are now not implementing them and they will panic in any case?
> Also, I personally lack a little bit of context to know if all the defined types are correct, although they look correct!

The PR is about returning the old behavior and this was the old behavior 🤷 

>  But for example, in this method the returned type does not seem to match with the behavior specified in the Godoc, no?

I understand your concerns and I have the same concerns. That method was in discussions between us long before. We still don't know whether to return an error or bool. I left it as it was. We can fix it when the time comes to implement it.

inancgumus:
@ka3de,

> I guess adding the return types for these methods do not harm, but I wonder if it's necessary or the best to do that in this PR considering that we are now not implementing them and they will panic in any case?
> Also, I personally lack a little bit of context to know if all the defined types are correct, although they look correct!

The PR is about returning the old behavior and this was the old behavior 🤷 

>  But for example, in this method the returned type does not seem to match with the behavior specified in the Godoc, no?

I understand your concerns and I have the same concerns. That method was in discussions between us long before. We still don't know whether to return an error or bool. I left it as it was. We can fix it when the time comes to implement it.

PR #747, opened 2023-02-02
Add example release notes

ankur22:


inancgumus:
@ankur22 What is this? :) Can we do it by directly editing the k6 release notes when we want to release them?

ankur22:
This was mainly to show what the readme style will look like based on k6's readme style. Happy to work with k6 on the process and style though 👍

PR #751, opened 2023-02-06
Map Page's Keyboard, Mouse, and Touchscreen properties

### LINKED ISSUE #749: Map Page's Keyboard, Mouse, and Touchscreen properties
### Brief summary

In v0.8.0, probably due to the change on how the mapping is done from xk6-browser components to Goja, the `keyboard` and `mouse` elements are no longer exposed in the JS API.

I think this also proves that we need more tests / examples for these functionalities from the JS API side.

### xk6-browser version

v0.8.0

### OS

Ubuntu 20.04.5 LTS

### Chrome version

109.0.5414.119 (Official Build) (64-bit)

_No response_

Run this script which actually works with v0.7.0 version, but not for v0.8.0:

```js
export default function () {
    const browser = chromium.launch({ headless: false, timeout: "10m" });
  
    page.goto("https:/duckduckgo.com/", { waitUntil: 'networkidle' })
    .then(() => {
        console.log(page.keyboard); // prints undefined
        console.log(page.mouse); // prints undefined
        
        page.keyboard.type('k6');
        page.keyboard.press('Enter');
        sleep(3);
    })
    .finally(() => {
        page.close()
        browser.close()
    });
}
```

### Expected behaviour

Error execution printing undefined for both `keyboard` and `mouse` elements of `page` and fails due to keyboard being undefined.
<details>
<summary>See error log.</summary>

```shell
INFO[0001] undefined                                     source=console
INFO[0001] undefined                                     source=console
ERRO[0001] Uncaught (in promise) TypeError: Cannot read property 'type' of undefined or null
running at file:///<redacted>/test.js:30:13(15)  executor=per-vu-iterations scenario=default
```
</details>

### Actual behaviour

The script should type in `k6` in the search bar, submit by pressing `Enter` through the `page.keyboard` element and finish its execution correctly.

ka3de:
Closes #749.

### DIFF
--- a/browser/mapping_test.go
+++ b/browser/mapping_test.go
@@ -32,7 +32,9 @@ func wildcards() map[string]string {
 // getters is a list of mappings that return internal
 // API component properties as Goja objects.
 func getters() map[string]string {
-	return map[string]string{}
+	return map[string]string{
+		"Page.getKeyboard": "keyboard",
+	}
 }
 
 // TestMappings tests that all the methods of the API (api/) are
@@ -115,7 +117,9 @@ func TestMappings(t *testing.T) {
 		"page": {
 			apiInterface: (*api.Page)(nil),
 			mapp: func() mapping {
-				return mapPage(vu.Context(), vu, &common.Page{})
+				return mapPage(vu.Context(), vu, &common.Page{
+					Keyboard: &common.Keyboard{},
+				})
 			},
 		},
 		"elementHandle": {
--- a/browser/mapping_test.go
+++ b/browser/mapping_test.go
@@ -33,8 +33,9 @@ func wildcards() map[string]string {
 // API component properties as Goja objects.
 func getters() map[string]string {
 	return map[string]string{
-		"Page.getKeyboard": "keyboard",
-		"Page.getMouse":    "mouse",
+		"Page.getKeyboard":    "keyboard",
+		"Page.getMouse":       "mouse",
+		"Page.getTouchscreen": "touchscreen",
 	}
 }
 
@@ -119,8 +120,9 @@ func TestMappings(t *testing.T) {
 			apiInterface: (*api.Page)(nil),
 			mapp: func() mapping {
 				return mapPage(vu.Context(), vu, &common.Page{
-					Keyboard: &common.Keyboard{},
-					Mouse:    &common.Mouse{},
+					Keyboard:    &common.Keyboard{},
+					Mouse:       &common.Mouse{},
+					Touchscreen: &common.Touchscreen{},
 				})
 			},
 		},
--- a/browser/mapping_test.go
+++ b/browser/mapping_test.go
@@ -17,22 +17,20 @@ import (
 	k6metrics "go.k6.io/k6/metrics"
 )
 
-// wildcards is a list of extra mappings for our API (api/).
-func wildcards() map[string]string {
+// customMappings is a list of custom mappings for our API (api/).
+// Some of them are wildcards, such as query to $ mapping; and
+// others are for publicly accessible fields, such as mapping
+// of page.keyboard to Page.getKeyboard.
+func customMappings() map[string]string {
 	return map[string]string{
+		// wildcards
 		"Page.query":             "$",
 		"Page.queryAll":          "$$",
 		"Frame.query":            "$",
 		"Frame.queryAll":         "$$",
 		"ElementHandle.query":    "$",
 		"ElementHandle.queryAll": "$$",
-	}
-}
-
-// getters is a list of mappings that return internal
-// API component properties as Goja objects.
-func getters() map[string]string {
-	return map[string]string{
+		// getters
 		"Page.getKeyboard":    "keyboard",
 		"Page.getMouse":       "mouse",
 		"Page.getTouchscreen": "touchscreen",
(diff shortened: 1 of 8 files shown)


browser/mapping_test.go
```diff
@@ -17,6 +17,28 @@ import (
...
+// getters is a list of mappings that return internal
+// API component properties as Goja objects.
+func getters() map[string]string {
+	return map[string]string{
+		"Page.getKeyboard":    "keyboard",
+		"Page.getMouse":       "mouse",
+		"Page.getTouchscreen": "touchscreen",
+	}
+}
+
```
inancgumus, line 41: Can we join these two as `customMappings()` or something? I think having two of these is unnecessary (`wildcards()` and `getters()`); it complicates the rest of the code too. A comment in the new `customMappings` function can explain these.

I'll adjust where these are used so you can batch-commit them on Github UI.

```suggestion
// customMappings is a list of custom mappings for our API (api/).
// Some of them are wildcards, such as query to $ mapping; and
// others are for publicly accessible fields, such as mapping
// of page.keyboard to Page.getKeyboard.
func customMappings() map[string]string {
	return map[string]string{
		// wildcards
		"Page.query":             "$",
		"Page.queryAll":          "$$",
		"Frame.query":            "$",
		"Frame.queryAll":         "$$",
		"ElementHandle.query":    "$",
		"ElementHandle.queryAll": "$$",

		// getters
		"Page.getKeyboard":       "keyboard",
		"Page.getMouse":          "mouse",
		"Page.getTouchscreen":    "touchscreen",
	}
}
```

browser/mapping_test.go
```diff
@@ -36,6 +58,7 @@ func TestMappings(t *testing.T) {
 			},
 		}
 		wildcards = wildcards()
+		getters   = getters()
```
inancgumus, line 61:
```suggestion
		customMappings = customMappings()
```

browser/mapping_test.go
```diff
@@ -54,18 +77,19 @@ func TestMappings(t *testing.T) {
...
-			// if the method is a wildcard method, it should not
-			// be mapped to the module. so we should not find it
-			// in the mapped methods.
-			if _, ok := mapped[m]; wok && ok {
+			wgm, wgok := isWildcardOrGetter(wildcards, getters, typ.Name(), m)
+			// if the method is a wildcard or getter method, it should
+			// not be mapped to the module. so we should not find it in
+			// the mapped methods.
+
+			if _, ok := mapped[m]; wgok && ok {
```
inancgumus, line 85:
```suggestion
			cm, cmok := isCustomMapping(customMappings, typ.Name(), m)
			// if the method is a custom mapping, it should not be
			// mapped to the module. so we should not find it in
			// the mapped methods.
			if _, ok := mapped[m]; cmok && ok {
```

browser/mapping_test.go
```diff
@@ -54,18 +77,19 @@ func TestMappings(t *testing.T) {
...
 				t.Errorf("method %s should not be mapped", m)
 			}
 			// change the method name if it is mapped to a wildcard
 			// method. these wildcard methods are not exist on our
 			// API. so we need to use the mapped method instead.
-			if wok {
-				m = wm
+			if wgok {
+				m = wgm
```
inancgumus, line 92:
```suggestion
			// change the method name if it is mapped to a custom
			// method. these custom methods are not exist on our
			// API. so we need to use the mapped method instead.
			if cmok {
				m = cm
			}
```

browser/mapping_test.go
```diff
@@ -158,11 +186,17 @@ func toFirstLetterLower(s string) string {
...
-	return s, ok
+
+	if s, ok := wildcards[name]; ok {
+		return s, ok
+	}
+	if s, ok := getters[name]; ok {
+		return s, ok
+	}
+
+	return "", false
```
inancgumus, line 201:
```suggestion
// isCustomMapping returns true if the method is a custom mapping
// and returns the name of the method to be called instead of the
// original one.
func isCustomMapping(customMappings map[string]string, typ, method string) (string, bool) {
	name := typ + "." + method

	if s, ok := customMappings[name]; ok {
		return s, ok
	}

	return "", false
}
```

inancgumus (changes requested):
Awesome work 👏  Cleanly organized commits and commit messages. Flawless implementation 💯 Only one suggestion which you can easily batch commit.

browser/mapping_test.go
```diff
@@ -17,6 +17,18 @@ import (
 	k6metrics "go.k6.io/k6/metrics"
 )
 
+// wildcards is a list of extra mappings for our API (api/).
```
inancgumus, line 20: It should be harmless moving these here. On the other hand, having these on the `mapping.go` file's top location is to demonstrate to someone who might be reading the code which custom mappings we have—They can learn by reading the tests too—so it's harmless to move them here. Just a note :)

browser/mapping_test.go
ka3de: Yes, I think it makes complete sense to merge them. Thanks for the suggestion and modifications. Applying them!

ka3de: Seems like GH is not allowing me to batch commit all these suggestions due to:
```
Applying suggestions on deleted lines is currently not supported. Please and try again. 
```
:thinking: 
But I will address them manually and push adding you as co-author.

ka3de: Done in [c05242a](https://github.com/grafana/xk6-browser/pull/751/commits/c05242a2311ae41694ebfba13a59591108b09e4e).

inancgumus (approved):
💯

ankur22 (approved):
👏 nicely done!

LGTM 🙂

PR #754, opened 2023-02-06
Fix to use a fresh VU context on each mapped method call

ankur22:
The initial context that we get from the vu in NewModuleInstance should not be cached since the context is closed per iteration. The best practice is to work with the context directly from vu when it's needed.

This became apparent when the initial context was used in a test API method (`Launch`) and the test was failing with `context canceled`.

Related: #704 #702 #673 #744

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -1,6 +1,7 @@
 package browser
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/dop251/goja"
@@ -12,6 +13,24 @@ import (
 	k6common "go.k6.io/k6/js/common"
 	k6modules "go.k6.io/k6/js/modules"
 )
+
+// moduleVU carries module specific VU information.
+//
+// Currently, it is used to carry the VU object to the
+// inner objects and promises.
+type moduleVU struct {
+	k6modules.VU
+}
+
+func (vu moduleVU) Context() context.Context {
+	// promises and inner objects need the VU object to be
+	// able to use k6-core specific functionality.
+	//
+	// We should not cache the context (especially the init
+	// context from the vu that is received from k6 in
+	// NewModuleInstance).
+	return k6ext.WithVU(vu.VU.Context(), vu.VU)
+}
 
 // mapping is a type of mapping between our API (api/) and the JS
 // module. It acts like a bridge and allows adding wildcard methods
@@ -27,7 +46,7 @@ type mapping = map[string]any
 // The motivation of this mapping was to support $ and $$ wildcard
 // methods.
 // See issue #661 for more details.
-func mapBrowserToGoja(vu k6modules.VU) *goja.Object {
+func mapBrowserToGoja(vu moduleVU) *goja.Object {
 	var (
 		rt          = vu.Runtime()
 		obj         = rt.NewObject()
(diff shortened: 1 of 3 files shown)


ka3de (approved):
I wonder if it would be better to define the `ctx` on the top level of each mapping function instead of in each map property that requires a promise. But overall LGTM.

inancgumus (commented):
@ankur22, could you show me a way to test the behavior that this PR fixes? I want to compare the behavior before (v0.8) and after this PR.

By the way, can we use the solution I implemented to do this [before](https://github.com/grafana/xk6-browser/blob/3a4bee6af98006b544ae0a5af1267d23b1c472da/browser/module.go#L38-L81)? Is there a specific reason not to do so?

ankur22:
> I wonder if it would be better to define the `ctx` on the top level of each mapping function instead of in each map property that requires a promise. But overall LGTM.

Then we will be using the stale context again.

ankur22:
> @ankur22, could you show me a way to test the behavior that this PR fixes? I want to compare the behavior before (v0.8) and after this PR.

I did try to think of way to create a test for this but I couldn't 🤔  I used the following to verify the close state of init context and the context from the vu:

```diff
diff --git a/browser/mapping.go b/browser/mapping.go
index 45f6776..43f1b83 100644
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -616,6 +616,7 @@ func mapBrowser(vu k6modules.VU, b api.Browser) mapping {
 
 // mapBrowserType to the JS module.
 func mapBrowserType(vu k6modules.VU, bt api.BrowserType) mapping {
+       ctx := vu.Context()
        rt := vu.Runtime()
        return mapping{
                "connect":                 bt.Connect,
@@ -623,6 +624,15 @@ func mapBrowserType(vu k6modules.VU, bt api.BrowserType) mapping {
                "launchPersistentContext": bt.LaunchPersistentContext,
                "name":                    bt.Name,
                "launch": func(opts goja.Value) *goja.Object {
+                       if ctx.Done() != nil {
+                               fmt.Println("init context has an error", ctx.Err())
+                       }
+
+                       ctx := k6ext.WithVU(vu.Context(), vu)
+                       if ctx.Done() != nil {
+                               fmt.Println("vu context does not have an error")
+                       }
+
                        m := mapBrowser(vu, bt.Launch(opts))
                        return rt.ToValue(m).ToObject(rt)
```

> By the way, can we use the solution I implemented to do this [before](https://github.com/grafana/xk6-browser/blob/3a4bee6af98006b544ae0a5af1267d23b1c472da/browser/module.go#L38-L81)? That way, every mapping method can access the current VU, no? Is there a specific reason not to do so?

I have a slight preference not to, but it would be useful to put a comment somewhere explaining why we need to create a new context from the vu so i'll add it back in 🙂

ka3de:
> > I wonder if it would be better to define the `ctx` on the top level of each mapping function instead of in each map property that requires a promise. But overall LGTM.
> 
> Then we will be using the stale context again. Try the diff i suggested in [#754 (comment)](https://github.com/grafana/xk6-browser/pull/754#issuecomment-1420436064)
You are right, I misunderstood the implementation and thought that at that scope the `vu` was "live", but it would be a stale `ctx` also. Got it :+1:

inancgumus:
@ka3de [That trick here](https://github.com/grafana/xk6-browser/pull/754#pullrequestreview-1286582668) will allow us to use a fresh VU context without needing to attach the VU to the context.

inancgumus (approved):
LGTM 👍

ankur22:
Correction: I've updated the diff in the https://github.com/grafana/xk6-browser/pull/754#issuecomment-1420436064 to a test that actually exhibits the closed `context` on the init `context` and the unclose `context` when grabbing a fresh copy from the `vu`.

PR #755, opened 2023-02-06
Refactor force browser close on panic

ankur22:
We're currently storing the browser pid in the context. This is an issue because:

1. When working on https://github.com/grafana/xk6-browser/pull/744 it wasn't cleanly possible to panic at the mapping layer and kill all the browser processes since the pid was in a child context which we didn't easily have access to.
2. If a user launches more than one browser, and a panic occurs, only the last launched browser will be closed since that's the only pid in the context.

This change adds a `register` which is internally updated by `BrowserProcess`. When there is a need to panic (with `k6ext.Panic`) we can use the `register`'s `ForceProcessShutdown`, which will loop over all the registered `pid`s and force kill the processes.


browserprocess/browser_process.go
```diff
@@ -69,13 +69,6 @@ func New(
 	}
 
 	go func() {
-		defer func() {
```
ankur22: Unregister doesn't give us anything, so it's been removed.

browserprocess/register_test.go
```diff
@@ -1,57 +0,0 @@
-package browserprocess
```
ankur22: This became a bit meaningless after removing `unregister`. Testing this is a little tricky, but let me try again 🤔

browserprocess/browser_process.go
```diff
@@ -1,4 +1,6 @@
-package common
+// Package browserprocess is in charge of the interaction
+// with the OS process that is running the browser.
+package browserprocess
```
inancgumus, line 3: One suggestion: `browserprocess.BrowserProcess` stutters. Would it be better to rename it to `osext`? `osext.Process`?

That would also be consistent with the comment you wrote before the package definition: _"the interaction with the OS process"_. In the future, we can add more operating-system-related stuff into the same package without creating many packages. That is how it's done in the Go standard library as well. [See](https://pkg.go.dev/os).

inancgumus (commented):
Nice idea 👍  I'll take a look at this in detail later. Here's one suggestion after skimming the PR.

browserprocess/register.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/grafana/xk6-browser/log"
+)
+
+type processState struct {
+	pid int
+}
+
+var (
+	browserProcessRegister   = map[string]*processState{} //nolint:gochecknoglobals
```
ka3de, line 18: Can we simplify this map now that we don't need to store the `killed` status as `map[string]int` being: `pid+iID: pid` and get rid of the `processState` struct?

browserprocess/register.go
```diff
@@ -0,0 +1,59 @@
...
+// xk6-browser is having to shutdown due to an
+// internal error (and therefore a panic).
+func ForceProcessShutdown(ctx context.Context) {
+	browserProcessRegisterMu.Lock()
+	defer browserProcessRegisterMu.Unlock()
+
+	iID := GetIterationID(ctx)
+
+	for k, v := range browserProcessRegister {
+		if iID != "" && !strings.Contains(k, iID) {
```
ka3de, line 44: I'm a little bit skeptical about generating the keys for the register as a concatenation of `iterationID` and `pid`, and also about using `strings.Contains` to verify. I was experimenting with using to different data structures for the map from `iID->pid` and the list of `pid` but was resulting rather complex, I will try to provide an alternative. Did you consider other options?

ka3de (commented):
Thanks for working on this @ankur22 ! Added a couple of comments.

ka3de (commented):
I am providing a suggestion with which I was experimenting in order to avoid the usage of string concatenation for `iterationID`-`pid` relationship. Feel free to discard it or discuss just some options of it. We take advantage of the fact that when not using `test_browser`, we'll have one "default iteration ID" which is a void string. Or we could even fix that in the `GetIterationID` method.
Let me know if I misunderstood something.

browserprocess/register.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/grafana/xk6-browser/log"
+)
+
+type processState struct {
+	pid int
+}
+
+var (
+	browserProcessRegister   = map[string]*processState{} //nolint:gochecknoglobals
+	browserProcessRegisterMu = sync.Mutex{}               //nolint:gochecknoglobals
```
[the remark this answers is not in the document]
ankur22: Yeah, i did consider that, but the string concatenation was a lot easier to chuck in. I'll change it to this, thanks for the suggestion 🙂

browserprocess/register.go
ankur22: Good point, although it's needed in the [next PR](https://github.com/grafana/xk6-browser/pull/762) that i've created 😅 Maybe i'll revert it here still to show that it's not needed for this particular change 👍

browserprocess/register.go
ankur22: Yes, it's a smell. I like the approach you suggested 👍

browserprocess/browser_process.go
ankur22: Yeah, nice idea. I've amended it in a1f72e472bcca9d4522a99892e4498da5111f771.

tests/test_browser.go
```diff
@@ -82,6 +84,16 @@ func newTestBrowser(tb testing.TB, opts ...any) *testBrowser {
 		vu.CtxField = ctx
 	}
 
+	// This is used to help the browserprocess register
+	// distinguish between different error runs so that
+	// a panic in one test doesn't force close the browser
+	// in other running tests.
+	u, err := uuid.NewV4()
```
ankur22, line 91: I could use a global atomic counter instead of `uuid` so that we're not having to depend on another library, WDYT?

ka3de: I think the `uuid` library will be required anyway because it's included in k6, but I think an atomic counter will make it simpler so +1 for me.

osext/browser_process.go
```diff
@@ -23,20 +25,23 @@ type BrowserProcess struct {
...
+	UserDataDir *storage.Dir
 
 	logger *log.Logger
 }
 
-func NewBrowserProcess(
+// New creates a new osext. It will start
+// a child process where the internet browser will
+// run.
+func New(
```
ka3de, line 44: Considering that we have now renamed the pkg to `osext`, should we rename this to `NewBrowserProcess`?

browserprocess/browser_process.go
inancgumus: Thanks for the amend! What about `BrowserProcess` to `Process`? I don't foresee any other process types in the near future :)

ankur22: I like that. It's confusing otherwise. Resolved in 8af5625586f2ccd7a1a2f29451840fe955076f24

osext/browser_process.go
ankur22: Yep, good spot. Resolved in 2cdcab34d24729f1385b9a6660f27232204ab13e

ka3de:
Just one last comment briefly discussed in https://github.com/grafana/xk6-browser/pull/755#pullrequestreview-1288724078 about the "default iteration ID" that we can make it more explicit for example by defining a constant in `osext/context.go` such as:
```
    defIterationID = "default" // Or something similar
```
And then modify the `GetIterationID` method in order to return the default value in case the given context does not contain a value for iteration ID key. But happy to leave it as is counting on the void string as default.

ka3de (approved):
LGTM.

ankur22:
> defIterationID = "default"

This is a good idea! Avoids possible confusion. Keep comments and suggestions coming if you think it could help 🙂 

Resolved in ed5498c9cb40e37b9e96e2611152e2faa4410319

ka3de (approved):
LGTM :tada:

inancgumus (commented):
LGTM 👍 Some questions and suggestions.

I also wonder about the reason of associating processIDs to runIDs? Could we store the process IDs in `processRegister map[]int`? In the end, we want to kill processes, and I wrap my head around why do we need to associate the process IDs with run IDs? 🤔😅

ankur22:
> I also wonder about the reason for associating processIDs with runIDs? Could we store the process IDs in `processRegister map[string]int` or just `[]int`? In the end, we want to kill processes, and I can't wrap my head around why we need to associate the process IDs with run IDs? 🤔😅 If that's possible, then maybe we can get rid of storing data in context.

For the integration tests it makes sense to work with the pid as the `runID`, since each test should be running its own browser and any panic/kill should be isolated to that test and browser instance.

When running `k6-browser` tests, the user could create two `browser` class instances, opening two separate browsers:

```javascript
  const browser = chromium.launch({headless: false});

  const browser2 = chromium.launch({headless: false});
```

If we were to panic and kill based on the `pid`, we'd only force close one browser, and not the other. This is why we have a `default` `runId`.

> Nit: I also prefer to use something other than package-level variables. Although more straightforward at the beginning, they are error-prone and can be overwritten by different parts of code. Do you think we should find another way?

Yes, i've been thinking about this too. I'm not a fan of doing it this way, but as far as I can tell there's no nice way to inject a `register` instance to both the `osext.Process` (which is registering the browser pids) and `k6ext.Panic` without affecting the current API and a lot of code.

The only way I could think of was with the context, but i think that a misuse of the context (although we're already doing it with `vu`).

Eventually though, we will be able to find a clean way when we get to the point of only panicking in the `mapping` layer, at that point it should be easy to inject an instance of `register` into the mapping layer as well into `osext.Process`.

tests/test_browser.go
```diff
@@ -35,6 +37,8 @@ type testBrowser struct {
 	api.Browser
 }
 
+var iID uint64 //nolint:gochecknoglobals
```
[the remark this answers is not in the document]
ankur22: I think the responsibility of the `runID` `uint64` and where it's incremented should be done in the tests. The whole `runID`"hack" is there due to the tests, so if we can keep as much of it with the `test` package then I think that's a better signal to future devs.

ankur22:
@inancgumus @ka3de take a look at https://github.com/grafana/xk6-browser/pull/762, which could affect this PR too 😅

inancgumus (changes requested):
Please see #779.

PR #759, opened 2023-02-07
Remove the browser closing state

ankur22:
The `closing` state is not needed:
1. Nothing else relies on the `closing` state.
2. Straight after `closing` is set, `closed` is set.

We still need the `closed` state as there is code that changes behaviour if the state is no longer `open`.


inancgumus (approved):
I feel much lighter, thanks 👍

ka3de (approved):
LGTM.

PR #760, opened 2023-02-07
Add implementation for adding cookies

### LINKED ISSUE #5: Implement `BrowserContext.addCookies(cookies)`
Add support for `BrowserContext.addCookies(cookies)`, allowing users to add default cookies to all pages created in browser context.

**Relevant links**:
- Playwright docs: https://playwright.dev/docs/api/class-browsercontext#browsercontextaddcookiescookies
- xk6-browser code to start: https://github.com/grafana/xk6-browser/blob/main/common/browser_context.go#L75
- Playwright code: https://github.com/microsoft/playwright/blob/master/src/server/chromium/crBrowser.ts#L382
- Playwright tests: https://github.com/microsoft/playwright/blob/master/tests/browsercontext-add-cookies.spec.ts

zucchinho:
Closes #5

Signed-off-by: zookee <z.persboy@gmail.com>

### DIFF
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -71,15 +71,15 @@ func NewBrowserContext(
 func (b *BrowserContext) AddCookies(cookies goja.Value) {
 	b.logger.Debugf("BrowserContext:AddCookies", "bctxid:%v", b.id)
 
-	cookieParams := []network.CookieParam{}
-	if cookies == nil || goja.IsUndefined(cookies) || goja.IsNull(cookies) {
+	var cookieParams []network.CookieParam
+	if !gojaValueExists(cookies) {
 		k6ext.Panic(b.ctx, "adding cookies: cookies value is not set")
 	}
 
 	rt := b.vu.Runtime()
 	err := rt.ExportTo(cookies, &cookieParams)
 	if err != nil {
-		k6ext.Panic(b.ctx, "adding cookies: %v", err)
+		k6ext.Panic(b.ctx, "adding cookies: %w", err)
 	}
 
 	// Create new array of pointers to items in cookieParams
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -71,24 +71,25 @@ func NewBrowserContext(
 func (b *BrowserContext) AddCookies(cookies goja.Value) {
 	b.logger.Debugf("BrowserContext:AddCookies", "bctxid:%v", b.id)
 
-	var cookieParams []network.CookieParam
+	cookieParams := []network.CookieParam{}
 	if cookies != nil && !goja.IsUndefined(cookies) && !goja.IsNull(cookies) {
-		err := b.vu.Runtime().ExportTo(cookies, &cookieParams)
+		rt := b.vu.Runtime()
+		err := rt.ExportTo(cookies, &cookieParams)
 		if err != nil {
 			k6ext.Panic(b.ctx, "adding cookies: %v", err)
 		}
-	}
 
-	// Create new array of pointers to items in cookieParams
-	var cookieParamsPointers []*network.CookieParam
-	for i := 0; i < len(cookieParams); i++ {
-		cookieParamsPointers = append(cookieParamsPointers, &cookieParams[i])
-	}
+		// Create new array of pointers to items in cookieParams
+		var cookieParamsPointers []*network.CookieParam
+		for i := 0; i < len(cookieParams); i++ {
+			cookieParamsPointers = append(cookieParamsPointers, &cookieParams[i])
+		}
 
-	action := storage.SetCookies(cookieParamsPointers).WithBrowserContextID(b.id)
+		action := storage.SetCookies(cookieParamsPointers).WithBrowserContextID(b.id)
 
-	if err := action.Do(cdp.WithExecutor(b.ctx, b.browser.conn)); err != nil {
-		k6ext.Panic(b.ctx, "adding cookies: %w", err)
+		if err := action.Do(cdp.WithExecutor(b.ctx, b.browser.conn)); err != nil {
+			k6ext.Panic(b.ctx, "adding cookies: %w", err)
+		}
 	}
 }
 
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -72,24 +72,26 @@ func (b *BrowserContext) AddCookies(cookies goja.Value) {
 	b.logger.Debugf("BrowserContext:AddCookies", "bctxid:%v", b.id)
 
 	cookieParams := []network.CookieParam{}
-	if cookies != nil && !goja.IsUndefined(cookies) && !goja.IsNull(cookies) {
-		rt := b.vu.Runtime()
-		err := rt.ExportTo(cookies, &cookieParams)
-		if err != nil {
-			k6ext.Panic(b.ctx, "adding cookies: %v", err)
-		}
+	if cookies == nil || goja.IsUndefined(cookies) || goja.IsNull(cookies) {
+		k6ext.Panic(b.ctx, "adding cookies: cookies value is not set")
+	}
 
-		// Create new array of pointers to items in cookieParams
-		var cookieParamsPointers []*network.CookieParam
-		for i := 0; i < len(cookieParams); i++ {
-			cookieParamsPointers = append(cookieParamsPointers, &cookieParams[i])
-		}
+	rt := b.vu.Runtime()
+	err := rt.ExportTo(cookies, &cookieParams)
+	if err != nil {
+		k6ext.Panic(b.ctx, "adding cookies: %v", err)
+	}
+	b.logger.Infof("BrowserContext:AddCookies", "cookies: %#v", cookieParams)
 
-		action := storage.SetCookies(cookieParamsPointers).WithBrowserContextID(b.id)
+	// Create new array of pointers to items in cookieParams
+	var cookieParamsPointers []*network.CookieParam
+	for i := 0; i < len(cookieParams); i++ {
+		cookieParamsPointers = append(cookieParamsPointers, &cookieParams[i])
+	}
 
-		if err := action.Do(cdp.WithExecutor(b.ctx, b.browser.conn)); err != nil {
-			k6ext.Panic(b.ctx, "adding cookies: %w", err)
-		}
+	action := storage.SetCookies(cookieParamsPointers).WithBrowserContextID(b.id)
+	if err := action.Do(cdp.WithExecutor(b.ctx, b.browser.conn)); err != nil {
+		k6ext.Panic(b.ctx, "adding cookies: %w", err)
 	}
 }
 
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -72,7 +72,6 @@ func (b *BrowserContext) AddCookies(cookies goja.Value) {
 	b.logger.Debugf("BrowserContext:AddCookies", "bctxid:%v", b.id)
 
 	err := b.addCookies(cookies)
-
 	if err != nil {
 		k6ext.Panic(b.ctx, "adding cookies: %w", err)
 	}
@@ -478,7 +477,7 @@ func (b *BrowserContext) addCookies(cookies goja.Value) error {
 
 	action := storage.SetCookies(cookieParamsPointers).WithBrowserContextID(b.id)
 	if err := action.Do(cdp.WithExecutor(b.ctx, b.browser.conn)); err != nil {
-		return err
+		return fmt.Errorf("unable to execute SetCookies action: %w", err)
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


ankur22 (commented):
Thanks for the change @zucchinho!

This is a great start. Can you please:
1. Rebase against the latest version of `xk6-browser` `main` branch?
2. Add some tests. You can place them in the `tests` directory.

ankur22 (changes requested):
Changes requested in my previous comment. Thanks!

zucchinho:
> Thanks for the change @zucchinho!
> 
> This is a great start. Can you please:
> 
> 1. Rebase against the latest version of `xk6-browser` `main` branch?
> 2. Add some tests. You can place them in the `tests` directory.

Hey, thanks for the quick review. I've written a test locally, but i'm having trouble with the `goja` library which marshals the JSON. There's not a great amount of documentation for it, but afaict, something like this _should_ mirror what is happening in the JS module itself (which works fine when I run a test using the local build). 

```
		vm := goja.New()
		arr, err := vm.RunString(`
			[
				{
					name: "test_cookie_name",
					value: "test_cookie_value",
					url: "https://test.go"
				}
			];
		`)
		require.NoError(t, err)
		tc.bc.AddCookies(arr)
```

but I'm getting

```
--- FAIL: TestBrowserContextAddCookies (0.00s)
    --- FAIL: TestBrowserContextAddCookies/happy_path (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x2 addr=0x30 pc=0x100ff65e0]
```

i'm almost definitely missing something obvious here. any legup appreciated, not at all familiar with `goja`

zucchinho:
nvm, it was just the runtime wasn't set on the browser. Using the `newTestBrowser` function to init the browser fixed things. I've add the tests, but there's not really anything to assert as yet, since the `Cookies()` method to get the cookies isn't implemented yet.

common/browser_context.go
```diff
@@ -71,24 +71,25 @@ func NewBrowserContext(
 func (b *BrowserContext) AddCookies(cookies goja.Value) {
 	b.logger.Debugf("BrowserContext:AddCookies", "bctxid:%v", b.id)
 
-	var cookieParams []network.CookieParam
+	cookieParams := []network.CookieParam{}
```
ankur22, line 74: Can we catch the errors earlier so that the happy path is left aligned as much as possible. What i mean by that is doing something like:

```
if cookies == nil || goja.IsUndefined(cookies) || goja.IsNull(cookies) {
   // k6ext.Panic(
}

// happy path
```

ankur22 (changes requested):
Thanks for the tests! 😄 I've added some comments that should help you fill in the `TODO`s.

zucchinho:
Cool, i've made the suggested changes. Thanks for the tip on the static html server, I hadn't considered that, seems a better assertion than just using an adjacent method on the test object, neat. 

I've grouped the error test cases into a table, and ended up leaving the happy_path as a separate test case, as with the new assertions there wasn't enough overlap to warrant it.

common/browser_context.go
inancgumus: There is also `gojaValueExists` helper :)

ankur22 (changes requested):
Nicely done, and thank you for taking the suggestions and comments on board 🙂 I've left a few more comments.

common/browser_context.go
```diff
@@ -72,24 +72,26 @@ func (b *BrowserContext) AddCookies(cookies goja.Value) {
...
-		var cookieParamsPointers []*network.CookieParam
-		for i := 0; i < len(cookieParams); i++ {
-			cookieParamsPointers = append(cookieParamsPointers, &cookieParams[i])
-		}
+	rt := b.vu.Runtime()
+	err := rt.ExportTo(cookies, &cookieParams)
+	if err != nil {
+		k6ext.Panic(b.ctx, "adding cookies: %v", err)
+	}
+	b.logger.Infof("BrowserContext:AddCookies", "cookies: %#v", cookieParams)
```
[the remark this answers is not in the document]
zucchinho: sorry, didn't mean to leave this in! sloppy. i'll just remove it

common/browser_context.go
inancgumus: Btw, @zucchinho, This `var cookieParams []network.CookieParam` was OK; no need to change that.

common/browser_context.go
```diff
@@ -72,24 +72,26 @@ func (b *BrowserContext) AddCookies(cookies goja.Value) {
...
 	cookieParams := []network.CookieParam{}
-	if cookies != nil && !goja.IsUndefined(cookies) && !goja.IsNull(cookies) {
-		rt := b.vu.Runtime()
-		err := rt.ExportTo(cookies, &cookieParams)
-		if err != nil {
-			k6ext.Panic(b.ctx, "adding cookies: %v", err)
-		}
+	if cookies == nil || goja.IsUndefined(cookies) || goja.IsNull(cookies) {
+		k6ext.Panic(b.ctx, "adding cookies: cookies value is not set")
+	}
```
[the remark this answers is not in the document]
inancgumus: Replug 😅 Nice to use this: https://github.com/grafana/xk6-browser/pull/760#discussion_r1100164309.

common/browser_context.go
```diff
@@ -63,9 +66,33 @@ func NewBrowserContext(
...
 }
 
-// AddCookies is not implemented.
+// AddCookies adds cookies into this browser context.
+// All pages within this context will have these cookies installed.
 func (b *BrowserContext) AddCookies(cookies goja.Value) {
-	k6ext.Panic(b.ctx, "BrowserContext.addCookies(cookies) has not been implemented yet")
+	b.logger.Debugf("BrowserContext:AddCookies", "bctxid:%v", b.id)
+
+	cookieParams := []network.CookieParam{}
```
inancgumus, line 74: This `var cookieParams []network.CookieParam` was OK; no need to change that.

tests/browser_context_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"testing"
+
+	"github.com/dop251/goja"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBrowserContextAddCookies(t *testing.T) {
+	opts := defaultLaunchOpts()
+	tb := newTestBrowser(t, withFileServer(), opts)
```
inancgumus, line 14: The following is the same (the default options are used by default):

```suggestion
	tb := newTestBrowser(t, withFileServer())
```

tests/browser_context_test.go
```diff
@@ -0,0 +1,82 @@
...
+		cmd := fmt.Sprintf(`
+			[
+				{
+					name: "%v",
+					value: "%v",
+					url: "%v"
+				}
+			];
+		`, testCookieName, testCookieValue, tb.URL(""))
+		cookies, err := tb.vu.Runtime().RunString(cmd)
```
inancgumus, line 30: Could you use the helper function [`testBrowser.runJavaScript`](https://github.com/grafana/xk6-browser/blob/main/tests/test_browser.go#L232) here?

tests/browser_context_test.go
```diff
@@ -0,0 +1,82 @@
...
+	})
+
+	type errorTestCase struct {
+		description string
+		cookiesCmd  string
+	}
+
+	for _, testCase := range []errorTestCase{
+		{
+			description: "nil_cookies",
```
inancgumus, line 57: I liked the test case names; short and sweet 👍

tests/browser_context_test.go
```diff
@@ -0,0 +1,82 @@
...
+			cookiesCmd:  "null;",
+		},
+		{
+			description: "goja_undefined_cookies",
+			cookiesCmd:  "undefined;",
+		},
+	} {
+
+		t.Run(testCase.description, func(t *testing.T) {
+			var cookies goja.Value = nil
```
inancgumus, line 71: The following is the same:

```suggestion
			var cookies goja.Value
```

tests/browser_context_test.go
```diff
@@ -0,0 +1,82 @@
...
+			if testCase.cookiesCmd != "" {
+				var err error
+				cookies, err = tb.vu.Runtime().RunString(testCase.cookiesCmd)
```
inancgumus, line 74: Could you use the helper function [testBrowser.runJavaScript](https://github.com/grafana/xk6-browser/blob/main/tests/test_browser.go#L232) also here?

inancgumus (changes requested):
Hi, @zucchinho! Thanks for putting up the effort by implementing this feature 👏  Mostly LGTM, only a couple of suggestions to simplify the code.

zucchinho:
i think i have addressed all the comments above :)

tests/browser_context_test.go
```diff
@@ -0,0 +1,180 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/dop251/goja"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBrowserContextAddCookies(t *testing.T) {
```
ankur22, line 12: 👏 This is great! Just one comment more: could you make these tests run in parallel? Take a look at [this](https://github.com/zucchinho/xk6-browser/blob/2aa3ad11b531c0613e39d8a6a70f3d7120605bb2/tests/locator_test.go#L20-L21) and [this](https://github.com/zucchinho/xk6-browser/blob/2aa3ad11b531c0613e39d8a6a70f3d7120605bb2/tests/locator_test.go#L182-L184) as a guide.

ankur22 (changes requested):
I think we're on the home stretch. I left one comment about running tests in parallel, but otherwise i think it looks good 👍

zucchinho:
any ideas on the failing tests? don't seem related to my changes 🤔

inancgumus:
> any ideas on the failing tests? don't seem related to my changes 🤔

Don't worry. We have some flaky tests. I reran the tests.

tests/browser_context_test.go
inancgumus: Yes, the browsers will run in parallel.

common/browser_context.go
```diff
@@ -63,9 +66,53 @@ func NewBrowserContext(
 	return &b
 }
 
-// AddCookies is not implemented.
+// AddCookies adds cookies into this browser context.
+// All pages within this context will have these cookies installed.
 func (b *BrowserContext) AddCookies(cookies goja.Value) {
-	k6ext.Panic(b.ctx, "BrowserContext.addCookies(cookies) has not been implemented yet")
+	b.logger.Debugf("BrowserContext:AddCookies", "bctxid:%v", b.id)
```
inancgumus, line 72: This is optional for tackling right now. You might want to take a look if you have the time 🙇 

Instead of panicking in multiple places, it would be better to add another method called `addCookies`, return errors from it, call it from `AddCookies`, and panic only once in `AddCookies`. And we can eliminate prefixing the panic error only once with the following: `"adding cookies: "` in `AddCookies`.

tests/browser_context_test.go
```diff
@@ -0,0 +1,180 @@
...
+		cmd := fmt.Sprintf(`
+			[
+				{
+					name: "%v",
+					value: "%v",
+					url: "%v"
+				}
+			];
+		`, testCookieName, testCookieValue, tb.URL(""))
+		cookies, err := tb.runJavaScript(cmd)
```
inancgumus, line 29: No need for `Sprintf` here as `runJavascript` also acts like `Sprintf`.

inancgumus (approved):
I've added some optional suggestions. Take a look at them if you have the time.

Thanks for your contribution 🙇

tests/browser_context_test.go
zucchinho: i more mean in terms of resources available to the CI agent, is it actually more performant to run a separate instance of a browser for each test, even if they are running in parallel? or just spin up a single instance of a browser and then sequentially run the tests against that instance? most of the test execution time will be the instantiation of the browser. i'm happy to make the change anyway

zucchinho: oh looking closer at the examples you linked, i think i understand a bit better what you mean. we'd like the run the happy_path in parallel with all of the error cases, which are run sequentially?

ankur22: Yeah, you are right to raise concern whether there's any benefit to running them in parallel. At the moment it helps in a couple of ways:
- We think in general it does help speed the whole test suite.
- It isolates tests in their own browser instance. This prevents tests from interfering with each other due to any caching or config changes.

We parallelise all tests (including sub tests within a table test), unless there's a very good reason not to (at least for now).

zucchinho: no problem, have done that

ankur22 (approved):
LGTM 🎉

PR #762, opened 2023-02-07
Fix zombie browser by only kill

ankur22:
While testing I noticed that zombie browser processes were being left behind. It turns out the we don't need to use process [Release](https://pkg.go.dev/os#Process.Release) since we're using `Wait`. `Release` will prevent `Kill` from working since the `pid` in the process structure is overwritten, signalling that the process has already been released. Using `Kill` alone now kills the browser processes and no more zombie browser processes when a panic forces the browser to shutdown.

### DIFF
--- a/osext/register.go
+++ b/osext/register.go
@@ -46,6 +46,5 @@ var Kill = func(pid int) { //nolint:gochecknoglobals
 		return
 	}
 	// no need to check the error since we're already dying.
-	_ = p.Release()
 	_ = p.Kill()
 }


common/browser.go
```diff
@@ -409,6 +409,11 @@ func (b *Browser) Close() {
 	b.conn.IgnoreIOErrors()
 	b.browserProc.GracefulClose()
 
+	if b.browserProc.Killed(b.ctx) {
```
ankur22, line 412: This i think is racey. What if the kill process takes a bit longer, which would mean that we can still block on the cdp write/read command.

osext/register.go
```diff
@@ -36,15 +36,21 @@ func ForceProcessShutdown(ctx context.Context) {
...
-		// no need to check the error for waiting the process to release
-		// its resources or whether we could kill it as we're already
-		// dying.
-		_ = p.Release()
-		_ = p.Kill()
+		Kill(pid)
 	}
 }
+
+// Kill will look for and kill the process with the
```
ankur22, line 43: This change will also mean we shouldn't need the `runID` work that is in https://github.com/grafana/xk6-browser/pull/755.

ka3de:
Nice catch and thanks for the explanation! I was trying to come up with some alternative, as exposing the `kill` function seems a little bit hacky, but couldn't come up with a valid one, and I've been forced in the past to do this due to tests needing to overwrite some functionality also, so LGTM.

inancgumus (commented):
Hey, @ankur22!

I ran all the tests (including `examples/`) several times (with `-count=100`) **with/without using `osext.ForceProcessShutdown(ctx)` in `k6ext.Panic`** (comment it out) and no zombie process left behind. I've also tested using JavaScript tests that launches many browser without closing them. I'm not sure about the exact scenario that reproduces this.

Could you show me the steps so I can reproduce this?

ankur22:
> Hey, @ankur22!
> 
> I ran all the tests (including `examples/`) several times (with `-count=100`) **with/without using `osext.ForceProcessShutdown(ctx)` in `k6ext.Panic`** (comment it out) and no zombie process left behind. I've also tested using JavaScript tests that launches many browser without closing them. I'm not sure about the exact scenario that reproduces this.
> 
> Could you show me the steps so I can reproduce this?

😓 I wish i had saved the test, I cannot recreate this issue now without adding some temporary code to demo the issue 😅 

So the issue is when we have promisified methods. For this demo I've just picked a random unimplemented method `goBack` and promisified it like so:

```diff
diff --git a/browser/mapping.go b/browser/mapping.go
index 7cc761d..aca5c8a 100644
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -428,8 +428,13 @@ func mapPage(vu moduleVU, p api.Page) mapping {
                        return rt.ToValue(mfrs).ToObject(rt)
                },
                "getAttribute": p.GetAttribute,
-               "goBack":       p.GoBack,
-               "goForward":    p.GoForward,
+               "goBack": func(opts goja.Value) *goja.Promise {
+                       return k6ext.Promise(vu.Context(), func() (any, error) {
+                               resp := p.GoBack(opts)
+                               return mapResponse(vu, resp), nil
+                       })
+               },
+               "goForward": p.GoForward,
                "goto": func(url string, opts goja.Value) *goja.Promise {
                        return k6ext.Promise(vu.Context(), func() (any, error) {
                                resp, err := p.Goto(url, opts)
```

If you now run the following:

```

export default function () {
  const browser = chromium.launch({headless: false});

  const browser2 = chromium.launch({headless: false});

  page.goto('https://test.k6.io/flip_coin.php')
  .then(() => {
    return page2.goBack()
  })
  .finally(() => {
    page.close();
    browser.close();

    page2.close();
    browser2.close();
  })
}
```

Without the removal of `Release` you should end up with zombie browser processes. With the fix, the browser processes should be closed and so no zombie browser processes.

---

Before this fix, when running tests (either integration tests or normally running tests), I was sometimes left with zombie browser processes (at least once a week I have to restart my laptop). I'm hoping that this will fix will also resolve that issue too 🤞 I only wish I had recorded those cases in an issue instead of ignoring them 🤦

PR #772, opened 2023-02-08
Remove deprecated default browser flags

### LINKED ISSUE #771: Remove deprecated default browser flags
In [cfd6a83](https://github.com/grafana/xk6-browser/tree/cfd6a83dfc42b6bbba87dc4ecd1804b6e4c5eca6) we have a [list of flags](https://github.com/grafana/xk6-browser/blob/cfd6a83dfc42b6bbba87dc4ecd1804b6e4c5eca6/chromium/browser_type.go#L277-L321) that are set by default when launching a Chromium browser. Some of these flags are no longer [valid](https://peter.sh/experiments/chromium-command-line-switches/) as they've been deprecated / overlapped. More specifically:

- `disable-client-side-phishing-detection`: Deprecated, see [issue](https://bugs.chromium.org/p/chromium/issues/detail?id=360220) in Chromium project. In this case maybe we should evaluate the usage of `safebrowsing-enable-enhanced-protection`.
- `safebrowsing-disable-auto-update`: Deprecated, see [issue](https://bugs.chromium.org/p/chromium/issues/detail?id=354743) in Chromium project. Overlapped by `disable-background-networking`, which is already set as default argument in our implementation.
- `disable-sync`: No longer valid (could not find much information about this).
- `enable-use-zoom-for-dsf`: In our implementation this is only set for macOS, but this seems to be set now by default in [macOS](https://bugs.chromium.org/p/chromium/issues/detail?id=716231), [Windows](https://bugs.chromium.org/p/chromium/issues/detail?id=613414) and [Linux](https://bugs.chromium.org/p/chromium/issues/detail?id=640256).

Related: #550.

ka3de:
Closes #771.


chromium/browser_type.go
```diff
@@ -282,7 +282,6 @@ func prepareFlags(lopts *common.LaunchOptions, k6opts *k6lib.Options) (map[strin
 		"disable-background-timer-throttling":                true,
 		"disable-backgrounding-occluded-windows":             true,
 		"disable-breakpad":                                   true,
-		"disable-client-side-phishing-detection":             true,
```
ankur22, line 285: Maybe we need an issue to track `safebrowsing-enable-enhanced-protection`?

ankur22 (approved):
Thanks for tidying these up. Have you compared against PW? Not they should be the source of truth, but might be worth taking a quick look.

chromium/browser_type.go
ka3de: Absolutely. Was on my TODO list :) I will create an issue for it and relate it with #771.

ka3de:
> Thanks for tidying these up. Have you compared against PW? Not they should be the source of truth, but might be worth taking a quick look.

Actually I have not, but I think it's a good idea, I will do and update if there is anything that stands out in comparison.

ka3de:
> > Thanks for tidying these up. Have you compared against PW? Not they should be the source of truth, but might be worth taking a quick look.
> 
> Actually I have not, but I think it's a good idea, I will do and update if there is anything that stands out in comparison.

Coming back to this, there are a few differences on the flags used, I think we can modify #773 and expand its scope to consider all the differences between PW and k6 browser. What do you think @ankur22 ?

inancgumus (approved):

PR #776, opened 2023-02-10
Fix test for preset flags for chrome

ankur22:


chromium/browser_type_test.go
```diff
@@ -111,17 +109,6 @@ func TestBrowserTypePrepareFlags(t *testing.T) {
 			changeK6Opts:  &k6lib.Options{},
 			expChangedVal: nil,
 		},
-		{
-			flag:       "enable-use-zoom-for-dsf",
```
ankur22, line 115: Oops, we all missed this 😅

ka3de: Oh tricky case, I was wondering why it was not failing in CI or in my local machine, then a saw the skip when `runtime.GOOS != "darwin"`.

ka3de (approved):
LGTM.

chromium/browser_type_test.go
ankur22: That'll teach me for not running tests 😅

inancgumus (approved):

inancgumus:
@ankur22, since you're not around, I'm merging this as it's a constant pain for me while testing locally.

PR #779, opened 2023-02-15
Clear zombie browser processes on panic

inancgumus:
I'm strongly against exposing global variables as we did in #762 and #755 😒😭 

Our goal was to _fix_ [the regression](https://github.com/grafana/xk6-browser/pull/744#issuecomment-1418901227), preserve the v0.8.0 behavior, and move on with #744 as we [discussed](https://github.com/grafana/xk6-browser/pull/744#issuecomment-1418986802).

However, we ended up going with #762 and #755 instead and fixed another problem: a multi-browser close issue when one of them panics in the same script (_note that nobody has reported such a problem and usage so far_). While the PRs did fix the multi-browser close issue, they introduced global variables 😒

I'm so grateful for your (@ankur22) contributions to that effort! 🙇 I believe we can do better than relying on global variables. That's not the best solution for us in the short term or the long term.

## Solution

Earlier, I proposed a [solution](https://github.com/grafana/xk6-browser/pull/744#issuecomment-1425427019). This one is similar.

After having many problems introducing global variables on a work I've been working on lately [here](https://github.com/grafana/xk6-browser/pull/778), I took this issue upon myself to work on a different solution that doesn't rely on global variables, so we can move on with #744.

It solves the "multi-browser close on panic" issue in a way that's compatible with our current mapping layer and without using global variables in a cleaner way 🎉 Plus, it's even extensible! If you have a context with a `Pids() int` method, you can use `k6ext.Panic` to kill those processes without even knowing about the mapping layer.

## Testing

I ran the following JavaScript test that uses **multiple VUs and browsers** and no zombies are left 🥳 🧟 

```js

export const options = {
  vus: 2,
  iterations: 2,
}

export default function () {
  const browser = chromium.launch({headless: true});

  const browser2 = chromium.launch({headless: true});

  page.goto('https://test.k6.io/flip_coin.php').then(() => {
    return page.goBack()
  }).finally(() => {
    page.close();
    page2.close();
    browser.close();
    browser2.close();
  })
}
```

I also ran Go tests on this solution, and I'm happy to say that I was able to eliminate the need to kill any zombie browser processes in tests 🧟‍♀️🛑 . Even if we do run into issues down the line, we can fix the tests after moving all the panicking code to the mapping layer (which is possible now with this PR). **We should not tackle this edge case now: It's a very low priority.**

## PS

* I'm not saying it's the perfect solution; far from it. But it's better than relying on global variables. We can perfect it in time.
* Here, I've also applied the `Release` [fix](https://github.com/grafana/xk6-browser/pull/762). And I'd be happy to get it from the other PR.
* I'm OK with separating the browser process management into a new package, similar to what we did in #755. That would be a nice addition.

@grafana/k6-browser

### DIFF
--- a/browser/module.go
+++ b/browser/module.go
@@ -2,8 +2,6 @@
 package browser
 
 import (
-	"sync"
-
 	"github.com/dop251/goja"
 
 	"github.com/grafana/xk6-browser/common"
@@ -17,8 +15,7 @@ type (
 	// RootModule is the global module instance that will create module
 	// instances for each VU.
 	RootModule struct {
-		muBrowserProcessIDs sync.RWMutex
-		browserProcessIDs   []int
+		pidRegistry *pidRegistry
 	}
 
 	// JSModule exposes the properties available to the JS script.
@@ -41,7 +38,9 @@ var (
 
 // New returns a pointer to a new RootModule instance.
 func New() *RootModule {
-	return &RootModule{}
+	return &RootModule{
+		pidRegistry: &pidRegistry{},
+	}
 }
 
 // NewModuleInstance implements the k6modules.Module interface to return
@@ -50,32 +49,13 @@ func (m *RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
 	return &ModuleInstance{
 		mod: &JSModule{
 			Chromium: mapBrowserToGoja(moduleVU{
-				VU:   vu,
-				root: m,
+				VU:          vu,
+				pidRegistry: m.pidRegistry,
 			}),
 			Devices: common.GetDevices(),
 			Version: version,
 		},
 	}
-}
-
-// pids returns the launched browser process IDs.
-func (m *RootModule) pids() []int {
-	m.muBrowserProcessIDs.RLock()
-	defer m.muBrowserProcessIDs.RUnlock()
-
-	pids := make([]int, len(m.browserProcessIDs))
-	copy(pids, m.browserProcessIDs)
-
-	return pids
-}
-
-// addPid keeps track of the launched browser process IDs.
-func (m *RootModule) addPid(pid int) {
-	m.muBrowserProcessIDs.Lock()
-	defer m.muBrowserProcessIDs.Unlock()
-
-	m.browserProcessIDs = append(m.browserProcessIDs, pid)
 }
 
 // Exports returns the exports of the JS module so that it can be used in test
--- a/browser/module.go
+++ b/browser/module.go
@@ -2,6 +2,8 @@
 package browser
 
 import (
+	"sync"
+
 	"github.com/dop251/goja"
 
 	"github.com/grafana/xk6-browser/common"
@@ -14,7 +16,10 @@ const version = "0.8.1"
 type (
 	// RootModule is the global module instance that will create module
 	// instances for each VU.
-	RootModule struct{}
+	RootModule struct {
+		muBrowserProcessIDs sync.RWMutex
+		browserProcessIDs   []int
+	}
 
 	// JSModule exposes the properties available to the JS script.
 	JSModule struct {
@@ -41,14 +46,36 @@ func New() *RootModule {
 
 // NewModuleInstance implements the k6modules.Module interface to return
 // a new instance for each VU.
-func (*RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
+func (m *RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
 	return &ModuleInstance{
 		mod: &JSModule{
-			Chromium: mapBrowserToGoja(moduleVU{vu}),
-			Devices:  common.GetDevices(),
-			Version:  version,
+			Chromium: mapBrowserToGoja(moduleVU{
+				VU:   vu,
+				root: m,
+			}),
+			Devices: common.GetDevices(),
+			Version: version,
 		},
 	}
+}
+
+// pids returns the launched browser process IDs.
+func (m *RootModule) pids() []int {
+	m.muBrowserProcessIDs.RLock()
+	defer m.muBrowserProcessIDs.RUnlock()
+
+	pids := make([]int, len(m.browserProcessIDs))
+	copy(pids, m.browserProcessIDs)
+
+	return pids
+}
+
+// addPid keeps track of the launched browser process IDs.
+func (m *RootModule) addPid(pid int) {
+	m.muBrowserProcessIDs.Lock()
+	defer m.muBrowserProcessIDs.Unlock()
+
+	m.browserProcessIDs = append(m.browserProcessIDs, pid)
 }
 
 // Exports returns the exports of the JS module so that it can be used in test
--- a/browser/modulevu.go
+++ b/browser/modulevu.go
@@ -15,7 +15,7 @@ import (
 type moduleVU struct {
 	k6modules.VU
 
-	root *RootModule
+	*pidRegistry
 }
 
 func (vu moduleVU) Context() context.Context {
@@ -26,9 +26,4 @@ func (vu moduleVU) Context() context.Context {
 	// context from the vu that is received from k6 in
 	// NewModuleInstance).
 	return k6ext.WithVU(vu.VU.Context(), vu.VU)
-}
-
-// Pids returns the launched browser process IDs across all VUs.
-func (vu moduleVU) Pids() []int {
-	return vu.root.pids()
 }
--- a/browser/modulevu.go
+++ b/browser/modulevu.go
@@ -10,10 +10,12 @@ import (
 
 // moduleVU carries module specific VU information.
 //
-// Currently, it is used to carry the VU object to the
-// inner objects and promises.
+// Currently, it is used to carry the VU object to the inner objects and
+// promises.
 type moduleVU struct {
 	k6modules.VU
+
+	root *RootModule
 }
 
 func (vu moduleVU) Context() context.Context {
@@ -24,4 +26,9 @@ func (vu moduleVU) Context() context.Context {
 	// context from the vu that is received from k6 in
 	// NewModuleInstance).
 	return k6ext.WithVU(vu.VU.Context(), vu.VU)
+}
+
+// Pids returns the launched browser process IDs across all VUs.
+func (vu moduleVU) Pids() []int {
+	return vu.root.pids()
 }
--- a/browser/modulevu.go
+++ b/browser/modulevu.go
@@ -1,0 +1,27 @@
+package browser
+
+import (
+	"context"
+
+	"github.com/grafana/xk6-browser/k6ext"
+
+	k6modules "go.k6.io/k6/js/modules"
+)
+
+// moduleVU carries module specific VU information.
+//
+// Currently, it is used to carry the VU object to the
+// inner objects and promises.
+type moduleVU struct {
+	k6modules.VU
+}
+
+func (vu moduleVU) Context() context.Context {
+	// promises and inner objects need the VU object to be
+	// able to use k6-core specific functionality.
+	//
+	// We should not cache the context (especially the init
+	// context from the vu that is received from k6 in
+	// NewModuleInstance).
+	return k6ext.WithVU(vu.VU.Context(), vu.VU)
+}
--- a/browser/pidregistry.go
+++ b/browser/pidregistry.go
@@ -1,0 +1,28 @@
+package browser
+
+import "sync"
+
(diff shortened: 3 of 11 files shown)


ka3de:
Hey @inancgumus thanks for taking a deep look on this!
Overall I like the approach, avoiding the usage of global vars, although the approach taken into #755 seemed a little bit more explicit in order to know where the list of pids were maintained.
I have a question about this implementation. Maybe I'm missing something but, what is the reason to store the `pids` in the root module implementation? Could we store them directly in the `moduleVU` struct?

inancgumus:
Hey @ka3de,

> Hey @inancgumus thanks for taking a deep look on this!

Thanks for your review 🙇 

> Overall I like the approach, avoiding the usage of global vars, although the approach taken into #755 seemed a little bit more explicit in order to know where the list of pids were maintained.

We can still have a type where we store pids and put the type in the root module if we fancy that approach. Eventually, the extension wide state can go into the root module. This PR aligns with [the goal we set earlier](https://github.com/grafana/xk6-browser/issues/688#issuecomment-1408543524), it is the whole idea and then it has changed shape:

> The idea is that we will bubble up errors all the way to the mapping layer, where it can decide on whether to panic or to return an error to goja.
> This would mean that we have more control over when to panic, and to be more aware of what we want to panic on, as well as using good GO error handling practices throughout the codebase.

> I have a question about this implementation. Maybe I'm missing something but, what is the reason to store the `pids` in the root module implementation? Could we store them directly in the `moduleVU` struct?

```
k6 register:          each VU run:

                ----> ModuleInstance
               /
RootModule ----
               \
                ----> ModuleInstance
```

`RootModule` is created once by k6-core when you run the `k6` command. On each VU run, a new `moduleVU` is created. If we stored pids in `moduleVU`, in a multi-VU run, only the panicking VU's pids would be cleared.

inancgumus:
> This is a nice way of handling the multi browser pid issue, and it has eliminated the need for the global variable which is good. 👏

Thanks for the review, @ankur22 🙇

Let's not forget our ultimate goal: [To move panicking into the mapping layer](https://github.com/grafana/xk6-browser/issues/688#issuecomment-1408543524). That's why we've created #744 and #755. Soon, no code in the rest of the codebase will panic, and they return errors. Ultimately, everything related to managing PIDs will be driven by the mapping layer. This PR is a good first step in that direction.

> I still prefer the solution presented in #755 as it does a better job of decoupling the process and its lifecycle away from the rest of the code.

I understand where you're coming from, but I don't think that alone justifies the burden of global variables. Why don't we just keep _decoupling_ by moving every `common` type to its own package with its own global variables instead of using fields and types?

#779 decouples without using global variables and puts the responsibility on the mapping layer (our end goal). After moving `k6ext.Panic` into the mapping layer, it'll be fully decoupled. We can do the same in #779, too, by moving the relevant parts to another package. But that would be unnecessary since, in the end, the mapping layer will be responsible for managing PIDs.


> Although there is a global var, it is only accessible within the package it is in.

While the variables may be accessible only within the package, the other code drives their state and behavior.

Another problem is they still remain in memory for the entire lifetime of the code, which can cause issues when working with multiple `RootModule`s (for some odd reason: testing or benchmarking or something else). In contrast, #779 **encapsulates** everything it needs within itself (`RootModule`), making its behavior more predictable and testable.


> I also prefer the approach of overriding the Kill method in the test, instead of adding test code in production.

I don't understand what you mean here, the necessity and its priority.

You can still have a similar `Kill` method with this solution if you want. And `Kill` won't be necessary soon. as the `common` methods will no longer call `Panic`. Instead, they will return errors. That's been the goal.

inancgumus:
@grafana/k6-browser

We can now solve the issues in a cleaner way that aligns with [our goal](https://github.com/grafana/xk6-browser/issues/688#issuecomment-1408543524). I've added this PR because I strongly believe #755 wouldn't be a good solution for the project. I can close it if you don't find it valuable, though. No hard feelings. Let's see it as an experiment and a learning journey.

ka3de:
> `RootModule` is created once by k6-core when you run the `k6` command. On each VU run, a new `moduleVU` is created. If we stored pids in `moduleVU`, in a multi-VU run, only the panicking VU's pids would be cleared. The beauty of storing pids in the root module is that we can create many `RootModule`s in tests (which we'll do eventually to test things out), and each one will have its own state rather than depending on globals.

Thanks for the explanation! Ok, your last point is an argument in favor of this solution for me. Still, I would prefer to be more explicit, by maybe creating a dedicated register for browser PIDs and passing this one along instead of a reference to the whole root module? I would also like to try to avoid [this hack](https://github.com/grafana/xk6-browser/blob/860b178419ddb8ef502527aa308709d9777def90/k6ext/panic.go#L36-L38). I will review this more carefully.

ankur22 (approved):
Thanks. I'm happy with what the team decides.

inancgumus:
> Still, I would prefer to be more explicit, by maybe creating a dedicated register for browser PIDs and passing this one along instead of a reference to the whole root module?

Yes, that might make sense. We can easily do it by creating a type in the `browser` package like `Process` to store pids in it. WDYT?

> I would also like to try to avoid [this hack](https://github.com/grafana/xk6-browser/blob/860b178419ddb8ef502527aa308709d9777def90/k6ext/panic.go#L36-L38). I will review this more carefully.

It's something we might soon remove after we remove panicking in the `common` code and do that in the mapping layer instead. I've explained that [here](https://github.com/grafana/xk6-browser/pull/779#issuecomment-1437963455).

inancgumus:
@ka3de 👍 

> Still, I would prefer to be more explicit, by maybe creating a dedicated register for browser PIDs and passing this one along instead of a reference to the whole root module?

Yes, that might make sense. We can easily do it by creating a type in the `browser` package like `Process` to store pids in it. WDYT?

> I would also like to try to avoid [this hack](https://github.com/grafana/xk6-browser/blob/860b178419ddb8ef502527aa308709d9777def90/k6ext/panic.go#L36-L38). I will review this more carefully.

It's something we might soon remove after we remove panicking in the `common` code and do that in the mapping layer instead. I've explained that [here](https://github.com/grafana/xk6-browser/pull/779#issuecomment-1437963455).

browser/module.go
```diff
@@ -14,7 +16,10 @@ const version = "0.8.0"
 type (
 	// RootModule is the global module instance that will create module
 	// instances for each VU.
-	RootModule struct{}
+	RootModule struct {
+		muBrowserProcessIDs sync.RWMutex
```
ka3de, line 20: Should this be `browserProcessIDsMu` or `browserPIDsMu` instead? I believe we usually use the `mu` to identify the mutex as suffix instead of prefix.

browser/module.go
```diff
@@ -41,16 +46,38 @@ func New() *RootModule {
...
+	defer m.muBrowserProcessIDs.RUnlock()
+
+	pids := make([]int, len(m.browserProcessIDs))
+	copy(pids, m.browserProcessIDs)
+
+	return pids
+}
+
+// addPid keeps track of the launched browser process IDs.
+func (m *RootModule) addPid(pid int) {
```
ka3de, line 74: Should this be `addPID` ?

browser/modulevu.go
```diff
@@ -0,0 +1,34 @@
...
+	// able to use k6-core specific functionality.
+	//
+	// We should not cache the context (especially the init
+	// context from the vu that is received from k6 in
+	// NewModuleInstance).
+	return k6ext.WithVU(vu.VU.Context(), vu.VU)
+}
+
+// Pids returns the launched browser process IDs across all VUs.
+func (vu moduleVU) Pids() []int {
```
ka3de, line 32: Should this be `PIDs`?

k6ext/panic.go
```diff
@@ -33,21 +33,23 @@ func Panic(ctx context.Context, format string, a ...any) {
...
 
-	pid := GetProcessID(ctx)
-	if pid == 0 {
-		// this should never happen unless a programmer error
-		panic("no browser process ID in context")
-	}
-	p, err := os.FindProcess(pid)
-	if err != nil {
-		// optimistically return and don't kill the process
+	pidder, ok := GetVU(ctx).(interface {
```
ka3de, line 36: I was trying to improve this hack, but as you said @inancgumus I think the best way to fix it is in the long run, by moving panics only to the mapping layer were we would have access directly to the `moduleVU` which maintains the `pid` list and therefore will be able to avoid this casting.

ka3de (commented):
Made a few comments, no blockers. Also, in regards of the discussion of keeping the `PIDs` in a more explicit manner. Should we consider defining a struct such as:
```go
// PIDRegister registers the process ID for all browser instances.
type PIDRegister struct {
	pidsMu sync.RWMutex
	pids   []int
}

// PIDs returns the launched browser process IDs.
func (r *PIDRegister) PIDs() []int {
	r.pidsMu.RLock()
	defer r.pidsMu.RUnlock()

	pids := make([]int, len(r.pids))
	copy(pids, r.pids)

	return pids
}

// AddPid keeps track of the launched browser process IDs.
func (r *PIDRegister) AddPid(pid int) {
	r.pidsMu.Lock()
	defer r.pidsMu.Unlock()

	r.pids = append(r.pids, pid)
}
```
So then it's used in root module:
```go
type RootModule struct {
    register *PIDRegister
}
```
And in `moduleVU` without having to pass the reference to the root module:
```go
type moduleVU struct {
	k6modules.VU

	register *Register
}
```
?

browser/module.go
inancgumus: Both are valid conventions: prefixed or suffixed. Sure, we can call it `browserPidsMu` (Pid is a convention in Go—stdlib). WDYT?

browser/module.go
inancgumus: Pid is a valid convention in Go. See [this](https://pkg.go.dev/os#Process) for an example.

```go
type Process struct {
	Pid [int](https://pkg.go.dev/builtin#int)
	// contains filtered or unexported fields
}
```

browser/module.go
ka3de: Got it, intuitively I would have thought that by other conventions should be `PID`, but you are right `Pid` is used in stdlib, so LGTM.

inancgumus:
@ka3de, thanks for the detailed example 🙇 

Yes, that makes sense. Should we call it `State`? So we can put other variables in the `State` type related that stores data about the whole extension.

k6ext/panic.go
inancgumus: Exactly! Don't worry about it. I'll put a `TODO` comment there so we won't forget.

ka3de:
> Yes, that makes sense. Should we call it `State` to make it generic? So we can put other related variables in the `State` type that stores data about the whole extension.

Not sure about `state`, because the way I see it, in this case the state is more the whole `moduleVU` struct, so maybe I would be specific about the purpose in relation with maintaining the PIDs list. But not strong opinion :+1:

inancgumus:
> Not sure about `state`, because the way I see it, in this case the state is more the whole `moduleVU` struct, so maybe I would be specific about the purpose in relation with maintaining the PIDs list. But no strong opinion 👍

No worries. We can start with the `PidRegistry` type. WDYT?

inancgumus:
> Not sure about `state`, because the way I see it, in this case the state is more the whole `moduleVU` struct, so maybe I would be specific about the purpose in relation with maintaining the PIDs list. But no strong opinion 👍

@ka3de, no worries. We can start with the `PidRegistry` type. WDYT?

inancgumus:
@ka3de Please check [the last commit](https://github.com/grafana/xk6-browser/pull/779/commits/b3a98ddeb7928180fcbb366c7123ba8269696292) 🙇

browser/pidregistry.go
```diff
@@ -0,0 +1,28 @@
...
+
+// registerPid registers the launched browser process ID.
+func (r *pidRegistry) registerPid(pid int) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	r.ids = append(r.ids, pid)
+}
+
+// PIDs returns the launched browser process IDs.
```
ka3de, line 19: s/PIDS/Pids ?

ka3de (approved):
LGTM. :rocket:

browser/pidregistry.go
inancgumus: Yeah, I've noticed it now 😅 Thanks!

inancgumus: Resolved in https://github.com/grafana/xk6-browser/pull/779/commits/cedf4200402d79a514ca3040bea6241e93bd97d0.

PR #782, opened 2023-02-16
Refactor/mapping to first letter lower

inancgumus:
This is getting hairy so a map makes sense to me.

### DIFF
--- a/browser/mapping_test.go
+++ b/browser/mapping_test.go
@@ -172,16 +172,19 @@ func toFirstLetterLower(s string) string {
 	// Special cases.
 	// Instead of loading up an acronyms list, just do this.
 	// Good enough for our purposes.
-	switch s {
-	default:
-		return strings.ToLower(s[:1]) + s[1:]
-	case "URL":
-		return "url"
-	case "JSON":
-		return "json"
-	case "JSONValue":
-		return "jsonValue"
+	special := map[string]string{
+		"JSON":      "json",
+		"JSONValue": "jsonValue",
+		"URL":       "url",
 	}
+	if v, ok := special[s]; ok {
+		return v
+	}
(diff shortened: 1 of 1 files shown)


ka3de:
I don't particularly see the need for this change, but don't have strong opinion against it. Do you think it makes it more readable with a map? From the performance perspective, the switch implementation is better, but the impact is negligible, especially being in a test, so happy with any version.

inancgumus:
@ka3de Yes, I think it's more readable and manageable than a bunch of if statements. It's a small change and I don't have a strong opinion either, can revert it if you guys don't like it, no worries.

browser/mapping_test.go
```diff
@@ -172,16 +172,15 @@ func toFirstLetterLower(s string) string {
...
-		return "jsonValue"
+	special := map[string]string{
+		"JSON":      "json",
+		"JSONValue": "jsonValue",
+		"URL":       "url",
 	}
+	if v, ok := special[s]; ok {
+		return v
+	}
+	return strings.ToLower(s[:1]) + s[1:]
```
ankur22, line 183: Isn't there a chance that `s` will be of length `0`?

andrewslotin:
While the performance-wise maps and `switch` statements are pretty much equal, IIRC a method that contains `switch` only is inlined. Is this the same for map?

ankur22 (approved):

inancgumus:
Hey, @andrewslotin,

> While the performance-wise maps and `switch` statements are pretty much equal, IIRC a method that contains `switch` only is inlined. Is this the same for map?

Yes, it's the same as long as the function's complexity is low.

ka3de (approved):
LGTM.

PR #783, opened 2023-02-16
Fix unmap internal methods

inancgumus:
We shouldn't expose some of the methods as they are internal to the extension, have no meaning for the users, and Playwright doesn't support them.
    
Refactoring to the tests should be done with the fix together so as not to break the tests. Some parts could be in another commit, but the commit itself is short enough anyway.

Updates: #784


ka3de (approved):
As this is not documented in our API spec I think it makes sense and should be safe.
LGTM.

ankur22 (approved):
LGTM

PR #796, opened 2023-02-27
Return an error from BrowserContext.SetExtraHTTPHeaders and handle panic in mapping layer

ankur22:
At the moment the codebase uses panic too liberally and we don't have control over when we actually want to abort all test runs vs when we only want to fail the current test iteration.

The idea is that we will bubble up errors all the way to the mapping layer, where it can decide on whether to panic or to return an error to goja.

This would mean that we have more control over when to panic, and to be more aware of what we want to panic on, as well as using good GO error handling practices throughout the codebase.

This is the first PR that demonstrates how we can catch internal errors. There is another PR where we had most of the discussion and several other issues were found and resolved ([#744](https://github.com/grafana/xk6-browser/pull/744)). #744 was abandoned in favour of a new PR as the previous issue would require a lot of rework.

The script i've used to test this is:

```javascript

  const browser = chromium.launch({headless: false})
  context.setExtraHTTPHeaders()

  try {
    await page.goto('https://test.k6.io/flip_coin.php')
  } finally {
    page.close()
    browser.close()
  }
}
```

The output of this when ran with this should be:

```bash
panic: GoError: BrowserContext.setExtraHTTPHeaders(headers) has not been implemented yet: internal error
```

### DIFF
--- a/api/browser_context.go
+++ b/api/browser_context.go
@@ -22,7 +22,7 @@ type BrowserContext interface {
 	Route(url goja.Value, handler goja.Callable)
 	SetDefaultNavigationTimeout(timeout int64)
 	SetDefaultTimeout(timeout int64)
-	SetExtraHTTPHeaders(headers map[string]string)
+	SetExtraHTTPHeaders(headers map[string]string) error
 	SetGeolocation(geolocation goja.Value)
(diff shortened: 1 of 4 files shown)


inancgumus (approved):
LGTM 👍

ka3de (approved):
LGTM.

PR #797, opened 2023-02-27
Fix multi browser close vu was missing in Context

inancgumus:
Fixes a tiny but important regression in #779.

We should put the VU with `pidRegistry` into the `Context` so the panicking goroutines can grab it.


ankur22 (approved):

ka3de (approved):
LGTM.

PR #799, opened 2023-02-27
Add a testing functionality to detect redundant mappings

inancgumus:
This test will fail if a method is mapped and not exists on the API surface (api/). The test excludes custom mappings.

Realized it we were missing this when I was working on a bug.


ankur22 (approved):

ka3de (approved):
LGTM.

PR #800, opened 2023-02-27
Implement `BrowserType.connect`

### LINKED ISSUE #17: Implement `BrowserType.connect(wsEndpoint[, options])`
Add support for `BrowserType.connect(wsEndpoint[, options])`, to connect to an existing browser instance at specified CDP WS endpoint.

**Relevant links**:
- Playwright docs: https://playwright.dev/docs/api/class-browsertype#browser-type-connect
- xk6-browser code starting point: https://github.com/grafana/xk6-browser/blob/main/chromium/browser_type.go#L56
- Playwright code: https://github.com/microsoft/playwright/blob/master/packages/playwright-core/src/server/chromium/crBrowser.ts#L50
- Playwright tests: https://github.com/microsoft/playwright/blob/master/tests/browsertype-connect.spec.ts

ka3de:
This PR implements the `BrowserType.Connect` method for chromium browser type in order to, instead of launching a browser from our implementation, establish the WS connection to a remote running browser.

The approach taken has been to minimize changes in the current implementation/flow. For that:
- A new interface `BrowserProcessMetadata` has been introduced in order to abstract the elements of the current `BrowserProcess` implementation that differ for a local/remote browser process **[*]**.
- Refactor has been made in order to minimize the interactions with internal `BrowserProcess` properties from other components (mainly from `Browser` implementation).
- A new constructor has been added to comply with the differences for a new `BrowserProcess` struct that references a remote browser.
- Implementation of `BrowserType.Connect` method following behavior defined by the current `launch` method.
- Refactor in order to extract common behavior for `launch` and `connect` methods.

**[*]** There were alternatives other than creating a new interface to abstract these differences, for example to add a `flag` in `BrowserProcess` that would indicate if the `os.Process` property held by `BrowserProcess` was valid or not (due to being remote). But I believe the new interface definition abstracts this in a better way and prevents possible panics due to `NPE`, especially considering that there is not an abstraction in place for the whole `BrowserProcess` element, and this is reachable from other elements implementations, such as `Browser`.

Closes: grafana/xk6-browser#17.

### DIFF
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -126,46 +126,6 @@ func (b *BrowserType) link(
 	return p, nil
 }
 
-// ExecutablePath returns the path where the extension expects to find the browser executable.
-func (b *BrowserType) ExecutablePath() (execPath string) {
-	if b.execPath != "" {
-		return b.execPath
-	}
-	defer func() {
-		b.execPath = execPath
-	}()
-
-	for _, path := range [...]string{
-		// Unix-like
-		"headless_shell",
-		"headless-shell",
-		"chromium",
-		"chromium-browser",
-		"google-chrome",
-		"google-chrome-stable",
-		"google-chrome-beta",
-		"google-chrome-unstable",
-		"/usr/bin/google-chrome",
-
-		// Windows
-		"chrome",
-		"chrome.exe", // in case PATHEXT is misconfigured
-		`C:\Program Files (x86)\Google\Chrome\Application\chrome.exe`,
-		`C:\Program Files\Google\Chrome\Application\chrome.exe`,
-		filepath.Join(os.Getenv("USERPROFILE"), `AppData\Local\Google\Chrome\Application\chrome.exe`),
-
-		// Mac (from https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Mac/857950/)
-		"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
-		"/Applications/Chromium.app/Contents/MacOS/Chromium",
-	} {
-		if _, err := exec.LookPath(path); err == nil {
-			return path
-		}
-	}
-
-	return ""
-}
-
 func (b *BrowserType) initContext() context.Context {
 	ctx := k6ext.WithVU(b.vu.Context(), b.vu)
 	ctx = k6ext.WithCustomMetrics(ctx, b.k6Metrics)
@@ -294,6 +254,46 @@ func (b *BrowserType) allocate(
 	}
 
 	return common.NewLocalBrowserProcess(bProcCtx, path, args, env, dataDir, bProcCtxCancel, logger) //nolint: wrapcheck
+}
+
+// ExecutablePath returns the path where the extension expects to find the browser executable.
+func (b *BrowserType) ExecutablePath() (execPath string) {
+	if b.execPath != "" {
+		return b.execPath
+	}
+	defer func() {
+		b.execPath = execPath
+	}()
+
+	for _, path := range [...]string{
+		// Unix-like
+		"headless_shell",
+		"headless-shell",
+		"chromium",
+		"chromium-browser",
+		"google-chrome",
+		"google-chrome-stable",
+		"google-chrome-beta",
+		"google-chrome-unstable",
+		"/usr/bin/google-chrome",
+
+		// Windows
+		"chrome",
+		"chrome.exe", // in case PATHEXT is misconfigured
+		`C:\Program Files (x86)\Google\Chrome\Application\chrome.exe`,
+		`C:\Program Files\Google\Chrome\Application\chrome.exe`,
+		filepath.Join(os.Getenv("USERPROFILE"), `AppData\Local\Google\Chrome\Application\chrome.exe`),
+
+		// Mac (from https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Mac/857950/)
+		"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
+		"/Applications/Chromium.app/Contents/MacOS/Chromium",
+	} {
+		if _, err := exec.LookPath(path); err == nil {
+			return path
+		}
+	}
+
+	return ""
 }
 
 // parseArgs parses command-line arguments and returns them.
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -58,20 +58,46 @@ func NewBrowserType(vu k6modules.VU) api.BrowserType {
 	return &b
 }
 
-// Connect attaches k6 browser to an existing browser instance.
-func (b *BrowserType) Connect(wsEndpoint string, opts goja.Value) api.Browser {
+func (b *BrowserType) init(
+	opts goja.Value, isRemoteBrowser bool,
+) (context.Context, *common.LaunchOptions, *log.Logger, error) {
 	ctx := b.initContext()
 
-	var err error
-	var logger *log.Logger
-	if logger, err = makeLogger(ctx); err != nil {
-		k6ext.Panic(ctx, "setting up logger: %w", err)
+	logger, err := makeLogger(ctx)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("error setting up logger: %w", err)
 	}
-	launchOpts := common.NewLaunchOptions(k6ext.OnCloud(), true)
-	if err := launchOpts.Parse(ctx, logger, opts); err != nil {
-		k6ext.Panic(ctx, "parsing launch options: %w", err)
+
+	launchOpts := common.NewLaunchOptions(k6ext.OnCloud(), isRemoteBrowser)
+	if err = launchOpts.Parse(ctx, logger, opts); err != nil {
+		return nil, nil, nil, fmt.Errorf("error parsing launch options: %w", err)
 	}
 	ctx = common.WithLaunchOptions(ctx, launchOpts)
+
+	if err := logger.SetCategoryFilter(launchOpts.LogCategoryFilter); err != nil {
+		return nil, nil, nil, fmt.Errorf("error setting category filter: %w", err)
+	}
+	if launchOpts.Debug {
+		_ = logger.SetLevel("debug")
+	}
+
+	return ctx, launchOpts, logger, nil
+}
+
+func (b *BrowserType) initContext() context.Context {
+	ctx := k6ext.WithVU(b.vu.Context(), b.vu)
+	ctx = k6ext.WithCustomMetrics(ctx, b.k6Metrics)
+	ctx = common.WithHooks(ctx, b.hooks)
+	ctx = common.WithIterationID(ctx, fmt.Sprintf("%x", b.randSrc.Uint64()))
+	return ctx
+}
+
+// Connect attaches k6 browser to an existing browser instance.
+func (b *BrowserType) Connect(wsEndpoint string, opts goja.Value) api.Browser {
+	ctx, launchOpts, logger, err := b.init(opts, true)
+	if err != nil {
+		k6ext.Panic(ctx, "initializing browser type: %w", err)
+	}
 
 	bp, err := b.connect(ctx, wsEndpoint, launchOpts, logger)
 	if err != nil {
@@ -85,14 +111,9 @@ func (b *BrowserType) Connect(wsEndpoint string, opts goja.Value) api.Browser {
 	return bp
 }
 
-func (b *BrowserType) connect(ctx context.Context, wsURL string, opts *common.LaunchOptions, logger *log.Logger) (*common.Browser, error) {
-	if err := logger.SetCategoryFilter(opts.LogCategoryFilter); err != nil {
-		return nil, fmt.Errorf("%w", err)
-	}
-	if opts.Debug {
-		_ = logger.SetLevel("debug")
-	}
-
+func (b *BrowserType) connect(
+	ctx context.Context, wsURL string, opts *common.LaunchOptions, logger *log.Logger,
+) (*common.Browser, error) {
 	browserProc, err := b.link(ctx, wsURL, opts, logger)
 	if browserProc == nil {
 		return nil, fmt.Errorf("connecting to browser: %w", err)
@@ -126,29 +147,13 @@ func (b *BrowserType) link(
 	return p, nil
 }
 
-func (b *BrowserType) initContext() context.Context {
-	ctx := k6ext.WithVU(b.vu.Context(), b.vu)
-	ctx = k6ext.WithCustomMetrics(ctx, b.k6Metrics)
-	ctx = common.WithHooks(ctx, b.hooks)
-	ctx = common.WithIterationID(ctx, fmt.Sprintf("%x", b.randSrc.Uint64()))
-	return ctx
-}
-
 // Launch allocates a new Chrome browser process and returns a new api.Browser value,
 // which can be used for controlling the Chrome browser.
 func (b *BrowserType) Launch(opts goja.Value) (_ api.Browser, browserProcessID int) {
-	ctx := b.initContext()
-
-	var err error
-	var logger *log.Logger
-	if logger, err = makeLogger(ctx); err != nil {
-		k6ext.Panic(ctx, "setting up logger: %w", err)
+	ctx, launchOpts, logger, err := b.init(opts, false)
+	if err != nil {
+		k6ext.Panic(ctx, "initializing browser type: %w", err)
 	}
-	launchOpts := common.NewLaunchOptions(k6ext.OnCloud(), false)
-	if err := launchOpts.Parse(ctx, logger, opts); err != nil {
-		k6ext.Panic(ctx, "parsing launch options: %w", err)
-	}
-	ctx = common.WithLaunchOptions(ctx, launchOpts)
 
 	bp, pid, err := b.launch(ctx, launchOpts, logger)
 	if err != nil {
@@ -165,13 +170,6 @@ func (b *BrowserType) Launch(opts goja.Value) (_ api.Browser, browserProcessID i
 func (b *BrowserType) launch(
 	ctx context.Context, opts *common.LaunchOptions, logger *log.Logger,
 ) (_ *common.Browser, pid int, _ error) {
-	if err := logger.SetCategoryFilter(opts.LogCategoryFilter); err != nil {
-		return nil, 0, fmt.Errorf("%w", err)
-	}
-	if opts.Debug {
-		_ = logger.SetLevel("debug")
-	}
-
 	envs := make([]string, 0, len(opts.Env))
 	for k, v := range opts.Env {
 		envs = append(envs, fmt.Sprintf("%s=%s", k, v))
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -68,7 +68,13 @@ func (b *BrowserType) init(
 		return nil, nil, nil, fmt.Errorf("error setting up logger: %w", err)
 	}
 
-	launchOpts := common.NewLaunchOptions(k6ext.OnCloud(), isRemoteBrowser)
+	var launchOpts *common.LaunchOptions
+	if isRemoteBrowser {
+		launchOpts = common.NewRemoteBrowserLaunchOptions()
+	} else {
+		launchOpts = common.NewLaunchOptions()
+	}
+
 	if err = launchOpts.Parse(ctx, logger, opts); err != nil {
 		return nil, nil, nil, fmt.Errorf("error parsing launch options: %w", err)
 	}
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -40,7 +40,6 @@ type BrowserType struct {
 	k6Metrics *k6ext.CustomMetrics
 	execPath  string // path to the Chromium executable
 	randSrc   *rand.Rand
-	logger    *log.Logger
 }
 
 // NewBrowserType registers our custom k6 metrics, creates method mappings on
@@ -120,16 +119,17 @@ func (b *BrowserType) Launch(opts goja.Value) (_ api.Browser, browserProcessID i
 	ctx := b.initContext()
 
 	var err error
-	if b.logger, err = makeLogger(ctx); err != nil {
+	var logger *log.Logger
+	if logger, err = makeLogger(ctx); err != nil {
 		k6ext.Panic(ctx, "setting up logger: %w", err)
 	}
 	launchOpts := common.NewLaunchOptions(k6ext.OnCloud())
-	if err := launchOpts.Parse(ctx, b.logger, opts); err != nil {
+	if err := launchOpts.Parse(ctx, logger, opts); err != nil {
 		k6ext.Panic(ctx, "parsing launch options: %w", err)
 	}
 	ctx = common.WithLaunchOptions(ctx, launchOpts)
 
-	bp, pid, err := b.launch(ctx, launchOpts)
+	bp, pid, err := b.launch(ctx, launchOpts, logger)
 	if err != nil {
 		err = &k6ext.UserFriendlyError{
 			Err:     err,
@@ -142,13 +142,13 @@ func (b *BrowserType) Launch(opts goja.Value) (_ api.Browser, browserProcessID i
 }
 
 func (b *BrowserType) launch(
-	ctx context.Context, opts *common.LaunchOptions,
+	ctx context.Context, opts *common.LaunchOptions, logger *log.Logger,
 ) (_ *common.Browser, pid int, _ error) {
-	if err := b.logger.SetCategoryFilter(opts.LogCategoryFilter); err != nil {
+	if err := logger.SetCategoryFilter(opts.LogCategoryFilter); err != nil {
 		return nil, 0, fmt.Errorf("%w", err)
 	}
 	if opts.Debug {
-		_ = b.logger.SetLevel("debug")
+		_ = logger.SetLevel("debug")
 	}
 
 	envs := make([]string, 0, len(opts.Env))
@@ -168,7 +168,7 @@ func (b *BrowserType) launch(
 	go func(c context.Context) {
 		defer func() {
 			if err := dataDir.Cleanup(); err != nil {
-				b.logger.Errorf("BrowserType:Launch", "cleaning up the user data directory: %v", err)
+				logger.Errorf("BrowserType:Launch", "cleaning up the user data directory: %v", err)
 			}
 		}()
 		// There's a small chance that this might be called
@@ -179,19 +179,17 @@ func (b *BrowserType) launch(
 		<-c.Done()
 	}(ctx)
 
-	browserProc, err := b.allocate(ctx, opts, flags, envs, dataDir, b.logger)
+	browserProc, err := b.allocate(ctx, opts, flags, envs, dataDir, logger)
 	if browserProc == nil {
 		return nil, 0, fmt.Errorf("launching browser: %w", err)
 	}
-
-	browserProc.AttachLogger(b.logger)
 
 	// If this context is cancelled we'll initiate an extension wide
 	// cancellation and shutdown.
 	browserCtx, browserCtxCancel := context.WithCancel(ctx)
 	b.Ctx = browserCtx
 	browser, err := common.NewBrowser(browserCtx, browserCtxCancel,
-		browserProc, opts, b.logger)
+		browserProc, opts, logger)
 	if err != nil {
 		return nil, 0, fmt.Errorf("launching browser: %w", err)
 	}
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -60,9 +60,10 @@ func NewBrowserType(vu k6modules.VU) api.BrowserType {
 }
 
 // Connect attaches k6 browser to an existing browser instance.
-func (b *BrowserType) Connect(opts goja.Value) {
+func (b *BrowserType) Connect(wsEndpoint string, opts goja.Value) api.Browser {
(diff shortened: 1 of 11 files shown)


ka3de:
I have a comment about propagating the `ErrInternal` to the mapping layer, which I was trying to do for `connect` implementation [here](https://github.com/grafana/xk6-browser/blob/46ccdd1d749b5dc3a03956a8cd1bc6455f44c86a/chromium/browser_type.go#L103-L109).
The problem is that we are extensively using the `UserFriendlyError` (sometimes not entirely correctly IMO), which is later on [parsed](https://github.com/grafana/xk6-browser/blob/46ccdd1d749b5dc3a03956a8cd1bc6455f44c86a/k6ext/panic.go#L24-L33) from the `k6ext.Panic` implementation, but only one error can be wrapped with go v1.19, creating a "conflict" between `ErrInternal` to be handled in the [mapping](https://github.com/grafana/xk6-browser/blob/46ccdd1d749b5dc3a03956a8cd1bc6455f44c86a/browser/mapping.go#L677) layer, and `UserFriendlyError` to be handled in the panic implementation.

Should `ErrInternal` replace the `UserFriendlyError` implementation? Should it "absorb" its fields (e.g.: timeout)?
Should we upgrade to go v1.20 which allows to [wrap multiple errors](https://tip.golang.org/doc/go1.20#errors)?

WDYT @inancgumus @ankur22 ?

ankur22:
> Should `ErrInternal` replace the `UserFriendlyError` implementation? Should it "absorb" its fields (e.g.: timeout)?

I've been mulling over what makes an error internal again. Thinking back to how I might design errors in a backend service:

- if a user error (4**) occurred once, it would occur again the next time, so the user needs to change the input.
- If the error was unexpected then the chances are that the action can be retried, and it will work the next time, which are usually internal errors (5**).

So, it feels that we should be thinking about errors as whether they are temporary or permanent.

In xk6-browser:
- I think if a user sees a permanent error we should abort the whole test run, otherwise what's the point in a continuously failing test run where all the iterations are failing?
- If a user sees a temporary error then we should only end the iteration and not the whole test run.

So the question now is, are the `UserFriendlyError`s temporary or not?

There are two actions from my viewpoint:

1. We change `ErrInternal` to `ErrPermanent`;
2. We don't worry about making sure that `Permanent` errors are wrapped in `UserFriendlyError`.

> Should we upgrade to go v1.20 which allows to [wrap multiple errors](https://tip.golang.org/doc/go1.20#errors)?

I think we should get the conversation started and see if others agree.

ankur22:
> Should `ErrInternal` replace the `UserFriendlyError` implementation? Should it "absorb" its fields (e.g.: timeout)?

I've been mulling over what makes an error internal again. Thinking back to how I might design errors in a backend service:

- if a user error (4**) occurred once, it would occur again the next time, so the user needs to change the input.
- If the error was unexpected then the chances are that the action can be retried, and it will work the next time, which are usually internal errors (5**).

So, it feels that we should be thinking about errors as whether they are temporary or permanent.

In xk6-browser:
- I think if a user sees a permanent error we should abort the whole test run, otherwise what's the point in a continuously failing test run where all the iterations are failing?
- If a user sees a temporary error then we should only end the iteration and not the whole test run.

So the question now is, are the `UserFriendlyError`s temporary or not?

There are two actions from my viewpoint:

1. We change `ErrInternal` to `ErrPermanent`;
2. We don't worry about making sure that `Permanent` errors are wrapped in `UserFriendlyError`.

> Should we upgrade to go v1.20 which allows to [wrap multiple errors](https://tip.golang.org/doc/go1.20#errors)?

I think we should get the conversation started and see if others agree.

ka3de:
> - I think if a user sees a permanent error we should abort the whole test run, otherwise what's the point in a continuously failing test run where all the iterations are failing?
> - If a user sees a temporary error then we should only end the iteration and not the whole test run.

I agree on this approach, there can be cases where there is not a clear cut, but agree on the concept.

> So the question now is, are the `UserFriendlyErrors` temporary or not?

I believe right now we are using it interchangeably. Personally I'm not sure the `UserFriendlyError` is bringing us much value. To me it seems like the `UserFriendlyError` was created to be a sort of "catch all" errors, but I don't like the idea of parsing the error in the "panic layer" and building the error message based on its different fields. I think it adds complexity, especially in an implementation where only one error can be wrapped.

Would it make sense to remove this error, and change the implementations to provide context to the errors where these are defined (as you would normally do)? We could then define specific errors for specific use cases, for example a timeout error. Or for example in our case a `Fatal` error, which is meant to end all iterations.

In regards of this PR, do you think it makes sense to take this discussion away from it, move forward to it and then tackle all this error/panic handling in a separate PR, possibly the one related with issue #688 ?

inancgumus:
> In regards of this PR, do you think it makes sense to take this discussion away from it, move forward to it and then tackle all this error/panic handling in a separate PR, possibly the one related with issue #688 ?

@ka3de Agreed. Let's discuss this in #688.

inancgumus (commented):
Hey, @ka3de, thanks for the PR, a long-awaited feature! 👏  I'll review this tomorrow 👍 Before that, I did the following and came up with `panic: GoError: creating a new page: canceled` errors. You might be interested.

```diff
diff --git a/tests/test_browser.go b/tests/test_browser.go
index eef335f..856f51a 100644
--- a/tests/test_browser.go
+++ b/tests/test_browser.go
@@ -118,6 +118,7 @@ func newTestBrowser(tb testing.TB, opts ...any) *testBrowser {
 	if !ok {
 		panic(fmt.Errorf("testBrowser: unexpected browser %T", b))
 	}
+	b = bt.Connect(cb.WsURL(), rt.ToValue(launchOpts))
 
 	tb.Cleanup(func() {
 		select {
```

ka3de:
> Before that, I did the following and came up with panic: GoError: creating a new page: canceled errors. You might be interested.

Thanks for the heads up! I didn't want to convert the `BrowserType` tests into "test everything", but that's a good way to try this without creating a bunch of new tests :+1: will take a look. Also I think the work for #675 should help as also test this better at some point.

inancgumus:
@ka3de, sure! 👍 But I wasn't suggesting testing it like that 😅 I've shared the diff in case we discover some edge cases where `connect` might not work as expected since that's what people will experience.

ka3de:
> But I wasn't suggesting testing it like that and putting it into this PR

Yep, sure, I understood the purpose :smile:

common/browser_process_meta.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"errors"
+	"os"
+
+	"github.com/grafana/xk6-browser/storage"
+)
+
+// ErrUnknownRemoteBrowserProcessPid indicates that the pid for the browser process is unknown
+// as the browser is being executed in a remote machine.
+var ErrUnknownRemoteBrowserProcessPid = errors.New("browser process is remote and pid unknown")
```
inancgumus, line 12: Do we need to export these types as they are internal to the `common` package?

common/browser_process_meta.go
```diff
@@ -0,0 +1,72 @@
...
+
+// BrowserProcessMeta handles the metadata associated with
+// a browser process, especifically, the OS process handle
+// and the associated browser data directory.
+type BrowserProcessMeta interface {
+	Pid() (int, error)
```
inancgumus, line 18: Returning an error from a getter field does not seem to be the right design here. Since this method is primarily needed by `*Browser.Close()` (?). The other usage where we check this error is `*Browser.launch()`, but it doesn't seem necessary since `launch` always returns a PID, and we won't call `launch` when using `Connect()`. It seems we're unnecessarily checking an error in the latter.

Also, only the local browser meta uses `Cleanup()`. Could we instead inject the clean-up as a function value to change its behavior? Even further, do we know if we need a cleanup method on the remote browser?

chromium/browser_type.go
```diff
@@ -195,7 +195,12 @@ func (b *BrowserType) launch(
 		return nil, 0, fmt.Errorf("launching browser: %w", err)
 	}
 
-	return browser, browserProc.Pid(), nil
+	pid, err = browserProc.Pid()
+	if err != nil {
+		return nil, 0, fmt.Errorf("launching browser: %w", err)
+	}
```
inancgumus, line 219: Wouldn't this always return a valid PID?

chromium/browser_type.go
```diff
@@ -123,7 +123,7 @@ func (b *BrowserType) Launch(opts goja.Value) (_ api.Browser, browserProcessID i
 	if logger, err = makeLogger(ctx); err != nil {
 		k6ext.Panic(ctx, "setting up logger: %w", err)
 	}
-	launchOpts := common.NewLaunchOptions(k6ext.OnCloud())
+	launchOpts := common.NewLaunchOptions(k6ext.OnCloud(), false)
```
inancgumus: I suggest making this a constant to make it more readable. Maybe with: `LocalBrowserMode`?

common/browser_test.go
```diff
@@ -24,7 +24,7 @@ func TestBrowserNewPageInContext(t *testing.T) {
 	newTestCase := func(id cdp.BrowserContextID) *testCase {
 		ctx, cancel := context.WithCancel(context.Background())
 		logger := log.NewNullLogger()
-		b := newBrowser(ctx, cancel, nil, NewLaunchOptions(false), logger)
+		b := newBrowser(ctx, cancel, nil, NewLaunchOptions(false, false), logger)
```
inancgumus, line 27: As suggested in a previous comment, I believe making these constants can make it more readable.

tests/test_browser.go
```diff
@@ -187,6 +195,13 @@ func (b *testBrowser) staticURL(path string) string {
 	return b.URL("/" + testBrowserStaticDir + "/" + path)
 }
 
+// WsURL returns the Websocket URL that the test browser is listening on for CDP clients.
+func (b *testBrowser) WsURL() string {
+	b.t.Helper()
```
inancgumus, line 200: Do you think making this a helper necessary? I believe we can access its field in tests instead of via an exported method. WDYT?

ka3de:
> I did the following and came up with panic: GoError: creating a new page: canceled errors. You might be interested.

Hey @inancgumus , so I ran the tests with this "hack" to use the `connect` method in the `test browser`, but I was getting a different error. Basically I was getting a failure in the `TestBrowserLogIterationID` [here](https://github.com/grafana/xk6-browser/blob/7ee3d9248bc5029c07cd5c5f115a83d038b558e6/tests/browser_test.go#L153-L180).

I was taking a look and this is "expected" due to the hack. What is happening is that the "log hook" set up [here](https://github.com/grafana/xk6-browser/blob/7ee3d9248bc5029c07cd5c5f115a83d038b558e6/tests/test_browser.go#L108) for the test browser will end up collecting the logs for the "initial browser" [created](https://github.com/grafana/xk6-browser/blob/7ee3d9248bc5029c07cd5c5f115a83d038b558e6/chromium/browser_type.go#L153) through `launch` method, with an "initial" `iterationID` created [here](https://github.com/grafana/xk6-browser/blob/7ee3d9248bc5029c07cd5c5f115a83d038b558e6/chromium/browser_type.go#L91); but it will also collect logs for the "second browser" that is [created](https://github.com/grafana/xk6-browser/blob/7ee3d9248bc5029c07cd5c5f115a83d038b558e6/chromium/browser_type.go#L97) through the `connect` method, which will create a new `iterationID` here again, and that context will be the one set up, and [override](https://github.com/grafana/xk6-browser/blob/7ee3d9248bc5029c07cd5c5f115a83d038b558e6/chromium/browser_type.go#L125), the previous one in the `BrowserType` here and eventually in the `testBrowser` [here](https://github.com/grafana/xk6-browser/blob/7ee3d9248bc5029c07cd5c5f115a83d038b558e6/tests/test_browser.go#L134).

If you check the logs, there are references for both generated iteration IDs, one for the "initial browser" through `launch` and one for the "second" browser through the `connect`. I don't think this requires a fix, as is due to the "hack", but let me know if I misunderstood something.

inancgumus (commented):
Nice work 👏  Overall, LGTM. I have some suggestions and mostly questions to consider.

inancgumus:
> I don't think this requires a fix, as is due to the "hack", but let me know if I misunderstood something.

👍

common/browser.go
```diff
@@ -424,7 +426,13 @@ func (b *Browser) Close() {
 	select {
 	case <-b.browserProc.processDone:
 	case <-time.After(timeout):
-		b.logger.Debugf("Browser:Close", "killing browser process with PID %d after %s", b.browserProc.Pid(), timeout)
+		pid, err := b.browserProc.Pid()
+		if err != nil {
+			// An error here most probably means that Pid is unknown due to browser
+			// being executed on a remote machine, so just log and continue closing.
+			b.logger.Debugf("Browser:Close", "getting browser pid: %v", err)
```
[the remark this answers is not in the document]
ka3de: Don't we have to still close the connection [here](https://github.com/grafana/xk6-browser/blob/7ee3d9248bc5029c07cd5c5f115a83d038b558e6/common/browser.go#L447), even if the process is remote?

chromium/browser_type.go
```diff
@@ -160,7 +158,7 @@ func (b *BrowserType) launch(
 	if err != nil {
 		return nil, 0, fmt.Errorf("%w", err)
 	}
-	dataDir := b.storage
+	dataDir := &storage.Dir{}
```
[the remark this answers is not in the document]
ka3de: I agree, although the problem is that `allocate` method expects a `*storage.Dir`. That's why it was initialized like this. We can change it and then reference it when calling `allocate` such as `&dataDir`, but I don't think it is would be better than current implementation. WDYT?

common/browser_process_meta.go
ka3de: So you mean all the types defined for `BrowserProcessMeta`? It's not necessary to have them exported, I did it to be kind of consistent with other implementations, for example `BrowserProcess` is also internal to `common` pkg. But I guess in this case it's even more internal, so it probably makes sense to not export this.
I will change it.

common/browser_process_meta.go
ka3de: Agree that returning an error from a getter might not be the most elegant solution, but the idea was to provide an abstraction that would be safe to use for future cases also. So, with the current implementation we could make the changes that you mention, but we take the risk of "misusing" the `Pid` method in the future, assuming that it will always return a valid pid when in some cases, when using connect, it will not.
In any case, I don't have strong opinion, if you think this is unnecessary and that we can "take that risk" and handle it ourselves in future implementations and tests, that's ok! I will change it.

Regarding cleanup, the argument is similar as the previous point, the idea was to provide a safe abstraction that would work minimizing the changes on the current implementation. When you say passing `cleanup` as function, you mean to the browser process constructors?

chromium/browser_type.go
ka3de: Knowing the current implementation behind `allocate` and `NewLocalBrowserProcess` yes, but handling the error seems the proper way if we maintain this API for `Pid`.
Would you ignore the error handling based on the knowledge of the mentioned methods?

chromium/browser_type.go
ka3de: Agree, let me change it!

tests/test_browser.go
ka3de: Agree, this is unnecessary and does not make sense for a direct getter. Changing it :+1:

chromium/browser_type.go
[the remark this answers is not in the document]
inancgumus: `&dataDir` means sharing the value and it'd be fine. Resolving :)

common/browser_process_meta.go
inancgumus: I would take that risk since I don't think it's a big one, but I'd leave the decision to you 👍

chromium/browser_type.go
inancgumus: I would eliminate the interface and never need to check for errors.

common/browser_options.go
```diff
@@ -127,15 +129,42 @@ func (l *LaunchOptions) Parse(ctx context.Context, logger *log.Logger, opts goja
 	return nil
 }
 
-func (l *LaunchOptions) shouldIgnoreOnCloud(opt string) bool {
-	if !l.onCloud {
-		return false
+func (l *LaunchOptions) shouldIgnoreOpt(opt string) bool {
```
ankur22, line 152: I think we only need `shouldIgnoreIfBrowserIsRemote`, those options will also be irrelevant in the "other" environment.

common/browser_options.go
```diff
@@ -56,13 +57,14 @@ type LaunchPersistentContextOptions struct {
 }
 
 // NewLaunchOptions returns a new LaunchOptions.
-func NewLaunchOptions(onCloud bool) *LaunchOptions {
+func NewLaunchOptions(onCloud, isRemoteBrowser bool) *LaunchOptions {
```
ankur22: Could create a new function `NewRemoteBrowserLaunchOptions` where it will set `isRemoteBrowser` to `true`.

chromium/browser_type.go
```diff
@@ -60,9 +60,70 @@ func NewBrowserType(vu k6modules.VU) api.BrowserType {
...
+
+	return browser, nil
+}
+
+func (b *BrowserType) link(
+	ctx context.Context, wsURL string,
+	opts *common.LaunchOptions, logger *log.Logger,
+) (_ *common.BrowserProcess, rerr error) {
+	bProcCtx, bProcCtxCancel := context.WithTimeout(ctx, opts.Timeout)
+	defer func() {
```
ankur22, line 146: I'm not sure i understand why we need to defer this, could we check the error straight after `NewRemoteBrowserProcess`?

tests/test_browser.go
ka3de: Changed, see [here](https://github.com/grafana/xk6-browser/blob/c421358a0be874c99dc930ddddefaaa1ee7e12f5/tests/browser_type_test.go#L18).

chromium/browser_type.go
ka3de: I changed this in a new commit since it involved some refactor. See [c421358](https://github.com/grafana/xk6-browser/pull/800/commits/c421358a0be874c99dc930ddddefaaa1ee7e12f5).

common/browser_test.go
ka3de: See [c421358](https://github.com/grafana/xk6-browser/pull/800/commits/c421358a0be874c99dc930ddddefaaa1ee7e12f5).

common/browser_process_meta.go
ka3de: Changed, see [here](https://github.com/grafana/xk6-browser/blob/feat/17-browsertype-connect/common/browser_process_meta.go).

ankur22 (approved):
Nicely done 🙂 I just have some minor suggestions.

The one refactor suggestion (which might be too late now) is that it could have been a good opportunity to move `browserprocess` into its own package like we tried [here](https://github.com/grafana/xk6-browser/pull/755).

common/browser_process_meta.go
ka3de: It's ok for me! Will change this to NOT return an error, unless @ankur22 disagrees.

ankur22: Yep, let's try without it.

ka3de:
> The one suggestion (which might be too late now) is that it could have been a good opportunity to move browserprocess into its own package like we tried https://github.com/grafana/xk6-browser/pull/755.

I guess this can be done.. But I would prefer to tackle it in a separate PR if we really want to do it :sweat_smile: 
In this sense, I tried to "decouple" a little bit more the `Browser` from the `BrowserProcess`, as there were some references from the first one to internal properties of the second one, but yes, it's not much decoupled.

common/browser_options.go
inancgumus: > The one suggestion (which might be too late now) is that it could have been a good opportunity to move `browserprocess` into its own package like we tried [here](https://github.com/grafana/xk6-browser/pull/755).

@grafana/k6-browser 

I wouldn't do that in this PR as it's unrelated to its purpose. Another PR is welcome if we export only a few fields, unlike before, and develop a nicer solution (no pun intended! :)).

inancgumus:
> Ankur: The one suggestion (which might be too late now) is that it could have been a good opportunity to move `browserprocess` into its own package like we tried [here](https://github.com/grafana/xk6-browser/pull/755).

I wouldn't do that in this PR as it's unrelated to its purpose. Another PR is welcome if we export only a few fields, unlike before, and develop a nicer solution (no pun intended! :)). FYI: @grafana/k6-browser

chromium/browser_type.go
```diff
@@ -68,7 +68,12 @@ func (b *BrowserType) init(
 		return nil, nil, nil, fmt.Errorf("error setting up logger: %w", err)
 	}
 
-	launchOpts := common.NewLaunchOptions(k6ext.OnCloud(), isRemoteBrowser)
+	cloudMode := common.CloudModeOff
```
ankur22, line 71: Why do we need to differentiate the two? Aren't then the same now?

common/browser_process_meta.go
ka3de: Changed. See: [510346a](https://github.com/grafana/xk6-browser/pull/800/commits/510346a0285e638acc116ea3e0d7e0638ffa0fc3) and [f98a35f](https://github.com/grafana/xk6-browser/pull/800/commits/f98a35f4c89c66066b37f2683a519a8383fc2c2b).

inancgumus: @ka3de, thanks 🙇 

> When you say passing cleanup as function, you mean to the browser process constructors?

I mean injecting a cleanup function to the browser process instead of making this an interface.

We have two operating modes: remote and local; and I'm not sure it deserves an interface. Interface=slow and brings more complexity. Concrete=faster, direct, and easy to jump in the code editor. So to wrap up, another idea, `*BrowserProcess.Cleanup()` can decide whether to call the storage's `Cleanup` function without needing an interface.

common/browser_options.go
ka3de: But you say this based on the preliminary proposed solution for running browsers in the cloud right? Should we assume already that the browser is gonna be remote and therefore apply the changes?

chromium/browser_type.go
ka3de: Same comment as in https://github.com/grafana/xk6-browser/pull/800#discussion_r1121560746:

> But you say this based on the preliminary proposed solution for running browsers in the cloud right? Should we assume already that the browser is gonna be remote and therefore apply the changes?

common/browser_options.go
ka3de: Is another option, personally I prefer to keep a single "generic" constructor to which pass the appropriate flags, but no strong opinion. If we merge both flags and create this new constructor we can get rid of the constants defined in [e24732b](https://github.com/grafana/xk6-browser/pull/800/commits/e24732bd0a56656bc7ea885d305847c83a640f81). But again, I prefer to keep a single constructor.

common/browser_options.go
ankur22: Also based on what options have been ignored in the remote setup, should also be ignored in the cloud, no?

chromium/browser_type.go
ka3de: I kept this to mimic the way `launch` [implements](https://github.com/grafana/xk6-browser/blob/e24732bd0a56656bc7ea885d305847c83a640f81/chromium/browser_type.go#L242-L247) it. I guess this was done like that in order to not be able to handle explicitly the error returned from `NewBrowserProcess`, and just return directly with the call to the method. So, no strong opinion here, I thing the defer makes it a little bit more complicated IMO. Should we change both implementations?

common/browser_options.go
inancgumus: +1 for adding another constructor: Clearer, more readable, and we can eliminate the constants.

ankur22: I prefer a new constructor, but happy to leave it as is.

common/browser_options.go
inancgumus: Remote/cloud is the same thing, at least for now. If we want to let users connect to remote browsers, that's a different story, and then we would need to separate cloud/remote run meanings. I remember a few users asked for this feature once.

chromium/browser_type.go
ankur22: Yes, i agree, the defer makes it slightly more complicated. I would prefer a simpler approach.

inancgumus: The `defer` ensures that we cancel the context, and I'm on the side of not removing it and leaving it as it is.

ka3de: Do you prefer this:
```go
bProcCtx, bProcCtxCancel := context.WithTimeout(ctx, opts.Timeout)
p, err := common.NewRemoteBrowserProcess(bProcCtx, wsURL, bProcCtxCancel, logger)
if err != nil {
    bProcCtxCancel()
}

return p, err //nolint: wrapcheck
```
Or this:
```go
bProcCtx, bProcCtxCancel := context.WithTimeout(ctx, opts.Timeout)
p, err := common.NewRemoteBrowserProcess(bProcCtx, wsURL, bProcCtxCancel, logger)
if err != nil {
    bProcCtxCancel()
}

return p, nil
```
I think this last one is probably more idiomatic, but no strong opinions and first one is 1 line less :)

inancgumus: The idiomatic one is the `defer` approach :)

ankur22: @inancgumus I think the defer in the `launch` [implementation](https://github.com/grafana/xk6-browser/blob/e24732bd0a56656bc7ea885d305847c83a640f81/chromium/browser_type.go#L242-L247) could be left as is since there are multiple failure paths, but the defer here makes it a little more complicated. There's only one place where the error could be non-nil, which is after `NewRemoteBrowserProcess` is called, so it would make things simpler if we just checked the error there.

Happy to leave it as though if there's a strong opinion on keeping defer.

common/browser_process_meta.go
ka3de: I explored this option initially (mentioned in the PR description), but I discarded it in favor of an interface because it provides a better abstraction and more safety measures, in this case for the `os.Process` and `dataDir`, which is important considering that we do not have any abstraction for `BrowserProcess`, so any other component can reach the internal `os.Process` property of browser process, which in some situations will be `nil`. So I'm in favor of current implementation for these reasons, but let's discuss.

chromium/browser_type.go
ka3de: > The idiomatic one is the `defer` approach :)

Really? I find it quite complicated to grasp on first sight :sweat_smile:

common/browser_options.go
ankur22: I would also be in favour of removing cloud and keeping remote, as we should refrain from using `cloud`.

chromium/browser_type.go
ka3de: Resolving this for clarity as it's the same discussion as the one held in https://github.com/grafana/xk6-browser/pull/800#discussion_r1121487396.

chromium/browser_type.go
inancgumus: @ankur22, ah, right, `defer` is unnecessary there 😅

common/browser_process_meta.go
inancgumus: Alright, resolving if everyone is OK. @ankur22 @ka3de

chromium/browser_type.go
ka3de: Done, see [here](https://github.com/grafana/xk6-browser/blob/e1ca87ca0963b2528aaf4d8f1c01e0fcfc39031a/chromium/browser_type.go#L141-L153).

inancgumus (approved):
LGTM, and again, nice work 👏

common/browser_options.go
ka3de: I'm changing this, but the code in `BrowserType` is a little bit "uglier" when calling to `init` method due to the lack of constants, as we have to pass `true` or `false` values. Another alternative is to create the `launchOptions` in `connect` and `launch` methods and then provide that to `init` method so its modified in it. But don't really like it. Will push the first approach and we can discuss.

ka3de: See: [4d5a38f](https://github.com/grafana/xk6-browser/pull/800/commits/4d5a38fdaba893610c859a8d948318401f3cbeb4)

inancgumus: If we want to go this way, I suggest moving option parsing to `Connect` and `Launch` and passing the parsed launch options to `init()`. This way:
1. We can eliminate the ugly _local/remote_ bool.
2. More importantly. We'll move Goja stuff to the mapping layer (#271), and I wouldn't bury it in `init()`. Let's leave it at the exported methods while changing this part of the code.

inancgumus:
@ka3de Do you think [this](https://github.com/grafana/xk6-browser/blob/08d1d41189e7aa7b839b40916bdcd12b3f352618/common/browser.go#L417-L421) becomes more readable and easy to follow if we align it to the left? I haven't tried the following yet.

```go
	if !b.launchOpts.isRemoteBrowser {
		err := cdpbrowser.Close().Do(cdp.WithExecutor(b.ctx, b.conn))
		if werr := new(websocket.CloseError); !errors.As(err, &werr) {
			k6ext.Panic(b.ctx, "closing the browser: %v", err)
		}
	}
```

ka3de:
> @ka3de Do you think [this](https://github.com/grafana/xk6-browser/blob/08d1d41189e7aa7b839b40916bdcd12b3f352618/common/browser.go#L417-L421) becomes more readable and easy to follow if we align it to the left? I haven't tried the following yet and I leave the exact implementation to you.

That looks good, changed in [554c64f](https://github.com/grafana/xk6-browser/pull/800/commits/554c64f7c3ab0b00bdacf348f34997d760d1f78d).

Issue #687, opened 2023-01-03
Display a friendly error when a browser is re-closed


sniku:
### Brief summary

The browser panics and dies when chrome is closed by the user by mistake too early.

```
ERRO[0003] panic: close of closed channel
goroutine 70 [running]: VUs, 0 complete and 0 interrupted iterations
runtime/debug.Stack()----------------------------] 1 VUs  00m02.6s/10m0s  0/1 iters, 1 per VU
	runtime/debug/stack.go:24 +0x65
go.k6.io/k6/js/common.RunWithPanicCatching.func1()
	go.k6.io/k6@v0.42.1-0.20221223154116-08609b50de8f/js/common/util.go:82 +0x191
panic({0x144cc80, 0x192fc00})
	runtime/panic.go:884 +0x212
github.com/dop251/goja.(*Runtime).runWrapped.func1()
	github.com/dop251/goja@v0.0.0-20221222152515-69dc13e01785/runtime.go:2445 +0xdc
panic({0x144cc80, 0x192fc00})
	runtime/panic.go:884 +0x212
github.com/dop251/goja.(*vm).handleThrow(0xc00261e240, {0x144cc80, 0x192fc00})
	github.com/dop251/goja@v0.0.0-20221222152515-69dc13e01785/vm.go:725 +0x418
github.com/dop251/goja.(*vm).try.func1()
	github.com/dop251/goja@v0.0.0-20221222152515-69dc13e01785/vm.go:744 +0x45
panic({0x144cc80, 0x192fc00})
	runtime/panic.go:884 +0x212
github.com/dop251/goja.(*vm).handleThrow(0xc00261e240, {0x144cc80, 0x192fc00})
	github.com/dop251/goja@v0.0.0-20221222152515-69dc13e01785/vm.go:725 +0x418
github.com/dop251/goja.(*vm).runTryInner.func1()
	github.com/dop251/goja@v0.0.0-20221222152515-69dc13e01785/vm.go:767 +0x45
panic({0x144cc80, 0x192fc00})
	runtime/panic.go:884 +0x212
github.com/grafana/xk6-browser/common.(*Connection).IgnoreIOErrors(0x0?)
	github.com/grafana/xk6-browser@v0.7.0/common/connection.go:550 +0x25
github.com/grafana/xk6-browser/common.(*Browser).Close(0xc0026690e0)
	github.com/grafana/xk6-browser@v0.7.0/common/browser.go:415 +0xea
[...]

PR #803, opened 2023-03-02
Add browser close friendly err (#687)

inancgumus:
This PR does the following:
* Allows executing `Browser.Close` only once.
* Warns users with "a friendly message" when it's called multiple times.

Notes:
* `Close()` is thread-safe as only the same VU or testBrowser use it.

You can try the result with this script:

```js

    await page
        .goto('https://test.k6.io/', { waitUntil: 'networkidle' })
        .finally(() => {
            page.close();
            browser.close();
        });

    await page
        .goto('https://test.k6.io/contacts.php', { waitUntil: 'networkidle' })
        .finally(() => {
            page.close();
            browser.close();
        });
}
```

It should print:

```
WARN[0002] Please call browser.close only once, and do not use the browser after calling close.
  category="Browser:Close" elapsed="0 ms" goroutine=84
```

### DIFF
--- a/common/browser.go
+++ b/common/browser.go
@@ -62,6 +62,9 @@ type Browser struct {
 	sessionIDtoTargetIDMu sync.RWMutex
 	sessionIDtoTargetID   map[target.SessionID]target.ID
 
+	// Used to display a warning when the browser is reclosed.
+	closed bool
+
 	vu k6modules.VU
 
 	logger *log.Logger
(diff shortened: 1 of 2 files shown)


common/browser.go
```diff
@@ -62,6 +62,9 @@ type Browser struct {
 	sessionIDtoTargetIDMu sync.RWMutex
 	sessionIDtoTargetID   map[target.SessionID]target.ID
 
+	// Used to display a warning when the browser is reclosed.
+	closed bool
```
inancgumus, line 66: I would use `atomic.Bool` and then I changed my mind as a single VU enters here.

common/browser.go
```diff
@@ -391,6 +394,15 @@ func (b *Browser) newPageInContext(id cdp.BrowserContextID) (*Page, error) {
 
 // Close shuts down the browser.
 func (b *Browser) Close() {
+	if b.closed {
+		b.logger.Warnf(
+			"Browser:Close",
+			"Please call browser.close only once, and do not use continue using the browser after calling close.",
```
ka3de, line 400: There is a typo in this message right?
I guess it should be:
```
Please call browser.close only once, and do not use the browser after calling close.
```
Maybe even omitting the last part of `calling close` or changing it for `after it`? I'm not native English speaker, so just suggesting :smile:

inancgumus: @ka3de Rephrased like a native speaker here: https://github.com/grafana/xk6-browser/pull/803/commits/250b1e79838c9d636b5e8061ef35b180b1809ca9 😅

ka3de (approved):
Thanks for working on these optimizations!
LGTM.

ankur22 (approved):
LGTM

PR #810, opened 2023-03-06
Rename ErrInternal to ErrFatal and use abort

ankur22:
`Fatal` better describes the error, and it will help distinguish from other internal errors (which are generally temporary). `Fatal` is a permanent error, which should end the whole test run not just the current iteration. If we ran the same iteration again then there's a 100% chance that we will run into the same error, e.g. trying to work with an unimplemented API is a `Fatal` error.

Here are two test scripts I used to test this change:

When ran with `xk6 run -i 10 examples/test.js` before the change, this test will iterate 10 times, failing each iteration. After the change it will stop on the first iteration.
```js

  const browser = chromium.launch({headless: false})
  context.cookies()
}
```

This test will stop the whole test run before the change and after the change.
```js

  const browser = chromium.launch({headless: false})
  context.setExtraHTTPHeaders()
}
```

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -596,7 +596,7 @@ func mapBrowserContext(vu moduleVU, bc api.BrowserContext) mapping {
 			ctx := vu.Context()
 			return k6ext.Promise(ctx, func() (result any, reason error) {
 				err := bc.SetExtraHTTPHeaders(headers)
-				panicIfInternalError(ctx, err)
+				panicIfFatalError(ctx, err)
 				return nil, err //nolint:wrapcheck
 			})
 		},
@@ -678,8 +678,8 @@ func mapBrowserType(vu moduleVU, bt api.BrowserType) mapping {
 	}
 }
 
-func panicIfInternalError(ctx context.Context, err error) {
-	if errors.Is(err, k6error.ErrInternal) {
+func panicIfFatalError(ctx context.Context, err error) {
+	if errors.Is(err, k6error.ErrFatal) {
 		k6ext.Panic(ctx, err.Error())
 	}
 }
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -680,6 +680,6 @@ func mapBrowserType(vu moduleVU, bt api.BrowserType) mapping {
 
 func panicIfFatalError(ctx context.Context, err error) {
 	if errors.Is(err, k6error.ErrFatal) {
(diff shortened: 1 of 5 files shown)


browser/mapping.go
```diff
@@ -678,8 +683,8 @@ func mapBrowserType(vu moduleVU, bt api.BrowserType) mapping {
 	}
 }
 
-func panicIfInternalError(ctx context.Context, err error) {
-	if errors.Is(err, k6error.ErrInternal) {
-		k6ext.Panic(ctx, err.Error())
+func panicIfFatalError(ctx context.Context, err error) {
+	if errors.Is(err, k6error.ErrFatal) {
+		go k6ext.Panic(ctx, err.Error())
```
ka3de, line 688: Could you clarify why do we need to run this in a separate goroutine? I didn't quite understand it from the commit message.
In any case, should we add a comment here explaining the reason?

ka3de (commented):
Thanks for refactoring this after initial discussions @ankur22 ! :+1: 
I added just one question/comment.

browser/mapping.go
ankur22: Good point. Added a fixup with a comment in 350dd34542d57341a2d2754db92270e5efbbfb0e.

ankur22: This isn't relevant since we're now working with [abort](https://github.com/grafana/k6/blob/0b1fe61c5762c83c52a2378f6cdf9c0c3ff1f53d/js/modules/k6/execution/execution.go#L168-L175).

k6ext/panic.go
```diff
@@ -8,14 +8,32 @@ import (
...
+
+// Panic will cause a panic with the given error which will stop
+// the current iteration. Before panicking, it will find the
 // browser process from the context and kill it if it still exists.
 // TODO: test.
 func Panic(ctx context.Context, format string, a ...any) {
+	sharedPanic(ctx, func(rt *goja.Runtime, a ...any) { k6common.Throw(rt, fmt.Errorf(format, a...)) }, a...)
+}
+
+func sharedPanic(ctx context.Context, fail func(rt *goja.Runtime, a ...any), a ...any) {
```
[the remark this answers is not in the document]
ankur22: Yeah, good suggestions. I was wondering how I could make it look a bit clearer, and these suggestions do the trick :thumbsup:

inancgumus (approved):
Nice job 👏

PR #819, opened 2023-03-09
Transition from Browser.Launch to Browser.Connect

### LINKED ISSUE #808: Transition from `Browser.Launch` to `Browser.Connect` when browser WS URL is given by environment
In some circumstances, when the browser lifecycle is handled externally to xk6-browser extension, we should be able to transition from `Browser.Launch` to `Browser.Connect` transparently. This is done to increase compatibility for scripts written to run "locally", where the `Browser.Launch` method will be called and the browser is started by xk6-browser itself; and for cases on which we want to run the same script, maintaining the `launch` method usage and minimizing changes in the script, but in an environment where the browser is handled externally.

The initial idea to detect when to perform this translation is through the existence of an specific environment variable. We should consider therefore the deprecation of the current cloud [environment variable](https://github.com/grafana/xk6-browser/blob/34e1ad07ac294f0a451ac8355aa030bd85a167b6/k6ext/cloud.go#L7).

Additionally and as part of this issue we should consider how will the WS URL be handled by the environment. Will it be defined in the environment variable? Or will the environment variable define an endpoint that will possibly trigger the external procedure to start up the browser and then return its WS URL? Similarly we should consider a hook action when closing the browser connection which could be useful for the environment to handle the actual process closing procedure and cleanup.

ka3de:
This PR modifies the `BrowserType.Launch` mapping function in order to transition to `BrowserType.Connect` if the browser is detected to be remote and the CDP WS URL is given by the environment. This is done to increase compatibility for using `launch` method in the JS API and transparently transition to `connect` if the browser is remotely handled and this set up is indicated by the environment.

Closes #808.

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 
 	"github.com/dop251/goja"
 
@@ -31,11 +32,14 @@ type mapping = map[string]any
 // See issue #661 for more details.
 func mapBrowserToGoja(vu moduleVU) *goja.Object {
 	var (
-		rt          = vu.Runtime()
-		obj         = rt.NewObject()
-		browserType = chromium.NewBrowserType(vu)
+		rt  = vu.Runtime()
+		obj = rt.NewObject()
+		// TODO: Use k6 LookupEnv instead of OS package methods.
+		// See https://github.com/grafana/xk6-browser/issues/822.
+		wsURL, isRemoteBrowser = k6ext.IsRemoteBrowser(os.LookupEnv)
+		browserType            = chromium.NewBrowserType(vu)
 	)
-	for k, v := range mapBrowserType(vu, browserType) {
+	for k, v := range mapBrowserType(vu, browserType, wsURL, isRemoteBrowser) {
 		err := obj.Set(k, rt.ToValue(v))
 		if err != nil {
 			k6common.Throw(rt, fmt.Errorf("mapping: %w", err))
@@ -706,7 +710,7 @@ func mapBrowser(vu moduleVU, b api.Browser) mapping {
 }
 
 // mapBrowserType to the JS module.
-func mapBrowserType(vu moduleVU, bt api.BrowserType) mapping {
+func mapBrowserType(vu moduleVU, bt api.BrowserType, wsURL string, isRemoteBrowser bool) mapping {
 	rt := vu.Runtime()
 	return mapping{
 		"connect": func(wsEndpoint string, opts goja.Value) *goja.Object {
@@ -718,6 +722,14 @@ func mapBrowserType(vu moduleVU, bt api.BrowserType) mapping {
 		"launchPersistentContext": bt.LaunchPersistentContext,
 		"name":                    bt.Name,
 		"launch": func(opts goja.Value) *goja.Object {
+			// If browser is remote, transition from launch
+			// to connect and avoid storing the browser pid
+			// as we have no access to it.
+			if isRemoteBrowser {
+				m := mapBrowser(vu, bt.Connect(wsURL, opts))
+				return rt.ToValue(m).ToObject(rt)
+			}
+
 			b, pid := bt.Launch(opts)
 			// store the pid so we can kill it later on panic.
 			vu.registerPid(pid)
--- a/browser/mapping_test.go
+++ b/browser/mapping_test.go
@@ -117,7 +117,7 @@ func TestMappings(t *testing.T) {
 		"browserType": {
 			apiInterface: (*api.BrowserType)(nil),
 			mapp: func() mapping {
-				return mapBrowserType(moduleVU{VU: vu}, &chromium.BrowserType{})
+				return mapBrowserType(moduleVU{VU: vu}, &chromium.BrowserType{}, "", false)
 			},
 		},
 		"browser": {
--- a/k6ext/env.go
+++ b/k6ext/env.go
@@ -1,0 +1,40 @@
+package k6ext
+
+import (
+	"crypto/rand"
+	"math/big"
+	"strings"
+)
+
+type envLookupper func(key string) (string, bool)
+
+// IsRemoteBrowser returns true and the corresponding CDP
+// WS URL if this one is set through the K6_BROWSER_WS_URL
+// environment variable. Otherwise returns false.
+// If K6_BROWSER_WS_URL is set as a comma separated list of
+// URLs, this method returns a randomly chosen URL from the list
+// so connections are done in a round-robin fashion for all the
+// entries in the list.
+func IsRemoteBrowser(envLookup envLookupper) (wsURL string, isRemote bool) {
+	wsURL, isRemote = envLookup("K6_BROWSER_WS_URL")
+	if !isRemote {
+		return "", false
+	}
+	if !strings.ContainsRune(wsURL, ',') {
+		return wsURL, isRemote
+	}
+
+	// If last parts element is a void string,
+	// because WS URL contained an ending comma,
+	// remove it
+	parts := strings.Split(wsURL, ",")
+	if parts[len(parts)-1] == "" {
+		parts = parts[:len(parts)-1]
+	}
+
+	// Choose a random WS URL from the provided list
+	i, _ := rand.Int(rand.Reader, big.NewInt(int64(len(parts))))
+	wsURL = parts[i.Int64()]
+
+	return wsURL, isRemote
+}
--- a/tests/browser_type_test.go
+++ b/tests/browser_type_test.go
@@ -3,6 +3,9 @@ package tests
 import (
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/xk6-browser/browser"
 	"github.com/grafana/xk6-browser/chromium"
 	"github.com/grafana/xk6-browser/k6ext/k6test"
 )
@@ -17,4 +20,39 @@ func TestBrowserTypeConnect(t *testing.T) {
 
 	b := bt.Connect(tb.wsURL, nil)
 	b.NewPage(nil)
+}
+
+func TestBrowserTypeLaunchToConnect(t *testing.T) {
+	vu := k6test.NewVU(t)
+	tb := newTestBrowser(t)
+	bp := newTestBrowserProxy(t, tb)
+
+	// Export WS URL env var
+	// pointing to test browser proxy
+	t.Setenv("K6_BROWSER_WS_URL", bp.wsURL())
+
+	// We have to call launch method through JS API in Goja
+	// to take mapping layer into account, instead of calling
+	// BrowserType.Launch method directly
+	rt := vu.Runtime()
+	root := browser.New()
+	mod := root.NewModuleInstance(vu)
+	jsMod, ok := mod.Exports().Default.(*browser.JSModule)
+	require.Truef(t, ok, "unexpected default mod export type %T", mod.Exports().Default)
+
+	vu.MoveToVUContext()
+
+	require.NoError(t, rt.Set("chromium", jsMod.Chromium))
+	_, err := rt.RunString(`
+		const b = chromium.launch();
+		b.close();
+	`)
+	require.NoError(t, err)
+
+	// Verify the proxy, which's WS URL was set as
+	// K6_BROWSER_WS_URL, has received a connection req
+	require.True(t, bp.connected)
+	// Verify that no new process pids have been added
+	// to pid registry
+	require.Len(t, root.PidRegistry.Pids(), 0)
 }
--- a/browser/module.go
+++ b/browser/module.go
@@ -15,7 +15,7 @@ type (
 	// RootModule is the global module instance that will create module
 	// instances for each VU.
 	RootModule struct {
-		pidRegistry *pidRegistry
+		PidRegistry *pidRegistry
 	}
 
 	// JSModule exposes the properties available to the JS script.
@@ -39,7 +39,7 @@ var (
 // New returns a pointer to a new RootModule instance.
 func New() *RootModule {
 	return &RootModule{
-		pidRegistry: &pidRegistry{},
+		PidRegistry: &pidRegistry{},
 	}
 }
 
@@ -50,7 +50,7 @@ func (m *RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
 		mod: &JSModule{
 			Chromium: mapBrowserToGoja(moduleVU{
 				VU:          vu,
-				pidRegistry: m.pidRegistry,
+				pidRegistry: m.PidRegistry,
 			}),
 			Devices: common.GetDevices(),
 			Version: version,
--- a/k6ext/cloud.go
+++ b/k6ext/cloud.go
@@ -1,9 +1,0 @@
-package k6ext
-
-import "os"
-
-// OnCloud returns true if xk6-browser runs in the cloud.
-func OnCloud() bool {
-	_, ok := os.LookupEnv("K6_CLOUDRUN_INSTANCE_ID")
-	return ok
-}
--- a/k6ext/env_test.go
+++ b/k6ext/env_test.go
@@ -1,0 +1,81 @@
+package k6ext
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIsRemoteBrowser(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name           string
+		envLookup      envLookupper
+		expIsRemote    bool
+		expValidWSURLs []string
+	}{
+		{
+			name: "browser is not remote",
+			envLookup: func(key string) (string, bool) {
+				return "", false
+			},
+			expIsRemote: false,
+		},
+		{
+			name: "single WS URL",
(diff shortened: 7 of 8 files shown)


ka3de:
A note on an alternative implementation:
I experimented first by parsing the environment variable from within the `launch` mapping function, but that required to set / unset the environment variable in the test, which is undesirable as all tests run in parallel and it can conflict with other tests, not just with the test cases for the mapping function test (this could have been fixed by not using `t.Parallel` in this single test).
So I believe is better to provide this directly to the `BrowserType` mapping function, and detect the environment set up in the `mapBrowserToGoja` function, which is the entry point for the whole mapping layer.

tests/browser_type_test.go
```diff
@@ -18,3 +22,46 @@ func TestBrowserTypeConnect(t *testing.T) {
...
+
+func TestBrowserTypeLaunchToConnect(t *testing.T) {
+	wsURLEnvVar := "K6_BROWSER_WS_URL"
+
+	setUp := func(wsURL string) {
+		os.Setenv(wsURLEnvVar, wsURL) //nolint:errcheck,gosec
+	}
+	tearDown := func() {
+		os.Unsetenv(wsURLEnvVar) //nolint:errcheck
+	}
```
inancgumus, line 34: There is [`SetEnv`](https://pkg.go.dev/testing#T.Setenv) for this:

```suggestion
	t.SetEnv(wsURLEnvVar, wsURL)
```

tests/browser_type_test.go
```diff
@@ -18,3 +22,46 @@ func TestBrowserTypeConnect(t *testing.T) {
...
+		t.Fatalf("unexpected default mod export type %T", mod.Exports().Default)
+	}
+
+	vu.MoveToVUContext()
+
+	require.NoError(t, rt.Set("chromium", jsMod.Chromium))
+	_, err := rt.RunString(`
+		const b = chromium.launch();
+		b.close();
+	`)
```
inancgumus, line 61: 👍

browser/mapping_test.go
```diff
@@ -193,6 +193,91 @@ func TestMappings(t *testing.T) {
 	}
 }
 
+type mockBrowserType struct {
```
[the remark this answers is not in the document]
ankur22: I think this has value, as it's testing a "hidden" behaviour, which can only be asserted by looking at the internals of what the `launch` method is doing based on some input params, otherwise the result of the `launch` is the same, a `browser`, so difficult to assert that nay changed.

We could improve the testability by using a behavioural pattern that is a bit more explicit in what it does based on the input, but that also complicates the implementation.

browser/mapping_test.go
```diff
@@ -193,6 +193,91 @@ func TestMappings(t *testing.T) {
...
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			launch := mapBrowserType(
+				mVU, tc.fields.bt, tc.fields.wsURL, tc.fields.isRemoteBrowser,
+			)["launch"]
+			launchFn, ok := launch.(func(goja.Value) *goja.Object)
+			if !ok {
```
ankur22, line 269: Could you use `require.True` instead?

tests/browser_type_test.go
```diff
@@ -18,3 +22,46 @@ func TestBrowserTypeConnect(t *testing.T) {
...
+	setUp := func(wsURL string) {
+		os.Setenv(wsURLEnvVar, wsURL) //nolint:errcheck,gosec
+	}
+	tearDown := func() {
+		os.Unsetenv(wsURLEnvVar) //nolint:errcheck
+	}
+
+	vu := k6test.NewVU(t)
+	tb := newTestBrowser(t)
+	bp := newTestBrowserProxy(t, tb)
```
ankur22, line 38: I understand that the `browserProxy` helps assert that a connection has been made instead of launching a new browser, but i wonder if we could've launched a `newTestBrowser`; retrieved the ws url; set that as the env var; call `chromium.launch();`; and assert that there are no pids registered?

It could still be worth ensuring that a pid is not registered (if that is possible).

ankur22 (approved):
Nicely done 🙂 I left some minor things to consider, but they're not critical so you can leave them too.

browser/mapping_test.go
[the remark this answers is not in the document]
inancgumus: I took out a piece of code temporarily:

https://github.com/grafana/xk6-browser/blob/b2a884cf80c2fb09ac0f33859632bbdb81325491/browser/mapping.go#L722-L725

And then executed [the integration test](https://github.com/grafana/xk6-browser/blob/b2a884cf80c2fb09ac0f33859632bbdb81325491/tests/browser_type_test.go) and it failed:

```
=== RUN   TestBrowserTypeLaunchToConnect
    browser_type_test.go:66: 
                Error Trace:    /Users/inanc/grafana/k6browser/main/tests/browser_type_test.go:66
                Error:          Should be true
                Test:           TestBrowserTypeLaunchToConnect
```

I thought the error message could be improved. Maybe something like:

```
When "K6_BROWSER_WS_URL" environment variable is provided, the launch method in the mapping layer should call BrowserType.Connect.
```

Anyway. I've checked the test code:

https://github.com/grafana/xk6-browser/blob/b2a884cf80c2fb09ac0f33859632bbdb81325491/tests/browser_type_test.go#L57-L66

Where I see it calls `chromium.Launch` and it does not connect. The problem is obvious. I'll first check the mapping layer since the test code has ``jsMod.Chromium.Launch`. So I looked at the mapping layer, found [the launch mapping](https://github.com/grafana/xk6-browser/blob/b2a884cf80c2fb09ac0f33859632bbdb81325491/browser/mapping.go#LL718), and fixed the problem.

I quickly located the problem with the integration test without leaning on the unit test. The unit test has a value, but I don't think it's necessary in this situation. It just told me `Connect` is not called. I already know that from the integration test.

The actual problem arose from the fact that the `launch -> connect` logic was put in the mapping layer. This made testing difficult, and the unit test became brittle: The test can fail with even a small change in the tested code. Instead, placing this logic in the inner layers could make testing easier.

Then again, I'm happy to go with what you guys decide.

ka3de:
Thanks a lot for your reviews and comments @inancgumus and @ankur22 !
Let me review and address them :+1:

tests/browser_type_test.go
inancgumus: We can do this by putting a field on the `JSModule` and doing `require.Len(t, jsMod.PidRegistry.Pids(), 0)`?

browser/mapping.go
```diff
@@ -31,11 +31,12 @@ type mapping = map[string]any
...
-		rt          = vu.Runtime()
-		obj         = rt.NewObject()
-		browserType = chromium.NewBrowserType(vu)
+		rt                     = vu.Runtime()
+		obj                    = rt.NewObject()
+		wsURL, isRemoteBrowser = k6ext.IsRemoteBrowser()
+		browserType            = chromium.NewBrowserType(vu)
 	)
-	for k, v := range mapBrowserType(vu, browserType) {
+	for k, v := range mapBrowserType(vu, browserType, wsURL, isRemoteBrowser) {
```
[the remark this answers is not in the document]
ka3de: The distinction that I had in mind was in the case that there is an exported `K6_BROWSER_WS_URL` variable, but with value `""` (void string). In that case `isRemoteBrowser` will be `true` but `wsURL` will be `""`. What should we do in that case? Personally I thought that it was important to make this distinction because if the env var is set we can assume that the browser was expected to be remote (even if maybe there was an error setting the value), and therefore continue with the `connect` process, which will eventually fail. Otherwise, we only call `launch` if the env var is unset, not if its value is a void string.

What is your take on this?

browser/mapping_test.go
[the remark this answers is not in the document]
ka3de: If you guys are happy with the integration test, I think we can remove this unit test at the mapping level.

k6ext/env.go
```diff
@@ -0,0 +1,40 @@
...
+	"crypto/rand"
+	"math/big"
+	"os"
+	"strings"
+)
+
+// envResolve is defined as global var so we can overwrite it in
+// unit tests and therefore not manipulate the environment, which
+// allows us to run the test in parallel safely.
+var envResolve envResolver = os.LookupEnv //nolint:gochecknoglobals
```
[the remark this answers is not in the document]
ka3de: Done. See [here](https://github.com/grafana/xk6-browser/blob/359eeb2f0cacbcf8c2917859d9b769fb59b2a254/k6ext/env.go#L13-L14).
I did not change the issue as it already provides a GH search that highlights all the `os.LookupEnv` usage in xk6-browser.

k6ext/env.go
```diff
@@ -0,0 +1,40 @@
...
+
+type envResolver func(key string) (string, bool)
```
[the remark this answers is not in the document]
ka3de: I though about this, but I saw pros and cons. With this implementation we would support parallelism in test cases and avoid the global var, but we would "complicate" more the "runtime" code.
If you think that's the better option I'm happy to change it! :+1:  Just pointing out my though process for this decision.

tests/browser_type_test.go
ka3de: Great! Didn't know that, thanks! :+1: 
Ammending this to the original commit and adding you as co-author as more changes are required.
See [here](https://github.com/grafana/xk6-browser/blob/5d42c87d06684561b97f8bb9fa3213b47ea0b8b3/tests/browser_type_test.go#L30-L32).

browser/mapping_test.go
ka3de: Agree, that's better! :+1: 
Waiting to change this because based on this [discussion](https://github.com/grafana/xk6-browser/pull/819#discussion_r1151840768) we might remove the test completely.

browser/mapping_test.go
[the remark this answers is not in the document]
inancgumus: I'm happy with the integration test 😊

ka3de:
> Have you explored the [ws package](https://github.com/grafana/xk6-browser/tree/main/tests/ws)? I guess it didn't work for this.

I did experiment with it, basically replacing the `testBrowserProxy` by:
```go
        server := ws.NewServer(t)
	url, _ := url.Parse(server.ServerHTTP.URL)
	wsURL := fmt.Sprintf("ws://%s/", url.Host)
```
But this did not work, as, correct me if I'm wrong, the WS server provides a simple WS stub in order to test connections, which works very good for our `connection` implementation, but if you want it to replace the actual CDP WS you have to implement the handler yourself, such as in `[session_test.go](https://github.com/grafana/xk6-browser/blob/a01218eaee2f3bda9fbff9d730d6bb3ffe4d906f/common/session_test.go#L49-L81)`.
To me this seemed unnecessary for the integration test when I can wrap up the whole browser functionality in a small proxy, and I though it could also provide more value for other tests.

What is your take on this? Should we try to increment the functionalities of the WS server stub for the integration test in this case?

k6ext/env.go
```diff
@@ -0,0 +1,40 @@
...
+		// because WS URL contained an ending comma,
+		// remove it
+		if parts[len(parts)-1] == "" {
+			parts = parts[:len(parts)-1]
+		}
+		i, _ := rand.Int(rand.Reader, big.NewInt(int64(len(parts))))
+		wsURL = parts[i.Int64()]
+	}
+
+	return
```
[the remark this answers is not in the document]
inancgumus: Thanks for considering my suggestion! Instead of saying what the rand does, I meant explaining "why" we put rand there: Selecting WS URLs in a round-robin fashion since we want to connect to a browser evenly.". Something like that. But it's fine. Don't worry about rechanging the code.

Btw, please add the return values to the last return statement:

```go
return wsURL, isRemote
```

[Here](https://github.com/golang/go/issues/21291) is some more food thought about the why.

k6ext/env.go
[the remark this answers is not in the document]
inancgumus: I think this is perfectly fine to "complicate" the code a little instead having a global nasty environment variable. I don't worry too much about the current solution because we'll soon get rid of it.

tests/browser_type_test.go
inancgumus: That's why I like code reviews :) To learn from each other's knowledge.

inancgumus:
I see, that makes sense. And I agree this proxy can be valuable for other tests too. I'd love to see we improve what we have if it lacks some functionality. Maybe we can combine the proxy's power with the ws package. I have a weak opinion on this and trust your judgment.

k6ext/env.go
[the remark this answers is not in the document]
ka3de: Got it! I was not sure whether to mention `round-robin` here, as that seems to be "away" of the competencies of this method, but I guess it's already part of its intention, so adding it to the doc comment.
In regards of the `return` statement. Thanks for the input! Was not aware of that. I see some discussions ongoing, so not sure what would be the most idiomatic way as of now, but changing it to be more explicit as we agree on this.
See [203c192](https://github.com/grafana/xk6-browser/pull/819/commits/203c192cb8511e1d914e76f1a53287ff05aaf82b)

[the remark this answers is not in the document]
inancgumus: I believe, when you read a piece of code, it's better to know why it's there. So thanks, adding the comment.

The idiomatic way is to **never** use a naked return unless a function is teo or three lines of code.

tests/browser_type_test.go
ka3de: Got it. Nevermind, I think we can change this now also.
See [ec5399b](https://github.com/grafana/xk6-browser/pull/819/commits/ec5399bfd420498d6e731f9abd77a4fe60c8b3b9) and [953dee6](https://github.com/grafana/xk6-browser/pull/819/commits/953dee6ee05e6377217d5bfb41ea08441b49bf7b).

k6ext/env.go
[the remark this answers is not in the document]
ka3de: Got it. Nevermind, I think we can change this now also.
See [ec5399b](https://github.com/grafana/xk6-browser/pull/819/commits/ec5399bfd420498d6e731f9abd77a4fe60c8b3b9) and [953dee6](https://github.com/grafana/xk6-browser/pull/819/commits/953dee6ee05e6377217d5bfb41ea08441b49bf7b).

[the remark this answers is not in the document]
inancgumus: Thanks for considering and applying the change 👍

tests/browser_type_test.go
inancgumus: Yes, that would be a much better solution 👍

ka3de: Changed in [97c05d1](https://github.com/grafana/xk6-browser/pull/819/commits/97c05d19646c3247835f13cf997cbe6f4b4857d9) and [937ae00](https://github.com/grafana/xk6-browser/pull/819/commits/937ae00303651ef22d56dc3f4f7ab1d6404065ef).

browser/mapping_test.go
ka3de: Because the mentioned [discussion](https://github.com/grafana/xk6-browser/pull/819#discussion_r1151840768) was resolved, let me know here @ankur22  if you agree on removing the unit test now that we have a working integration test, or if we should address this comment on its implementation :+1:

tests/browser_type_test.go
ankur22: Yeah, this is a nice compliment 🙂 

I'm still not 100% sold on the idea of needing a proxy websocket server for this change. Why can't we work with something like:

```go
func TestBrowserTypeLaunchToConnect(t *testing.T) {
	vu := k6test.NewVU(t)
	tb := newTestBrowser(t)

	// Export WS URL env var
	// pointing to test browser proxy
	t.Setenv("K6_BROWSER_WS_URL", tb.wsURL)
```

The downside is that we can't assert using `require.True(t, bp.connected)`, but we're still able to assert that no new pid were registered, which is enough to suggest that `connect` was used, right?

browser/mapping_test.go
ka3de: Done :+1: See the current commit for the mapping layer modifications: [926f986](https://github.com/grafana/xk6-browser/pull/819/commits/926f986f88c2a83c941b0c8ad4d0027b75b9fc11).

tests/browser_type_test.go
ka3de: I guess we can, but for me the biggest verification that `launch` transitioned into `connect` is to verify that the "launched browser" was in fact not launched, but that it did `connect` to the WS URL that was set. Initially I tried doing this just using the `testBrowser` and verifying that the "new browser" WS URL was matching the one given. But this was "a mess" to do having Goja abstractions in between (which is necessary because the transition is mainly in the mapping layer). That's when I came with the idea to verify this connection from the other side, hence the proxy. Also I was hoping it could be useful for other test cases, but it's true that right now it's only serving this case.

I feel the pid registry len is a weeker/detail point to verify this functionality, but I'm happy to remove it if we think this is good enough :+1:

ankur22: Yeah, i do agree that the pid registry is a weaker method of asserting that no browser was launched. The combo of the unit test and a simpler integration test would've been good enough IMO. Happy to leave the proxy in for now, and if any other ideas crop up we can explore those at a later date.

ka3de: Do we agree on merging this then as it is?

Issue #823, opened 2023-03-10
Allow making a logger from `FieldLogger` instead of concrete `*Logrus`


inancgumus:
This is a continuation of #818. We've decided to use [a hack](https://github.com/grafana/xk6-browser/issues/818#issuecomment-1460081431) to stay compatible with k6-core until the next cycle.

PR #824, opened 2023-03-10
Refactor logger.New to accept FieldLogger

inancgumus:
While doing this I've also refactored the code a little bit. Let me know if you think the otherwise.

Closes: #823

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -31,21 +31,23 @@ func NewNullLogger() *Logger {
 }
 
 // New creates a new logger.
-func New(logger *logrus.Logger, iterID string) *Logger {
-	var defLogger bool
+func New(logger logrus.FieldLogger, iterID string) *Logger {
+	var ll *logrus.Logger
+
 	if logger == nil {
-		defLogger = true
-		logger = logrus.New()
-	}
-	l := &Logger{
-		Logger: logger,
-		iterID: iterID,
-	}
-	if defLogger {
-		l.Warnf("Logger", "no logger supplied, using default")
+		ll = logrus.New()
+		ll.Warn("Logger", "no logger supplied, using default")
+	} else if l, ok := logger.(*logrus.Logger); ok {
+		ll = l
+	} else {
+		// panic so that we know when to upgrade. see issue #818.
+		panic("logger is not a logrus.Logger")
 	}
 
-	return l
+	return &Logger{
+		Logger: ll,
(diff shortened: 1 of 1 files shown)


log/logger.go
```diff
@@ -31,21 +31,23 @@ func NewNullLogger() *Logger {
...
 	if logger == nil {
-		defLogger = true
-		logger = logrus.New()
+		ll = logrus.New()
+		ll.Warn("Logger", "no logger supplied, using default")
+	} else if l, ok := logger.(*logrus.Logger); ok {
+		ll = l
+	} else {
+		// panic so that we know when to upgrade. see issue #818.
+		panic("logger is not a logrus.Logger")
```
ankur22, line 44: Could we work with the default (create a new logger like we do in the `nil` case) logger instead of panicking?

inancgumus: Sure, but does it provide the same kind of security? I fear we might easily miss it.

inancgumus: Ah, I see what you mean. You want to keep working with the `*logrus.Logger` no matter what. That's also an option, yes. WDYT? @na--?

ankur22: Yeah, that makes sense. Could you try working with `ErrFatal` instead of `panic`?

inancgumus: @ankur22 Please take a look: https://github.com/grafana/xk6-browser/pull/824/commits/3b2de1b241017bad1707289ca974a7264be3463b.

log/logger.go
```diff
@@ -31,21 +31,21 @@ func NewNullLogger() *Logger {
...
-	var defLogger bool
+func New(logger logrus.FieldLogger, iterID string) *Logger {
+	ll := logrus.New()
+
 	if logger == nil {
-		defLogger = true
-		logger = logrus.New()
+		ll.Warn("Logger", "no logger supplied, using default")
+	} else if l, ok := logger.(*logrus.Logger); !ok {
+		ll.Warn("Logger", "invalid logger type, using default")
```
ankur22, line 40: Isn't this a NPE?

inancgumus: No, why?

ankur22 (approved):
LGTM

log/logger.go
inancgumus: Yep! The previous one was so :)

ka3de (approved):
LGTM.

Issue #804, opened 2023-03-02
Edge-case: Mapping fails if the extension code returns nil values


inancgumus:
### Brief summary

Tried to map even if an interim step in the map method gets a `nil` value.

For example, the following will panic if `p.Query` returns `nil`.

```go
func mapPage(vu moduleVU, p api.Page) mapping {
	...
	maps["$"] = func(selector string) *goja.Object {
		eh := p.Query(selector)
		// 
		// this will panic if Query returns nil.
		// 
		ehm := mapElementHandle(vu, eh)
		return rt.ToValue(ehm).ToObject(rt)
	}
	...
}
```

### Steps to reproduce the problem

You can see it in action by running the following script:

<details>
<summary>Test Script</summary>

```js
import { check } from 'k6';
import { chromium } from 'k6/x/browser';

export const options = {
  thresholds: {
    checks: ["rate==1.0"]
  }
}

export default async function() {
  const browser = chromium.launch({
    headless: true,
  });
  const context = browser.newContext();
  const page = context.newPage();

  try {
    await page.goto('https://test.k6.io/');
    check(page, {
      'Title with CSS selector':
        p => p.$('NOT EXIST').textContent() == 'test.k6.io',
    });
  } finally {
    page.close();
    browser.close();
  }
}
```

</details>

<details>
<summary>Environmental details</summary>

### Expected behaviour

No panic and log a friendly error message.

### Actual behaviour

Nil pointer error.

### xk6-browser version

0.8.1

### OS

macOS 13.1 (22C65)

### Chrome version

98.0.4753.0

</details>

## Possible problematic mappings

The following mapping gets a value from the extension methods and tries to map. This can be problematic if the returned value is `nil` and the extension method is not panicking.

```[tasklist]
- [x] https://github.com/grafana/xk6-browser/pull/820
- [x] evaluateHandle: `mapJSHandle.evaluateHandle`, `mapFrame.evaluateHandle`, `mapPage.evaluateHandle`, `mapWorker.evaluateHandle`
- [x] `mapJSHandle.getProperties`
- [x] `mapElementHandle.contentFrame`
- [x] `mapElementHandle.ownerFrame`
- [x] `mapFrame.childFrames`
- [x] `mapFrame.frameElement`
- [x] `mapBrowserContext.newPage`
```

PR #820, opened 2023-03-09
Add error handling to wildcard selectors

inancgumus:
This allows us to:
- Better handle extension errors.
- Leaves the decision of **properly handling errors** to Goja when an element is not found. This is how it's done in the k6-core and other extensions.

Returning an error from the mapping layer also allows us to prevent multiple unnecesary panics.

<details>
<summary>You can use this script to test the behavior (main and this branch)</summary>

```js
import { check } from 'k6';
import { chromium } from 'k6/x/browser';

export const options = {
  thresholds: {
    checks: ["rate==1.0"]
  }
}

export default async function() {
  const browser = chromium.launch({
    headless: __ENV.XK6_HEADLESS ? true : false,
  });
  const context = browser.newContext();
  const page = context.newPage();

  try {
    await page.goto('https://test.k6.io/');

    check(page, {
      'Title with CSS selector':
        p => {
          console.log('checking page textcontext');
          return p.$('header h1.titlex').textContent() == 'test.k6.io'
        },
    });
  } finally {
    page.close();
    browser.close();
  }
}
```

</details>

<details>
<summary>Error output (current v0.8.0)</summary>

```
ERRO[0001] panic: runtime error: invalid memory address or nil pointer dereference
goroutine 68 [running]:
runtime/debug.Stack()
        runtime/debug/stack.go:24 +0x64
go.k6.io/k6/js/common.RunWithPanicCatching.func1()
        go.k6.io/k6@v0.42.1-0.20230130080633-582ec4d3940c/js/common/util.go:82 +0x190
panic({0x101be1820, 0x102a6d620})
        runtime/panic.go:884 +0x1f4
github.com/dop251/goja.(*Runtime).runWrapped.func1()
        github.com/dop251/goja@v0.0.0-20230128084908-78b980256d04/runtime.go:2480 +0x168
panic({0x101be1820, 0x102a6d620})
        runtime/panic.go:884 +0x1f4
github.com/dop251/goja.(*vm).handleThrow(0x140005fc240, {0x101be1820, 0x102a6d620})
        github.com/dop251/goja@v0.0.0-20230128084908-78b980256d04/vm.go:730 +0x334
github.com/dop251/goja.(*vm).try.func1()
        github.com/dop251/goja@v0.0.0-20230128084908-78b980256d04/vm.go:749 +0x48
panic({0x101be1820, 0x102a6d620})
        runtime/panic.go:884 +0x1f4
[...]

PR #826, opened 2023-03-14
Fix Edge-case: Mapping fails if the extension code returns nil values #804

inancgumus:
The error output is similar to #820.

Please let me know if you see other edge cases, so I can amend them.

Here are some test snippets you can try.

<details>
<summary>evaluateHandle</summary>

Output:

```
ERRO[0001] Uncaught (in promise) GoError: evaluating handle: result is nil
        at github.com/grafana/xk6-browser/browser.mapPage.func2 (native)
        at file:///Users/inanc/grafana/k6browser/tests/evaluate_handle.js:20:54(31)  executor=per-vu-iterations scenario=default
```

Script:

```js
await page.goto('https://test.k6.io/', { waitUntil: 'load' });
const pageTitleHandle = await page.evaluateHandle(() => {
  return document.querySelector('nothing');
});
```

</details>

<details>
<summary>waitForSelector</summary>

Output:

```
ERRO[0001] Uncaught (in promise) GoError: querying selector "x"
        at github.com/grafana/xk6-browser/browser.mapPage.func13 (native)
        at file:///Users/inanc/grafana/k6browser/tests/waitforselector.js:20:23(32)  executor=per-vu-iterations scenario=default
```

Script:

```js
await page.goto('https://test.k6.io/', { waitUntil: 'load' });
const a = await page.waitForSelector('x');
```

</details>

Closes: https://github.com/grafana/xk6-browser/issues/804
Related: https://github.com/grafana/xk6-browser/issues/688 #820

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -662,10 +662,12 @@ func mapBrowserContext(vu moduleVU, bc api.BrowserContext) mapping {
 
 			return rt.ToValue(mpages).ToObject(rt)
 		},
-		"newPage": func() *goja.Object {
-			page := bc.NewPage()
-			m := mapPage(vu, page)
-			return rt.ToValue(m).ToObject(rt)
+		"newPage": func() (mapping, error) {
+			page, err := bc.NewPage()
+			if err != nil {
+				return nil, err //nolint:wrapcheck
+			}
+			return mapPage(vu, page), nil
 		},
 	}
 }
@@ -689,10 +691,12 @@ func mapBrowser(vu moduleVU, b api.Browser) mapping {
 			m := mapBrowserContext(vu, bctx)
 			return rt.ToValue(m).ToObject(rt)
 		},
-		"newPage": func(opts goja.Value) *goja.Object {
-			page := b.NewPage(opts)
-			m := mapPage(vu, page)
-			return rt.ToValue(m).ToObject(rt)
+		"newPage": func(opts goja.Value) (mapping, error) {
+			page, err := b.NewPage(opts)
+			if err != nil {
+				return nil, err //nolint:wrapcheck
+			}
+			return mapPage(vu, page), nil
 		},
 	}
 }
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -208,10 +208,12 @@ func mapElementHandle(vu moduleVU, eh api.ElementHandle) mapping {
 				return nil, err //nolint:wrapcheck
 			})
 		},
-		"contentFrame": func() *goja.Object {
-			f := eh.ContentFrame()
-			mf := mapFrame(vu, f)
-			return rt.ToValue(mf).ToObject(rt)
+		"contentFrame": func() (mapping, error) {
+			f, err := eh.ContentFrame()
+			if err != nil {
+				return nil, err //nolint:wrapcheck
+			}
+			return mapFrame(vu, f), nil
 		},
 		"dblclick":      eh.Dblclick,
 		"dispatchEvent": eh.DispatchEvent,
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -230,10 +230,12 @@ func mapElementHandle(vu moduleVU, eh api.ElementHandle) mapping {
 		"isEnabled":     eh.IsEnabled,
 		"isHidden":      eh.IsHidden,
 		"isVisible":     eh.IsVisible,
-		"ownerFrame": func() *goja.Object {
-			f := eh.OwnerFrame()
-			mf := mapFrame(vu, f)
-			return rt.ToValue(mf).ToObject(rt)
+		"ownerFrame": func() (mapping, error) {
+			f, err := eh.OwnerFrame()
+			if err != nil {
+				return nil, err //nolint:wrapcheck
+			}
+			return mapFrame(vu, f), nil
 		},
 		"press":                  eh.Press,
 		"screenshot":             eh.Screenshot,
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -171,12 +171,17 @@ func mapJSHandle(vu moduleVU, jsh api.JSHandle) mapping {
 			}
 			return mapJSHandle(vu, h), nil
 		},
-		"getProperties": func() *goja.Object {
+		"getProperties": func() (mapping, error) {
+			props, err := jsh.GetProperties()
+			if err != nil {
+				return nil, err //nolint:wrapcheck
+			}
+
 			dst := make(map[string]any)
-			for k, v := range jsh.GetProperties() {
+			for k, v := range props {
 				dst[k] = mapJSHandle(vu, v)
 			}
-			return rt.ToValue(dst).ToObject(rt)
+			return dst, nil
 		},
 		"getProperty": func(propertyName string) *goja.Object {
 			var (
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -322,9 +322,12 @@ func mapFrame(vu moduleVU, f api.Frame) mapping {
 		},
 		"fill":  f.Fill,
 		"focus": f.Focus,
-		"frameElement": func() *goja.Object {
-			eh := mapElementHandle(vu, f.FrameElement())
-			return rt.ToValue(eh).ToObject(rt)
+		"frameElement": func() (mapping, error) {
+			fe, err := f.FrameElement()
+			if err != nil {
+				return nil, err //nolint:wrapcheck
+			}
+			return mapElementHandle(vu, fe), nil
 		},
 		"getAttribute": f.GetAttribute,
 		"goto": func(url string, opts goja.Value) *goja.Promise {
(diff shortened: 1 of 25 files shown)


common/element_handle.go
```diff
@@ -755,20 +755,21 @@ func (h *ElementHandle) Click(opts goja.Value) error {
...
 		err  error
 	)
 	action := dom.DescribeNode().WithObjectID(h.remoteObject.ObjectID)
 	if node, err = action.Do(cdp.WithExecutor(h.ctx, h.session)); err != nil {
-		k6ext.Panic(h.ctx, "getting remote node %q: %w", h.remoteObject.ObjectID, err)
+		return nil, fmt.Errorf("getting remote node %q: %w", h.remoteObject.ObjectID, err)
 	}
 	if node == nil || node.FrameID == "" {
-		return nil
+		return nil, fmt.Errorf("element is not an iframe")
```
inancgumus, line 774: TODO: Better error message is needed.

common/execution_context.go
```diff
@@ -319,8 +319,12 @@ func (e *ExecutionContext) EvalHandle(
 	}
 	res, err := e.eval(apiCtx, opts, js.ToString().String(), evalArgs...)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("evaluating handle: %w", err)
+	}
+	if res == nil {
```
ankur22, line 324: if we ran:

```javascript
function() {
  console.log("hello world)
}
```

This wouldn't return a result, so would `res` be `nil` in that case?

Is it possible to create a test for this scenario?

common/frame.go
```diff
@@ -758,11 +758,13 @@ func (f *Frame) EvaluateHandle(pageFunc goja.Value, args ...goja.Value) (handle
 	}
 	f.executionContextMu.RUnlock()
 	if err != nil {
-		k6ext.Panic(f.ctx, "evaluating handle: %w", err)
+		// EvalHandle has a rich error, nolinting.
+		return nil, err //nolint:wrapcheck
```
ankur22, line 767: IMO we should wrap this error with the method name to give the error more context here, instead of wrapping it in `EvalHandle`.

If another method calls `EvalHandle` then `evaluating handle:` might be the wrong context for that caller.

common/js_handle.go
```diff
@@ -101,14 +101,10 @@ func (h *BaseJSHandle) Evaluate(pageFunc goja.Value, args ...goja.Value) any {
...
-func (h *BaseJSHandle) EvaluateHandle(pageFunc goja.Value, args ...goja.Value) api.JSHandle {
+func (h *BaseJSHandle) EvaluateHandle(pageFunc goja.Value, args ...goja.Value) (api.JSHandle, error) {
 	rt := h.execCtx.vu.Runtime()
 	args = append([]goja.Value{rt.ToValue(h)}, args...)
-	res, err := h.execCtx.EvalHandle(h.ctx, pageFunc, args...)
-	if err != nil {
-		k6ext.Panic(h.ctx, "%w", err)
-	}
-	return res
+	return h.execCtx.EvalHandle(h.ctx, pageFunc, args...)
```
ankur22, line 107: Same here, I think we should wrap the error with more context here.

common/worker.go
```diff
@@ -69,9 +69,9 @@ func (w *Worker) Evaluate(pageFunc goja.Value, args ...goja.Value) any {
 }
 
 // EvaluateHandle evaluates a page function in the context of the web worker and returns a JS handle.
-func (w *Worker) EvaluateHandle(pageFunc goja.Value, args ...goja.Value) api.JSHandle {
+func (w *Worker) EvaluateHandle(pageFunc goja.Value, args ...goja.Value) (api.JSHandle, error) {
 	// TODO: implement
-	return nil
+	return nil, nil
```
ankur22, line 74: Could we use `ErrFatal` instead of `nil` here?

browser/mapping.go
```diff
@@ -248,10 +247,12 @@ func mapElementHandle(vu moduleVU, eh api.ElementHandle) mapping {
 		"type":                   eh.Type,
 		"uncheck":                eh.Uncheck,
 		"waitForElementState":    eh.WaitForElementState,
-		"waitForSelector": func(selector string, opts goja.Value) *goja.Object {
-			eh := eh.WaitForSelector(selector, opts)
-			ehm := mapElementHandle(vu, eh)
-			return rt.ToValue(ehm).ToObject(rt)
```
ankur22, line 245: Why don't we need to use `rt.ToValue(ehm).ToObject(rt)`?

ankur22 (approved):
LGTM. I left some minor comments, so up to you whether you need to action them or not.

common/frame.go
inancgumus: You mean something like `EvaluateHandle: evaluating handle:`?

common/js_handle.go
inancgumus: `BaseJSHandle` is an interim method. I don't know what we should add here to enhance the returned error. Would you happen to have a suggestion?

common/worker.go
inancgumus: Makes sense 👍

browser/mapping.go
inancgumus: Goja handles it automatically. We don't need to return a `goja.Object` or `value`. The current mapping layer implementation unnecessarily uses objects.

common/execution_context.go
inancgumus: Yes, `res` would be `nil` in that case. It's expected since `EvaluateHandle` should return a DOM handle. What would it return for that function?

inancgumus: Here's a test: https://github.com/grafana/xk6-browser/pull/826/commits/c78b301967870a6033636b4bfe7d9e607ea8f78d

common/worker.go
inancgumus: Fatalized here: https://github.com/grafana/xk6-browser/pull/826/commits/04b2e667d203a5b95cf097fc087b480548dcd0e2

common/element_handle.go
```diff
@@ -982,28 +982,31 @@ func (h *ElementHandle) OwnerFrame() api.Frame {
 	}
 	res, err := h.evalWithScript(h.ctx, opts, fn)
 	if err != nil {
-		k6ext.Panic(h.ctx, "getting document element: %w", err)
+		err = fmt.Errorf("getting document element: %w", err)
 	}
-	if res == nil {
-		return nil
+	if res == nil || err != nil {
+		return nil, err
```
ka3de, line 993: I find this part a little bit confusing. Would it be better to be more explicit, such as:
```go
        res, err := h.evalWithScript(h.ctx, opts, fn)
	if err != nil {
		return nil, fmt.Errorf("getting document element: %w", err)
	}
	if res == nil {
		return nil, nil
	}
```
?

ka3de (approved):
Thanks for working on this! Seems like we still have things to learn about Goja :smile: 
Made a small comment, but overall LGTM.

common/element_handle.go
inancgumus: > Thanks for working on this! Seems like we still have things to learn about Goja 😄

Yeah 😄 Thanks!

> I find this part a little bit confusing.

Explicitified here: https://github.com/grafana/xk6-browser/pull/826/commits/d4aec1fafd7cb909d761a5ba54bb54b5339214ee

common/frame.go
ankur22: I agree that `EvaluateHandle: evaluating handle:` is not a good approach.

This is very minor so, don't worry about changing the code if it doesn't make sense. What i meant was that instead of wrapping the error in `EvalHandle` like so:

```go
return nil, fmt.Errorf("evaluating handle: %w", err)
```

why not just return `err` and allow this method (`EvaluateHandle`) to wrap the error with:

```go
 fmt.Errorf("evaluating handle: %w", err)
```

Another example is, in `EvalHandle` when `res` is `nil` it is currently wrapped like so:

```go
return nil, fmt.Errorf("evaluating handle: result is nil")
```

When it could be:

```go
return nil, fmt.Errorf("result is nil")
```

And in `EvaluateHandle` we would wrap the `err` with:

```go
fmt.Errorf("evaluating handle: %w", err).
```

Like i said, super minor so don't worry about this 🙂

common/js_handle.go
ankur22: Ah, ok, that's fine in that case. I wouldn't worry about changing anything here. It was minor anyway 🙂

ankur22 (approved):

common/frame.go
inancgumus: @ankur22, oh, I see what you mean now 🙂 Thanks for explaining 🙇

I couldn't see a difference in adding `"evaluating handle"` in `EvalHandle` vs. in`Page|Frame|JSHandle (ElementHandle)` since the final error message would be the same. So I've decided to go this way: [f040f90](https://github.com/grafana/xk6-browser/pull/826/commits/f040f9019edaeca5e017cbe577742d1059cee3de) to differentiate the error messages depending on where `EvalHandle` is called. Please check it out and let me know; thanks!

ankur22: 🎉 Nice! That's exactly what i was thinking. Thanks 🙂

ankur22 (approved):

PR #832, opened 2023-03-17
Fix killing a browser process on panic if one of them is missing

inancgumus:
The current code returns when it can't find a process with PID, instead of trying to kill the rest. This PR fixes that sneaky bug.

Related: #779


ka3de (approved):
Nice catch.
LGTM.

ankur22 (approved):
LGTM

Issue #833, opened 2023-03-20
`close` on `page` should close the `page` and not the whole `browserContext`


ankur22:
### Brief summary

When a test is ran with multiple `page`s, if one `page` is closed with `page.close`, it closes the `browserContext`, so all pages are closed, even though the user may only have wanted the current `page` to close. This results in an error when other `page`s from the now closed `browserContext` are used later in the test.

### xk6-browser version

48ee41a59e0f0561dbcae39b91dc6fba338422a4

### OS

Mac

### Chrome version

Version 110.0.5481.177 (Official Build) (arm64)

### Docker version and image (if applicable)

NA

### Steps to reproduce the problem

When the following test is ran, it closes the whole `browserContext` when we want it to only close the current `page`:

```js
import { chromium } from 'k6/x/browser';

export default async function () {
  const browser = chromium.launch({headless: true});
  const context = browser.newContext();

  try {
    await runTest(context, 'https://grafana.com', 'https://play.grafana.org/', 'grafana_screenshot.png')
    await runTest(context, 'https://k6.io', '/jobs/', 'k6_screenshot.png')
  } finally {
    browser.close();
  }
}

async function runTest(context, url, clickHref, screenshotFile) {
  try{
    const page = context.newPage();
    await page.goto(url, { waitUntil: 'networkidle' })

    await Promise.all([
      page.waitForNavigation({ waitUntil: 'networkidle' }),
      page.locator('a[href="' + clickHref + '"]').click(),
    ]);

    page.screenshot({ path: screenshotFile });
  } finally {
    page.close();
  }
}
```

### Expected behaviour

No errors.

### Actual behaviour

The following error and the second navigation on a new page is not ran:

```bash
ERRO[0006] Uncaught (in promise) ReferenceError: page is not defined
```

PR #834, opened 2023-03-20
Fix close page so that it closes the current page and not the whole browser context

ankur22:
This changes the behaviour of what `page.close` does so now it will close the `page` and leave the other related objects intact.

Before this change when `page.close` was called, it would close the whole `browserContext`. In most cases this wasn't an issue. In very few cases this can be an issue, such as with the attached script:

```js

  const browser = chromium.launch({headless: true});

  try {
    await runTest(context, 'https://grafana.com', 'https://play.grafana.org/', 'grafana_screenshot.png')
    await runTest(context, 'https://k6.io', '/jobs/', 'k6_screenshot.png')
  } finally {
    browser.close();
  }
}

async function runTest(context, url, clickHref, screenshotFile) {
  
  try{
    await page.goto(url, { waitUntil: 'networkidle' })

    await Promise.all([
      page.waitForNavigation({ waitUntil: 'networkidle' }),
      page.locator('a[href="' + clickHref + '"]').click(),
    ]);

    page.screenshot({ path: screenshotFile });
  } finally {
    page.close();
  }
}
```

This would result in `ERRO[0005] Uncaught (in promise) GoError: creating new page in browser context: missing browser context: 1B027BB7E43D45C6533FA188B6850E1A`.

There is also a `browserContext.close` API, and it seems that a correct approach when it comes to cleaning up a `browser`, `browserContext` and `page`, is to call their respective `close` functions instead of relying on `page.close` to close the `browserContext` too.

We're also exploring a more concise way of writing tests, where the end-user doesn't need to be aware of a `browser` or a `browserContext`, and the lifecycle of these would be handled by xk6-browser itself. This means that the `page` object is what users will work with and that is all they need to cleanup once the test completes. This change also helps us work towards that end goal.

Closes: https://github.com/grafana/xk6-browser/issues/833

### DIFF
--- a/common/page.go
+++ b/common/page.go
@@ -410,10 +410,12 @@ func (p *Page) Close(opts goja.Value) error {
 		// When a close target command is sent to the browser via CDP,
 		// the browser will start to cleanup and the first thing it
 		// will do is return a target.EventDetachedFromTarget, which in
-		// our implementation will close the connection associated to
-		// this session, which can result in the context being closed
-		// while we're waiting for the response to come back from the
-		// browser for this current command (it's racey).
+		// our implementation will close the session connection (this
+		// does not close the CDP websocket, just removes the session
+		// so no other CDP calls can be made with the session ID).
+		// This can result in the session's context being closed while
+		// we're waiting for the response to come back from the browser
+		// for this current command (it's racey).
 		if errors.Is(err, context.Canceled) {
 			return nil
 		}


common/page.go
```diff
@@ -401,10 +401,27 @@ func (p *Page) Click(selector string, opts goja.Value) error {
...
+		// our implementation will close the connection associated to
+		// this session, which can result in the context being closed
+		// while we're waiting for the response to come back from the
+		// browser for this current command (it's racey).
+		if errors.Is(err, context.Canceled) {
+			return nil
+		}
+
+		return fmt.Errorf("closing a page: %w", err)
+	}
```
inancgumus, line 422: Should we add a timeout here in case the closed browser connection hogs the CDP send action here?

browser/mapping.go
```diff
@@ -439,7 +439,9 @@ func mapPage(vu moduleVU, p api.Page) mapping {
 				return nil, err //nolint:wrapcheck
 			})
 		},
-		"close":                   p.Close,
+		"close": func(opts goja.Value) error {
+			return p.Close(opts) //nolint:wrapcheck
+		},
```
inancgumus, line 444: Can we simply do:

```suggestion
		"close": p.Close(opts),
```

tests/browser_test.go
```diff
@@ -28,12 +28,14 @@ func TestBrowserNewPage(t *testing.T) {
 	l = len(b.Contexts())
 	assert.Equal(t, 2, l, "expected there to be 2 browser context, but found %d", l)
 
-	p.Close(nil)
+	err := p.Close(nil)
+	require.NoError(t, err)
 	l = len(b.Contexts())
-	assert.Equal(t, 1, l, "expected there to be 1 browser context after first page close, but found %d", l)
-	p2.Close(nil)
+	assert.Equal(t, 2, l, "expected there to be 2 browser context after first page close, but found %d", l)
```
inancgumus, line 34:
```suggestion
	assert.Equal(t, 2, l, "expected there to be 2 browser contexts after first page close, but found %d", l)
```

tests/browser_test.go
```diff
@@ -28,12 +28,14 @@ func TestBrowserNewPage(t *testing.T) {
...
+	err = p2.Close(nil)
+	require.NoError(t, err)
 	l = len(b.Contexts())
-	assert.Equal(t, 0, l, "expected there to be 0 browser context after second page close, but found %d", l)
+	assert.Equal(t, 2, l, "expected there to be 2 browser context after second page close, but found %d", l)
```
inancgumus, line 38:
```suggestion
	assert.Equal(t, 2, l, "expected there to be 2 browser contexts after second page close, but found %d", l)
```

tests/browser_test.go
```diff
@@ -28,12 +28,14 @@ func TestBrowserNewPage(t *testing.T) {
 	l = len(b.Contexts())
 	assert.Equal(t, 2, l, "expected there to be 2 browser context, but found %d", l)
 
-	p.Close(nil)
+	err := p.Close(nil)
+	require.NoError(t, err)
```
inancgumus, line 32: Should we reduce the number-of-lines noise?

```suggestion
	require.NoError(t, p.Close(nil))
```

tests/page_test.go
```diff
@@ -679,6 +679,15 @@ func TestPageURL(t *testing.T) {
 	assert.Regexp(t, "http://.*/get", p.URL())
 }
 
+func TestPageClose(t *testing.T) {
```
inancgumus, line 682: Should we also add a test for pages created via a browser context?

inancgumus (approved):
Nicely done 👍 LGTM with some suggestions.

common/page.go
```diff
@@ -401,10 +401,27 @@ func (p *Page) Click(selector string, opts goja.Value) error {
...
+		// When a close target command is sent to the browser via CDP,
+		// the browser will start to cleanup and the first thing it
+		// will do is return a target.EventDetachedFromTarget, which in
+		// our implementation will close the connection associated to
+		// this session, which can result in the context being closed
+		// while we're waiting for the response to come back from the
+		// browser for this current command (it's racey).
+		if errors.Is(err, context.Canceled) {
+			return nil
+		}
```
ka3de, line 419: The comment seems a little bit misleading for me, especially due to the usage of `connection` and `context`. Which connection are we referring to here? It's not our "CDP connection" implementation, right? Also, I would try to clarify that the context is the page context passed to the CDP action (right?) as otherwise can be misleading for the page's `BrowserContext`.

browser/mapping.go
ankur22: I tried this, but then realised that the opts haven't been defined. So reverting this to what it was.

common/page.go
```diff
@@ -401,10 +401,27 @@ func (p *Page) Click(selector string, opts goja.Value) error {
...
+	action := target.CloseTarget(p.targetID)
+	err := action.Do(cdp.WithExecutor(p.ctx, p.session))
+	if err != nil {
+		// When a close target command is sent to the browser via CDP,
+		// the browser will start to cleanup and the first thing it
+		// will do is return a target.EventDetachedFromTarget, which in
+		// our implementation will close the connection associated to
+		// this session, which can result in the context being closed
+		// while we're waiting for the response to come back from the
+		// browser for this current command (it's racey).
```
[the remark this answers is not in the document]
ankur22: Yeah, good point. The session connection doesn't actually close the CDP connection, it just removes the sessions from the current runtime. The context is related to the session's context, not the page's context or the browserContext. 

Resolved in cd8b5c082fd24ed2aae226482ce27597c69de83d

tests/page_test.go
ankur22: Resolved in afbb6991d1cd837e62d64d04d7bc3f92e24aeec5

common/page.go
ankur22: I'd prefer not to. It seems that a lot of the CDP commands we execute are done without a timeout, including the previous call to dispose of the browserContext.

ka3de (approved):
LGTM.

PR #842, opened 2023-04-03
Update to k6-core master and fix logCache logger

### LINKED ISSUE #841: Accept `FieldLogger` for test browser `logCache`
The goal of this issue is to fix the test browser `logCache` implementation to accept a `logurs.FieldLogger` (as used in the current k6 implementation as of [9a635a5](https://github.com/grafana/k6/commit/9a635a5c6602ee34877375ec00678c0956b60184)) meanwhile the bigger refactor defined in #818 is done.

Related: #818 
Related: #823

ka3de:
This PR fixes an incompatibility detected with k6 core usage of `state.Logger`, which has been modified in grafana/k6#2960.

This is done following the same approach as in #824 meanwhile a bigger refactor is applied, as defined in #818.

Closes #841 .


tests/logrus_hook.go
```diff
@@ -74,10 +74,18 @@ var _ logrus.Hook = &logCache{}
 
 // attachLogCache sets logger to DebugLevel, attaches a LogCache hook and
 // returns it.
-func attachLogCache(logger *logrus.Logger) *logCache {
+func attachLogCache(fl logrus.FieldLogger) *logCache {
+	var logger *logrus.Logger
+	if ll, ok := fl.(*logrus.Logger); ok {
+		logger = ll
+	} else {
+		logger = logrus.New()
```
ankur22, line 82: Would it be a good idea to log a warning as we do in https://github.com/grafana/xk6-browser/pull/824 when `fl` is not the expected type?

ankur22 (approved):
LGTM, just wondering whether it's worth logging something when anything unexpected happens.

tests/logrus_hook.go
ankur22: Tbh, I don't fully understand how it's used, so it's why i'm leaning towards making things clearer with a log message. But, if you know what's going on then i'm happy to leave it as is.

ka3de: Actually, reviewing this to better explain the context, I think the implementation is not exactly correct, as using the default logger here (in case the provided by `state.Logger` is not of type `*logrus.Logger`), will make the tests fail because that "default logger" is in the end not used by our implementation, so no hooks will be executed. This case is different from the one implemented in #824 where we actually replace one logger for another.

I think we should explicitly return an error if the given logger is not of type `*logrus.Logger` and fail the test, until we fix #818.
Let me make the changes for this :+1:

ankur22: Brilliant, i like it 🙂

ka3de: Pushed a new version in [d506083](https://github.com/grafana/xk6-browser/pull/842/commits/d506083ca71d697a3116c569133cfc603dd6c5ff).
Now, the tests will fail also if the given logger is not of type `*logrus.Logger`, but in this case the behavior is less misleading.
Currently the tests don't fail because we transform the "initial" `state.Logger` into a `*logrus.Logger` when our implementation of `log.New` is called, as modified in #824 .

tests/test_browser.go
```diff
@@ -105,10 +105,14 @@ func newTestBrowser(tb testing.TB, opts ...any) *testBrowser {
...
 		lc         *logCache
+		err        error
 	)
 
 	if enableLogCache {
-		lc = attachLogCache(state.Logger)
+		lc, err = attachLogCache(state.Logger)
+		if err != nil {
+			panic(fmt.Errorf("testBrowser: attaching logger: %w", err))
+		}
```
inancgumus, line 115: Instead of panicking here, it would make better sense to `t.Fail` like:

```go
attachLogCache(t, state.Logger)
// attachLogCache can now use `tb.Fatalf("unexpected logger type: %T", fl)`
```

The benefit is we won't get a huge panic output because of a few failing tests. And we can continue running the other tests even if those fail. And we don't need to use an error-handling code, as the testing package will handle it for us.

`Fatalf` will print out something like: `--- FAIL: TestBrowserLogIterationID: logrus_hook.go:81: unexpected logger type: *tests.MyLogger`.

inancgumus (commented):
LGTM 👍 With one suggestion.

tests/test_browser.go
ka3de: I used `panic` to follow the pattern applied in the rest of the `testBrowser` implementation (e.g.: [here](https://github.com/grafana/xk6-browser/blob/d506083ca71d697a3116c569133cfc603dd6c5ff/tests/test_browser.go#L99) and [here](https://github.com/grafana/xk6-browser/blob/d506083ca71d697a3116c569133cfc603dd6c5ff/tests/test_browser.go#L126)), but I personally prefer to use `tb.Fatalf` also if we agree on changing all `panic` occurrences. What do you think?

inancgumus: Got it 👍 I prefer `Fatal` in the rest too. At least, we can use `Fatal` in this PR. It's fine if you don't want to tackle the rest now.

inancgumus: Got it 👍 At least we can use `Fatal` in this PR. It's fine if you don't want to tackle the rest now (as they're unrelated to this PR).

inancgumus (approved):

tests/test_browser.go
ka3de: I prefer to fix the three `panic` usages to keep it consistent. Is it ok if we do that in this PR?

ka3de: Done in [68a366e](https://github.com/grafana/xk6-browser/pull/842/commits/68a366e77e5ea85d40422b82350a2158d1bb3ff2) and [8740dd4](https://github.com/grafana/xk6-browser/pull/842/commits/8740dd463833653966bfd4da875b7f9c90dbafcb).

inancgumus (commented):
Thanks, some suggestions.

tests/logrus_hook.go
```diff
@@ -74,10 +75,19 @@ var _ logrus.Hook = &logCache{}
...
 // returns it.
-func attachLogCache(logger *logrus.Logger) *logCache {
+func attachLogCache(fl logrus.FieldLogger) (*logCache, error) {
+	var ok bool
+	var logger *logrus.Logger
+	if logger, ok = fl.(*logrus.Logger); !ok {
+		// TODO: Fix this to always work with logrus.FieldLoger.
+		// See: https://github.com/grafana/xk6-browser/issues/818
+		return nil, fmt.Errorf("unexpected logger type: %T", fl)
+	}
```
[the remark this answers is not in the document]
ka3de: I can see that also, done in [1c23598](https://github.com/grafana/xk6-browser/pull/842/commits/1c23598faecc792a873b2281beaf983eb35fcfce). Let me know about the specific log error message.

tests/test_browser.go
```diff
@@ -105,10 +105,14 @@ func newTestBrowser(tb testing.TB, opts ...any) *testBrowser {
...
 		lc         *logCache
+		err        error
 	)
 
 	if enableLogCache {
-		lc = attachLogCache(state.Logger)
+		lc, err = attachLogCache(state.Logger)
+		if err != nil {
+			tb.Fatalf("testBrowser: attaching logger: %v", err)
+		}
```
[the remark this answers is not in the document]
ka3de: Done in [1c23598](https://github.com/grafana/xk6-browser/pull/842/commits/1c23598faecc792a873b2281beaf983eb35fcfce).

tests/logrus_hook.go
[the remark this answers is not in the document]
inancgumus: It's fine as it is. Thanks for applying my suggestion 👍

PR #843, opened 2023-04-03
Remove the build github action

ankur22:
xk6-browser is now an experiment module within k6. This means that we no longer need to build a dedicated version of xk6-browser binary and anyone wishing to download a pre-built binary should head over to the release section of the k6 repo.


ka3de:
I think we can also remove the [build-release-sh](https://github.com/grafana/xk6-browser/blob/e7174062ccd51ddb19022a094663feca544fb999/build-release.sh) file as it was only referenced from the `build` GH action.

ka3de (approved):
LGTM.

inancgumus (commented):
LGTM but why does it fail? :)

ankur22:
> LGTM but why does it fail? :)

The fix for the build is in https://github.com/grafana/xk6-browser/pull/842.

inancgumus (approved):

PR #846, opened 2023-04-04
Fix metrics registry incompatibility with k6

ka3de:
An incompatibility was detected with k6 after grafana/k6#2999. This commit updates k6 dependency to latest version and fixes the present incompatibility, which mainly affects structures defined in k6 browser tests.


inancgumus (approved):
🙌

ankur22 (approved):
LGTM 🙂

PR #850, opened 2023-04-07
Remove multierror and refactor to Go errors.Join

### LINKED ISSUE #738: Replace custom multierror implementation
As part of the [tasks](https://github.com/grafana/k6-cloud/issues/752) done prior to the merge with k6 core in #734 we got rid of Hashicorp's [multierror](https://github.com/hashicorp/go-multierror) dependency in favor of our custom implementation that fulfilled our current use case.

This implementation should be replaced by something simpler, and we should also reevaluate if there is a need for `multierror` in `remote_object` implementation, which currently is the only component that uses it, or overall in xk6-browser codebase.
A possible candidate to replace the current implementation would be to use Go's support for [wrapping multiple errors](https://tip.golang.org/doc/go1.20#errors) introduced in v1.20.

inancgumus:
This should be merged after upgrading Go's previous version on CI (currently 1.19) to 1.20. Currently, the tests fail because of that.

Updates #738.

### DIFF
--- a/common/remote_object.go
+++ b/common/remote_object.go
@@ -3,6 +3,7 @@ package common
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math"
 	"regexp"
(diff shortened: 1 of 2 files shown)


ka3de (commented):
It's great to get rid of this code! :clap: 
One question. Should we update `go.mod`'s go version to `1.20` besides the go versions in CI also?

ankur22 (approved):
Nice!

inancgumus:
> Should we update go.mod's go version to 1.20 besides the go versions in CI also?

@ka3de, that makes sense. But k6 still uses Go 1.19 in `go.mod` 🤔

inancgumus:
@ankur22 I'll remove the last commit after merging #1216.

ankur22 (approved):
LGTM

ankur22 (approved):
🚀

Issue #831, opened 2023-03-16
Evaluate browser metrics against webvital metrics


ankur22:
- Ensure we’re not working with duplicate metrics (prefixed browser_). We should gather the measurements from one source and not both for the same metric.
- We need to take a look at the existing browser metrics and evaluate whether they are still needed when we have the core webvitals.
- Work with Faro team to ensure that our core webvitals measurement values are the same as the webvitals measured by Faro.

Current browser and webvital metrics from the summary at the end of the test:

```bash
    browser_dom_content_loaded.......: avg=108.7ms  min=10µs     med=952µs    max=802.96ms p(75)=1.88ms   p(90)=330.51ms p(95)=566.73ms
     browser_first_meaningful_paint...: avg=1.27s    min=132.07ms med=1.27s    max=2.41s    p(75)=1.84s    p(90)=2.18s    p(95)=2.3s    
     browser_first_paint..............: avg=122.45ms min=112.84ms med=122.45ms max=132.07ms p(75)=127.26ms p(90)=130.15ms p(95)=131.11ms
     browser_loaded...................: avg=312.03ms min=74µs     med=1.73ms   max=1.39s    p(75)=357.65ms p(90)=933.52ms p(95)=1.16s   
     data_received....................: 18 MB 3.2 MB/s
     data_sent........................: 57 kB 10 kB/s
     http_req_connecting..............: avg=6.55ms   min=0s       med=0s       max=176ms    p(75)=0s       p(90)=29.4ms   p(95)=37.09ms 
     http_req_duration................: avg=129.33ms min=341µs    med=120.22ms max=476.59ms p(75)=212.86ms p(90)=241.5ms  p(95)=281.22ms
     http_req_receiving...............: avg=75.53ms  min=0s       med=50.5ms   max=440ms    p(75)=117ms    p(90)=172.1ms  p(95)=192.55ms
     http_req_sending.................: avg=0s       min=0s       med=0s       max=0s       p(75)=0s       p(90)=0s       p(95)=0s      
     http_req_tls_handshaking.........: avg=3.92ms   min=0s       med=0s       max=90ms     p(75)=0s       p(90)=18.3ms   p(95)=22ms    
     http_reqs........................: 130   23.199591/s
     iteration_duration...............: avg=5.6s     min=5.6s     med=5.6s     max=5.6s     p(75)=5.6s     p(90)=5.6s     p(95)=5.6s    
     iterations.......................: 1     0.178458/s
[...]

PR #851, opened 2023-04-12
Remove existing browser namespaced metrics

ankur22:
Now that we have web vitals, we have evaluated the existing browser metrics:

- [browser_first_paint](https://github.com/grafana/xk6-browser/issues/831#issuecomment-1496313265)
- [browser_dom_content_loaded](https://github.com/grafana/xk6-browser/issues/831#issuecomment-1497094423)
- [browser_loaded](https://github.com/grafana/xk6-browser/issues/831#issuecomment-1497100020)

Our conclusion is that none of these metrics gives us more value or answers questions that we don't already have answers for from the web vitals. The web vitals concentrate on measuring and showing users how useable the website it, rather than these existing metrics which only show us how long it took to download the html file, it's dependencies and when the first element was rendered (usually the background colour).

closes: https://github.com/grafana/xk6-browser/issues/831

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -12,7 +12,6 @@ import (
 	"github.com/grafana/xk6-browser/log"
 
 	k6modules "go.k6.io/k6/js/modules"
-	k6metrics "go.k6.io/k6/metrics"
 
 	"github.com/chromedp/cdproto/cdp"
 	"github.com/chromedp/cdproto/network"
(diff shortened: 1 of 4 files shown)


inancgumus (approved):
Less code! 🥳

ka3de (approved):
LGTM.

Issue #736, opened 2023-01-27
`waitFor` on `Locator` errors when waiting on `detached`


ankur22:
### Brief summary

We have a scenario where we're trying to wait for a given selector to be attached then detached in succession; the application is a SPA.

Has there been any progress into https://github.com/grafana/xk6-browser/issues/472? Waiting for the locator to be attached appears to be working just fine, but waiting for detached always throws some sort of error.

Tested against: [dbede12](https://github.com/grafana/xk6-browser/commit/dbede120c63df43995813a847a25b0e66e289592)

The waitFor method on locator only seems to work with the visible state.

In the below code it works against localhost:8080, which is a test server that can be found here with instructions on how to run it locally: https://github.com/ankur22/testserver.

```go
import { sleep } from 'k6';
import launcher from 'k6/x/browser';

export default function () {
  const browser = launcher.launch('chromium', {
    headless: false,
  });
  const context = browser.newContext();
  const page = context.newPage();
  const res = page.goto('http://localhost:8080/other');
  const l = page.locator('#input-text-hidden');
  l.waitFor({
    state: 'hidden',
  });
  const l2 = page.locator('#input-text-test');
  l2.fill('wait complete');

  sleep(5);
  browser.close();
}
```

In the test above against the test page, the expected output is that wait complete is filled in one of the input text boxes, but this doesn't happen and instead times out.

### xk6-browser version

https://github.com/grafana/xk6-browser/commit/dbede120c63df43995813a847a25b0e66e289592

### OS

Unknown

### Chrome version

Unknown

### Docker version and image (if applicable)

NA

### Steps to reproduce the problem

See summary

### Expected behaviour

Waiting for the `locator` to be detached works.

### Actual behaviour

Waiting for the `locator` to be detached returns an error.

inancgumus:
@ankur22, the test script is for an old version:

```js
  const browser = launcher.launch('chromium', {
    headless: false,
  });
```

Shouldn't it be the following?

```js
import { chromium } from 'k6/x/browser';
...
const browser = chromium.launch({
  headless: false,
});
```

---

With the script fixed, I tested this using the testserver and received the following error:

```bash
ERRO[0030] GoError: waiting for "#input-text-hidden": websocket: close 1006 (abnormal closure): unexpected EOF
        at reflect.methodValueCall (native)
        at file:///Users/inanc/grafana/k6browser/tests/issue736.js:12:12(32)
```

Is this the error you mentioned in the issue?

---

[...]

Issue #472, opened 2022-07-27
The `waitFor` method on `locator` only seems to work with the `visible` state.


ankur22:
Tested against: https://github.com/grafana/xk6-browser/commit/dbede120c63df43995813a847a25b0e66e289592

The `waitFor` method on `locator` only seems to work with the `visible` state.

In the below code it works against `localhost:8080`, which is a test server that can be found here with instructions on how to run it locally: https://github.com/ankur22/testserver.

```javascript
import { sleep } from 'k6';
import launcher from 'k6/x/browser';

export default function () {
  const browser = launcher.launch('chromium', {
    headless: false,
  });
  const context = browser.newContext();
  const page = context.newPage();
  const res = page.goto('http://localhost:8080/other');
  const l = page.locator('#input-text-hidden');
  l.waitFor({
    state: 'hidden',
  });
  const l2 = page.locator('#input-text-test');
  l2.fill('wait complete');

  sleep(5);
  browser.close();
}
```

In the test above against the test page, the expected output is that `wait complete` is filled in one of the input text boxes, but this doesn't happen and instead times out.

PR #852, opened 2023-04-13
Fix Locator.WaitFor for `detached` and `hidden` states

ka3de:
This PR fixes the `Locator.WaitFor` functionality when the state defined is either `hidden` or `detached`. Previously, waiting for these two states would result in an infinite pooling loop and eventually a TO error.

Besides the integration tests included in the PR, this changes can be tested also using this [test browser](https://github.com/ankur22/testserver) implemented by @ankur22 and executing the following script:
```js

        headless: false,
    })

    try {
        await page.goto('http://localhost/other', { waitUntil: 'networkidle' })
        
        page.locator("#attach-detach").waitFor({state: "detached"});
        console.log("detached");

        page.locator('#input-text-hidden').waitFor({state: "hidden"});
        console.log("hidden");
    } finally {
        page.close()
        browser.close()
    }
}
```
Take into account that **this test will require user interaction** to click on the `Detach` button on the top left of the page in order to remove the element that we are waiting for from the DOM.

Closes: #736.
Closes: #472.

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -490,6 +490,18 @@ func (f *Frame) waitForSelector(selector string, opts *FrameWaitForSelectorOptio
 	return handle, nil
 }
 
+func (f *Frame) waitFor(selector string, opts *FrameWaitForSelectorOptions) error {
+	f.log.Debugf("Frame:waitFor", "fid:%s furl:%q sel:%q", f.ID(), f.URL(), selector)
+
+	document, err := f.document()
+	if err != nil {
+		return err
+	}
+
+	_, err = document.waitForSelector(f.ctx, selector, opts)
+	return err
+}
+
 // AddScriptTag is not implemented.
 func (f *Frame) AddScriptTag(opts goja.Value) {
 	k6ext.Panic(f.ctx, "Frame.AddScriptTag() has not been implemented yet")
--- a/common/js/injected_script.js
+++ b/common/js/injected_script.js
@@ -889,7 +889,7 @@ class InjectedScript {
         case "attached":
           return element ? element : continuePolling;
         case "detached":
-          return !element ? undefined : continuePolling;
+          return !element ? true : continuePolling;
         case "visible":
           return visible ? element : continuePolling;
         case "hidden":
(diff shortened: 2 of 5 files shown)


ankur22 (approved):
Nicely done! Thanks for adding the succinct commit messages which helped explain the change.

My only feedback is on what i've already commented on regarding whether to extend the `waitForSelector` to add a new behaviour. If possible, it would be good to separate the behaviour for `waitFor` so that it's clear that it doesn't require the `elementHandle`.

ka3de:
> Agreed. Is this something that we can tackle now or maybe best suited for another PR?

If you mean deleting this line specifically from the `injected_script.js`, I can take care of that in this same PR if you agree. As I mentioned, I think it can be misleading considering that waiting for a `detached` element is valid, and in that case the user would see that log line.

> This is a tricky situation if we try to work with the current methods we have. Since locator.waitFor doesn't have a return value, could we create a new waitFor method in frame.go so that we don't check whether handle is nil?

It can make sense, although we would have the same situation in `ElementHandle.waitForSelector` [here](https://github.com/grafana/xk6-browser/blob/257bc9885bdc27da99cbe6b7f2206576fae88391/common/element_handle.go#L697).
Should we add a new `waitFor` method at the `Frame` level but keep the `waitForSelector` at the `ElementHandle` level? I believe that's a good compromise to not "duplicate" too much code but have more clear "decision branches" in the code. WDYT?

inancgumus (approved):
Nice fix 👍

ankur22:
> If you mean deleting this line specifically from the `injected_script.js`, I can take care of that in this same PR if you agree. As I mentioned, I think it can be misleading considering that waiting for a `detached` element is valid, and in that case the user would see that log line.

Actually, let's leave it for now... we might lose some valuable information if we delete this log.

> See the following diff for the changes to apply from this PRs current version:

Yep, that's exactly what i was trying to get across, nice 👍

ka3de:
> Actually, let's leave it for now... we might lose some valuable information if we delete this log.

:+1: 

> Yep, that's exactly what i was trying to get across, nice +1

Ok! Let me apply these changes, I think you are right,  they simplify the implementation at the `Frame` level.

ankur22 (approved):
LGTM

ka3de:
Rebasing main and merging.

PR #866, opened 2023-05-02
Disable timeout on browser process execution

### LINKED ISSUE #864: Do not apply browser `timeout` option as default navigation timeout
Currently the timeout option defined at the browser level applies to two different functionalities:
- Sets the TO to apply for some "browser initialization" actions, such as TO when launching the browser instance.
- Sets the default navigation TO. This can be overwritten at the browserContext and page levels.

This double purpose is wrong, as a the use cases are quite different and a single value for both does not make sense. At the same time, though, we still believe that defining a TO for the browser initialization actions provides value and is required. Taking this into account our decisions have been:
- The browser level timeout option will apply only for the browser initialization actions. Not for the default navigation TO.
- The browser level timeout option will be moved to be defined as an ENV variable, instead of being defined at the script level.
- The default navigation TO will be set by our own implementation with an initial sensible value. This value will be able to be overwritten at the browserContext and page levels, as it's currently possible.

Related: https://github.com/grafana/xk6-browser/issues/856.

### LINKED ISSUE #865: `browser.timeout` option incorrectly applies as the global test timeout
### Brief summary

Currently (v0.9.0 commit 480c0c1) the `BrowserType.Launch` method accepts a `timeout` option which is defined in the documentation as:
```
timeout | string | '30s' | Default timeout to use for various actions and navigation.
```
But in practice this `timeout` also behaves as the total test timeout. This is because during the browser initialization process, in `BrowserType.allocate` [here](https://github.com/grafana/xk6-browser/blob/480c0c1f74ef377570be4a6a720ddd7a1cd447e7/chromium/browser_type.go#L243), a new `context` with timeout is created, based on the `browser.options.timeout`. Eventually, this `context` is passed into the browser process "execution command" [here](https://github.com/grafana/xk6-browser/blob/480c0c1f74ef377570be4a6a720ddd7a1cd447e7/common/browser_process.go#L157), which in practice will kill the process once the `context` is Done, which will happen after the set `timeout` elapses.

This `context` is also misused when passed into `BrowserProcess.handleClose` [here](https://github.com/grafana/xk6-browser/blob/480c0c1f74ef377570be4a6a720ddd7a1cd447e7/common/browser_process.go#L64) during browser process initialization. In this case the `BrowserProces.cancel()` [action](https://github.com/grafana/xk6-browser/blob/480c0c1f74ef377570be4a6a720ddd7a1cd447e7/common/browser_process.go#L101) does not finish the test, but is still a misleading action based on this `context` timeout.

### xk6-browser version

v0.9.0

### OS

Ubuntu 20.04.5 LTS

### Chrome version

109.0.5414.119 (Official Build) (64-bit)

_No response_

Execute the following tests:

```js
import { chromium } from 'k6/experimental/browser';
import { sleep } from 'k6';

    headless: true,
    timeout: 2000,
  });

  try {
    sleep(3);
  } finally {
    page.close();
    browser.close();
  }
}
```

```js
import { chromium } from 'k6/experimental/browser';
import { sleep } from 'k6';

    headless: true,
    // No timeout set defaults to 30s
  });

  try {
    sleep(31);
  } finally {
    page.close();
    browser.close();
  }
}
```

### Expected behaviour

The test finishes correctly after ~3s and ~31s respectively.

### Actual behaviour

The test finishes with error messages:
```shell
ERRO[0004] communicating with browser: websocket: close 1006 (abnormal closure): unexpected EOF  category=cdp elapsed="0 ms" goroutine=101
ERRO[0004] process with PID 15009 unexpectedly ended: signal: killed  category=browser elapsed="5 ms" goroutine=95
ERRO[0005] Uncaught (in promise) GoError: internal error while removing binding from page: websocket: close 1006 (abnormal closure): unexpected EOF
running at github.com/grafana/xk6-browser/api.Page.Close-fm (native)
default at file:///<path/to/script>/script.js:22:4(29)  executor=per-vu-iterations scenario=default
```
due to the process being killed "by force" when browser timeout finishes.

ka3de:
This PR modifies the browser launching/connecting process to not pass a context with timeout to the actual browser process cmd handle and/or implementation closing handle. Currently this was done using as timeout the value set for the browser.timeout option, which should serve a different purpose.

Instead, it makes the browser process handle and its [closing handling method](https://github.com/grafana/xk6-browser/blob/480c0c1f74ef377570be4a6a720ddd7a1cd447e7/common/browser_process.go#L64) depend on the VU context.

Closes #865.
Closes #864.

### DIFF
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -120,7 +120,7 @@ func (b *BrowserType) Connect(wsEndpoint string, opts goja.Value) api.Browser {
 func (b *BrowserType) connect(
 	ctx context.Context, wsURL string, opts *common.LaunchOptions, logger *log.Logger,
 ) (*common.Browser, error) {
-	browserProc, err := b.link(ctx, wsURL, opts, logger)
+	browserProc, err := b.link(ctx, wsURL, logger)
 	if browserProc == nil {
 		return nil, fmt.Errorf("connecting to browser: %w", err)
 	}
@@ -140,10 +140,9 @@ func (b *BrowserType) connect(
 }
 
 func (b *BrowserType) link(
-	ctx context.Context, wsURL string,
-	opts *common.LaunchOptions, logger *log.Logger,
+	ctx context.Context, wsURL string, logger *log.Logger,
 ) (*common.BrowserProcess, error) {
-	bProcCtx, bProcCtxCancel := context.WithTimeout(ctx, opts.Timeout)
+	bProcCtx, bProcCtxCancel := context.WithCancel(ctx)
 	p, err := common.NewRemoteBrowserProcess(bProcCtx, wsURL, bProcCtxCancel, logger)
 	if err != nil {
 		bProcCtxCancel()
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -240,7 +240,7 @@ func (b *BrowserType) allocate(
 	flags map[string]any, env []string, dataDir *storage.Dir,
 	logger *log.Logger,
 ) (_ *common.BrowserProcess, rerr error) {
-	bProcCtx, bProcCtxCancel := context.WithTimeout(ctx, opts.Timeout)
+	bProcCtx, bProcCtxCancel := context.WithCancel(ctx)
 	defer func() {
 		if rerr != nil {
 			bProcCtxCancel()


ankur22 (approved):
LGTM

inancgumus (approved):

PR #872, opened 2023-05-03
Remove support for browser proxy option

ka3de:
A decision was made to remove proxy from browser option as, even though the option is parsed and accepted, it has no implication on the rest of the code as we are missing the actual implementation. Therefore the option will not be parsed from now on until we complete the implementation so it can have an effect on k6 browser execution.


ankur22 (approved):
Nice! LGTM 😃

PR #881, opened 2023-05-12
Add request failed metric

ankur22:
### Description of changes

When a request completes we want to measure how many of these failed. This is based on the status code of the response. If the code is outside of the 200-399 range, it is deemed to have failed. The default status code range was taken from k6.

The opposite is also possible to determine, we can work out which requests succeeded since the value of the metric will be 0.

We're using a custom metric named `browser_http_req_failed`, instead of the builtin k6 `http_req_failed`. There was a concern that if the browser module worked with the builtin k6 metric, then the results would be skewed for both modules and therefore meaningless. This is due to the fact that the implementation of http could differ. Users are also likely to want to understand their API http failures differently to their browser level http failures.

The shape of the data is free to be moulded the best way that works for the browser module use case. I tested to see whether the [cache](https://github.com/grafana/xk6-browser/blob/015ee2e95fc1e70ba3c7a4ec310c86c43c55fd7a/common/network_manager.go#L220-L222) values varied, which they did, so I've left them in. We can adjust the metrics and labels as we see fit once we have a better idea of what users require. 

### Testing

To test this change, influxdb will need to installed or another suitable backend time series database. Once installed, run a test that you know will navigate to a website which has failing dependencies (e.g. an image that doesn't load). You should see something like this in the `http_req_failed` table:

Screenshot 2023-05-12 at 14 11 27

You should also see metric in the end summary:

```bash
browser_http_req_failed............: 0.00%  ✓ 0         ✗ 218
```

### DIFF
--- a/common/network_manager.go
+++ b/common/network_manager.go
@@ -265,7 +265,7 @@ func (m *NetworkManager) emitResponseMetrics(resp *Response, req *Request) {
 					Time:       wallTime,
 				},
 				{
-					TimeSeries: k6metrics.TimeSeries{Metric: state.BuiltinMetrics.HTTPReqFailed, Tags: tags},
+					TimeSeries: k6metrics.TimeSeries{Metric: m.customMetrics.BrowserHTTPReqFailed, Tags: tags},
 					Value:      failed,
 					Time:       wallTime,
 				},
--- a/common/network_manager.go
+++ b/common/network_manager.go
@@ -178,6 +178,7 @@ func (m *NetworkManager) emitResponseMetrics(resp *Response, req *Request) {
 		fromCache, fromPreCache, fromSvcWrk bool
 		url                                 = req.url.String()
 		wallTime                            = time.Now()
+		failed                              float64
 	)
 	if resp != nil {
 		status = resp.status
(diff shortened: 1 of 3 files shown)


common/network_manager.go
```diff
@@ -187,6 +190,11 @@ func (m *NetworkManager) emitResponseMetrics(resp *Response, req *Request) {
 		fromSvcWrk = resp.fromServiceWorker
 		wallTime = resp.wallTime
 		url = resp.url
+		// Assuming that a failure is when status
+		// is between 200 and 399 (inclusive).
```
ka3de, line 194:
```suggestion
		// Assuming that a failure is when status
		// is not between 200 and 399 (inclusive).
```

ka3de (commented):
Just a small comment. Also there is a conflict in `k6ext/metrics.go`, but the rest, as for [10c9708](https://github.com/grafana/xk6-browser/pull/881/commits/10c97080097d26555a8178832dde440cb1cf6b23) and [015ee2e](https://github.com/grafana/xk6-browser/pull/881/commits/015ee2e95fc1e70ba3c7a4ec310c86c43c55fd7a) LGTM.

dgzlopes (approved):

ka3de (approved):
LGTM.

inancgumus (approved):
👍

Issue #884, opened 2023-05-15
Rename the browser web vital metrics


ankur22:
Here is an example of the web vital metrics when they're emitted: `webvital_cumulative_layout_shift`

There are three issues with it:
1. It is missing the `browser` namespace as discussed in #883 
2. web vital is two words, so it should be `web_vital`
3. `cumulative_layout_shift` says what it is, which is great, but we will assume that people who are looking to measure and improve their frontend web experience will likely have the context of what these web vital metrics are, so we want to go with the abbreviated name e.g. `cls`. This also helps when it comes to the display of the metrics in the summary.

So we need to rename the web vitals from `webvital_xx_xx_xx` to `browser_web_vital_xxx`.

PR #885, opened 2023-05-15
Rename web vital metrics

ankur22:
### Description of changes

We're renaming them from `webvital_xxx_xxx_xxx` to `browser_web_vital_xxx`. This change benefits us in three ways:
1. Browser metrics will be namespaced with `browser_`;
2. _Webvital_ is actually two words, so _web vital_;
3. We're assuming that users will have the context of web vital metrics, so we can use the abbreviations which also helps with the display of them at the end of the test run.

Closes: https://github.com/grafana/xk6-browser/issues/884

### Checklist
```[tasklist]
```

### Optional tasks
```[tasklist]
- [X] Update k6 documentation -- https://github.com/grafana/k6-docs/pull/1178
```

### DIFF
--- a/examples/multiple-scenario.js
+++ b/examples/multiple-scenario.js
@@ -17,7 +17,7 @@ export const options = {
     },
   },
   thresholds: {
-    webvital_first_contentful_paint: ['max < 1000'],
+    browser_web_vital_fcp: ['max < 1000'],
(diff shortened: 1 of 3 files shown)


ka3de (approved):
LGTM.

Issue #888, opened 2023-05-17
Parse instance scenarios into a flat WS URL list


inancgumus:
The k6-browser can connect to browsers in a round-robin fashion by retrieving data from the `K6_INSTANCE_SCENARIOS` environment variable.

## Problem

Currently, we [consider](https://github.com/grafana/xk6-browser/blob/dc0a4789f06778d03b5c2e5aad17814e1555bcca/k6ext/env.go#L18) a list of WS URLs.

However, instead of receiving a flat list of WS URLs, we will receive the following minified JSON via the `K6_INSTANCE_SCENARIOS` environment variable:

```bash
K6_INSTANCE_SCENARIOS="[{\"id\":\"one\",\"browsers\":[{\"handle\":\"ws://1...\"},{\"handle\":\"ws://2...\"}]},{\"id\":\"two\",\"browsers\":[{\"handle\":\"ws://3...\"}]}]"
```

The pre-minified JSON looks like the following (where `id` is a scenario ID/name):

```json
[
    {
        "id": "one",
        "browsers": [
            { "handle": "ws://1..." },
            { "handle": "ws://2..." }
        ]
    },
    {
        "id": "two",
        "browsers": [
            { "handle": "ws://3..." }
        ]
    }
]
```

## Solution

Currently, we don't support other browsers than Chromium. And we don't need to respect the scenario details to pool WS URLs in each scenario. So we can parse the source JSON into the following, a flat list of WS URLs by joining all scenarios into one as follows:

```json
[ "ws://1...", "ws://2...", "ws://3..." ]
```

With this, we can still use the same feature set in the k6-browser without changing anything else. And we can be ready when we support other browsers (since we'll be getting WS URLs in scenarios based on environmental details like browser types).

## Suggestion

Change the `IsRemoteBrowser` function to parse the scenario-based JSON data into a flat list of WS URLs.

ankur22:
Now we are reading the WS URL in the `mapBrowserToGoja` function. The scope is per `VU` lifecycle? So we're parsing the map once every `vu` when it's probably ok to parse once and use when we need.

A possible solution to this could be:
1. Parse the ENV variable once. So maybe when we create the `RootModule`?
2. Keep the list of WS URLS in the `RootModule`
3. Pass a reference to the list to the `VUmodule` when called to `NewInstanceModule` (same way we do for `pidRegistry`)
[...]

PR #889, opened 2023-05-17
Refactor isRemoteBrowser check

ankur22:
### Description of changes

This refactors the `IsRemoteBrowser` detection of the code so that it is done once when creating the `RootModule`. We now rely on a `remoteRegistry` which can be accessed by the mapping when a new vu is initialised. Eventually when there are multiple remote chrome instances, this will select one of many ws urls and return it for use in the current vu/iteration.


inancgumus (commented):
Nice, but I remember @ka3de has been refactoring the similar parts in #876?

ankur22:
> Nice, but I remember @ka3de has been refactoring the similar parts in #876?

I'll take a look at that PR first. Thanks for the heads up.

browser/remoteregistry.go
```diff
@@ -0,0 +1,18 @@
...
+import "github.com/grafana/xk6-browser/env"
+
+type remoteRegistry struct {
+	isRemoteBrowser bool
+	wsURL           string
+}
+
+func newRemoteRegistry(envLookup env.LookupFunc) *remoteRegistry {
+	r := &remoteRegistry{}
+	r.wsURL, r.isRemoteBrowser = env.IsRemoteBrowser(envLookup)
```
ankur22, line 12: I've left us relying on `env`.

I would like to repurpose the `env` package to a `remoteRegistry` package, where this file could be merged with the existing `env.go` file. This would help encapsulate `remoteRegistry` so that we only expose `IsRemoteBrowser`, and allow it to retrieve and process the env var without leaking the unexported logic. At the moment anything in the `browser` pkg can access the unexported fields of the `remoteRegistry` struct. WDYT?

browser/remoteregistry.go
```diff
@@ -0,0 +1,18 @@
+package browser
+
+import "github.com/grafana/xk6-browser/env"
+
+type remoteRegistry struct {
+	isRemoteBrowser bool
+	wsURL           string
```
ka3de, line 7: One question.
We currently support the `K6_BROWSER_WS_URL` to be a comma separated list of WS URLS. In that case, the `env.IsRemoteBrowser` method returns one WS URL randomly from the list in each call.
Considering this, shouldn't we modify this method to return always the full list (if it is a list)? And then change the `IsRemoteBrowser` implementation from the new `remoteRegistry` to return one random entry for each call? This way we would be able to round robin across the list entries, otherwise all connections to the remote browser will be done to the same instance, which does not seem to fulfill our use case to support a WS URL list.

which, in that case, are used to round robin through them per each iteration. Shouldn't we change the method `env.IsRemoteBrowser` to return the whole list of WS URLs (in case it is a list) and store that in

browser/remoteregistry.go
ka3de: I think that can make sense yes :+1:

browser/remoteregistry.go
ankur22: This PR intention was to only refactor what we currently have by moving it to the `rootModule`. I'll make the change in this [comment](https://github.com/grafana/xk6-browser/pull/889#discussion_r1197697833), and i think this suggestion may occur naturally though. If it doesn't then i'll take it into account in a subsequent PR for the next iteration that involves https://github.com/grafana/xk6-browser/issues/888.

ankur22: Ah, i get what you're saying now. I'll ensure to make that suggested change 👍

browser/remoteregistry.go
ankur22: Resolved in 0a49751d5998dea4efb9958013c91f15fd76a37b, 670d249561b60eed255174080782a5ab28da6151 and df397e5ba2dee445b06b454408526175b26facc7

inancgumus (approved):
Nice 👏

registry/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package registry provides types and methods that store
+// information that currently span the whole test run.
+package registry
```
ka3de, line 3: I like this!
Considering this. Do you think we should move also the `pidRegistry` implementation to this package?

browser/modulevu.go
```diff
@@ -16,6 +17,7 @@ type moduleVU struct {
 	k6modules.VU
 
 	*pidRegistry
+	remoteRegistry *registry.RemoteRegistry
```
ka3de, line 20: What do you think about just using composition as we do with `pidRegistry`? E.g.:
```suggestion
	*registry.RemoteRegistry
```
I like the idea of making the `moduleVU` group methods that belong to that specific VU/iteration execution and being able to call them directly. E.g.: `vu.IsRemoteBrowser()` instead of `vu.remoteRegistry.IsRemoteBrowser()`. But I understand this is pretty personal opinion.

registry/doc.go
ankur22: Yeah, happy to do this in another PR 👍

ka3de (approved):
LGTM :clap:

PR #902, opened 2023-05-23
Refactor registry back into browser package

ankur22:
### Description of changes

The `browser` package doesn't quite reflect what is does, so initially the change was made to rename it to `module`. This affects the import in k6. For now it's a lot simpler to just keep it as `browser`.

It also makes more sense to have the registry types close to where they're being used, which is in the `browser` package, so `pidRegistry` and `remoteRegistry` have been moved back into a new `registry.go` file in the `browser` package.

Changes based on: https://github.com/grafana/xk6-browser/pull/899#pullrequestreview-1436923320

### Checklist
```[tasklist]
```


module/module.go
```diff
@@ -27,15 +27,15 @@ type (
...
+	// JSModuleInstance represents an instance of the JS module.
+	JSModuleInstance struct {
 		mod *JSModule
 	}
 )
 
 var (
 	_ k6modules.Module   = &RootModule{}
-	_ k6modules.Instance = &ModuleInstance{}
+	_ k6modules.Instance = &JSModuleInstance{}
```
ankur22, line 37: I had to rename this to avoid the stuttering which was also causing a lint error.

browser/module.go
```diff
@@ -27,34 +26,34 @@ type (
 		Version  string
 	}
 
-	// ModuleInstance represents an instance of the JS module.
-	ModuleInstance struct {
+	// JSModuleInstance represents an instance of the JS module.
+	JSModuleInstance struct {
```
ankur22, line 30: I had to rename this to avoid the stuttering which was also causing a lint error.

inancgumus (approved):
👍

ka3de (approved):
LGTM :tada:

PR #904, opened 2023-05-23
Fix race condition in web vital unit test

ankur22:
### Description of changes

This returns when all the conditions are met from within the goroutine so that the expected map is not written to once the context is cancelled.

### Checklist
```[tasklist]
```


ka3de (approved):
LGTM.

Issue #822, opened 2023-03-10
Use k6 LookupEnv instead of os.GetEnv and os.LookupEnv


inancgumus:
k6-core [upgraded to a new API](https://github.com/grafana/k6/pull/2890) in v0.43 to get the environment variables without depending on the operating system. We can switch to it to stay compatible and prevent possible clashes. It can also be helpful in testing.

We can update [`os.Getenv`](https://github.com/search?q=repo%3Agrafana%2Fxk6-browser%20getenv&type=code) and [`os.LookupEnv`](https://github.com/search?q=repo%3Agrafana%2Fxk6-browser%20lookupenv&type=code). _Click on the links to see where we use them_.

**Breaking change:** The following are all the environment variables and options that the k6-browser module supports. The outdated XK6_* will be removed and aligned with the new browser options we have.

```go
// Execution specific.
const (
	// InstanceScenarios is an environment variable that can be used to
	// define the extra scenarios details to use when running remotely.
	InstanceScenarios = "K6_INSTANCE_SCENARIOS"

	// WebSocketURLs is an environment variable that can be used to
	// define the WS URLs to connect to when running remotely.
	WebSocketURLs = "K6_BROWSER_WS_URL"

	// BrowserArguments is an environment variable that can be used to
	// pass extra arguments to the browser process.
	BrowserArguments = "K6_BROWSER_ARGS"

	// BrowserExecutablePath is an environment variable that can be used
	// to define the path to the browser to execute.
	BrowserExecutablePath = "K6_BROWSER_EXECUTABLE_PATH"

	// BrowserEnableDebugging is an environment variable that can be used to
	// define if the browser should be launched with debugging enabled.
	BrowserEnableDebugging = "K6_BROWSER_DEBUG"

	// BrowserHeadless is an environment variable that can be used to
	// define if the browser should be launched in headless mode.
	BrowserHeadless = "K6_BROWSER_HEADLESS"

	// BrowserIgnoreDefaultArgs is an environment variable that can be
	// used to define if the browser should ignore default arguments.
	BrowserIgnoreDefaultArgs = "K6_BROWSER_IGNORE_DEFAULT_ARGS"

	// BrowserGlobalTimeout is an environment variable that can be used
	// to set the global timeout for the browser.
	BrowserGlobalTimeout = "K6_BROWSER_TIMEOUT"
)

// Logging and debugging.
const (
	// EnableProfiling is an environment variable that can be used to
[...]

PR #906, opened 2023-05-24
Parse the scenarios json from the `K6_INSTANCE_SCENARIOS` env var

ankur22:
### Description of changes

`newRemoteRegistry` now needs to parse two env vars. The first is the one that we have already defined, `K6_BROWSER_WS_URL`, the new one is `K6_INSTANCE_SCENARIOS`. `K6_INSTANCE_SCENARIOS` defines a JSON object of the form:

```json
[
    {
        "id": "one",
        "browsers": [
            { "handle": "ws://1..." }
        ]
    },
]
```

Each scenario will generally point to a unique ws url, which is from a unique instance of chrome. Currently we are only working with chrome, and all chrome instances are setup in the same way. In the future we will want to be able to launch different chrome instances each with their own unique characteristics as well as other web browsers.

`K6_INSTANCE_SCENARIOS` takes precedence over `K6_BROWSER_WS_URL`. `K6_BROWSER_WS_URL` could still be useful for debugging purposes.

Closes: https://github.com/grafana/xk6-browser/issues/888

### Checklist
```[tasklist]
```

### DIFF
--- a/browser/registry.go
+++ b/browser/registry.go
@@ -2,6 +2,8 @@ package browser
 
 import (
 	"crypto/rand"
+	"encoding/json"
+	"fmt"
 	"math/big"
 	"strings"
 	"sync"
@@ -73,6 +75,41 @@ func newRemoteRegistry(envLookup env.LookupFunc) *remoteRegistry {
 	r.wsURLs = parts
 
 	return r
+}
+
+// newRemoteRegistry will create a new RemoteRegistry. This will
+// parse the K6_BROWSER_WS_URL env var to retrieve the defined
+// list of WS URLs from the scenarios object.
+//
+// TODO: Name will change at the end of the PR.
+func newRemoteRegistryFromScenarios(envLookup env.LookupFunc) (*remoteRegistry, error) {
+	r := &remoteRegistry{}
+
+	scenariosJSON, isRemote := envLookup("K6_BROWSER_WS_URL")
+	if !isRemote {
+		return r, nil
+	}
+
+	var scenarios []struct {
+		ID       string `json:"id"`
+		Browsers []struct {
+			Handle string `json:"handle"`
+		} `json:"browsers"`
+	}
+
+	err := json.Unmarshal([]byte(scenariosJSON), &scenarios)
+	if err != nil {
+		return nil, fmt.Errorf("parsing K6_BROWSER_WS_URL: %w", err)
+	}
+
+	for _, s := range scenarios {
+		for _, b := range s.Browsers {
+			r.wsURLs = append(r.wsURLs, b.Handle)
+			r.isRemote = true
+		}
+	}
+
+	return r, nil
 }
 
 // isRemoteBrowser returns a WS URL and true when a WS URL is defined,
--- a/browser/registry.go
+++ b/browser/registry.go
@@ -49,23 +49,22 @@ type remoteRegistry struct {
 //
 // K6_BROWSER_WS_URL can be defined as a single WS URL or a
 // comma separated list of URLs.
-func newRemoteRegistry(envLookup env.LookupFunc) *remoteRegistry {
+func newRemoteRegistry(envLookup env.LookupFunc) (*remoteRegistry, error) {
 	r := &remoteRegistry{}
 
 	isRemote, wsURLs, err := checkForScenarios(envLookup)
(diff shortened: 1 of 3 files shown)


browser/registry.go
```diff
@@ -69,10 +85,44 @@ func newRemoteRegistry(envLookup env.LookupFunc) *remoteRegistry {
...
+
+	err := json.Unmarshal([]byte(scenariosJSON), &scenarios)
+	if err != nil {
+		return false, nil, fmt.Errorf("parsing K6_INSTANCE_SCENARIOS: %w", err)
+	}
+
+	var wsURLs []string
+	for _, s := range scenarios {
+		for _, b := range s.Browsers {
+			if b.Handle == "" {
```
inancgumus, line 114: To be on the safe side?

```suggestion
			if strings.TrimSpace(b.Handle) == "" {
```

inancgumus (approved):
LGTM 👍 Thanks!

We're panicking in `New() *RootModule`. And, it might not align with the expectations of the k6-core about how a module should behave when constructing it.

One suggestion to improve the situation is to handle initializations in `NewModuleInstance` where we have a `VU`. This way, we can handle any unexpected issues properly—with a k6 panic. To implement this, we can make use of `sync.Once` for these global variables. 

It might be unnecessary now. We can address this concern later on.

Squashing this PR might help. For example, ["Add Some"](https://github.com/grafana/xk6-browser/pull/906/commits/28f6723044740bb5028b11b97581bfa53a490cfb) 😄

ankur22:
> LGTM 👍 Thanks!
> 
> We're panicking in `New() *RootModule`. And, it might not align with the expectations of the k6-core about how a module should behave when constructing it.
> 
> One suggestion to improve the situation is to handle initializations in `NewModuleInstance` where we have a `VU`. This way, we can handle any unexpected issues properly—with a k6 panic. To implement this, we can make use of `sync.Once` for these global variables.
> 
> It might be unnecessary now. We can address this concern later on.

Yeah, that's something I wasn't sure of, so I just used `panic`. I'll double check that though 👍 

> Squashing this PR might help. For example, ["Add Some"](https://github.com/grafana/xk6-browser/pull/906/commits/28f6723044740bb5028b11b97581bfa53a490cfb) 😄

🤦 thanks!

ka3de (approved):
LGTM.

browser/registry_test.go
```diff
@@ -33,70 +33,110 @@ func TestPidRegistry(t *testing.T) {
...
+			envVarValue: ``,
 		},
 	}
 
 	for _, tc := range testCases {
-		tc := tc
-
 		t.Run(tc.name, func(t *testing.T) {
-			t.Parallel()
+			t.Setenv(tc.envVarName, tc.envVarValue)
```
inancgumus, line 131: @ankur22, thanks for making the latest changes about `panic`, LGTM!

One last suggestion. Haven't we defined the `env.Lookup` func also for testing? 🤔 WDYT about reenabling `t.Parallel` and passing a fake environment value provider function here?

ankur22: I made a trade off here for simplicity over speed. Since the `newRemoteRegistry` now looks for two env vars, the `envLookup` in the tests would also need to be able to handle both env vars, so the simple _"single WS URL"_ would have an `envLookup` method:

```
func(key string) (string, bool) {
  if key != "K6_BROWSER_WS_URL" {
	  return "", false
  }
  return "WS_URL", true
}
```

The speed gain by parallelising this specific unit test doesn't seem worth it.

If there's a strong opinion though i'll refactor the test in another PR, if that's ok?

inancgumus: I understand the motivation, but I'm not sure a simple if case adds much complexity 🤔 I'm fine either way.

ankur22: Thanks @inancgumus. In that case I'll add a note to refactor this test in [#822](https://github.com/grafana/xk6-browser/issues/822) and maybe we can tackle this in the next couple of weeks 🙂

inancgumus: :D Sure.

inancgumus: Fixed [here](https://github.com/grafana/xk6-browser/pull/918/commits/ad81e72a044af49b63bed8918f946966b8740000).

PR #910, opened 2023-05-30
Abstract `browser` and `browserType` from JS API

ka3de:
### Description of changes

This PR abstracts `browser` and `browserType` implementations from the JS API. This will allow reducing the boilerplate code typically present in k6 browser scripts in order to initialize and close the browser object. Instead this will be handled transparently for the user.

With this change we are also moving into a tighter integration with k6 by:
- Enforcing separation of browser scope based on k6 scenarios and iterations.
- Integrating k6 browser options into k6 scenario definitions.

### Checklist
```[tasklist]
```

### Optional tasks
```[tasklist]
- [ ] Update k6 documentation https://github.com/grafana/k6-docs/issues/1194
```


browser/mapping.go
```diff
@@ -708,6 +743,54 @@ func mapBrowser(vu moduleVU, b api.Browser, wsURL string, isRemoteBrowser bool)
...
+		vu.State().Iteration,
+	)
+
+	var (
+		ok  bool
+		err error
+		b   api.Browser
+	)
+
+	if b, ok = vu.getBrowser(id); ok {
```
ankur22, line 765: ~~It looks like we're building a new `browser` object on every iteration, so this would always return `nil` and `false`, no? If that's the case then maybe we don't need this check.~~

Now i see why -- calls to the other browser APIs will retrieve the associated `browser` object (e.g. i can call `newContext` multiple times and only the first call to `newContext` will create the `browser` object).

browser/registry.go
```diff
@@ -90,3 +91,27 @@ func (r *remoteRegistry) isRemoteBrowser() (string, bool) {
 
 	return wsURL, true
 }
+
+// browserRegistry stores browser instances indexed per
+// iteration as identified by VUID-scenario-iterationID.
+type browserRegistry struct {
```
ankur22, line 97: The reason we need the `browserRegistry` in the `RootModule` is because there's no other way to store the references to `browser` and `browserProcess` at the iteration scope, so that we can later retrieve them and do the necessary cleanup, correct?

browser/mapping.go
```diff
@@ -677,28 +676,64 @@ func mapBrowserContext(vu moduleVU, bc api.BrowserContext) mapping {
 }
 
 // mapBrowser to the JS module.
-func mapBrowser(vu moduleVU, b api.Browser, wsURL string, isRemoteBrowser bool) mapping {
+func mapBrowser(vu moduleVU, wsURL string, isRemoteBrowser bool) mapping { //nolint:funlen
 	rt := vu.Runtime()
+	bt := chromium.NewBrowserType(vu)
```
inancgumus, line 681: Question :) Why don't we initialize the browser once in here and then use in the mapped methods below?

browser/mapping.go
```diff
@@ -708,6 +743,54 @@ func mapBrowser(vu moduleVU, b api.Browser, wsURL string, isRemoteBrowser bool)
...
+		b, pid, err = bt.Launch()
+		if err != nil {
+			return nil, err //nolint:wrapcheck
+		}
+		vu.registerPid(pid)
+	}
+
+	vu.setBrowser(id, b)
+
+	go func(ctx context.Context) {
```
inancgumus, line 785: IDK if it's possible, but can we pass the context to `bt.Connect` and `bt.Launch` so that we don't have to track it in a new goroutine here? Launching a goroutine also pollutes the logs and makes it hard to track.

examples/browser_on.js
```diff
@@ -24,30 +24,33 @@ export default function() {
...
-    });
-  });
-
-  check(browser, {
-    'should be connected before ending iteration': browser.isConnected(),
-  });
-
-  // Disconnect from the browser instance.
-  browser.close();
+  // TODO: Reset test after browser.on for 'disconnect' issue is
```
ankur22, line 27: Is it worth creating an issue for this?

ankur22 (approved):
LGTM 👏

chromium/browser_type.go
```diff
@@ -92,7 +92,7 @@ func (b *BrowserType) init(
 }
 
 func (b *BrowserType) initContext() context.Context {
-	ctx := k6ext.WithVU(b.vu.Context(), b.vu)
+	ctx := k6ext.WithVU(context.Background(), b.vu)
```
inancgumus, line 97: Can we initialize this context before calling `initContext` and then calling `initContext(ctx)`? So that we can have an explicit lifecycle, manage it from tests, etc.

inancgumus (commented):
Overall, it's LGTM, thanks! I believe we need more tests for this change. Especially for reusing of browsers: `getOrInitBrowser`. It would also be great if we can have tests for the contexts as they were the primary suspects of earlier bugs.

inancgumus (approved):

browser/registry.go
ka3de: Correct, the necessary cleanup, and be able to reuse it for multiple `browser` method calls in the same iteration.

chromium/browser_type.go
ka3de: Partially related with this comment and [this comment](https://github.com/grafana/xk6-browser/pull/910#discussion_r1210096528), I remember experimenting with passing `context` to `Launch` and `Connect` methods unsuccessfully, because the calling point was other and `ctx` was being closed, but now, calling from the mapped methods this is working, which improves our implementation and allows us to get rid of https://github.com/grafana/xk6-browser/pull/910/commits/b08de580ac67cb56d0ed4634f7c21647a160e007. See https://github.com/grafana/xk6-browser/pull/910/commits/a23399ed896732557fb0a4ced33c46754739b039.
Thanks for bringing this up!

browser/mapping.go
ka3de: We could potentially wait for the VU `context`'s "done" in `bt.Launch` and `bt.Connect` implementations, but, because we have to call `vu.deleteBrowser` once the browser is closed, we would have to export that method and also the iteration ID generation "logic" used as index for the registry to `chromium` pkg, which does not seem ideal either.

What is your take on this?

In any case this goroutine will be removed once we have a reliable event from k6 to wait for at the end of the iteration.

browser/mapping.go
inancgumus: Thanks for the explanation 🙇

chromium/browser_type.go
inancgumus: Nice, thanks! Should we also pass the context to `getOrInitBrowser` and leave the context initialization to `getOrInitBrowser` for the same reason?

browser/mapping.go
inancgumus: Since this is will be removed soon, I don't mind keeping it as it is 👍 Creating an issue about this would help, thanks!

chromium/browser_type.go
ka3de: Yep, I did this kind of naively, my bad, it makes sense. See [4464d33](https://github.com/grafana/xk6-browser/pull/910/commits/4464d33337ac1ff9f65144449f0b53d88f9b233a).

ka3de:
Merging this as it is. We will improve it once we have the k6 event system.

PR #915, opened 2023-05-31
Remove the web vital rate metric time series metrics

ankur22:
### Description of changes

There are multiple reasons for making this change:

1. We're reducing the number of time series.
2. The rating should live as close to the value that it represents.
3. The rating in the summary is noise that doesn't add value unless we break them down by url.
4. This shouldn't increase the cardinality of the metric that we're labelling with the rating. It's a better trade off than having more time series metrics.
5. The summary at the end will be a bit more focused and shorter.

### Test

You should be able to run any example test (e.g. the fillform.js test). The web vital summary output should now look like this (note that we no longer have the rating metrics):

```bash
     browser_web_vital_cls..............: avg=0.000029 min=0        med=0.000029 max=0.000057 p(90)=0.000052 p(95)=0.000055
     browser_web_vital_fcp..............: avg=288.9ms  min=131.79ms med=240ms    max=494.9ms  p(90)=443.92ms p(95)=469.41ms
     browser_web_vital_fid..............: avg=500µs    min=100µs    med=500µs    max=899.99µs p(90)=819.99µs p(95)=859.99µs
     browser_web_vital_inp..............: avg=136ms    min=16ms     med=136ms    max=256ms    p(90)=232ms    p(95)=244ms   
     browser_web_vital_lcp..............: avg=313.35ms min=131.79ms med=313.35ms max=494.9ms  p(90)=458.59ms p(95)=476.74ms
     browser_web_vital_ttfb.............: avg=219.13ms min=105.09ms med=212.29ms max=340ms    p(90)=314.45ms p(95)=327.22ms
```

If you run it against influxdb or another backend you should find the new `rating` label like so:

Screenshot 2023-05-31 at 17 43 56

### Checklist
```[tasklist]
- [ ] Update k6 documentation
```

### DIFF
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -306,11 +306,6 @@ func (fs *FrameSession) parseAndEmitWebVitalMetric(object string) error {
 		return fmt.Errorf("metric not registered %q", wv.Name)
 	}
 
-	metricRating, ok := fs.k6Metrics.WebVitals[k6ext.ConcatWebVitalNameRating(wv.Name, wv.Rating)]
-	if !ok {
-		return fmt.Errorf("metric not registered %q", k6ext.ConcatWebVitalNameRating(wv.Name, wv.Rating))
-	}
-
 	value, err := wv.Value.Float64()
 	if err != nil {
 		return fmt.Errorf("value couldn't be parsed %q", wv.Value)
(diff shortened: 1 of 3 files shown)


common/frame_session.go
```diff
@@ -322,6 +317,8 @@ func (fs *FrameSession) parseAndEmitWebVitalMetric(object string) error {
 		tags = tags.With("url", wv.URL)
 	}
 
+	tags = tags.With("rate", wv.Rating)
```
ka3de, line 320: Should it be `rate` or `rating`?

ankur22: Yeah, `rating` is better, not sure why i chose `rate` considering we're already using `rating` elsewhere. Done in 76f3b0fc45b343ad666fb1ff480ccfb09397d7fe

ka3de: To be honest I was looking for some reference on Web Vitals doc for it, but found none, but still, as you say, we have been using `rating` so it seems more correct :+1:

inancgumus (approved):
LGTM 👍

PR #917, opened 2023-06-01
Add unquoting to the received scenario data

inancgumus:
This is needed because we receive the remote's sanitized (quoted) scenario JSON data.

Related: #906 and #888.


ankur22 (approved):
LGTM 👍

PR #918, opened 2023-06-01
Refactor to k6 lookupEnv and abstract env. var. lookup

inancgumus:
This PR refactors the direct usage of `os.LookupEnv` to k6 `LookupEnv`. Please have a look at the commit descriptions for the changes needed.

These new changes provide enhanced testability and consistency throughout the module.

The following are all the environment variables and options that the k6-browser module supports. The outdated `XK6_*` are removed and aligned with the new browser options we have.

```go
// Execution specific.
const (
	// InstanceScenarios is an environment variable that can be used to
	// define the extra scenarios details to use when running remotely.
	InstanceScenarios = "K6_INSTANCE_SCENARIOS"

	// WebSocketURLs is an environment variable that can be used to
	// define the WS URLs to connect to when running remotely.
	WebSocketURLs = "K6_BROWSER_WS_URL"

	// BrowserArguments is an environment variable that can be used to
	// pass extra arguments to the browser process.
	BrowserArguments = "K6_BROWSER_ARGS"

	// BrowserExecutablePath is an environment variable that can be used
	// to define the path to the browser to execute.
	BrowserExecutablePath = "K6_BROWSER_EXECUTABLE_PATH"

	// BrowserEnableDebugging is an environment variable that can be used to
	// define if the browser should be launched with debugging enabled.
	BrowserEnableDebugging = "K6_BROWSER_DEBUG"

	// BrowserHeadless is an environment variable that can be used to
	// define if the browser should be launched in headless mode.
	BrowserHeadless = "K6_BROWSER_HEADLESS"

	// BrowserIgnoreDefaultArgs is an environment variable that can be
	// used to define if the browser should ignore default arguments.
	BrowserIgnoreDefaultArgs = "K6_BROWSER_IGNORE_DEFAULT_ARGS"

	// BrowserGlobalTimeout is an environment variable that can be used
	// to set the global timeout for the browser.
	BrowserGlobalTimeout = "K6_BROWSER_TIMEOUT"
)

// Logging and debugging.
const (
	// EnableProfiling is an environment variable that can be used to
	// enable profiling for the browser. It will start up a debugging
	// server on ProfilingServerAddr.
	EnableProfiling = "K6_BROWSER_ENABLE_PPROF"

	// ProfilingServerAddr is the address of the profiling server.
	ProfilingServerAddr = "localhost:6060"

	// LogCaller is an environment variable that can be used to enable
	// the caller function information in the browser logs.
	LogCaller = "K6_BROWSER_LOG_CALLER"

	// LogLevel is an environment variable that can be used to set the
	// log level for the browser logs.
	LogLevel = "K6_BROWSER_LOG"

	// LogCategoryFilter is an environment variable that can be used to
	// filter the browser logs based on their category. It supports
	// regular expressions.
	LogCategoryFilter = "K6_BROWSER_LOG_CATEGORY_FILTER"
)
```

### DIFF
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
-	"os"
 	"os/exec"
 	"path/filepath"
 	"sort"
@@ -62,7 +61,7 @@ func (b *BrowserType) init(
 ) (context.Context, *common.BrowserOptions, *log.Logger, error) {
 	ctx = b.initContext(ctx)
 
-	logger, err := makeLogger(ctx)
+	logger, err := makeLogger(ctx, b.envLookupper)
 	if err != nil {
 		return nil, nil, nil, fmt.Errorf("error setting up logger: %w", err)
 	}
@@ -444,12 +443,12 @@ func setFlagsFromK6Options(flags map[string]any, k6opts *k6lib.Options) error {
 }
 
 // makeLogger makes and returns an extension wide logger.
-func makeLogger(ctx context.Context) (*log.Logger, error) {
+func makeLogger(ctx context.Context, envLookup env.LookupFunc) (*log.Logger, error) {
 	var (
 		k6Logger = k6ext.GetVU(ctx).State().Logger
 		logger   = log.New(k6Logger, common.GetIterationID(ctx))
 	)
-	if el, ok := os.LookupEnv("XK6_BROWSER_LOG"); ok {
+	if el, ok := envLookup("XK6_BROWSER_LOG"); ok {
 		if logger.SetLevel(el) != nil {
 			return nil, fmt.Errorf(
 				"invalid log level %q, should be one of: panic, fatal, error, warn, warning, info, debug, trace",
@@ -457,7 +456,7 @@ func makeLogger(ctx context.Context) (*log.Logger, error) {
 			)
 		}
 	}
-	if _, ok := os.LookupEnv("XK6_BROWSER_CALLER"); ok {
+	if _, ok := envLookup("XK6_BROWSER_CALLER"); ok {
 		logger.ReportCaller()
 	}
 
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -448,7 +448,7 @@ func makeLogger(ctx context.Context, envLookup env.LookupFunc) (*log.Logger, err
 		k6Logger = k6ext.GetVU(ctx).State().Logger
 		logger   = log.New(k6Logger, common.GetIterationID(ctx))
 	)
-	if el, ok := envLookup("XK6_BROWSER_LOG"); ok {
+	if el, ok := envLookup(env.LogLevel); ok {
 		if logger.SetLevel(el) != nil {
 			return nil, fmt.Errorf(
 				"invalid log level %q, should be one of: panic, fatal, error, warn, warning, info, debug, trace",
@@ -456,7 +456,7 @@ func makeLogger(ctx context.Context, envLookup env.LookupFunc) (*log.Logger, err
 			)
 		}
 	}
-	if _, ok := envLookup("XK6_BROWSER_CALLER"); ok {
+	if _, ok := envLookup(env.LogCaller); ok {
 		logger.ReportCaller()
 	}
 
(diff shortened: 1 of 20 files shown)


tests/test_browser.go
```diff
@@ -77,6 +80,16 @@ func newTestBrowser(tb testing.TB, opts ...any) *testBrowser {
...
 			samples = opt
+		case withLookupFunc:
+			lookupFunc = func(key string) (string, bool) {
+				v, ok := opt(key)
+				if ok {
+					return v, ok
+				}
+				// return from the real environment lookup function
+				// so that we can debug (or other things) when we want it.
+				return os.LookupEnv(key)
```
ankur22: The behaviour could be quite confusing if an env var is set in the os (maybe for a test script test run) and then the unit/integration tests pick that up. If we don't need this now, then it might be better not to look up the os env.

chromium/browser_type.go
```diff
@@ -179,8 +179,9 @@ func (b *BrowserType) launch(
 	if err != nil {
 		return nil, 0, fmt.Errorf("%w", err)
 	}
+
 	dataDir := &storage.Dir{}
-	if err := dataDir.Make("", flags["user-data-dir"]); err != nil {
+	if err := dataDir.Make(b.tmpdir(), flags["user-data-dir"]); err != nil {
```
ankur22, line 183: Was this a bug? Seems like we weren't working with `TEMPDIR` at until this change 🙂

ankur22 (approved):
LGTM 🙂

chromium/browser_type.go
inancgumus: Most probably 😅

tests/test_browser.go
ka3de: To be honest I find this a little bit confusing also. If I explicitly set the `withLookupFunc` option for `testbrowser` I would expect it to just use the function implementation that I passed, not default to O.S. env lookup if one variable is not set in my "mock" implementation. I think this can lead to confusing executions, but no strong opinion.

env/env.go
```diff
@@ -62,3 +64,21 @@ const (
 
 // LookupFunc defines a function to look up a key from the environment.
 type LookupFunc func(key string) (string, bool)
+
+// EmptyLookup is a LookupFunc that always returns "" and false.
+func EmptyLookup(key string) (string, bool) { return "", false }
+
+// Lookup is a LookupFunc that uses os.LookupEnv.
+func Lookup(key string) (string, bool) { return os.LookupEnv(key) }
```
ka3de, line 72: I wonder if this should be called `OSLookup`?

k6ext/k6test/vu.go
```diff
@@ -57,10 +57,11 @@ func (v *VU) AssertSamples(assertSample func(s k6metrics.Sample)) int {
 // so that the test can read the metrics being emitted to the channel.
 type WithSamplesListener chan k6metrics.SampleContainer
 
-// WithLookupFunc a custom lookup function that returns test values.
-type WithLookupFunc env.LookupFunc
-
```
ka3de: Why did we remove this and now just accept a "direct" `env.LookupFunc`? Isn't it more idiomatic to keep it this way? Did we remove it because it was making declaration too long? E.g: `withLookupFunc(env.ConstLookup(env.BrowserHeadless, "0"))` ?

ka3de (approved):
Great work @inancgumus ! :clap: Made a couple of comments, but overall LGTM.

tests/test_browser.go
inancgumus: If we remove this, it'll be harder to see logs and debug, at least for me 😄 I'm open to other suggestions 🙇 

In practice, we won't pass an environment variable other than `K6_BROWSER_LOG`. So, my other suggestion is to only accept `K6_BROWSER_LOG` here instead of accepting every environment variable. WDYT?

env/env.go
inancgumus: I thought about it too, but I wanted to keep it generic and added documentation instead, saying that it uses `os.LookupEnv`.

k6ext/k6test/vu.go
inancgumus: Idiomatic Go is about being concise. From the Effective Go: _"A helpful doc comment can often be more valuable than an extra long name."_.

`WithLookupFunc` makes the usage unnecessarily longer and causes unnecessary roundtrips between types (`WithLookupFunc` to `env.LookupFunc`). `NewVU` checks the options by their types, and there is an already `env.LookupFunc` type. Instead of using `withLookupFunc` everywhere, I used the documentation to make it clear where to use what and how.

https://github.com/grafana/xk6-browser/blob/e5fbe425bbb3a2c918a5f29c133e350045a4b951/k6ext/k6test/vu.go#L60-L64

Issue #925, opened 2023-06-08
Enforce 1 BrowserContext per iteration


ka3de:
As part of the modifications to be made in order to integrate k6-browser more into the k6 "ecosystem/APIs", we should enforce the usage of a single "concurrent" browser context per iteration. Any interaction requiring more than one active browser context per iteration should be reorganized in order to take advantage of k6 scenarios features.

This issue should address:
- Limit `browser` implementation to hold for a single browser context.
- Remove exposed methods to create new browser contexts.
- Expose a new method in order to configure the iteration's browser context.

One possible outcome for the JS API that addresses these changes would be:
```js
import { browser } from 'k6/x/browser';

export const options = {
  scenarios: {
    ui: {
      executor: 'shared-iterations',
      options: {
        browser: {
            type: 'chromium',
        },
      },
    },
  },
};
  

export default async function () {
  // Note: If set, Browser Context must be set before
  // creating any new page. If not set, the page will
  // use the default Browser Context options
  browser.setupContext({
    isMobile: true,
    viewport: {
      'width': 768,
      'height': 1024,
    },
  })

  const page = browser.newPage();

  // Note: Uncommeting this would produce an error like:
  // "browser context can only be initialized once per iteration"
  // browser.setupContext({
  //   isMobile: false,
  // });

  try {
    await page.goto('https://test.k6.io/')
  } finally {
    page.close();
  } 
}
```
```[tasklist]
### Tasks
- [x] Can we setup properties on the fly (such as context;grantPermissions) with all context properties?
- [x] Limit browser implementation to hold a single browserContext [PR](https://github.com/grafana/xk6-browser/pull/929)
- [x] ~~Prevent more than one context from being created in a single iteration [PR](https://github.com/grafana/xk6-browser/pull/937)~~
- [x] ~~Expose a new method in order to configure the iteration's browser context only when the iteration starts.~~
- [x] Remove exposed methods to create new browser contexts. [PR](https://github.com/grafana/xk6-browser/pull/945)
- [x] Update Type Definitions [Issue](https://github.com/grafana/xk6-browser/issues/955) [PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66111)
- [x] Update k6-docs [PR](https://github.com/grafana/k6-docs/pull/1242)
```

inancgumus:
Please merge this issue's PR to https://github.com/grafana/k6-docs/pull/1208.

ankur22:
[...]

PR #929, opened 2023-06-12
Limit browser implementation to hold a single browserContext

ankur22:
### Description of changes

This PR is one of a few that will be used to help enforce a single `browserContext` per `browser` per iteration. The reason for this change can be found in this directly linked [issue](https://github.com/grafana/xk6-browser/issues/925) and this overarching [issue](https://github.com/grafana/k6-cloud/issues/1077). The summary is that we want to be able to predict the resource requirements of how many browsers a test run may need, and the first step towards this goal is to be able to predict the number of `browserContext`s a test run will need.

At the moment (in this PR) we enforce a single `browserContext` per iteration, which also means we allow the user to `close()` the existing `browserContext` and create a new one with `NewContext`. Any objects still associated to the previous `browserContext` will be closed and the user will not be able to work with those objects. In a later PR we will change this behaviour so once a `browserContext` is setup, no other `browserContext` can be created, even if the existing one is closed.

```js

  // const context2 = context.newContext(); // If uncommented, it would cause an error.

  context.close(); // page cannot be used after this call

  // await page.goto("https://test.k6.io"); // If uncommented, it would cause an error.

  const context2 = context.newContext();

  // await page.goto("https://test.k6.io"); // If uncommented, it would cause an error.

  await page2.goto("https://test.k6.io");
  ...
```

### Checklist
```[tasklist]
- [ ] Update k6 documentation (if relevant)
- [ ] Generate TS definitions (if relevant)
```

### DIFF
--- a/common/browser.go
+++ b/common/browser.go
@@ -50,7 +50,6 @@ type Browser struct {
 	// A *Connection is saved to this field, see: connect().
 	conn connection
 
-	contextMu      sync.RWMutex
 	context        *BrowserContext
 	defaultContext *BrowserContext
 
@@ -136,8 +135,6 @@ func (b *Browser) disposeContext(id cdp.BrowserContextID) error {
 		return fmt.Errorf("disposing browser context ID %s: %w", id, err)
 	}
 
-	b.contextMu.Lock()
-	defer b.contextMu.Unlock()
 	b.context = nil
 
 	return nil
@@ -147,9 +144,6 @@ func (b *Browser) disposeContext(id cdp.BrowserContextID) error {
 // If the browser context is not found, the default BrowserContext is returned.
 // If the existing browser context id doesn't match an error is returned.
 func (b *Browser) getDefaultBrowserContextOrMatchedID(id cdp.BrowserContextID) (*BrowserContext, error) {
-	b.contextMu.RLock()
-	defer b.contextMu.RUnlock()
-
 	if b.context == nil {
 		return b.defaultContext, nil
 	}
@@ -384,10 +378,7 @@ func (b *Browser) onDetachedFromTarget(ev *target.EventDetachedFromTarget) {
 }
 
 func (b *Browser) newPageInContext(id cdp.BrowserContextID) (*Page, error) {
-	b.contextMu.RLock()
-	browserCtx := b.context
-	b.contextMu.RUnlock()
-	if browserCtx == nil || browserCtx.id != id {
+	if b.context == nil || b.context.id != id {
 		return nil, fmt.Errorf("missing browser context: %s", id)
 	}
 
@@ -400,7 +391,7 @@ func (b *Browser) newPageInContext(id cdp.BrowserContextID) (*Page, error) {
 
 	waitForPage, removeEventHandler := createWaitForEventHandler(
 		ctx,
-		browserCtx, // browser context will emit the following event:
+		b.context, // browser context will emit the following event:
 		[]string{EventBrowserContextPage},
 		func(e any) bool {
 			tid := <-targetID
@@ -501,9 +492,6 @@ func (b *Browser) Close() {
 
 // Contexts returns list of browser contexts.
 func (b *Browser) Contexts() []api.BrowserContext {
-	b.contextMu.RLock()
-	defer b.contextMu.RUnlock()
-
 	if b.context == nil {
 		return []api.BrowserContext{}
 	}
@@ -519,10 +507,7 @@ func (b *Browser) IsConnected() bool {
 
 // NewContext creates a new incognito-like browser context.
 func (b *Browser) NewContext(opts goja.Value) (api.BrowserContext, error) {
-	b.contextMu.RLock()
-	browserCtx := b.context
-	b.contextMu.RUnlock()
-	if browserCtx != nil {
+	if b.context != nil {
 		return nil, errors.New("close the existing browser context before creating a new one")
 	}
 
@@ -538,9 +523,7 @@ func (b *Browser) NewContext(opts goja.Value) (api.BrowserContext, error) {
 		k6ext.Panic(b.ctx, "parsing newContext options: %w", err)
 	}
 
-	b.contextMu.Lock()
-	defer b.contextMu.Unlock()
-	browserCtx, err = NewBrowserContext(b.ctx, b, browserContextID, browserCtxOpts, b.logger)
+	browserCtx, err := NewBrowserContext(b.ctx, b, browserContextID, browserCtxOpts, b.logger)
 	if err != nil {
 		return nil, fmt.Errorf("new context: %w", err)
 	}
--- a/common/browser.go
+++ b/common/browser.go
@@ -142,17 +142,12 @@ func (b *Browser) disposeContext(id cdp.BrowserContextID) error {
 
 // getDefaultBrowserContextOrMatchedID returns the BrowserContext for the given browser context ID.
 // If the browser context is not found, the default BrowserContext is returned.
-// If the existing browser context id doesn't match an error is returned.
-func (b *Browser) getDefaultBrowserContextOrMatchedID(id cdp.BrowserContextID) (*BrowserContext, error) {
-	if b.context == nil {
-		return b.defaultContext, nil
+func (b *Browser) getDefaultBrowserContextOrMatchedID(id cdp.BrowserContextID) *BrowserContext {
+	if b.context == nil || b.context.id != id {
+		return b.defaultContext
 	}
 
-	if b.context.id != id {
-		return nil, fmt.Errorf("missing browser context. Have: %s, want: %s", b.context.id, id)
-	}
-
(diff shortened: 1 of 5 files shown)


common/browser.go
```diff
@@ -513,6 +519,13 @@ func (b *Browser) IsConnected() bool {
 
 // NewContext creates a new incognito-like browser context.
 func (b *Browser) NewContext(opts goja.Value) (api.BrowserContext, error) {
+	b.contextMu.RLock()
+	browserCtx := b.context
+	b.contextMu.RUnlock()
+	if browserCtx != nil {
+		return nil, errors.New("close the existing browser context before creating a new one")
```
ankur22, line 526: Let me know how you feel about this. I could change this in this PR (was going to do it in a later PR) so that we don't allow more than one `browserContext` to be setup as we originally [wanted](https://github.com/grafana/k6-cloud/issues/1077#issuecomment-1570001658).

common/browser.go
```diff
@@ -50,8 +50,7 @@ type Browser struct {
 	// A *Connection is saved to this field, see: connect().
 	conn connection
 
-	contextsMu     sync.RWMutex
```
ankur22, line 53: I've done a stress test on this mutex removal change, and it all seems to work still. I think it's safe to remove since as far as i can tell there's no need to sync the `browserContext` between goroutines. `browser.context` is set when `browser.NewContext` is called, and unset when `browser.Close` is called.

ankur22:
I haven't added a docs issue to this PR, as there are still two more PRs to create (as directed by the original [issue](https://github.com/grafana/xk6-browser/issues/925)). Once the final PR is done I will work on the docs.

ka3de (commented):
Overall LGTM :+1: Just a couple of small comments.

common/browser.go
```diff
@@ -137,23 +136,29 @@ func (b *Browser) disposeContext(id cdp.BrowserContextID) error {
...
+func (b *Browser) getDefaultBrowserContextOrMatchedID(id cdp.BrowserContextID) (*BrowserContext, error) {
+	b.contextMu.RLock()
+	defer b.contextMu.RUnlock()
+
+	if b.context == nil {
+		return b.defaultContext, nil
+	}
+
+	if b.context.id != id {
+		return nil, fmt.Errorf("missing browser context. Have: %s, want: %s", b.context.id, id)
```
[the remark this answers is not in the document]
ankur22: Good question, I'm not sure what the right way forward is. When i ran stress tests it didn't panic.

I prefer to be more defensive at this early stage of this change, but we're in a good place to be able to test this change in the next several weeks and if it panics we can investigate and switch to a warning.

WDYT?

common/browser.go
inancgumus: 😱🤞🙃

common/browser.go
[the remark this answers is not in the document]
inancgumus: I agree with you both. By the way, could we make this error message slightly better (all lowercase and like a sentence) as follows? I believe the `getDefaultBrowserContextOrByID()` name was fine, but I'm fine with both.

```suggestion
		return nil, fmt.Errorf("missing browser context %s, current context is %s", id, b.context.id)
```

common/browser.go
```diff
@@ -225,10 +224,12 @@ func (b *Browser) onAttachedToTarget(ev *target.EventAttachedToTarget) {
...
 
-	var (
-		targetPage = ev.TargetInfo
-		browserCtx = b.getDefaultBrowserContextOrByID(targetPage.BrowserContextID)
-	)
+	targetPage := ev.TargetInfo
+
+	browserCtx, err := b.getDefaultBrowserContextOrMatchedID(targetPage.BrowserContextID)
+	if err != nil {
+		k6ext.Panic(b.ctx, "attaching target to browserContext: %v", err)
```
inancgumus, line 231: WDYT? To be in line with other error messages.

```suggestion
		k6ext.Panic(b.ctx, "attaching target to browser context: %v", err)
```

common/browser.go
```diff
@@ -377,10 +378,7 @@ func (b *Browser) onDetachedFromTarget(ev *target.EventDetachedFromTarget) {
 }
 
 func (b *Browser) newPageInContext(id cdp.BrowserContextID) (*Page, error) {
-	b.contextsMu.RLock()
-	browserCtx, ok := b.contexts[id]
-	b.contextsMu.RUnlock()
-	if !ok {
+	if b.context == nil || b.context.id != id {
 		return nil, fmt.Errorf("missing browser context: %s", id)
```
inancgumus, line 382: Do you think we should make this in line with the other error message? https://github.com/grafana/xk6-browser/pull/929/files#r1228001140

tests/keyboard_test.go
```diff
@@ -12,9 +12,8 @@ import (
 )
 
 func TestKeyboardPress(t *testing.T) {
-	tb := newTestBrowser(t)
-
 	t.Run("all_keys", func(t *testing.T) {
+		tb := newTestBrowser(t)
```
inancgumus, line 16: What is the reason for moving a test browser inline? Just wondering :)

inancgumus (commented):
Looks good! Minor skippable suggestions suggestions.

common/browser.go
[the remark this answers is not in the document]
ka3de: > I prefer to be more defensive at this early stage of this change, but we're in a good place to be able to test this change in the next several weeks and if it panics we can investigate and switch to a warning. If we work with warn, then we might miss it and not get around to investigating the issue in time.
> 
> WDYT?

Ok, I'm ok leaving it as error at least by now :+1:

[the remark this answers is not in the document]
ankur22: It didn't take long to `panic` 🙂 So instead i'm switching back to working with the `defaultContext`. It occurs as soon as we try to connect multiple instances of k6 to a single browser.

My guess as to why a `browserContext` cannot be found when we receive a `onAttachedToTarget` event is because each of the k6 instances mirrors the `page`s from all other k6 instances, but we're not mirroring the creation of the `browserContexts`. It would be great if we didn't also have to create the `page` that is associated with a different k6 instance and not the current one 🤔 

I ran a quick experiment where i returned `nil` from `getDefaultBrowserContextOrMatchedID`:

```go
func (b *Browser) getDefaultBrowserContextOrMatchedID(id cdp.BrowserContextID) (*BrowserContext, error) {
	if b.context == nil {
		return nil, nil
	}

	if b.context.id != id {
		return nil, nil
	}

	return b.context, nil
}
```

and the caller would have to check for a nil `browserCtx`:

```go
	if browserCtx == nil {
		return // Ignore requests to attach target for other k6 instances
	}
```

Which actually all worked! So maybe there is a possibility of making this less complex and not needing to track `page`s from other k6 instances in the current k6 instance. I'll create an issue to look into this as it's not a priority for this current PR and issue that it is trying to resolve.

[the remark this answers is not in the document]
inancgumus: > It would be great if we didn't also have to create the page that is associated with a different k6 instance and not the current one 🤔

That is how CDP works, but we can internally put a mechanism to manage that.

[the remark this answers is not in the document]
ankur22: > I'm not sure how returning nil is a better solution 😄 But I agree that making it straightforward to handle all these would be great.
Returning `nil` is just the hack, but essentially we wouldn't need to create a `page` if a `nil` `browserContext` was returned to the caller of `getDefaultBrowserContextOrMatchedID`, but i'm hoping that there are other things we can skip too, but yeah, this requires some deeper investigation.

tests/keyboard_test.go
ankur22: We can't share a `browser` instance in tests now, since we can only create a single `browserContext` in a `browser`, that's the reason for the changes in some of the tests.

inancgumus: Ah, right. Thanks for explaining!

common/browser.go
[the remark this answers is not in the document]
ankur22: Reverted back to original logic: f6109efe8a0c752c6c13a2f6bb4e9e0d581db1ba

common/browser.go
ankur22: I've reverted the code so we don't panic now, thanks for the suggestion though 👍

inancgumus: :) Sure thing. Glad we got panic earlier. 🙇 to you both!

common/browser.go
inancgumus: Thanks!

common/browser.go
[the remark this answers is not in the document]
ka3de: To be honest, I don't think our previous implementation was correct anyway, but having https://github.com/grafana/xk6-browser/issues/933 open will address finding the proper solution for this. Resolving this discussion.

ka3de (approved):
LGTM.

inancgumus (approved):
LGTM 👍

Issue #898, opened 2023-05-22
Remove the `api` package to use concrete `common` types


ankur22:
### Feature Description

In Go, generally, interfaces are small and focused, as well as defined where they're being used. This also helps in testing, where we can define smaller focused interfaces which only mock methods that are used by the test, rather than a large interface where everything needs to be mocked even if majority of the methods aren't being used by the test.

The `api` package was useful when we required the concrete types to comply with the interface since we were returning the interfaces to the goja layer directly. E.g.:

```go
// Ensure Browser implements the EventEmitter and Browser interfaces.
var (
	_ EventEmitter = &Browser{}
	_ api.Browser  = &Browser{}
)
```

Now that we're working with the mapping layer which explicitly maps the implementation to a goja type, this api compliance requirement is no longer required.

When working on recent investigative issues, we've found that this interface api compliance has become a hinderance.

This issue is too explore whether we can remove the `api` package and what the consequences of that would be. Something to consider are:
- Is is straight forward to find and replace the interface type with the concrete type?
- [How the testing of the mapping layer is affected.](https://github.com/grafana/xk6-browser/pull/1006#discussion_r1307244661)
- Will the removal affect future work when working with other web browser?

## POC

https://github.com/grafana/xk6-browser/commits/refactor/api-package-poc

---

```[tasklist]
### Primary tasks
- [x] #935
- [ ] #1055
- [ ] #1057
```

```[tasklist]
### [#1055: Remove the `api` package](https://github.com/grafana/xk6-browser/pull/1055)
- [x] Postfix the `api` types with `API`
- [x] Move the `api` types into the `common` pkg
- [x] Remove the `api` pkg for good
```

```[tasklist]
### [#1057: Refactor the `common` API types to concrete](https://github.com/grafana/xk6-browser/pull/1057)
- [x] Remove the `common` pkg API type dependencies
- [x] Use only concrete  types throughout the module
- [x] Move the interfaces into the mapping layer and unexport
```

### Already existing or connected issues / PRs (optional)

Related issues:
- #271

PR #935, opened 2023-06-14
Refactor integration test browser

inancgumus:
Updates: #898.

This PR improves [the test browser](https://github.com/grafana/xk6-browser/blob/b0c50e06d1e78d7e97ae222c272c05ee998172c3/tests/test_browser.go) into [this one](https://github.com/grafana/xk6-browser/blob/0fc72a91b8ebaeeea92b624c48c2ea6b2b37d7df/tests/test_browser.go) as follows. It's a nice first small step in removing the `api` package, and starting with the core of tests seemed to me an effective idea. While on it, I also made a refactoring that had been on my mind for quite a while.

* Removes the `api` package dependency by using concrete types. [See it here altogether](https://github.com/grafana/xk6-browser/pull/935/files/290b9129e896457bcb3347df0d379cbe9f21ac51..80631844a4fd908463d08f54d0010b124042ebb4).
* Refactors the test browser options to Rob Pike's self-referential functions pattern. [See it here altogether](https://github.com/grafana/xk6-browser/pull/935/files/7eeed2e25d0628b55019bdde96da3b0b0b263ca1..0be2d8873c5e6e0247ac7f1c76441bae6bec6f1b).
* Refactors the options to adjust the test browser behavior directly. I can remove [this change](https://github.com/grafana/xk6-browser/pull/935/files/930640f2131258a9f9e849ce959bfa7be46e34fc..0fc72a91b8ebaeeea92b624c48c2ea6b2b37d7df). That part was an experiment, and it turned out to be useful. Please let me know which one you prefer: `testBrowserOptions` or `testBrowser`.
* Unexports the exported identifiers to keep it self-reliant.
* Makes use of the test browser helpers in the tests.
* Fixes the linter errors.
* Removes remnants.


tests/test_browser.go
```diff
@@ -18,168 +21,159 @@ import (
...
-			samples = opt
-		case env.LookupFunc:
-			lookupFunc = opt
-		}
-	}
-
-	vu := setupHTTPTestModuleInstance(tb, samples)
+	tbr := &testBrowser{t: tb}
+	tbr.applyDefaultOptions()
+	tbr.applyOptions(opts...) // apply pre-init stage options.
```
inancgumus, line 75: We can also join these together like this: `applyDefaultOptions(opts...)` calls `applyOptions()` underneath. In this case, it'd make more sense to name `applyDefaultOptions` to `applyPreInitOptions`.

ankur22 (approved):
LGTM. I left a couple of nit picks comments 🙂 

Would be nice next time if large refactors could be broken down into smaller PRs 😅

tests/test_browser.go
```diff
@@ -423,3 +410,27 @@ func setupHTTPTestModuleInstance(tb testing.TB, samples chan k6metrics.SampleCon
...
+) {
+	tb.Helper()
+
+	vu := setupHTTPTestModuleInstance(tb, opts.samples)
+	registry := k6metrics.NewRegistry()
+	k6m := k6ext.RegisterCustomMetrics(registry)
+	vu.CtxField = k6ext.WithCustomMetrics(vu.Context(), k6m)
+	vu.InitEnvField.LookupEnv = opts.lookupFunc
+
+	ctx, cancel := context.WithCancel(vu.Context())
```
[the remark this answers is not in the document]
inancgumus: `newBrowserTypeWithVU()` is like `context.WithCancel`; similar to returning a context and a cancel function. I thought `tb.Cleanup` makes the main `newTestBrowser()` clear that the browser stops when the test finishes. Otherwise, it becomes buried in `newBrowserTypeWithVU`, which can make it harder to see when it stops. Since `newTestBrowser` manages the lifecycle, also [here](https://github.com/grafana/xk6-browser/blob/0fc72a91b8ebaeeea92b624c48c2ea6b2b37d7df/tests/test_browser.go#L93-L101).

tests/test_browser.go
```diff
@@ -246,18 +247,25 @@ type testBrowserOptions struct {
...
 		samples:    make(chan k6metrics.SampleContainer, 1000),
 		lookupFunc: env.Lookup,
 	}
+	tbo.apply(opts...)
+
+	return tbo
+}
+
+// apply applies the given options to the testBrowserOptions.
+func (tbo *testBrowserOptions) apply(opts ...func(*testBrowserOptions)) {
```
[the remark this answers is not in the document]
inancgumus: As I explain [here](https://github.com/grafana/xk6-browser/blob/0fc72a91b8ebaeeea92b624c48c2ea6b2b37d7df/tests/test_browser.go#L73-L79), some options (i.e. `withLogCache`) require initialized VU to work, and some don't (i.e. `withSamples`).

inancgumus:
@ankur22, thanks for your review 🙇 

> Would be nice next time if large refactors could be broken down into smaller PRs 😅

Sorry for the hassle 😢 I agree there are a lot of commits 😅 While the PR itself isn't that large (just around 300 LOC), I understand that breaking down large refactors into smaller PRs can help with better review and comprehension. We should also think about how to make them atomic while doing so since breaking them down into many PRs can hurt comprehensibility again. I think the issue might be more about the number of commits and how the changes are structured. Let's figure out the best way forward in a small async-meeting. I'm confident we can come up with a solution that works for everyone. Thanks for bringing it up! 🙇

inancgumus:
@ankur22, thanks for your review 🙇 

> Would be nice next time if large refactors could be broken down into smaller PRs 😅

Sorry for the hassle 😢 I understand the challenge with the commit count! While I've added descriptive links to help, they may not be enough 😢 Despite this PR being smaller than usual (about 300 LOC), I acknowledge that breaking larger refactors into smaller atomic PRs could aid comprehension. We also need to ensure clarity is maintained due to too many PRs. How about discussing the best approach in a brief meeting with you? I'm sure we can devise a suitable solution for all. Thanks for your input! 🙇

tests/test_browser.go
```diff
@@ -18,168 +21,159 @@ import (
...
+	metricsCtx := k6ext.WithCustomMetrics(
+		vu.Context(),
+		k6ext.RegisterCustomMetrics(k6metrics.NewRegistry()),
+	)
+	ctx, cancel := context.WithCancel(metricsCtx)
+	vu.CtxField = ctx
+	vu.InitEnvField.LookupEnv = tbr.lookupFunc
+
+	bt := chromium.NewBrowserType(vu)
+	vu.RestoreVUState()
```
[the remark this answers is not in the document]
inancgumus: Changed in a98cfff.

tests/test_browser.go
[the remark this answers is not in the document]
inancgumus: Changed in f5dd191.

k6ext/k6test/vu.go
```diff
@@ -33,9 +33,9 @@ type VU struct {
 // ToGojaValue is a convenience method for converting any value to a goja value.
 func (v *VU) ToGojaValue(i any) goja.Value { return v.Runtime().ToValue(i) }
 
-// MoveToVUContext moves the VU to VU context, adding a predefined k6 lib State and nilling the InitEnv
-// to simulate how that is done in the real k6.
-func (v *VU) MoveToVUContext() {
+// RestoreVUState restores the state field of the VU and nilling the InitEnv field.
+// This is done to simulate how k6 operates.
+func (v *VU) RestoreVUState() {
```
ka3de, line 38: Nit picking: Calling this `Restore` is confusing to me because, for what I understand, what we are doing here is making the VU "advance" from the `Init` state to the `Active`/`Execution` state, where the `vu.State()` is populated.
Personally I would prefer calling it `InitVUState()`, but no strong opinion.

ka3de (approved):
Nice refactor! :clap: Although quite lengthy :sweat_smile: 
I think readability of the `testBrowser` implementation is better now and helper methods are better organized. Thanks!
Just made a small comment.

inancgumus:
> Nice refactor! 👏 Although quite lengthy 😅

Thanks, @ka3de 🙇  You mean the number of commits since our usual PRs are larger? Then again, sorry for the hassle :( I couldn't keep it smaller to make it a meaningful unit of work. I'm happy to discuss this in a short meeting :)

> I think readability of the `testBrowser` implementation is better now and helper methods are better organized. Thanks!
> Just made a small comment.

Good to hear!

k6ext/k6test/vu.go
inancgumus: I agree, it's confusing. Thanks for the suggestion 👍 WDYT about `ActivateVU()` with a comment saying, _"ActiveVU mimicks the activation of VU, just like in k6"_?

ka3de: I still think `InitVUState()` represents better the action that is performed in the method, but `ActivateVU()` looks better to me than `RestoreVUState()`, if you prefer that.

inancgumus: Hmm, ok. I've reconsidered this. But, I concluded that it would be better to name it with the "why" instead of "how" because we can look into the method for the "how". I think I'll change it to `ActivateVU`. Thanks for the suggestion 🙇 :)

See: 48d9d7f.

PR #937, opened 2023-06-14
Add restriction on one context per iteration

ankur22:
### Description of changes

This change will restrict the creation and use of a single `browserContext` per iteration. Once the `browserContext` has been closed, a new one cannot be created in the same iteration. This essentially ends the browser test for the iteration, but allows the user to run other k6 tests.
    
The reason for this change is:
1. restricting the user to work with a single browser context per iteration will restrict the scenario to one type of `browserContext`. If the user wishes to work with different types of `browserContext`s, they should work with `scenarios` to create a new test that changes the type of `browserContext`. This helps separate the concerns of the different types of `browserContext`s.
2. Creating and closing `browserContext`s while also running async APIs will cause errors and issues for the user. They are likely to be a confusing experience. We want to avoid this.
3. The browser module processes background async operations when APIs are called. If a `browserContext` is closed, this will affect how we process those background operations which will inadvertently affect the test and the results.

```js

  // const context2 = context.newContext(); // If uncommented, it will cause an error.

  await page2.goto("https://test.k6.io");

  ...

  // The browser and context cannot be used after this call.
  // This will essentially end the browser test for the current
  // iteration, but allow the use of other k6 APIs.
  context.close();

  // const context2 = context.newContext(); // If uncommented, it will cause an error.

  const res = http.get('https://test.k6.io/news.php');

  check(res, {
    'status is 200': (r) => r.status === 200,
  });
```

### Checklist

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -774,13 +774,13 @@ func getOrInitBrowser(
 		vu.registerPid(pid)
 	}
 
-	vu.setBrowser(id, b)
-
 	go func(ctx context.Context) {
 		<-ctx.Done()
 		b.Close()
 		vu.deleteBrowser(id)
 	}(vu.Context())
+
+	vu.setBrowser(id, b)
 
 	return b, nil
 }
--- a/common/browser.go
+++ b/common/browser.go
@@ -47,6 +47,7 @@ type Browser struct {
 	// A *Connection is saved to this field, see: connect().
 	conn connection
 
+	contextCreated bool
 	context        *BrowserContext
 	defaultContext *BrowserContext
 
@@ -496,8 +497,8 @@ func (b *Browser) IsConnected() bool {
 
 // NewContext creates a new incognito-like browser context.
 func (b *Browser) NewContext(opts goja.Value) (api.BrowserContext, error) {
-	if b.context != nil {
-		return nil, errors.New("existing browser context must be closed before creating a new one")
+	if b.contextCreated {
+		return nil, fmt.Errorf("only one browser context can be setup per iteration")
 	}
 
 	action := target.CreateBrowserContext().WithDisposeOnDetach(true)
@@ -517,6 +518,7 @@ func (b *Browser) NewContext(opts goja.Value) (api.BrowserContext, error) {
 		return nil, fmt.Errorf("new context: %w", err)
 	}
 	b.context = browserCtx
+	b.contextCreated = true
 
 	return browserCtx, nil
 }
--- a/tests/browser_test.go
+++ b/tests/browser_test.go
@@ -27,7 +27,7 @@ func TestBrowserNewPage(t *testing.T) {
 	assert.Equal(t, 1, l, "expected there to be 1 browser context, but found %d", l)
 
 	_, err := b.Browser.NewPage(nil)
-	assert.EqualError(t, err, "new page: existing browser context must be closed before creating a new one")
+	assert.EqualError(t, err, "new page: only one browser context can be setup per iteration")
 
 	err = p1.Close(nil)
 	require.NoError(t, err)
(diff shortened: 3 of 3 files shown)


common/browser.go
```diff
@@ -500,8 +501,8 @@ func (b *Browser) IsConnected() bool {
 
 // NewContext creates a new incognito-like browser context.
 func (b *Browser) NewContext(opts goja.Value) (api.BrowserContext, error) {
-	if b.context != nil {
-		return nil, errors.New("existing browser context must be closed before creating a new one")
+	if b.contextCreated {
```
ankur22, line 504: Unlike the [POC](https://github.com/grafana/xk6-browser/compare/main...poc/1vu-1bc-browser-pool-bctx) where the check to see if a `browserContext` already exists is done in the mapping layer, in this PR it is being done in the business logic. It feels correct to do it here to avoid spilling business logic changes into the mapping layer -- which feels more like a translation layer.

It's worth noting that we can run other `browser` APIs which don't rely on the `browserContext` (such as `browser.version`) before a `browserContext` is created or after it is closed.

WDYT?

browser/mapping.go
```diff
@@ -774,14 +774,14 @@ func getOrInitBrowser(
...
 
-	vu.setBrowser(id, b)
-
 	go func(ctx context.Context) {
 		<-ctx.Done()
 		b.Close()
 		vu.deleteBrowser(id)
 	}(vu.Context())
 
+	vu.setBrowser(id, b)
```
ka3de, line 783: Just for curiosity. What is the reasoning for this change? Can we have an "race condition" between the `ctx.Done` and `vu.deleteBrowser` with the `vu.setBrowser` ?

ka3de (commented):
Overall LGTM. But I have one consideration. Would it make sense to restore the `contextCreated` flag on the `BrowserContext.Close()` method so we allow creating more than one browser context per iteration, but avoid having more than 1 concurrently? Which I believe was actually the goal.

Maybe this is messing things up now, but just wondering if there would be any cons doing that, or if that would allow for a better UX for some uses cases.

browser/mapping.go
ankur22: The reason for the change was to ensure that we still call `deleteBrowser` if a panic occurs in `setBrowser`. I know that should never happen. I just thought that the goroutine setup of the cleanup should be closer to where the browser process was initialised 🙂 There's no race conditions as far as i know.

ankur22:
> Overall LGTM. But I have one consideration. Would it make sense to restore the `contextCreated` flag on the `BrowserContext.Close()` method so we allow creating more than one browser context per iteration, but avoid having more than 1 concurrently? Which I believe was actually the goal.
> 
> Maybe this is messing things up now, but just wondering if there would be any cons doing that, or if that would allow for a better UX for some uses cases.

I think i may have misunderstood the original requirement 😅 I thought we wanted to prevent any more `browserContext`s from being created once the `browserContext` for the current iteration was closed. If that's not the case then this PR isn't needed.

ka3de:
> I think i may have misunderstood the original requirement sweat_smile I thought we wanted to prevent any more browserContexts from being created once the browserContext for the current iteration was closed. If that's not the case then this PR isn't needed.

Well I think the main requirement was that we wanted to have a reliable control of how many concurrent browser contexts are in a test execution. For that, a valid approach is the one implemented here, which limits to a single browser context per iteration, so knowing how many iterations are in a test we know how many browser contexts are concurrently "alive" (max), so I can have a capacity planning if I'm sharing _N_ browser instances.
But, if we consider that in an iteration you can create a browser context, close it and then create a new one, we would still have the same number of concurrent browser contexts during the execution.

But, yes, sorry for messing with this now, let's continue with this implementation as discussed and we can possibly extend the possibilities in the future if there are use cases that require it :+1:

inancgumus (approved):
🚀

ankur22:
NOTE: We're putting this PR on hold for now. We think that we don't need it, and want to work with what we have already which is to allow the sequential creation of `browserContext` per iteration, but still not parallel.

ankur22:
Closing this since we're happy with the current implementation and behaviour.

PR #938, opened 2023-06-16
Fix CI failures

inancgumus:
The solution in this PR still needs to be better, but it's better than the current workflows. When we run it multiple times, tests eventually pass. Feel free to push commits to this PR, if needed.

_PS: While testing this PR, GH has automatically closed the PR multiple times 😄_

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         go-version: [1.18.x]
-        platform: [ubuntu-latest]
+        platform: [ubuntu-latest-8-cores]
     runs-on: ${{ matrix.platform }}
     steps:
       - name: Setup Chrome
@@ -104,7 +104,7 @@ jobs:
       fail-fast: false
       matrix:
         go-version: [1.19.x]
-        platform: [ubuntu-latest]
+        platform: [ubuntu-latest-8-cores]
     runs-on: ${{ matrix.platform }}
     steps:
       - name: Setup Chrome
@@ -158,7 +158,7 @@ jobs:
       fail-fast: false
       matrix:
         go-version: [1.19.x]
-        platform: [ubuntu-latest]
+        platform: [ubuntu-latest-8-cores]
     runs-on: ${{ matrix.platform }}
     steps:
       - name: Setup Chrome
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,7 +55,7 @@ jobs:
           export GOMAXPROCS=1
           args=("-p" "1" "-race")
           export K6_BROWSER_HEADLESS=true
-          go test "${args[@]}" -timeout 5m ./...
+          go test "${args[@]}" -timeout 10m ./...
 
   test-tip:
     runs-on: ubuntu-latest
@@ -97,7 +97,7 @@ jobs:
           export GOMAXPROCS=1
           args=("-p" "1" "-race")
           export K6_BROWSER_HEADLESS=true
-          go test "${args[@]}" -timeout 5m ./...
+          go test "${args[@]}" -timeout 10m ./...
 
   test-current-cov:
     strategy:
@@ -141,7 +141,7 @@ jobs:
                   list=$(echo "$list" | cut -f1 -d ' ' | sort -u | paste -sd, -)
               fi
 
-              go test "${args[@]}" -timeout 5m --coverpkg="$list" -coverprofile=$(echo $pkg | tr / -).coverage $pkg
+              go test "${args[@]}" -timeout 10m --coverpkg="$list" -coverprofile=$(echo $pkg | tr / -).coverage $pkg
           done
           grep -h -v "^mode:" *.coverage >> coverage.txt
           rm -f *.coverage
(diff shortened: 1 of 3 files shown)


inancgumus:
hey

inancgumus:
hey

inancgumus:
open

inancgumus:
sdsadss

tests/browser_test.go
```diff
@@ -217,6 +217,9 @@ func TestBrowserLogIterationID(t *testing.T) {
 }
 
 func TestMultiBrowserPanic(t *testing.T) {
+	// TODO: This test never works on CI, fix it.
+	t.Skip("skipping, fix this")
```
ka3de, line 221: I wonder if we should use an ENV var, set by us in CI, as a flag in order to skip this test or not. In that way at least we will always run this test when testing locally. Do you think that is worth it?

chromium/browser_type.go
```diff
@@ -334,6 +334,7 @@ func parseArgs(flags map[string]any) ([]string, error) {
 func prepareFlags(lopts *common.BrowserOptions, k6opts *k6lib.Options) (map[string]any, error) {
 	// After Puppeteer's and Playwright's default behavior.
 	f := map[string]any{
+		"disable-gpu":                                        true, // fixes CI tests.
```
ka3de, line 337: Out of curiosity, how did you get to this?

ka3de (commented):
Overall I think the changes make sense in any case, although I'm still a little bit confused as to why CI tests started to fail, which makes me doubt whether it is due to a change in GH or a change that we introduced :thinking:

.github/workflows/test.yml
```diff
@@ -21,6 +21,19 @@ jobs:
         platform: [ubuntu-latest]
     runs-on: ${{ matrix.platform }}
     steps:
+      - name: Setup Chrome
```
ankur22, line 24: This is interesting! So we were relying on Chrome being installed on the CI test VMs/Containers up until we started seeing these issues?

.github/workflows/test.yml
```diff
@@ -51,14 +51,9 @@ jobs:
...
-          export GOMAXPROCS=2
-          args=("-p" "2" "-race")
-          # Run with less concurrency on Windows to minimize flakiness.
-          if [[ "${{ matrix.platform }}" == windows* ]]; then
-            unset args[2]
-            args[1]="1"
-            export GOMAXPROCS=1
-          fi
+          # Run with less concurrency to reduce CI flakiness.
+          export GOMAXPROCS=1
```
ankur22, line 55: I'm not sure about this change. I think it makes sense since when we see a flakey tests we don't have the resources to resolve them straight away, and they can be quite demanding tests to fix. At the same time, i think we've found some issues in CI that we don't see in our local test runs, which are generally race conditions or deadlocks. I'll go with what the consensus is on this one 🙂

.github/workflows/test.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           export GOMAXPROCS=1
           args=("-p" "1" "-race")
           export K6_BROWSER_HEADLESS=true
-          go test "${args[@]}" -timeout 5m ./...
+          go test "${args[@]}" -timeout 10m ./...
```
ankur22, line 58: Why do we need such a large timeout?

ankur22 (approved):
LGTM. I think it would be good to get some clarification on the changes so that we're aligned 🙂 

Thanks for tackling this issue!!

ankur22:
> which makes me doubt whether it is due to a change in GH or a change that we introduced

We're seeing this CI issue with our tagged releases too. I've tested `v0.10.0` and `v0.9.0` 😞

chromium/browser_type.go
inancgumus: TBH, I just experimented with it when I see it [here](https://bugs.chromium.org/p/chromium/issues/detail?id=383314), and I don't have the best answer.

.github/workflows/test.yml
inancgumus: Yeah, it's interesting, indeed. It didn't work when I removed it.

.github/workflows/test.yml
ka3de: That's a good point. I wonder if we should have one set of tests specifically run with higher concurrency, or maybe just leave the current setup and expect it to pass in any case.

inancgumus: This is the one that makes it consistently (eh, to some degree) pass. Otherwise, it constantly fails.

.github/workflows/test.yml
inancgumus: Since the concurrency is 1 now, and it was 2 before, I thought it would make sense to double the timeout. When I tested it, tests took about 2x longer.

tests/browser_test.go
inancgumus: Makes sense; I was thinking about the same, too. Especially for the `disable-gpu` one. We can do it whether we're running on Github by checking the `GITHUB_ACTIONS` environment variable.

.github/workflows/test.yml
inancgumus: When I'm testing this PR with several commits (in and out), I noticed, the nature of the tests don't matter for the outcome of the test runs.

inancgumus:
#947 is a better fix. Closing this one. We can reopen it if needed.

Issue #941, opened 2023-06-19
Deadlock in FrameManager's `frameNavigated` when we navigate to certain websites


ankur22:
### Brief summary

A deadlock occurs when we navigate to certain websites, and perform a click action which results in another navigation.

### xk6-browser version

v0.10.0

### OS

NA

### Chrome version

NA

### Docker version and image (if applicable)

NA

### Steps to reproduce the problem

When this script is ran, it deadlocks after the click occurs.

```go
import { chromium } from 'k6/experimental/browser';

export default async function () {
  const browser = chromium.launch({
    headless: false,
  });
  const page = browser.newPage();

  try {
    await page.goto('https://near.org/', { waitUntil: 'networkidle' });

    console.log('About to click on sign-in');

    await Promise.all([
      page.waitForNavigation({ waitUntil: 'networkidle' }),
      page.locator('button[class="sign-in"]').click(), // <- deadlocks on a navigation caused by this click
    ]);

    console.log('click on sign-in done');
  } finally {
    await page.close();
    await browser.close();
  }
}
```

The exact reason as to why this particular website is causing this issue is unknown, but there is a hierarchy of frames. When a navigation completes, we sometimes need to remove child frames. During this removal of child frames, we see a deadlock. The reason for this is due to the `m.framesMu` lock being held in `FrameManger.frameNavigated` [here](https://github.com/grafana/xk6-browser/blob/8b00867586b2f069f625a69b4880a585ec09533d/common/frame_manager.go#L226), and when it calls [removeFrameRecursively](https://github.com/grafana/xk6-browser/blob/8b00867586b2f069f625a69b4880a585ec09533d/common/frame_manager.go#L242) it deadlocks [here](https://github.com/grafana/xk6-browser/blob/8b00867586b2f069f625a69b4880a585ec09533d/common/frame_manager.go#L415) waiting for `m.framesMu`.

### Expected behaviour

The script completes after navigating to the sign in page.

### Actual behaviour

The script hangs after the navigation occurs to the sign in page.

PR #942, opened 2023-06-19
Fix deadlock by removing lock in frameNavigated

ankur22:
### Description of changes

In certain navigations, the current frame may have child frames. In these cases, the child frames need to be removed. frameNavigated holds on to a framesMu lock for the whole duration of the frameNavigated.

removeFramesRecursively, which is what removes the child frames, needs the same lock so that it can amend the synchronised variable.

It's also worth noting that this also prevent frameNavigated holding on to a lock when it calls ChildFrames, which requires another lock. It's generally unsafe to hold onto more than one lock at a time.

Closes: https://github.com/grafana/xk6-browser/issues/941

### Checklist

### DIFF
--- a/common/frame_manager.go
+++ b/common/frame_manager.go
@@ -238,9 +238,11 @@ func (m *FrameManager) frameNavigated(frameID cdp.FrameID, parentFrameID cdp.Fra
 		m.ID(), frameID, parentFrameID, documentID, name, url, initial)
 
 	if frame != nil {
+		m.framesMu.Unlock()
 		for _, child := range frame.ChildFrames() {
 			m.removeFramesRecursively(child.(*Frame))
 		}
+		m.framesMu.Lock()
 	}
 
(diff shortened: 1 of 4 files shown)


common/frame_manager.go
```diff
@@ -238,9 +238,11 @@ func (m *FrameManager) frameNavigated(frameID cdp.FrameID, parentFrameID cdp.Fra
 		m.ID(), frameID, parentFrameID, documentID, name, url, initial)
 
 	if frame != nil {
+		m.framesMu.Unlock()
```
ankur22, line 241: TODO: Create a e2e test for this. Need to work out how to create a frame in a frame, maybe with `iframe`s?

ka3de: Should we add it in this same PR?

ka3de (approved):
LGTM.

Issue #914, opened 2023-05-31
Web Vitals metrics are not consistently reported


ka3de:
### Brief summary

At least some Web Vitals metrics are not consistently reported on every page navigation. This usually affects particularly the metrics:
- `browser_web_vital_lcp`
- `browser_web_vital_fid`
- `browser_web_vital_cls`
and their corresponding rating metrics. See: #915.

Initial suspicion was a race condition between the call to Web Vitals event binding from the browser execution, and the page/browser closing when the test implementation just contains a single `page.goto` for example.

But even adding a `sleep` after the navigation does not seem to fix the issue.

### xk6-browser version

Commit 45b3e8b

### OS

Ubuntu 20.04.5 LTS

### Chrome version

113.0.5672.126 (Official Build) (64-bit)

### Steps to reproduce the problem

Build a custom k6 binary:
```
xk6 build --with github.com/grafana/xk6-browser@7cfb927
```
Run the following test:
```js
import { chromium } from 'k6/x/browser';

export const options = {
    scenarios: {
        ui: {
            executor: 'shared-iterations',
            options: {
                browser: {
                    type: 'chromium',
                },
            },
        },
    },
};

export default async function() {
    const browser = chromium.launch();
    const page = browser.newPage();

    try {
        await page.goto('https://test.k6.io/', { waitUntil: 'networkidle' });
    } finally {
        page.close();
    } 
}

```

### Expected behaviour

All Web Vitals metrics are reported.

### Actual behaviour

Only FCP and TTF Web Vitals metrics are reported:
![Screenshot from 2023-05-31 08-16-50](https://github.com/grafana/xk6-browser/assets/6529572/02b56bcd-42b0-4234-acad-2c5b5a600c19)

ka3de:
Take into account that `browser_web_vital_fid` will not be reported if the user never interacts with the page.

inancgumus:
Related knowledge source: https://github.com/GoogleChrome/web-vitals.

inancgumus:
Updated the issue description with some minor findings. Also updated the commit that we're making tests on: 45b3e8b.

inancgumus:
#943 aims the fix the first part of the problem. Feel free to assign yourself to this issue besides me for the next part :)🙇

inancgumus:
As far as I can see, this is how the metric emission works:

## Largest Contentful Paint (LCP):

Emitted: LCP is typically emitted when the largest visible element in the viewport becomes visible.
[...]

PR #943, opened 2023-06-20
Fix race condition in navigation (and lifecycle events)

inancgumus:
_TLDR: Skip the initial blank page navigation/lifecycle event if we want to navigate to a non-blank page so that we can wait for the navigation to complete. This way, we can receive custom metrics like Web Vitals. Custom metrics would be randomly skipped without this._

In order to make sure we receive custom metrics like Web Vitals, it is necessary to skip the initial blank page navigation event when navigating to a non-blank page. By doing this, we can wait for the navigation to complete fully. Without skipping this event, custom metrics could be randomly skipped.

<details>
<summary>Before this fix, this test's Web Vital metrics were flaky. Sometimes they show up, and sometimes, they don't.</summary>

```js

export const options = {
  scenarios: {
    ui: {
      vus: 1,
      iterations: 1,
      options: {
        browser: {
          type: 'chromium',
        },
      },
    },
  },
};

  try {
    // the above was returning prematurely and caused page.close to be
    // called before the navigation finished.
  } finally {
    page.close();
  } 
}
```

</details>

I'll go ahead and explain the problem using the provided script.

The issue of inconsistency arose because we were removing the Web Vital bindings using `page.Close` before the navigation had finished. The navigation process was prematurely interrupted due to the initial blank page lifecycle events. As a result, we were not able to receive all the events from the Web Vitals library since we didn't wait for the navigation to complete.

This pull request resolves the problem of unreliable behavior in emitting Web Vitals (or other custom metrics). However, it doesn't address the issue of displaying all Web Vital reports. According to [Google's Web Vital documentation](https://github.com/GoogleChrome/web-vitals#usage), most metrics are meant to be reported only occasionally.

P.S.: I attempted to create a test for this behavior, but I was unable to replicate it. Please let me know if you have any ideas. The existing tests _locally_ pass!

**Tip:** I noticed that reloading the page just before closing it leads to more emitted metrics. Another pull request could tackle this aspect. Update: See #949.

Updates: #914.

### DIFF
--- a/common/frame_manager.go
+++ b/common/frame_manager.go
@@ -592,10 +592,23 @@ func (m *FrameManager) NavigateFrame(frame *Frame, url string, parsedOpts *Frame
 	lifecycleEvtCh, lifecycleEvtCancel := createWaitForEventPredicateHandler(
 		timeoutCtx, frame, []string{EventFrameAddLifecycle},
 		func(data any) bool {
-			if le, ok := data.(LifecycleEvent); ok {
-				return le == parsedOpts.WaitUntil
+			le, ok := data.(FrameLifecycleEvent)
+			if !ok {
+				return false
 			}
-			return false
+			// skip the initial blank page if we are navigating to a non-blank page.
+			// otherwise, we will get a lifecycle event for the initial blank page
+			// and return prematurely before waiting for the navigation to complete.
+			if url != "about:blank" && le.URL == "about:blank" {
+				m.logger.Debugf(
+					"FrameManager:NavigateFrame:createWaitForEventPredicateHandler",
+					"fmid:%d fid:%v furl:%s url:%s waitUntil:%s event.lifecycle:%q event.url:%q skipping about:blank",
+					fmid, fid, furl, url, parsedOpts.WaitUntil, le.Event, le.URL,
+				)
+				return false
+			}
+
+			return le.Event == parsedOpts.WaitUntil
 		})
 	defer lifecycleEvtCancel()
 
--- a/common/frame_manager.go
+++ b/common/frame_manager.go
@@ -599,11 +599,11 @@ func (m *FrameManager) NavigateFrame(frame *Frame, url string, parsedOpts *Frame
 			// skip the initial blank page if we are navigating to a non-blank page.
 			// otherwise, we will get a lifecycle event for the initial blank page
 			// and return prematurely before waiting for the navigation to complete.
-			if url != "about:blank" && le.URL == "about:blank" {
+			if url != BlankPage && le.URL == BlankPage {
 				m.logger.Debugf(
 					"FrameManager:NavigateFrame:createWaitForEventPredicateHandler",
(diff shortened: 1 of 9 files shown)


common/frame_manager.go
```diff
@@ -592,10 +592,23 @@ func (m *FrameManager) NavigateFrame(frame *Frame, url string, parsedOpts *Frame
...
+			// skip the initial blank page if we are navigating to a non-blank page.
+			// otherwise, we will get a lifecycle event for the initial blank page
+			// and return prematurely before waiting for the navigation to complete.
+			if url != BlankPage && le.URL == BlankPage {
+				m.logger.Debugf(
+					"FrameManager:NavigateFrame:createWaitForEventPredicateHandler",
+					"fmid:%d fid:%v furl:%s url:%s waitUntil:%s event.lifecycle:%q event.url:%q skipping %s",
+					fmid, fid, furl, url, parsedOpts.WaitUntil, le.Event, le.URL, BlankPage,
+				)
+				return false
```
ka3de, line 608: Thanks for working on this @inancgumus !

I think this change makes complete sense, but I'm concerned because in practice I don't see a difference in the Web Vitals metrics being reported. For example, with the following test:
<details>
<summary>Simple test example.</summary>

```js

export const options = {
  scenarios: {
    ui: {
      options: {
        browser: {
            type: 'chromium',
        },
      },
    },
  },
};

  try {
    await page.goto('https://test.k6.io/')
  } finally {
    page.close();
  } 
}
```

</details>

I only see the `FCP` and `TTFB` being reported, but I don't see the `CLS` and `LCP` metrics, which I believe should always be reported after a navigation. And in this sense this is the same behavior we have in `main` right now. Did you observe something different? Am I doing or understanding something wrong? :thinking: 

In any case I understand the change is a step forward on improving this! :clap:

inancgumus: Sure thing ❤️ 

I don't see a difference, too... It was flaky when I started working on this, but now it constantly works on `main`. That's super strange! The only moving parts are the `test.k6.io` site and my network connection. There were network problems a few days back. Maybe the bug only surfaces in those cases. I'm still determining.

I noticed `CLS` and `LSP` are only fired if we make another navigation after the initial one: `page.goto` then `page.reload` or similar. Are we sure that they only get fired for the initial navigation?

When I look into the trace logs, I see that:
1. `NewBrowserContext` adds the Web Vital JSes (`web_vital_[iife|init].js`).
2. Then `NewPage` registers the bindings.
3. The initial frame navigation happens.
4. `beforeunload` and `pagehide` events happen. Usually, as I explained above, after the initial navigation, these events trigger the `CLS` and `LSP` metrics. But, for the initial navigation, they do not.

What this PR does is wait for the navigation to end prematurely, and it's a good first step towards fixing this general Web Vitals issue 🤞

inancgumus: @ka3de, also see [this fix](https://github.com/grafana/xk6-browser/pull/949).

ankur22 (approved):
Thanks @inancgumus LGTM 👍

PR #952, opened 2023-06-23
Fix web vital test bug while receiving metrics

inancgumus:
Since we now send metrics each time they fire up, we need to adjust the test accordingly. I thought we would need this fix, I couldn't resist doing it before my leave :) Sorry for the inconvenience :(

Please feel free to modify the PR commits.

Related: #914.


ankur22 (approved):
LGTM

PR #962, opened 2023-07-07
Add K6_BROWSER_EXECUTABLE_PATH in ci

ankur22:
### Description of changes

The main change (initially) was to work with the `K6_BROWSER_EXECUTABLE_PATH` env var on linux github actions to ensure it picks chrome and not chromium. The github runners have both chromium and google chrome installed. Google chrome is kept up to date whereas chromium isn't. We want to work with the latest version of google chrome (for now, in the future we may want to install the exact version we want to work with).

I also found that the e2e workflow wasn't correctly using the defined matrix, and was only working with `ubuntu-latest`. There are several commits which attempt to resolve this.

Two issues which were discovered but not resolved in the e2e workflow were:

1. the `examples/hosts.js` test script doesn't pass in the Windows stable e2e test. When tested in a Windows 11 computer, the test passed. I've opened a new issue (https://github.com/grafana/xk6-browser/issues/963) to track this.
2. the Windows tip e2e test fails when building a k6 binary with xk6 and the tip version of go. The tip version of go is installed successfully and we seem to be setting it up correctly (ensuring `GOROOT` is pointing to the tip version and not the stable version that is already on the system), but even still xk6 works with the stable version of go which ends up with an [error](https://github.com/grafana/xk6-browser/actions/runs/5881417653/job/15949854594). This hasn't been tested on a standalone windows computer. I've opened a new issue (https://github.com/grafana/xk6-browser/issues/1002) to track this.

Closes: https://github.com/grafana/xk6-browser/issues/963

### Checklist

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         go: [stable, tip]
         platform: [ubuntu-latest, windows-latest, macos-latest]
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.platform }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -23,13 +23,15 @@ jobs:
     runs-on: ${{ matrix.platform }}
     steps:
       - name: Checkout code
+        if: matrix.go != 'tip' || matrix.platform != 'windows-latest'
         uses: actions/checkout@v2
       - name: Install Go
+        if: matrix.go != 'tip' || matrix.platform != 'windows-latest'
         uses: actions/setup-go@v2
         with:
           go-version: 1.x
       - name: Install Go tip
-        if: matrix.go == 'tip'
+        if: matrix.go == 'tip' && matrix.platform != 'windows-latest'
         run: |
           go install golang.org/dl/gotip@latest
           gotip download
@@ -38,8 +40,10 @@ jobs:
           echo "$HOME/go/bin" >> "$GITHUB_PATH"
           echo "$HOME/sdk/gotip/bin" >> "$GITHUB_PATH"
       - name: Install xk6
+        if: matrix.go != 'tip' || matrix.platform != 'windows-latest'
         run: go install go.k6.io/xk6/cmd/xk6@master
       - name: Build extension
+        if: matrix.go != 'tip' || matrix.platform != 'windows-latest'
         run: |
           which go
           go version
@@ -49,6 +53,7 @@ jobs:
             --with github.com/grafana/xk6-browser=.
           ./k6extension version
       - name: Run E2E tests
+        if: matrix.go != 'tip' || matrix.platform != 'windows-latest'
         run: |
           set -x
           if [ "$RUNNER_OS" == "Linux" ]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e.yml
```diff
@@ -23,13 +23,15 @@ jobs:
     runs-on: ${{ matrix.platform }}
     steps:
       - name: Checkout code
+        if: matrix.go != 'tip' || matrix.platform != 'windows-latest'
```
ka3de, line 26: I guess we can skip this by now as long as we are testing with stable version.

ka3de (approved):
Thanks for looking into this! LGTM.

.github/workflows/e2e.yml
```diff
@@ -19,17 +19,19 @@ jobs:
...
-        platform: [ubuntu-latest-8-cores, windows-latest, macos-latest]
-    runs-on: ubuntu-latest-8-cores
+        platform: [ubuntu-latest, windows-latest, macos-latest]
+    runs-on: ${{ matrix.platform }}
     steps:
       - name: Checkout code
+        if: matrix.go != 'tip' || matrix.platform != 'windows-latest'
         uses: actions/checkout@v2
       - name: Install Go
+        if: matrix.go != 'tip' || matrix.platform != 'windows-latest'
```
inancgumus, line 29: I'm wondering if it's possible (an if to me) to put this condition at the job level. If these conditions are still [necessary](https://github.com/grafana/xk6-browser/pull/962/files#r1296911621).

inancgumus (dismissed):
Thanks for making things better 👍

.github/workflows/e2e.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       matrix:
         go: [stable, tip]
         platform: [ubuntu-latest, windows-latest, macos-latest]
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.platform }}
```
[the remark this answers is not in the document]
inancgumus: 💯

.github/workflows/test.yml
```diff
@@ -46,9 +46,7 @@ jobs:
             args[1]="1"
             export GOMAXPROCS=1
           fi
-          export K6_BROWSER_EXECUTABLE_PATH=/usr/bin/google-chrome
-          export K6_BROWSER_HEADLESS=true
-          go test "${args[@]}" -timeout 5m ./...
+          K6_BROWSER_EXECUTABLE_PATH=/usr/bin/google-chrome K6_BROWSER_HEADLESS=true go test "${args[@]}" -timeout 5m ./...
```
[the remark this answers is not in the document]
ankur22: I tried this change but there are tests that rely on an explicit `K6_BROWSER_HEADLESS=false` otherwise they run in headful mode e.g. [TestFrameNoPanicWithEmbeddedIFrame](https://github.com/grafana/xk6-browser/blob/3015c48d5e502561aa8fd9ebf0766b4824b7e04b/tests/frame_test.go#L72).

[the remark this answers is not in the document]
ka3de: Oh OK! :+1:

.github/workflows/e2e.yml
ankur22: I've created an issue to track this: https://github.com/grafana/xk6-browser/issues/1002

ka3de (dismissed):
LGTM.

.github/workflows/e2e.yml
ankur22: That's a nice idea. I tried working with `if` at various levels but it only works in each step. It's because `matrix` is evaluated after the `if` at the [job level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif).

Issue #960, opened 2023-06-30
WebVitals metrics measurements are incorrect for "slow" loading sites


ka3de:
### Brief summary

After modifications made in https://github.com/grafana/xk6-browser/pull/949 and https://github.com/grafana/xk6-browser/pull/943, and comparisons done in https://github.com/grafana/xk6-browser/issues/950, we concluded that, in some cases, k6 browser reports incorrect Web Vitals metrics measurements. This happens especially for sites that are more _dynamic_ and take longer to load, as in this cases k6 browser seems to report "intermediate" values corresponding to samples taken before the page is fully loaded, event for which k6 browser seems to not wait for long enough. This affects particularly [FCP](https://web.dev/fcp/) and [LCP](https://web.dev/lcp/) metric.

### xk6-browser version

commit: c959b57

### OS

Ubuntu 20.04.5 LTS

### Chrome version

113.0.5672.126 (Official Build) (64-bit)

### Steps to reproduce the problem

Run the following test:

```js
import { browser } from 'k6/x/browser';

export const options = {
  scenarios: {
    ui: {
      executor: 'shared-iterations',
      options: {
        browser: {
            type: 'chromium',
        },
      },
    },
  },
};
  

export default async function () {
  const page = browser.newPage();

  try {
    await page.goto('https://grafana.com')
  } finally {
    page.close();
  } 
}
```

### Expected behaviour

Web Vitals metrics should be similar to the ones reported by Google Lighthouse:
![Screenshot from 2023-06-30 11-43-34](https://github.com/grafana/xk6-browser/assets/6529572/c07ec510-0692-47f9-81c5-ac418e00a991)

### Actual behaviour

Web Vitals metrics reported by k6 browser:
![Screenshot from 2023-06-30 11-47-33](https://github.com/grafana/xk6-browser/assets/6529572/81b9572c-9277-4526-92a6-5c6cd7f6bae2)
Notice the differences in FCP and LCP.

ka3de:
[...]

PR #964, opened 2023-07-11
Revert "Report all web vital metric changes (#949)"

ka3de:
This reverts commit f25b1d427af756b58a1a424bf66f735b0d710ad5.

### Description of changes

In our effort to achieve more consistent Web Vitals reporting, a change was introduced in https://github.com/grafana/xk6-browser/pull/949 to force the Web Vitals library to report metrics [on every value change](https://github.com/GoogleChrome/web-vitals#report-the-value-on-every-change). But, in recent tests we have observed that this was an incorrect configuration, especially for [LCP](https://web.dev/lcp/#when-is-largest-contentful-paint-reported) metric. LCP measures the time it takes to render the largest [image or text block](https://web.dev/lcp/#what-elements-are-considered) visible within the viewport, relative to when the page [first started loading](https://w3c.github.io/hr-time/#timeorigin-attribute). Because there can only be one "largest" element in a page, LCP should only provide one sample per page load. Instead, if we use the `reportAllChanges` flag, LCP is reported multiple times when rendering intermediate elements that, up until that point, are the largest rendered in the page. This makes k6 browser record incorrect values for LCP on every new page.

Therefore, we should not use `reportAllChanges` flag for Web Vitals library.

Related: #960.

### DIFF
--- a/common/js/web_vital_init.js
+++ b/common/js/web_vital_init.js
@@ -13,15 +13,13 @@ function print(metric) {
 }
 
 function load() {
-  const reportAllChanges = {
-    reportAllChanges: true,
-  }
-  webVitals.onCLS(print, reportAllChanges);
-  webVitals.onFID(print, reportAllChanges);
-  webVitals.onLCP(print, reportAllChanges);
-  webVitals.onFCP(print, reportAllChanges);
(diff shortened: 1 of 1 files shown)


ankur22 (approved):
LGTM 🚀

PR #974, opened 2023-07-14
Improve browser not found in registry error message

ka3de:
### Description of changes

This PR improves the error message shown when a browser instance for an iteration, which should be automatically initialized as a consequence of the `IterStart` event generated from k6-core, is not found in the browser registry. Because this error might be produced due to the browser type option not being set in scenario definition (the only other possibility is a fault in k6-core/k6 browser implementation), it is better to provide more context information to the user.

### DIFF
--- a/browser/registry.go
+++ b/browser/registry.go
@@ -21,6 +21,13 @@ import (
 	k6modules "go.k6.io/k6/js/modules"
 )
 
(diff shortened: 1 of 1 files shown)


ankur22 (approved):
Nice! LGTM 🚀

Issue #971, opened 2023-07-13
Deadlock: Multi-VU many iterations


ankur22:
While running the following test script:

<details>
<summary>Test script</summary>

```js
import { check } from 'k6';
import { browser } from 'k6/experimental/browser';

export const options = {
  scenarios: {
    ui: {
      executor: 'shared-iterations',
      vus: 5,
      iterations: 1000,
      maxDuration: '40m',
      options: {
        browser: {
          type: 'chromium',
        },
      },
    },
  },
  thresholds: {
    checks: ["rate==1.0"]
  }
}

export default async function () {
  const context = browser.newContext();
  const page = context.newPage();

  try {
    // Goto front page, find login link and click it
    await page.goto('https://test.k6.io/', { waitUntil: 'networkidle' });
    await Promise.all([
      page.waitForNavigation(),
      page.locator('a[href="/my_messages.php"]').click(),
    ]);
    // Enter login credentials and login
    page.locator('input[name="login"]').type('admin');
    page.locator('input[name="password"]').type('123');
    // We expect the form submission to trigger a navigation, so to prevent a
    // race condition, setup a waiter concurrently while waiting for the click
    // to resolve.
    await Promise.all([
      page.waitForNavigation(),
      page.locator('input[type="submit"]').click(),
    ]);
    check(page, {
      'header': page.locator('h2').textContent() == 'Welcome, admin!',
    });
  } finally {
    page.close();
  }
}
```

</details>

The test has to be aborted since it goes over the 40m limit. Investigate why this happens and resolve it.

![Image](https://github.com/grafana/xk6-browser/assets/1112428/a002bb66-f3ce-4bf4-b7eb-557500f5337a)

This is the screenshot of the dashboard of the test run. It would seem that all but one of the VUs completed all the allotted iterations. One vu was stuck in a deadlock, which meant the test run was waiting for it to complete the iteration, which it never did, that eventually resulted in an abort.

---

```[tasklist]
### Tasks
- [x] Collect every instance of this issue occurring
- [x] Investigate the root cause
- [x] Fix
- [x] Test
- [x] Polish
- [x] Retest
- [ ] https://github.com/grafana/xk6-browser/pull/1219
- [ ] Release
- [ ] Update the release notes
```

ankur22:
## Investigation

### Rerun test with different params

The test was ran again, with the fields amended to the values shown in the table:

| Iterations | Max duration | Success (not aborted) |
| - | - | - |
[...]

PR #979, opened 2023-07-25
Temporary fix for deadlock on pagehide event evaluation when page.close is called

ankur22:
### Description of changes

When we run the k6 browser module against a remote chrome instance, we've found that 1 or more VUs deadlock when running the test in this [issue](https://github.com/grafana/xk6-browser/issues/971). What we have come to realise during the investigation is that the [evaluate](https://github.com/grafana/xk6-browser/blob/a8ebe8cc24f639e0553b1d969df97368db7b5ec5/common/page.go#L430) on the `pagehide` event when `page.close` is called seems to wait indefinitely for a response from chrome, that never seems to arrive.

This is a temporary fix where we use the page's `defaultTimeout` to timeout this `evaluate` request instead of waiting indefinitely. We still need to investigate the root cause of the issue. The downside of this fix is that when this `evaluate` does timeout then we may not receive all the web vital measurements for that iteration (the reason we added the `pagehide` event evaluate was to force the web vital measurements to be calculated and sent to the browser module: https://github.com/grafana/xk6-browser/pull/953).

Doesn't close issue https://github.com/grafana/xk6-browser/issues/971 but temporarily alleviates the issue.

### Checklist

### DIFF
--- a/api/frame.go
+++ b/api/frame.go
@@ -1,6 +1,10 @@
 package api
 
-import "github.com/dop251/goja"
+import (
+	"context"
+
+	"github.com/dop251/goja"
+)
 
 // Frame is the interface of a CDP target frame.
 type Frame interface {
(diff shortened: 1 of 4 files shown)


ka3de (approved):
LGTM.

ankur22:
> That's interesting! Have you tried using [ExecuteWithoutExpectationOnReply](https://github.com/grafana/xk6-browser/blob/78b14b64c16ef55506a7c73c2b4fa71765db88f0/common/session.go#L170)?

That is a good solution, and a cleaner way in avoiding having to wait on the response. I did consider a similar approach, but my concern with that was that we would hide the problem, but this is only meant as a temporary fix with a constant reminder (whenever we see the warning) that there is an underlying issue that needs to be dealt with.

PR #980, opened 2023-07-26
log serializer: fix loki log output

mstoykov:
logs going through loki use the `hooks` logrus concept.

The serialzer was not copying this which meant logs wouldn't have been send to loki.


ankur22 (approved):
Aha! Thanks for this! I've also found where the hook is [defined](https://github.com/grafana/k6/blob/c09919b4961f9d93a32f482089a068d36f667ee2/cmd/root.go#L214-L288) for anyone interested.

Issue #861, opened 2023-04-26
TestMultiConnectToSingleBrowser is flaky


inancgumus:
Investigate why the test fails to create the second page.

```
time="2023-04-26T06:02:15Z" level=error msg="communicating with browser: websocket: close 1006 (abnormal closure): unexpected EOF" category=cdp elapsed="0 ms" goroutine=983
time="2023-04-26T06:02:15Z" level=error msg="communicating with browser: websocket: close 1006 (abnormal closure): unexpected EOF" category=cdp elapsed="0 ms" goroutine=1165

--- FAIL: TestMultiConnectToSingleBrowser (0.19s)
    browser_test.go:249: 
        	Error Trace:	/home/runner/work/xk6-browser/xk6-browser/tests/browser_test.go:249
        	Error:      	Received unexpected error:
        	            	creating new page in browser context: creating a new blank page: Session with given id not found. (-32001)
        	Test:       	TestMultiConnectToSingleBrowser
        	Messages:   	failed to create page #2

time="2023-04-26T06:02:15Z" level=error msg="communicating with browser: websocket: close 1006 (abnormal closure): unexpected EOF" category=cdp elapsed="0 ms" goroutine=1299
time="2023-04-26T06:02:30Z" level=error msg="Failed to load resource: the server responded with a status of 401 (Unauthorized)" browser_source=network line_number=0 source=browser stacktrace="<nil>" url="http://127.0.0.1:34171/basic-auth/validuser/validpass"
time="2023-04-26T06:02:30Z" level=warning msg="url:http://127.0.0.1:34171/basic-auth/validuser/validpass method:GET err:fetching response body: context canceled" category="Response:bodySize:fetchBody" elapsed="0 ms" goroutine=8069
time="2023-04-26T06:02:30Z" level=warning msg="url:http://127.0.0.1:35603/static/empty.html method:GET err:fetching response body: context canceled" category="Response:bodySize:fetchBody" elapsed="0 ms" goroutine=8365
time="2023-04-26T06:02:32Z" level=warning msg="The specified value \"invalid date\" does not conform to the required format, \"yyyy-MM-dd\"." browser_source=rendering line_number=457 source=browser url=__xk6_browser_evaluation_script__
time="2023-04-26T06:02:34Z" level=error msg="communicating with browser: websocket: close 1006 (abnormal closure): unexpected EOF" category=cdp elapsed="0 ms" goroutine=10587
[...]

PR #984, opened 2023-07-27
Ensure msgId in CDP requests are unique at the connection scope

ankur22:
### Description of changes

A detailed description as to why we're doing this can be found [here](https://github.com/grafana/xk6-browser/issues/861#issuecomment-1653312043).

Currently the `msgId`s clash from one session to another as well as with`msgId`s that are sent without a session on the connection event loop.

Generally this isn't an issue, but recently we have found that chrome sometimes returns an error without the `sessionId` that was in the original request msg. When this occurs, we route those error responses to handlers that are waiting on the connection event loop. If there's a clash in `msgId` then the wrong handler will work with the wrong response, which could result in incorrect behaviour. The other issue with routing the response to the wrong handler is that the correct handler could end up waiting indefinitely (or timeout) which could pause the whole test iteration/vu.

This change will enforce a unique msgId at the connection level, which should avoid such scenarios.

Linked issue: https://github.com/grafana/xk6-browser/issues/861

### Checklist

### DIFF
--- a/common/connection.go
+++ b/common/connection.go
@@ -28,6 +28,19 @@ const wsWriteBufferSize = 1 << 20
 var _ EventEmitter = &Connection{}
 var _ cdp.Executor = &Connection{}
 
+// Each connection needs its own msgID. A msgID will be used by the
+// connection and associated sessions. When a CDP request is made to
+// chrome, it's best to work with unique ids to avoid the Execute
+// handlers working with the wrong response, or handlers deadlocking
+// when their response is rerouted to the wrong handler.
+type msgID struct {
+	id int64
+}
+
+func (m *msgID) new() int64 {
+	return atomic.AddInt64(&m.id, 1)
+}
+
 type executorEmitter interface {
 	cdp.Executor
 	EventEmitter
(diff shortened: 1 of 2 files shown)


common/connection.go
```diff
@@ -28,6 +28,19 @@ const wsWriteBufferSize = 1 << 20
...
+// chrome, it's best to work with unique ids to avoid the Execute
+// handlers working with the wrong response, or handlers deadlocking
+// when their response is rerouted to the wrong handler.
+type msgID struct {
+	id int64
+}
+
+func (m *msgID) new() int64 {
+	return atomic.AddInt64(&m.id, 1)
+}
```
ka3de, line 42: Do you think it would make sense to abstract this behind an interface such as:
```
type msgIDGenerator interface {
    newID() int64
}
```
So we would hold this type in both `Connection` and `Session` and guarantee there is no direct access to the underlying `id` var?

ka3de (approved):
LGTM :clap: Just a small consideration.

inancgumus (approved):
LGTM 🎉

PR #991, opened 2023-08-02
Fix for data directory not being cleared during browser.close() panic

ankur22:
### Description of changes

While testing for another change, it became apparent that the panic when working with browser.close() would prevent any of the remaining data directories from being cleaned up in the temporary directory. Instead of panicking, we're going to log an error and carry on with rest of the process to try and close the subprocess and eventually delete the data directory. This change does indeed bring about the behaviour we want even when the context has been closed.

### Checklist

- [x] Written tests for the changes

### DIFF
--- a/tests/browser_test.go
+++ b/tests/browser_test.go
@@ -69,7 +69,14 @@ func TestBrowserNewContext(t *testing.T) {
 func TestTmpDirCleanup(t *testing.T) {
 	t.Parallel()
 
-	const tmpDirPath = "./"
+	const tmpDirPath = "./1/"
+	err := os.Mkdir(tmpDirPath, os.ModePerm)
+	require.NoError(t, err)
+
+	defer func() {
+		err = os.Remove(tmpDirPath)
+		require.NoError(t, err)
+	}()
 
 	b := newTestBrowser(
 		t,
@@ -77,7 +84,7 @@ func TestTmpDirCleanup(t *testing.T) {
 		withEnvLookup(env.ConstLookup("TMPDIR", tmpDirPath)),
 	)
 	p := b.NewPage(nil)
-	err := p.Close(nil)
+	err = p.Close(nil)
 	require.NoError(t, err)
 
 	matches, err := filepath.Glob(tmpDirPath + "xk6-browser-data-*")
--- a/tests/browser_test.go
+++ b/tests/browser_test.go
@@ -98,6 +98,38 @@ func TestTmpDirCleanup(t *testing.T) {
 	assert.Empty(t, matches, "a dir shouldn't exist which matches the pattern `xk6-browser-data-*`")
 }
 
+func TestTmpDirCleanupOnContextClose(t *testing.T) {
+	t.Parallel()
+
+	const tmpDirPath = "./2/"
+	err := os.Mkdir(tmpDirPath, os.ModePerm)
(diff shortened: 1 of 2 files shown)


tests/browser_test.go
```diff
@@ -69,15 +69,22 @@ func TestBrowserNewContext(t *testing.T) {
 func TestTmpDirCleanup(t *testing.T) {
 	t.Parallel()
 
-	const tmpDirPath = "./"
+	const tmpDirPath = "./1/"
+	err := os.Mkdir(tmpDirPath, os.ModePerm)
+	require.NoError(t, err)
+
+	defer func() {
+		err = os.Remove(tmpDirPath)
```
inancgumus, line 77: From [the Go doc](https://pkg.go.dev/os#Remove): `os.Remove` remove removes the named file or (empty) directory. If there is an error, it will be of type *PathError. Should we use `os.RemoveAll` here?

We're also overwriting an existing error variable. There is no problem at this stage. However, if someone were to add another deferred function that also manipulates the same `err` variable, the order in which the deferred functions are defined would influence the behavior. Should we use another variable?

One last thing. If there's an assertion failure (e.g., require.NoError) before the deferred function is called, the cleanup code might not execute. This could leave the temporary directory on the file system. WDYT?

inancgumus (commented):
LGTM. Thanks for working on this and fixing the issues. And also adding one more test 👏  Could you consider my suggestions? My suggestions also apply to the second test. Thanks!

tests/browser_test.go
```diff
@@ -69,15 +69,22 @@ func TestBrowserNewContext(t *testing.T) {
 func TestTmpDirCleanup(t *testing.T) {
 	t.Parallel()
 
-	const tmpDirPath = "./"
+	const tmpDirPath = "./1/"
```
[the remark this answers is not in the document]
ankur22: Yeah, sounds good to me. Resolved in fd5d7e528d938939bd92f6f0322bf17c337586c4.

tests/browser_test.go
ankur22: > From [the Go doc](https://pkg.go.dev/os#Remove): os.Remove remove removes the named file or (empty) directory. Should we use os.RemoveAll here?

Yeah that's a good idea incase the test fails and the sub dir hasn't been removed.

> We're also overwriting an existing error variable. There is no problem at this stage. However, if someone were to add another deferred function that also manipulates the same err variable, the order in which the deferred functions are defined would influence the behavior. Should we use another variable?

👍 

> One last thing. If there's an assertion failure (e.g., require.NoError) before the deferred function is called, the cleanup code might not execute. This could leave the temporary directory on the file system. WDYT?

If there's an error on `MkdirTemp` then I would expect there to be no dir, but as a precaution i've moved the defer straight after calling `MkdirTemp`. Would it be better to work with `t.Cleanup` instead of defer though?

Resolved in: fd5d7e528d938939bd92f6f0322bf17c337586c4

inancgumus: 👍 Yes, `Cleanup` might make more sense since we might be queuing other `Cleanup` in our helpers like `testBrowser`.

ankur22: Resolved in: 1d4172704f7b1b0324bfccf4ec084c55e1296172

inancgumus (approved):
👍

PR #1003, opened 2023-08-17
Add a LookupBool utility function

inancgumus:
## What?

Defines a `LookupBool` helper.

## Why?

To make bool lookup queries less verbose in tests.

## Checklist

- [x] I have added tests for my changes (indirectly)

### DIFF
--- a/env/env.go
+++ b/env/env.go
@@ -1,7 +1,10 @@
 // Package env provides types to interact with environment setup.
 package env
 
-import "os"
+import (
+	"os"
+	"strconv"
+)
 
 // Execution specific.
 const (
@@ -66,7 +69,7 @@ const (
 type LookupFunc func(key string) (string, bool)
 
 // EmptyLookup is a LookupFunc that always returns "" and false.
-func EmptyLookup(key string) (string, bool) { return "", false }
+func EmptyLookup(_ string) (string, bool) { return "", false }
 
 // Lookup is a LookupFunc that uses os.LookupEnv.
 func Lookup(key string) (string, bool) { return os.LookupEnv(key) }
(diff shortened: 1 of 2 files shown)


env/env.go
```diff
@@ -82,3 +85,18 @@ func ConstLookup(k, v string) LookupFunc {
 		return EmptyLookup(key)
 	}
 }
+
+// LookupBool returns the result of Lookup as a bool.
+// Returns false if the key does not exist or the value
+// is not a bool.
+func LookupBool(key string) bool {
```
ka3de, line 92: I wonder if this should return two `bool` values, one to indicate whether the variable was actually set and one for its value? Take for instance the `K6_BROWSER_HEADLESS` environment variable, if we lookup for it, and it's actually not set, this would return `false`, so the browser would be launched in `headfull` mode, when instead it should have been launched as `headless` because it's the default mode.

inancgumus: Say we return two bools so we can use it like this:

```go
headless, exists := LookupBool("K6_BROWSER_HEADLESS)
if !exists {
    headless = true
}
```

WDYT about adding this (default behavior of our env. vars.) as a special condition to this function? I'd prefer the func to be pure of these concerns. Still, this function can be tightly knit to our environment variables and provide defaults when necessary without the other parts of the code duplicating this behavior.

inancgumus: I've applied your suggestion here: [7f3aa58](https://github.com/grafana/xk6-browser/pull/1003/commits/7f3aa5878aee80032f031decd48b48c55dd01049).

Then I've introduced the `IsBrowserHeadless` helper instead of modifying `LookupBool` here: [104245a](https://github.com/grafana/xk6-browser/pull/1003/commits/104245ace62813c16e8bb4389bf60c27fc9707c3).

WDYT?

ankur22 (approved):
LGTM

ka3de (approved):
Nice! LGTM.

Issue #965, opened 2023-07-12
`Response.AllHeaders()` does not include security-related headers


ka3de:
### Brief summary

Currently the k6 HTTP `Response` implementation contains the headers returned in the `Network.responseReceived` CDP [event](https://chromedevtools.github.io/devtools-protocol/tot/Network/#event-responseReceived). These headers are the ones that should be returned for example when calling the `Response.Headers()` method. Instead, the `Response.AllHeaders()` method should also include the security-related headers (e.g.: cookie headers) which can be parsed through the `Network.responseReceivedExtraInfo` CDP [event](https://chromedevtools.github.io/devtools-protocol/tot/Network/#event-responseReceivedExtraInfo) that contains all raw headers.

### xk6-browser version

v0.10.0

### OS

Ubuntu 20.04.5 LTS

### Chrome version

113.0.5672.126 (Official Build) (64-bit)

### Steps to reproduce the problem

Run the following script:
```js
import { check } from 'k6';
import { chromium } from 'k6/experimental/browser';

export const options = {
  scenarios: {
    ui: {
      executor: 'shared-iterations',
      options: {
        browser: {
            type: 'chromium',
        },
      },
    },
  },
  thresholds: {
    checks: ["rate==1.0"]
  }
}

export default async function() {
  const browser = chromium.launch();
  const context = browser.newContext();
  const page = context.newPage();

  try {
    // Goto front page, find login link and click it
    await page.goto('https://test.k6.io/', { waitUntil: 'networkidle' });
    await Promise.all([
      page.waitForNavigation(),
      page.locator('a[href="/my_messages.php"]').click(),
    ]);

    // Enter login credentials and login
    page.locator('input[name="login"]').type('admin');
    page.locator('input[name="password"]').type('123');
    // We expect the form submission to trigger a navigation, so to prevent a
    // race condition, setup a waiter concurrently while waiting for the click
    // to resolve.
    await Promise.all([
      page.waitForNavigation(),
      page.locator('input[type="submit"]').click(),
    ]).then(function(res) {
      console.log("headers: " + JSON.stringify(res[0].allHeaders()));
    });
    check(page, {
      'header': page.locator('h2').textContent() == 'Welcome, admin!',
    });
  } finally {
    page.close();
    browser.close();
  }
}
```

### Expected behaviour

The test output should show the `console.log()` message with all headers, including also the cookie related headers after the login performed in the page.

### Actual behaviour

The test output `console.log()` message does not include any cookie related header:
[...]

PR #1005, opened 2023-08-21
Add browserContext.cookies()

### LINKED ISSUE #6: Cookies: Retrieving and filtering by URLs
Add support for `BrowserContext.cookies([urls])`, to return all cookies or only cookies for specified URLs in browser context.

**Relevant links**:
- [Playwright docs](https://playwright.dev/docs/api/class-browsercontext#browser-context-cookies)
- Playwright code: [1](https://github.com/microsoft/playwright/blob/0c3703fda190f8fa1ea6e5e8bc51a23385f80602/packages/playwright-core/src/server/browserContext.ts#L271-L275) and [2](https://github.com/microsoft/playwright/blob/0c3703fda190f8fa1ea6e5e8bc51a23385f80602/packages/playwright-core/src/server/chromium/crBrowser.ts#L391-L403)
- [Playwright browserContext.cookies tests](https://github.com/microsoft/playwright/blob/0c3703fda190f8fa1ea6e5e8bc51a23385f80602/tests/library/browsercontext-cookies.spec.ts) 

```[tasklist]
### Tasks
- [x] #1008
- [x] #1005
- [x] #1030
- [x] #1025
- [x] #1016
- [x] #1039
- [x] grafana/k6-docs#1328
- [x] grafana/k6-DefinitelyTyped#14
- [x] grafana/k6#3352
```

inancgumus:
## What?

Adds `browserContext.cookies()` support.

There will be another PR to add `browserContext.cookies([urls])` support to filter by cookie URLs.

## Why?

1. To let users access cookies.
2. Providing a better Playwright compatibility.

## Checklist

The docs will be updated when #6 finishes.

## Related PR(s)/Issue(s)

Updates #6

### DIFF
--- a/examples/cookies.js
+++ b/examples/cookies.js
@@ -1,0 +1,50 @@
+import { check } from 'k6';
+import { browser } from 'k6/x/browser';
+
+export const options = {
+  scenarios: {
+    ui: {
+      executor: 'shared-iterations',
+      options: {
+        browser: {
+            type: 'chromium',
+        },
+      },
+    },
+  },
+  thresholds: {
+    checks: ["rate==1.0"]
+  }
+};
+
+export default async function () {
+  const page = browser.newPage();
+  const context = page.context();
+
+  try {
+    // get cookies from the browser context
+    check(context.cookies().length, {
+        'initial number of cookies should be zero': n => n === 0,
+    });
+
+    // add some cookies to the browser context
+    context.addCookies([{name: 'testcookie', value: '1', sameSite: 'Strict', domain: '127.0.0.1', path: '/'}]);
+    context.addCookies([{name: 'testcookie2', value: '2', sameSite: 'Lax', domain: '127.0.0.1', path: '/'}]);
+
+    check(context.cookies().length, {
+      'number of cookies should be 2': n => n === 2,
+    });
+
+    const cookies = context.cookies();
+    check(cookies[0], {
+      'cookie 1 name should be testcookie': c => c.name === 'testcookie',
+      'cookie 1 value should be 1': c => c.value === '1',
+    });
+    check(cookies[1], {
+      'cookie 2 name should be testcookie2': c => c.name === 'testcookie2',
+      'cookie 2 value should be 2': c => c.value === '2',
+    });
+  } finally {
+    page.close();
+  } 
+}
--- a/examples/cookies.js
+++ b/examples/cookies.js
@@ -46,5 +46,5 @@ export default async function () {
     });
   } finally {
     page.close();
-  } 
+  }
 }
--- a/tests/browser_context_test.go
+++ b/tests/browser_context_test.go
@@ -284,6 +284,28 @@ func TestBrowserContextCookies(t *testing.T) {
 				},
 			},
 		},
+		"cookie_with_path": {
+			setupHandler: okHandler,
+			documentCookiesSnippet: `
+				() => {
+					document.cookie = "name=value; path=/empty";
+					return document.cookie;
+				}
+			`,
+			wantDocumentCookies: "name=value",
+			wantContextCookies: []*api.Cookie{
+				{
+					Name:     "name",
+					Value:    "value",
+					Domain:   "127.0.0.1",
+					Expires:  -1,
+					HTTPOnly: false,
+					Path:     "/empty",
+					SameSite: "",
+					Secure:   false,
+				},
+			},
+		},
 	}
 	for name, tt := range tests {
 		tt := tt
--- a/tests/browser_context_test.go
+++ b/tests/browser_context_test.go
@@ -229,6 +229,28 @@ func TestBrowserContextCookies(t *testing.T) {
 			wantDocumentCookies: "",
 			wantContextCookies:  nil,
 		},
+		"cookie": {
+			setupHandler: okHandler,
+			documentCookiesSnippet: `
+				() => {
+					document.cookie = "name=value";
+					return document.cookie;
+				}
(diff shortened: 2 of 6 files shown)


tests/browser_context_test.go
```diff
@@ -189,3 +192,243 @@ func TestBrowserContextAddCookies(t *testing.T) {
...
+			setupHandler: okHandler,
+			documentCookiesSnippet: `
+				() => {
+					document.cookie = "name=value; domain=k6.io";
+					return document.cookie;
+				}
+			`,
+			wantDocumentCookies: "", // some cookies cannot be set (i.e. cookies using different domains)
+			wantContextCookies:  nil,
+		},
```
ka3de, line 319: This test case seems to test more the browser implementation than ours, but I guess it doesn't harm either.

examples/cookies.js
```diff
@@ -0,0 +1,50 @@
...
+      'cookie 1 value should be 1': c => c.value === '1',
+    });
+    check(cookies[1], {
+      'cookie 2 name should be testcookie2': c => c.name === 'testcookie2',
+      'cookie 2 value should be 2': c => c.value === '2',
+    });
+  } finally {
+    page.close();
+  } 
+}
```
ka3de, line 50: Should we add a newline at the end of the file?

ka3de (approved):
This is great! :clap:  I was just wondering where is the URLs filter, but I see this is just a part of #6.
LGTM.

examples/cookies.js
```diff
@@ -0,0 +1,50 @@
...
+  const page = browser.newPage();
+  const context = page.context();
+
+  try {
+    // get cookies from the browser context
+    check(context.cookies().length, {
+        'initial number of cookies should be zero': n => n === 0,
+    });
+
+    // add some cookies to the browser context
```
ka3de, line 30: Just one small comment. Would it make sense to write a test, or complement this test with cookies that are result of a "page navigation" instead of setting them through our own API? E.g.: Navigate to https://test.k6.io/my_messages.php , login and check for `sid` cookie.

tests/browser_context_test.go
```diff
@@ -189,3 +192,86 @@ func TestBrowserContextAddCookies(t *testing.T) {
 		})
 	}
 }
+
+func TestBrowserContextCookies(t *testing.T) {
+	t.Parallel()
+
+	// an empty page is required to set cookies. we're just using a
```
ankur22, line 199: Thanks for all these comments 🙂 

I feel that the code is self explanatory, and the comments just get in the way. The names of the table tests are perfect, and as more tests are added it helps build the picture of how this test works.

examples/cookies.js
inancgumus: You mean like #965?

tests/browser_context_test.go
ankur22: Yeah, it is. I actually quite like it. For someone who isn't clued up on cookies this is useful to see 🙂

tests/browser_context_test.go
```diff
@@ -340,6 +340,29 @@ func TestBrowserContextCookies(t *testing.T) {
 				},
 			},
 		},
+		"same_site_strict_cookie": {
+			setupHandler: func(w http.ResponseWriter, _ *http.Request) {
+				w.Header().Set("Set-Cookie", "name=value;SameSite=Strict")
+			},
+			documentCookiesSnippet: `
+				() => {
+					return document.cookie;
```
ankur22, line 349: Is there a way to set a cookie that isn't from the "same site", and then ensure that we don't see if in the response?

ankur22 (approved):
LGTM 👏 Nicely done 🙂 

I only left minor comments for you to consider 👍

tests/browser_context_test.go
inancgumus: You're right 👍 I'll remove this.

examples/cookies.js
inancgumus: Sure.

tests/browser_context_test.go
inancgumus: Let's keep it, then :)

tests/browser_context_test.go
inancgumus: The comments are a bit verbose, I agree 👍 They can also help other contributors. Let me know if you want me to remove some of them. Thanks!

tests/browser_context_test.go
inancgumus: This test is for testing whether we receive the `SameSite: api.CookieSameSiteStrict` field correctly. It would be nice to have such a test, but Chromium/Chrome won't let webpages set cookies in a different domain when webpages set this header. I'm afraid it would be like testing the browser's behavior as in [here](https://github.com/grafana/xk6-browser/pull/1005#discussion_r1304412292).

examples/cookies.js
inancgumus: It's a nice idea, so I added this: https://github.com/grafana/xk6-browser/pull/1005/commits/b4eeca4b3fa46ef6361af7a4f1168230b8f4ee59. Let me know what you think. Thanks!

tests/browser_context_test.go
ka3de: We can keep it :+1:

examples/cookies.js
ka3de: It's maybe not the best that it's not under the `cookies` example, but it's true that we would have to duplicate pretty much the whole `fillform` test, so should be OK :+1:

PR #1006, opened 2023-08-21
Add `page.on`

### LINKED ISSUE #835: Implement page.on('console') to be able to assert on logs
### Feature Description

There has been a request to be able to assert on logs. The use case is for when a user want to assert that something has occurred, but that something can only print to the console with `console.log`, and there is nothing on the site itself that can be used. When something is logged, we should be able to bring that to the attention of the user who can compare it against an expected string.

### Suggested Solution (optional)

`page.on('console')` -> https://playwright.dev/docs/api/class-page#events

### Already existing or connected issues / PRs (optional)

_No response_
```[tasklist]
### Tasks
- [ ] #1006
- [ ] grafana/k6-docs#1307
- [ ] #1015
- [ ] grafana/k6#3353
```

ka3de:
## What?

Adds support for `page.on` method. Currently `'console'` event is the only one allowed.

This will allow users to define a function to be executed every time a `console` API is called in the web page context. Such as:

- `console.log`
- `console.debug`
- `console.info`
- `console.error`
- `console.warn`
- `console.dir`
- `console.dirxml`
- `console.table`
- `console.trace`
- `console.clear`
- `console.assert`
- `console.profile`
- `console.profileEnd`
- `console.group`
- `console.groupCollapsed`
- `console.groupEnd`
- `console.count`

See full reference and details at [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/API/console) for `console` API.

The handler function defined by the user will receive a console message object as input for every call, such as:
```js
ConsoleMessage {
    args()    // returns the input arguments for the console API method call
    page()   // returns the page that produced this console message
    text()     // returns the text for the console message
    type()    // returns the type for the console message
}
```

## Why?

- Increases the compatibility with Playwright.
- Allows users to perform assertions on console events from the page context execution.

## Checklist

## Related PR(s)/Issue(s)

Updates #835.

### DIFF
--- a/api/console_message.go
+++ b/api/console_message.go
@@ -1,0 +1,19 @@
+package api
+
+// ConsoleMessage represents a page console message.
+type ConsoleMessage struct {
+	// Args represent the list of arguments passed to a console function call.
+	Args []JSHandle
+
+	// Page is the page that produced the console message, if any.
+	Page Page
+
+	// Text represents the text of the console message.
+	Text string
+
+	// Type is the type of the console message.
+	// It can be one of 'log', 'debug', 'info', 'error', 'warning', 'dir', 'dirxml',
+	// 'table', 'trace', 'clear', 'startGroup', 'startGroupCollapsed', 'endGroup',
+	// 'assert', 'profile', 'profileEnd', 'count', 'timeEnd'.
+	Type string
+}
--- a/common/page.go
+++ b/common/page.go
@@ -1168,7 +1168,7 @@ func (p *Page) consoleMsgFromConsoleEvent(e *cdpruntime.EventConsoleAPICalled) (
 }
 
 // executionContextForID returns the page ExecutionContext for the given ID.
-func (p *Page) executionContextForID( //nolint:unused
+func (p *Page) executionContextForID(
 	executionContextID cdpruntime.ExecutionContextID,
 ) (*ExecutionContext, error) {
 	p.frameSessionsMu.RLock()
--- a/common/page.go
+++ b/common/page.go
@@ -125,7 +125,6 @@ func NewPage(
 		workers:          make(map[target.SessionID]*Worker),
 		routes:           make([]api.Route, 0),
 		vu:               k6ext.GetVU(ctx),
-		tq:               taskqueue.New(k6ext.GetVU(ctx).RegisterCallback),
 		logger:           logger,
 	}
 
@@ -187,9 +186,12 @@ func (p *Page) initEvents() {
 		defer func() {
 			p.logger.Debugf("Page:initEvents:go:return",
 				"sid:%v tid:%v", p.session.ID(), p.targetID)
-			// The TaskQueue must be closed in order to let
-			// the event loop finish
-			p.tq.Close()
+			// TaskQueue is only initialized when calling page.on() method
+			// so users are not always required to close the page in order
+			// to let the iteration finish.
+			if p.tq != nil {
+				p.tq.Close()
+			}
 		}()
 
 		for {
@@ -798,6 +800,13 @@ func (p *Page) MainFrame() api.Frame {
 func (p *Page) On(event string, handler func(*api.ConsoleMessage) error) error {
 	if event != eventPageConsoleAPICalled {
 		return fmt.Errorf("unknown page event: %q, must be %q", event, eventPageConsoleAPICalled)
+	}
+
+	// Once the TaskQueue is initialized, it has to be closed so the event loop can finish.
+	// Therefore, instead of doing it in the constructor, we initialize it only when page.on()
+	// is called, so the user is only required to close the page it using this method.
+	if p.tq == nil {
+		p.tq = taskqueue.New(p.vu.RegisterCallback)
 	}
 
 	p.eventHandlersMu.Lock()
--- a/common/page.go
+++ b/common/page.go
@@ -25,13 +25,19 @@ import (
 	"github.com/dop251/goja"
 )
 
-const webVitalBinding = "k6browserSendWebVitalMetric"
+const (
+	webVitalBinding = "k6browserSendWebVitalMetric"
+
+	eventPageConsoleAPICalled = "console"
+)
 
 // Ensure page implements the EventEmitter, Target and Page interfaces.
 var (
 	_ EventEmitter = &Page{}
 	_ api.Page     = &Page{}
 )
+
+type consoleEventHandlerFunc func(*api.ConsoleMessage) error
 
 // Page stores Page/tab related context.
 type Page struct {
@@ -70,6 +76,10 @@ type Page struct {
 
 	backgroundPage bool
 
+	eventCh         chan Event
+	eventHandlers   map[string][]consoleEventHandlerFunc
+	eventHandlersMu sync.RWMutex
+
 	mainFrameSession *FrameSession
 	frameSessions    map[cdp.FrameID]*FrameSession
 	frameSessionsMu  sync.RWMutex
@@ -105,6 +115,8 @@ func NewPage(
 		timeoutSettings:  NewTimeoutSettings(bctx.timeoutSettings),
 		Keyboard:         NewKeyboard(ctx, s),
 		jsEnabled:        true,
+		eventCh:          make(chan Event),
+		eventHandlers:    make(map[string][]consoleEventHandlerFunc),
 		frameSessions:    make(map[cdp.FrameID]*FrameSession),
 		workers:          make(map[target.SessionID]*Worker),
 		routes:           make([]api.Route, 0),
@@ -732,6 +744,25 @@ func (p *Page) MainFrame() api.Frame {
 	}
 
 	return mf
+}
+
+// On subscribes to a page event for which the given handler will be executed
+// passing in the ConsoleMessage associated with the event.
+// The only accepted event value is 'console'.
+func (p *Page) On(event string, handler func(*api.ConsoleMessage) error) error {
+	if event != eventPageConsoleAPICalled {
+		return fmt.Errorf("unknown page event: %q, must be %q", event, eventPageConsoleAPICalled)
+	}
+
+	p.eventHandlersMu.Lock()
+	defer p.eventHandlersMu.Unlock()
+
+	if _, ok := p.eventHandlers[eventPageConsoleAPICalled]; !ok {
+		p.eventHandlers[eventPageConsoleAPICalled] = make([]consoleEventHandlerFunc, 0, 1)
+	}
+	p.eventHandlers[eventPageConsoleAPICalled] = append(p.eventHandlers[eventPageConsoleAPICalled], handler)
+
+	return nil
 }
 
 // Opener returns the opener of the target.
--- a/common/page.go
+++ b/common/page.go
@@ -14,6 +14,7 @@ import (
 
 	k6modules "go.k6.io/k6/js/modules"
 
+	"github.com/chromedp/cdproto"
 	"github.com/chromedp/cdproto/cdp"
 	"github.com/chromedp/cdproto/dom"
 	"github.com/chromedp/cdproto/emulation"
@@ -148,6 +149,8 @@ func NewPage(
 	p.Mouse = NewMouse(ctx, s, p.frameManager.MainFrame(), bctx.timeoutSettings, p.Keyboard)
 	p.Touchscreen = NewTouchscreen(ctx, s, p.Keyboard)
 
+	p.initEvents()
+
 	action := target.SetAutoAttach(true, true).WithFlatten(true)
 	if err := action.Do(cdp.WithExecutor(p.ctx, p.session)); err != nil {
 		return nil, fmt.Errorf("internal error while auto attaching to browser pages: %w", err)
@@ -163,6 +166,40 @@ func NewPage(
 	}
 
 	return &p, nil
+}
+
+func (p *Page) initEvents() {
+	p.logger.Debugf("Page:initEvents",
+		"sid:%v tid:%v", p.session.ID(), p.targetID)
+
+	events := []string{
+		cdproto.EventRuntimeConsoleAPICalled,
+	}
+	p.session.on(p.ctx, events, p.eventCh)
+
+	go func() {
+		p.logger.Debugf("Page:initEvents:go",
+			"sid:%v tid:%v", p.session.ID(), p.targetID)
+		defer p.logger.Debugf("Page:initEvents:go:return",
+			"sid:%v tid:%v", p.session.ID(), p.targetID)
+
+		for {
+			select {
+			case <-p.session.Done():
+				p.logger.Debugf("Page:initEvents:go:session.done",
+					"sid:%v tid:%v", p.session.ID(), p.targetID)
+				return
+			case <-p.ctx.Done():
+				p.logger.Debugf("Page:initEvents:go:ctx.Done",
+					"sid:%v tid:%v", p.session.ID(), p.targetID)
+				return
+			case event := <-p.eventCh:
+				if ev, ok := event.data.(*cdpruntime.EventConsoleAPICalled); ok {
+					p.onConsoleAPICalled(ev)
+				}
+			}
+		}
+	}()
 }
 
 func (p *Page) closeWorker(sessionID target.SessionID) {
@@ -1070,6 +1107,66 @@ func (p *Page) Workers() []api.Worker {
 	return workers
 }
 
+func (p *Page) onConsoleAPICalled(event *cdpruntime.EventConsoleAPICalled) {
+	// If there are no handlers for EventConsoleAPICalled, return
+	p.eventHandlersMu.RLock()
+	if _, ok := p.eventHandlers[eventPageConsoleAPICalled]; !ok {
+		p.eventHandlersMu.RUnlock()
+		return
+	}
+	p.eventHandlersMu.RUnlock()
+
+	m, err := p.consoleMsgFromConsoleEvent(event)
+	if err != nil {
+		p.logger.Errorf("Page:onConsoleAPICalled", "building console message: %v", err)
+		return
+	}
+
+	p.eventHandlersMu.RLock()
+	defer p.eventHandlersMu.RUnlock()
+	for _, h := range p.eventHandlers[eventPageConsoleAPICalled] {
+		h := h
+		if err := h(m); err != nil {
+			p.logger.Errorf("Page:onConsoleAPICalled", "executing handler: %v", err)
+		}
+	}
+}
+
+func (p *Page) consoleMsgFromConsoleEvent(e *cdpruntime.EventConsoleAPICalled) (*api.ConsoleMessage, error) {
+	execCtx, err := p.executionContextForID(e.ExecutionContextID)
+	if err != nil {
+		return nil, err
+	}
+
+	var (
+		l = p.logger.WithTime(e.Timestamp.Time()).
+			WithField("source", "browser").
+			WithField("browser_source", "console-api")
+
+		objects       = make([]any, 0, len(e.Args))
+		objectHandles = make([]api.JSHandle, 0, len(e.Args))
+	)
+
+	for _, robj := range e.Args {
+		i, err := parseRemoteObject(robj)
+		if err != nil {
+			handleParseRemoteObjectErr(p.ctx, err, l)
+		}
+
+		objects = append(objects, i)
+		objectHandles = append(objectHandles, NewJSHandle(
+			p.ctx, p.session, execCtx, execCtx.Frame(), robj, p.logger,
+		))
+	}
+
+	return &api.ConsoleMessage{
+		Args: objectHandles,
+		Page: p,
+		Text: textForConsoleEvent(e, objects),
+		Type: e.Type.String(),
+	}, nil
+}
+
 // executionContextForID returns the page ExecutionContext for the given ID.
 func (p *Page) executionContextForID( //nolint:unused
 	executionContextID cdpruntime.ExecutionContextID,
@@ -1093,4 +1190,31 @@ func (p *Page) sessionID() (sid target.SessionID) {
 		sid = p.session.ID()
 	}
 	return sid
+}
+
+// textForConsoleEvent generates the text representation for a consoleAPICalled event
+// mimicking Playwright's behavior.
+func textForConsoleEvent(e *cdpruntime.EventConsoleAPICalled, args []any) string {
+	if e.Type.String() == "dir" || e.Type.String() == "dirxml" ||
+		e.Type.String() == "table" {
+		if len(e.Args) > 0 {
+			// These commands accept a single arg
+			return e.Args[0].Description
+		}
+		return ""
+	}
+
+	// args is a mix of string and non strings, so using fmt.Sprint(args...)
+	// might not add spaces between all elements, therefore use a strings.Builder
+	// and handle format and concatenation
+	var b strings.Builder
+	for i, a := range args {
+		format := " %v"
+		if i == 0 {
+			format = "%v"
+		}
+		b.WriteString(fmt.Sprintf(format, a))
+	}
+
+	return b.String()
 }
--- a/common/page.go
+++ b/common/page.go
@@ -8,12 +8,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/grafana/xk6-browser/api"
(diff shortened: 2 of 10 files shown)


common/page.go
```diff
@@ -70,13 +78,19 @@ type Page struct {
...
+	eventHandlersMu sync.RWMutex
+
 	mainFrameSession *FrameSession
 	frameSessions    map[cdp.FrameID]*FrameSession
 	frameSessionsMu  sync.RWMutex
 	workers          map[target.SessionID]*Worker
 	routes           []api.Route
 	vu               k6modules.VU
 
+	tq *taskqueue.TaskQueue
```
ka3de, line 92: A couple of comments about the usage of the `TaskQueue`:

1. The `TaskQueue` must be closed in order to let the event loop finish and therefore the iteration to finish. Considering that now the browser is abstracted from the JS API, this will only happen if the actual chrome instance exits for some reason, or if page is explicitly closed in the test. Which means that users have to call `page.close()` to let the iteration finish. I don't see other event that would allow us to close the task queue safely (see point 2 also). One side effect of this is that requiring the explicit `page.close()` we can guarantee also the Web Vitals "force" measurements.
The only improvement that I can think of about this is to only initialize the `TasQueue` when/if `page.on` is called. Therefore the user would only be required to close the page if using this method. But this might make it more confusing overall than always requiring to close the page...
2. We had a brief discussion on making the `TaskQueue` a global entity, possibly being handled at the mapping layer/registry. In practice I think this is not an option, as there is not an event that we can hook to in order to close the queue. The only event that would make sense is the `IterEnd` event, but this will not be fired from k6 until the iteration ends, and the iteration will not end until the event loop ends, which requires the `TaskQueue` to be closed.
All in all I think it makes sense to use it locally in the `page` implementation, especially now that we only have one use case for it in the whole codebase, and also considering how easy it us to use it.

tests/page_test.go
```diff
@@ -703,6 +706,233 @@ func TestPageClose(t *testing.T) {
...
+				}
+				return nil
+			}
+
+			err := tb.vu.Loop.Start(func() error {
+				r := tb.vu.Loop.RegisterCallback()
+				go func() {
+					// Wait some time for the browser to
+					// generate the onConsoleAPICalled event
+					time.Sleep(200 * time.Millisecond)
```
inancgumus, line 914: This can be racy (although rare, but it can be environment-specific). What do you think about using a channel instead? Let the `eventHandler` close the chan and the goroutine wait for it to be closed before closing the task queue.

common/page.go
```diff
@@ -1039,8 +1116,74 @@ func (p *Page) Workers() []api.Worker {
...
+		p.tq.Queue(func() error {
+			if err := h(m); err != nil {
+				return fmt.Errorf("executing onConsoleAPICalled handler: %w", err)
+			}
+			return nil
+		})
+	}
+}
+
+func (p *Page) consoleMssgFromConsoleEvent(e *cdpruntime.EventConsoleAPICalled) (*ConsoleMessage, error) {
```
inancgumus, line 1148:
```suggestion
func (p *Page) consoleMsgFromConsoleEvent(e *cdpruntime.EventConsoleAPICalled) (*ConsoleMessage, error) {
```

common/page.go
```diff
@@ -1039,8 +1116,74 @@ func (p *Page) Workers() []api.Worker {
...
+func (p *Page) onConsoleAPICalled(event *cdpruntime.EventConsoleAPICalled) {
+	// If there are no handlers for EventConsoleAPICalled, return
+	p.eventHandlersMu.RLock()
+	if _, ok := p.eventHandlers[eventPageConsoleAPICalled]; !ok {
+		p.eventHandlersMu.RUnlock()
+		return
+	}
+	p.eventHandlersMu.RUnlock()
+
+	m, err := p.consoleMssgFromConsoleEvent(event)
```
inancgumus, line 1128:
```suggestion
	m, err := p.consoleMsgFromConsoleEvent(event)
```

common/page.go
inancgumus: I think this might be okay for now. 👍

Perhaps, if we consider even a single use case, there might be an opportunity later on to shift this task queue and event-loop-related logic into the mapping layer. For example, we [moved the promises (event-loop-related logic) to the mapping layer](https://github.com/grafana/xk6-browser/issues/683), which helped in simplifying our tests and code. We're also [looking into the idea of moving the Goja-related mechanics](https://github.com/grafana/xk6-browser/issues/271) away from the business logic to the mapping layer, aiming for improved maintainability. I hope we're continuing in the direction of progress.

inancgumus (approved):
Nice work 👏 Thank you for working on this challenging task and delivering it quite remarkably 🙇 I believe we might want to refactor the `TaskQueue` part later on, but it's fine for now.

common/page.go
ankur22: I think you've done a great job on incorporating the taskqueue into k6 browser 👏 Looks super useful in general.

My main worry is that we're now relying on the user to use `page.close`, which is probably an ok requirement, but it feels like a poor DX if k6 hangs and times out (i'm assuming it's a timeout) if the user forgets to work with `page.close`. It would be great if we could handle the lifecycle of the eventqueue ourselves, but considering the taskqueue must be closed before the iterEnd can be fired by k6, I can't see how this can be done 🤔 

Could we have a taskqueue for the whole test run and wait for the `Exit` event (although doesn't feel like the correct fit)? I'm presuming the `Exit` event will fire regardless of whether the taskqueue is closed 🤷

tests/page_test.go
```diff
@@ -703,6 +706,233 @@ func TestPageClose(t *testing.T) {
...
+			name:      "on console.count (default label)",
+			consoleFn: "() => console.count()", // default label
+			assertFn: func(cm *common.ConsoleMessage) bool {
+				return cm.Type() == "count" && //nolint:goconst
+					cm.Text() == "default: 1" &&
+					cm.Page().URL() == blankPage
+			},
+		},
+		{
+			name:      "on console.count (default label)",
```
ankur22, line 861: Is this the same as the test above?

examples/pageon.js
```diff
@@ -0,0 +1,38 @@
...
+  const page = browser.newPage();
+  
+  try {
+    await page.goto('https://test.k6.io/');
+
+    page.on('console', msg => {
+        check(msg, {
+            'assertConsoleMessageType': msg => msg.type() == 'log',
+            'assertConsoleMessageText': msg => msg.text() == 'this is a console.log message 42',
+            'assertConsoleMessageArgs': msg => msg.args()[0].jsonValue() == 'this is a console.log message',
```
ankur22, line 30: Is `msg.args()[1]` equal to `42`?

ankur22 (approved):
This is an awesome feature! Thanks for working on it.

I left some minor comments.

My only concern is around the requirement to need to work with `page.close` otherwise this is great 🙂 Maybe we can think over another solution or talk it through with k6 to see if they can help.

common/page.go
ka3de: We can reevaluate having the `TaskQueue` as a "global component", but again, I don't see how could we handle its closing in a better way.

> Could we have a taskqueue for the whole test run and wait for the Exit event

I believe here the problem is that the `Exit` event will not be fired until all iterations have ended, and the iteration won't end if the task queue is not closed. So we have the same issue as waiting for the `IterEnd`.

The only alternative that I could think of, to not have to use the `TaskQueue`, was to limit the `page.on` method to receive just a single `msg` per call. Or allow the user to define how many messages does he/she want to receive. In these cases I believe we could get rid of the task queue. However, I think these alternatives are worse than requiring the `page.close()` call.

tests/page_test.go
ka3de: I had a similar implementation previously, but using a `sync.WaitGroup`. The problem is that in that case the test could hang if the handler is not called for some reason, so decided to change it to use the current implementation, which pretty much waits fro a TO of 200ms. Do you think it would be better to wait indefinitely?

inancgumus: I think it would be better to let the tests time out due to `go test` time limit if the handlers do not get called rather than to have possible flaky tests. Have you considered using a timeout context instead?

api/console_message.go
```diff
@@ -0,0 +1,9 @@
+package api
+
+// ConsoleMessage represents a page console message.
+type ConsoleMessage interface {
```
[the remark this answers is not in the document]
ka3de: I knew this was gonna be a comment, I had doubts myself :smile: 
Because we've always had this idea of possibly providing other browser implementations I thought that it was better to stick with the current pattern of working with interfaces defined in the API until we take a final decision. But I'm OK moving this to a struct.

common/page.go
```diff
@@ -165,6 +168,40 @@ func NewPage(
...
+	go func() {
+		p.logger.Debugf("NewPage:initEvents:go",
+			"sid:%v tid:%v", p.session.ID(), p.targetID)
+		defer p.logger.Debugf("NewPage:initEvents:go:return",
+			"sid:%v tid:%v", p.session.ID(), p.targetID)
+
+		for {
+			select {
+			case <-p.session.Done():
+				p.logger.Debugf("Page:initEvents:go:session.done",
```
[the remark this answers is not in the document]
ka3de: I mimicked the pattern defined in `FrameSession` [here](https://github.com/grafana/xk6-browser/blob/6d2e50b184e85a0725261dbdf15ca144eeda7718/common/frame_session.go#L212-L226). Personally I would change all these logs to `Page:...` as this is a background goroutine, even if it's initialized from `NewPage` method. WDYT?

examples/pageon.js
ka3de: Yes, should we add it to the example?

common/page.go
```diff
@@ -1070,6 +1107,67 @@ func (p *Page) Workers() []api.Worker {
...
+	defer p.eventHandlersMu.RUnlock()
+	for _, h := range p.eventHandlers[eventPageConsoleAPICalled] {
+		if err := h(m); err != nil {
+			p.logger.Errorf("Page:onConsoleAPICalled", "executing handler: %v", err)
+		}
+	}
+}
+
+func (p *Page) consoleMssgFromConsoleEvent(e *cdpruntime.EventConsoleAPICalled) (*ConsoleMessage, error) {
+	// Find execution context for the event
```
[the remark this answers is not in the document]
ka3de: Removed it [here](https://github.com/grafana/xk6-browser/blob/f5abd5d865238b52bc046a18ed082dcdfe4ed138/common/page.go#L1149-L1154).

common/page.go
```diff
@@ -1070,6 +1107,67 @@ func (p *Page) Workers() []api.Worker {
...
+	execCtx, err := p.executionContextForID(e.ExecutionContextID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Parse event args
```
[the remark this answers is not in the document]
ka3de: Removed it [here](https://github.com/grafana/xk6-browser/blob/f5abd5d865238b52bc046a18ed082dcdfe4ed138/common/page.go#L1149-L1154).

common/page.go
ka3de: Force pushed with the change in https://github.com/grafana/xk6-browser/pull/1006/commits/12f6f4a6038e20c250fac2498f3cc37ecdf25b6a. See https://github.com/grafana/xk6-browser/blob/f5abd5d865238b52bc046a18ed082dcdfe4ed138/common/page.go#L1148.

common/page.go
ka3de: Force pushed with the change in https://github.com/grafana/xk6-browser/pull/1006/commits/12f6f4a6038e20c250fac2498f3cc37ecdf25b6a. See https://github.com/grafana/xk6-browser/blob/f5abd5d865238b52bc046a18ed082dcdfe4ed138/common/page.go#L1148.

common/page.go
```diff
@@ -1070,6 +1107,67 @@ func (p *Page) Workers() []api.Worker {
...
+	p.eventHandlersMu.RLock()
+	defer p.eventHandlersMu.RUnlock()
+	for _, h := range p.eventHandlers[eventPageConsoleAPICalled] {
+		if err := h(m); err != nil {
+			p.logger.Errorf("Page:onConsoleAPICalled", "executing handler: %v", err)
+		}
+	}
+}
+
+func (p *Page) consoleMssgFromConsoleEvent(e *cdpruntime.EventConsoleAPICalled) (*ConsoleMessage, error) {
```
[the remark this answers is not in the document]
ka3de: Force pushed with the change in https://github.com/grafana/xk6-browser/pull/1006/commits/12f6f4a6038e20c250fac2498f3cc37ecdf25b6a. See https://github.com/grafana/xk6-browser/blob/f5abd5d865238b52bc046a18ed082dcdfe4ed138/common/page.go#L1148.

tests/page_test.go
ka3de: Changed the test to wait for both handlers to be called or up to a more than reasonable timeout. Also added testing for messages multiplexing. Nice catch finding an issue with that!
See https://github.com/grafana/xk6-browser/pull/1006/commits/e4f4a30efb26d911a3704390cd547cfdbc48a237

examples/pageon.js
ka3de: Added it in [de28297](https://github.com/grafana/xk6-browser/pull/1006/commits/de282976e05a05de8453fef13d433d0da92a84e6).

tests/page_test.go
inancgumus: Nice 👍 I suggest heeding this:

[WaitGroup.Add](https://pkg.go.dev/sync#WaitGroup.Add):

> Note that calls with a positive delta that occur when the counter is zero must happen before a Wait. Calls with a negative delta, or calls with a positive delta that start when the counter is greater than zero, may happen at any time. Typically this means the calls to Add should execute before the statement creating the goroutine or other event to be waited for. If a WaitGroup is reused to wait for several independent sets of events, new Add calls must happen after all previous Wait calls have returned. See the WaitGroup example.

Since we're calling `wg.Add` **after** `wg.Wait()`, we might get across with unintended behavior:

```go
err := tb.vu.Loop.Start(func() error {
	// ...
	go func() {
		// ...
		go func() {
			wg.Wait() // <----
			close(done)
		}()

		select {
		case <-done:
		case <-time.After(testTO):
			assertTO = true
		}

		r(func() error {
			tb.Close()
			return nil
		})
	}()
	wg.Add(1) // <--- AFTER
	err := p.On("console", eventHandlerOne)
	if err != nil {
		return err
	}
	wg.Add(1) // <--- AFTER
	err = p.On("console", eventHandlerTwo)
	// ...
})
```

ka3de: Yep, just realized this, fixed in [f72e9c1](https://github.com/grafana/xk6-browser/pull/1006/commits/f72e9c1269cb84531a0c8f9e9126cfbc6f8a843f).

inancgumus: It's not unexpected to deal with this stuff when we incorporate [event-loop handling in the business logic](https://github.com/grafana/xk6-browser/pull/1006#discussion_r1305513979) :(

common/page.go
inancgumus: > We had a brief discussion on making the TaskQueue a global entity, possibly being handled at the mapping layer/registry. In practice I think this is not an option, as there is not an event that we can hook to in order to close the queue.
> I don't see how could we handle its closing in a better way.

What about adding the task queue to the registry, then in the mapping layer, creating a new task queue on a new page(s) call(s), queue `page.On` handlers there, and closing the task queue when `page.close` is called? We might want to make it page specific, though. And, this isn't lifting the `page.Close` problem, though. For that, it seems we need k6-core's help, as I cannot see another way to do this either :(

ka3de: > What about adding the task queue to the registry (as a field in module vu), then in the mapping layer, creating a new task queue on a new page(s) call(s), queue page.On handlers there, and closing the task queue when page.close is called?

I see pros and cons of this. From one side we would remove the usage of the `TaskQueue` (wich is an element mainly related with Goja) from our core implementation, which is something we are trying to do. But at the same time I think it would increase the complexity of the codebase. Especially considering how easy it has been to use the `TaskQueue` in our `page` implementation, as it basically adds "3" lines of code. For example, I think a similar situation is the one that we faced with the mapping for the `page.on` events. In the end this was fixed like [this](https://github.com/grafana/xk6-browser/blob/1de4e41f70cb5a3c9dfabd71f3640631a9696e2e/browser/mapping.go#L510-L517), which I think is quite complex to understand, but in this case there was no better solution.

For the options that we have considered, I would keep the usage of `TaskQueue` at the `page` implementation by now, unless we can find a global implementation that will cover more use cases (as now we have only one), or another implementation that would allow us to not require an explicit `page.close` call.

inancgumus: Yes, Goja is complex, and event-loop is even more; that's why I prefer to keep it away from the actual code. Not only does the code, but also complicates the tests a lot, which we see [here](https://github.com/grafana/xk6-browser/pull/1006#discussion_r1305498962) and [here](https://github.com/grafana/xk6-browser/pull/1006#discussion_r1305513979). It takes away the expressiveness of tests and adds a lot of boilerplate. As I said, I'm fine with this for now until we find a better solution.

common/page.go
```diff
@@ -153,6 +171,48 @@ func NewPage(
...
+		p.logger.Debugf("NewPage:initEvents:go",
+			"sid:%v tid:%v", p.session.ID(), p.targetID)
+		defer func() {
+			p.logger.Debugf("NewPage:initEvents:go:return",
+				"sid:%v tid:%v", p.session.ID(), p.targetID)
+			// TaskQueue is only initialized when calling page.on() method
+			// so users are not always required to close the page in order
+			// to let the iteration finish.
+			if p.tq != nil {
+				p.tq.Close()
```
ka3de, line 193: In regards of requiring to close the page so the `TaskQueue` is closed and therefore the iteration can end. I've changed this in [60a94f0](https://github.com/grafana/xk6-browser/pull/1006/commits/60a94f0b3a4c9f0be2826bb62eef47a583ffa422) so the `TaskQueue` is only initialized when calling `page.on()` method, and therefore only has to be closed when using this method. I think this is a better way, to not introduce more breaking changes by now, and this particular case can be documented in the `page.on` method description.
Also I think requiring the user to close the page only when using this method makes more sense, as `page.on` defines a function that will be executed for every event (in this case only `console` events) throughout the page's life, so it's easier to understand that the page has to be closed so the handler execution stops and the iteration can finish.

WDYT @inancgumus @ankur22 ?

inancgumus: I think this is an excellent idea from the usability perspective 👍 As you suggested, we can mention this in a notice box in the documentation 👍

api/console_message.go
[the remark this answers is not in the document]
ka3de: I have changed `ConsoleMessage` to be defined as a struct in API pkg here: [21c7a5d](https://github.com/grafana/xk6-browser/pull/1006/commits/21c7a5d055f8c11a732a291c0d641efde118fa91). I have also amended the posterior commits to adapt to this change. The only relevant modification happens in [94fbd6c](https://github.com/grafana/xk6-browser/pull/1006/commits/94fbd6c4fcada8568cce9b1914d0869a1fef99ff) as we have to define "allowed redundant mappings" in the mapping layer tests, as `page()`, `text()` and `type()` methods have been defined in the mapping layer explicitly in order to comply with Playwright's API.
cc: @inancgumus

common/page.go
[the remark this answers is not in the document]
ka3de: Changed this to `Page:...`. See [here](https://github.com/grafana/xk6-browser/blob/2548cc33185c0bc01dd88297a04ca8ba3f4efa88/common/page.go#L184-L206).

api/console_message.go
```diff
@@ -0,0 +1,16 @@
...
+// · Type is the type of the console message. It can be one of 'log',
+// 'debug', 'info', 'error', 'warning', 'dir', 'dirxml', 'table', 'trace',
+// 'clear', 'startGroup', 'startGroupCollapsed', 'endGroup', 'assert', 'profile',
+// 'profileEnd', 'count', 'timeEnd'.
+type ConsoleMessage struct {
+	Args []JSHandle
+	Page Page
+	Text string
+	Type string
+}
```
inancgumus, line 16: This comment will render the docs like this:

Screenshot 2023-08-28 at 13 01 05

I believe the following suggestion makes it better readable on the docs and in the IDEs.

```suggestion
// ConsoleMessage represents a page console message.
type ConsoleMessage struct {
	// Args represent the list of arguments passed to a console function call.
	Args []JSHandle

	// Page is the page that produced the console message, if any.
	Page Page

	// Text represents the text of the console message.
	Text string

	// Type is the type of the console message.
	//
	// It can be one of log, debug, info, error, warning, dir, dirxml,
	// table, trace, clear, startGroup, startGroupCollapsed, endGroup,
	// assert, profile, profileEnd, count, timeEnd.
	Type string
}
```

For example:
Screenshot 2023-08-28 at 13 04 27

api/console_message.go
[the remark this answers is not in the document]
inancgumus: I think these methods are not redundant and should be tested if they conform to an API. Since there is not such an API in the `api` package, we can solidify its protocol in tests instead. Like this:

```
type consoleMessageAPI interface {
	Args() []api.JSHandle
	Page() api.Page
	Text() string
	Type() string
}
```

Then in the test:

```
"mapConsoleMessage": {
	apiInterface: (*consoleMessageAPI)(nil), // <---
	mapp: func() mapping {
		return mapConsoleMessage(moduleVU{VU: vu}, &api.ConsoleMessage{})
	},
},
```

I believe this is also a good step in making progress for issue #898, and also provides a solution for my earlier question: "How the testing of the mapping layer is affected?".

inancgumus (commented):
One suggestion about documentation 👍

api/console_message.go
ka3de: I was doubtful what was better for Godoc as there are examples for both approaches, but this LGTM :+1: 
Changed in [eac5e96](https://github.com/grafana/xk6-browser/pull/1006/commits/eac5e96810feeb5828d47e3215532a2574d45c98).

api/console_message.go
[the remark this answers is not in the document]
ka3de: > Since no such an API exists in the api package, we can solidify the ConsoleMessage protocol in tests instead.

Changed in [3eff159](https://github.com/grafana/xk6-browser/pull/1006/commits/3eff159e72b1609ca8e117057f3fcab8b48075a3). I set this as an anonymous function by now, as this is the only occurrence. If we ever remove the whole API pkg we can define the set of interfaces separately just for the mapping layer testing.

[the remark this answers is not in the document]
inancgumus: Perfect!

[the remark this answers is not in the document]
inancgumus: Perfect! The only downside of an _anonymous interface_ is we'll get no type information when the mapping tests fail. I think we can declare a new type for `ConsoleMessage` in the same test to prevent this.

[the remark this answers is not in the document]
ka3de: > I realized now that the only downside of an anonymous interface is we'll get no type information when the mapping tests fail

I found it strange that the subtest name was not referenced on failures, so fixed small issues in #1014. With that there is also no need to reference the type name in the error messages, as it is already identified through the subtest name.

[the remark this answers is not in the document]
inancgumus: Thanks for fixing this. If we had declared a named type, this wouldn't be needed. It's okay this way, too, but I wonder what would happen if we wanted to add custom mappings for the anonymous type? A named type makes more sense.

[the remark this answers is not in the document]
ka3de: > If we had declared a named type, this wouldn't be needed, right? It's okay this way, too, but I wonder what would happen if we wanted to add custom mappings for the anonymous type?

I see this the other way around. If the table tests already define a name for the subtest, there is no need to use the type name, as that was only used to identify the failing test, but that is already referenced through the subtest name.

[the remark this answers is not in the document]
inancgumus: Yep, but custom mappings wouldn't get along nicely with that anonymous type. Anyway, we can take care of it when the time comes 👍

common/page.go
ankur22: I'm also happy with the current solution, thanks @ka3de!

ankur22 (approved):
LGTM! This is looking great. Thanks for taking the feedback onboard.

inancgumus (approved):
Awesome work 👏 Thanks for considering our suggestions and applying them nicely!

PR #1007, opened 2023-08-21
Add core to inject k6 object into window

ankur22:
## What?

The change injects `window.k6 = {};` into all browser contexts and subsequent `page`s that are created.

## Why?

In some cases we want to pass information to other apps to allow them to infer who is performing requests against a website. For now we're exposing an empty object, but we hope to extend the object with meta data that will be useful for other apps that have been instrumented on the website.

## Test

To test this change you can run the following k6 script, which will give you enough time to go into the chrome dev tools console and type in `window.k6` which should return an empty object (`{}`).

```js
import { browser } from 'k6/experimental/browser';
import { sleep } from 'k6';

export const options = {
  scenarios: {
    browser: {
      options: {
        browser: {
            type: 'chromium',
        },
      },
    },
  },
}

  await page.goto("https://test.k6.io");

  sleep(20);

  page.close();
  context.close();
}
```

Screenshot 2023-08-21 at 14 47 18
Screenshot 2023-08-21 at 14 47 22

## Checklist

Closes: https://github.com/grafana/xk6-browser/issues/1017


common/browser_context.go
```diff
@@ -26,6 +26,9 @@ import (
 var (
 	_ EventEmitter       = &BrowserContext{}
 	_ api.BrowserContext = &BrowserContext{}
+
+	//nolint:gochecknoglobals
+	k6ObjScript = "window.k6 = {};"
```
inancgumus, line 31: Is there a particular reason we're using a global variable? Doing this slows down our tests because they can't run at the same time. Just as a heads-up, the current test setup takes a second longer than it could.

You probably already have thought about this, but what if we move `k6ObjScript` into the `Browser` type and make the script content a constant? This way, we can speed up tests and the module stays tidy between runs. Another option could be to use `sync.Once` in `BrowserContext`. What do you think?

common/browser_context_test.go
```diff
@@ -27,18 +27,22 @@ func TestNewBrowserContext(t *testing.T) {
...
 		for _, script := range bc.evaluateOnNewDocumentSources {
 			switch script {
 			case js.WebVitalIIFEScript:
 				webVitalIIFEScriptFound = true
 			case js.WebVitalInitScript:
 				webVitalInitScriptFound = true
+			case k6ObjScript:
+				k6ObjScriptFound = true
 			default:
 				assert.Fail(t, "script is neither WebVitalIIFEScript nor WebVitalInitScript")
```
inancgumus, line 40:
```suggestion
				assert.Fail(t, "script is neither WebVitalIIFEScript, WebVitalInitScript, nor k6ObjScript")
```

inancgumus (commented):
Looks good to me overall, and I'm excited about this feature! 😊

I do have a few points for us to chat about. Also, it might be helpful to include some comments in the code to clarify the role of `k6ObjScript`.

common/browser_context.go
ka3de: I like @inancgumus considerations. I would also consider here for consistency that we do already have other scripts that are injected just like this one. E.g.: `js.WebVitalIIFEScript` and `js.WebVitalInitScript` and other. In these cases these are defined as global vars which content is loaded from an external JS file. 

We don't necessarily need to tackle this on this PR, but I think it would be good to be consistent on how these JS "injections" are done.

inancgumus: I assumed `k6ObjScript` is mutable unlike the embedded files, which is a Go feature that should be done that way.

ankur22: 🤦 I totally missed this and was only meant to a global while testing, it wasn't meant to stay as a global. Good spot!

> I assumed k6ObjScript is mutable, unlike the embedded files, which is a Go feature that should be done that way. I believe we can still fetch k6ObjScript's content from an embedded file to be consistent with WebVitalIIFEScript, and the other one.

Yeah, this is true (and why I was testing it as a global var), it was meant to be mutable so that we can add unique metadata which is associated to a new test run.

I like the idea of it being another file like we have for the other web vital js scripts that we're injecting in at the same time, just for consistency. We may want to consolidate them all into one large script in the future to reduce the chatter too. I'd rather tackle the issue of how we're going to mutate the `k6ObjScript` in another PR when we actually need to do that.

inancgumus: No worries! 🙇🙂

inancgumus: > Can't the current TestK6Object run in parallel? I don't see anything that wouldn't allow us to do so, I tested it and seems to be working fine.

Sure, it can, because we don't mutate it yet. But, it's meant to be mutated, like soon.

ankur22: Resolved in: 45bdce84c276c2e35c9ab34dfaced0cc39641543 & a6d5a15d83313186fbb0b7f819533d21eed6bd30

inancgumus: #1019 doesn't know about this PR's `TestK6Object` yet. I'm working on it off of the current `main` (447c8c2). I don't understand what you mean, sorry :(

ankur22: > https://github.com/grafana/xk6-browser/pull/1019 doesn't know about this PR's TestK6Object. I'm working on it from the current main (https://github.com/grafana/xk6-browser/commit/447c8c2ded4bff552f1fd21f9246f45d211065e6). It's not related to this PR in any way. I don't understand what you mean, sorry :(

In https://github.com/grafana/xk6-browser/pull/1019 I believe you've made `TestNewBrowserContext` parallel, which this PR amends, but i could make that test parallel in this PR, but I won't.

inancgumus: OK, sure. I made all tests parallel regardless of that particular test. But if I can make `TestNewBrowserContext` sequential in #1019, that's what you want.

inancgumus (approved):
Nice job, thanks!

ka3de (approved):
LGTM :clap:

common/browser_context.go
inancgumus: TestNewBrowserContext is intentionally not made parallel in https://github.com/grafana/xk6-browser/pull/1019 since it uses globals :)

ankur22: Aha, thanks for clarifying 👍

inancgumus: Jot it down here in case we might be referencing it later on 😅

PR #1009, opened 2023-08-23
Add `Page.executionContextForID`

ka3de:
## What?

Adds support to retrieve the associated execution context from within all page frames for a given ID.

## Why?

This is required to associate CDP runtime events for a given page, which include the associated execution context ID (e.g.: see [onConsoleAPICalled](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#event-consoleAPICalled)), with the execution context itself among the page frames and their sessions.

## Checklist

## Related PR(s)/Issue(s)

Related #1006 .

### DIFF
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -1084,3 +1084,16 @@ func (fs *FrameSession) updateViewport() error {
 
 	return nil
 }
+
+func (fs *FrameSession) executionContextForID(
+	executionContextID cdpruntime.ExecutionContextID,
+) (*ExecutionContext, error) {
+	fs.contextIDToContextMu.Lock()
+	defer fs.contextIDToContextMu.Unlock()
+
+	if exc, ok := fs.parent.contextIDToContext[executionContextID]; ok {
+		return exc, nil
+	}
+
+	return nil, fmt.Errorf("no execution context found for id: %v", executionContextID)
+}
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -1085,7 +1085,7 @@ func (fs *FrameSession) updateViewport() error {
 	return nil
 }
 
-func (fs *FrameSession) executionContextForID(
+func (fs *FrameSession) executionContextForID( //nolint:unused
 	executionContextID cdpruntime.ExecutionContextID,
 ) (*ExecutionContext, error) {
 	fs.contextIDToContextMu.Lock()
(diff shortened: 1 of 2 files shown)


inancgumus (approved):
LGTM, but I wish we had some tests to validate that this really works without deadlocks and data races. I'm saying this because execution contexts and frame sessions have been the most challenging parts of the module to debug, and they used to fail a lot.

ka3de:
> LGTM, but I wish we had some tests to validate that this really works without deadlocks and data races. I'm saying this because execution contexts and frame sessions have been the most challenging parts of the module to debug, and they used to fail a lot. It'd nice to have `nolint` for the current linter warnings and then we can lift them in the other `console.log` PR.

Just added the `nolint` instructions for the unused methods until we complete #1006. See [9318ba9](https://github.com/grafana/xk6-browser/pull/1009/commits/9318ba9af787263b2b586223cb150162a77c00cd).

In regards of tests, I thought about it, but a unit test was providing pretty much no value, and it would be fairly difficult to try to force a data race or a deadlock for frame attach events only in an integration test (as in this PR these new methods are not even called). Isn't this better suited for e2e tests once we also process events and have a better "platform" for them?

inancgumus (dismissed):

common/frame_session.go
```diff
@@ -1084,3 +1084,18 @@ func (fs *FrameSession) updateViewport() error {
...
+	executionContextID cdpruntime.ExecutionContextID,
+) (*ExecutionContext, error) {
+	fs.contextIDToContextMu.Lock()
+	defer fs.contextIDToContextMu.Unlock()
+
+	for id, exc := range fs.contextIDToContext {
+		if id == executionContextID {
+			return exc, nil
+		}
+	}
```
ankur22, line 1098: Why are we ranging over the `contextIDToContext` map?

ankur22 (dismissed):
LGTM.

Just one minor comment about ranging over the map instead of accessing with the key.

common/frame_session.go
inancgumus: Oh, I missed this 👍 I also wonder why :)

ka3de: Force pushed applying the change at [4136588](https://github.com/grafana/xk6-browser/pull/1009/commits/4136588f95c2462bb64caf9b05132b23f436af5f)

inancgumus (approved):

PR #1014, opened 2023-08-29
Fix mapping tests

ka3de:
## What?

Fixes the mapping layer tests in order to use the subtest `testing.T` reference so the subtest name is identified on failure and therefore there is no need to reference the specific type through reflection.

Previous error message on a subtest failure:
```
--- FAIL: TestMappings (0.00s) // --> Does not reference the specific subtest
    /home/ka3de/devel/go/src/github.com/grafana/xk6-browser/browser/mapping_test.go:111: method on for Browser not found
FAIL
FAIL	github.com/grafana/xk6-browser/browser	0.008s
FAIL
```

Current error message:
```
--- FAIL: TestMappings (0.00s)
    --- FAIL: TestMappings/browser (0.00s)
        /home/ka3de/devel/go/src/github.com/grafana/xk6-browser/browser/mapping_test.go:194: method "on" not found
FAIL
FAIL	github.com/grafana/xk6-browser/browser	0.007s
FAIL
```
## Why?

Previous behavior was not correct and required the actual type name specification in the error message through reflection in order to understand which was the failing test.

## Checklist

### DIFF
--- a/browser/mapping_test.go
+++ b/browser/mapping_test.go
@@ -78,10 +78,9 @@ func TestMappings(t *testing.T) {
 		t.Helper()
 
 		var (
-			typ     = reflect.TypeOf(tt.apiInterface).Elem()
-			typName = typ.Name()
-			mapped  = tt.mapp()
-			tested  = make(map[string]bool)
+			typ    = reflect.TypeOf(tt.apiInterface).Elem()
+			mapped = tt.mapp()
+			tested = make(map[string]bool)
 		)
 		for i := 0; i < typ.NumMethod(); i++ {
 			method := typ.Method(i)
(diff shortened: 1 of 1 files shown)


browser/mapping_test.go
```diff
@@ -74,7 +74,9 @@ func TestMappings(t *testing.T) {
 	// testMapping tests that all the methods of an API are mapped
 	// to the module. And wildcards are mapped correctly and their
 	// methods are not mapped.
-	testMapping := func(tt test) {
+	testMapping := func(t *testing.T, tt test) {
```
inancgumus, line 77: Nice catch 👏

browser/mapping_test.go
```diff
@@ -109,15 +109,15 @@ func TestMappings(t *testing.T) {
 				m = cm
 			}
 			if _, ok := mapped[m]; !ok {
-				t.Errorf("method %s not found", m)
+				t.Errorf("method %q not found", m)
```
inancgumus, line 112: Better 👍 Quoted-strings!

inancgumus (approved):
LGTM 👍 This was a nice catch. IDK how we missed it. Thanks!

One downside of the anonymous interface is it won't get along with custom mappings. As I said [here](https://github.com/grafana/xk6-browser/pull/1006#discussion_r1308384887), we can take care of it later.

ankur22 (commented):
LGTM

PR #1016, opened 2023-08-30
Filter cookies by URLs

inancgumus:
## What?

1. Adds support for filtering cookies by URLs.
2. Improves tests and examples.

## Why?

Users can now filter cookies by URLs. See the `example/cookies.js` file for an example.

## Checklist

## Related PR(s)/Issue(s)

Updates: #6

### DIFF
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -3,7 +3,9 @@ package common
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/grafana/xk6-browser/api"
@@ -511,7 +513,7 @@ func (b *BrowserContext) ClearCookies() error {
 // automatically taken from the browser context when it is created. And some of
 // them are set by the page, i.e., using the Set-Cookie HTTP header or via
 // JavaScript like document.cookie.
-func (b *BrowserContext) Cookies(urls ...string) ([]*api.Cookie, error) { //nolint:revive // TODO: remove
+func (b *BrowserContext) Cookies(urls ...string) ([]*api.Cookie, error) {
 	b.logger.Debugf("BrowserContext:Cookies", "bctxid:%v", b.id)
 
 	// get cookies from this browser context.
@@ -522,9 +524,8 @@ func (b *BrowserContext) Cookies(urls ...string) ([]*api.Cookie, error) { //noli
 		cdp.WithExecutor(b.ctx, b.browser.conn),
 	)
 	if err != nil {
-		return nil, fmt.Errorf("getting cookies from the browser context: %w", err)
+		return nil, fmt.Errorf("retrieving cookies: %w", err)
 	}
-
 	// return if no cookies found so we don't have to needlessly convert them.
 	// users can still work with cookies using the empty slice.
 	// like this: cookies.length === 0.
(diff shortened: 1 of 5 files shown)


inancgumus:
Before merging this, we should be fixing this bug.

```
--- FAIL: TestBrowserContextCookies (0.00s)
    --- FAIL: TestBrowserContextCookies/filter_no_cookies (0.22s)
        browser_context_test.go:545: 
            	Error Trace:	/home/runner/work/xk6-browser/xk6-browser/tests/browser_context_test.go:545
            	Error:      	Received unexpected error:
            	            	cannot recognize cookie values: could not convert struct value 0 to *cdp.TimeSinceEpoch for field Expires: could not convert 0 to cdp.TimeSinceEpoch
            	Test:       	TestBrowserContextCookies/filter_no_cookies
            	Messages:   	failed to add cookies to the browser context: []*api.Cookie{(*api.Cookie)(0xc000136a10), (*api.Cookie)(0xc000136af0)}
```

common/browser_context.go
```diff
@@ -546,6 +547,131 @@ func (b *BrowserContext) Cookies(urls ...string) ([]*api.Cookie, error) { //noli
 			SameSite: api.CookieSameSite(c.SameSite),
 		}
 	}
+	// filter cookies by the provided URLs.
+	cookies, err = filterCookies(cookies, urls...)
+	if err != nil {
+		return nil, fmt.Errorf("retrieving cookies: %w", err)
```
ankur22, line 553: How about:

```suggestion
		return nil, fmt.Errorf("filtering cookies: %w", err)
```

Just so that it's different from the error above and so possibly easier to debug?

common/browser_context.go
```diff
@@ -546,6 +547,131 @@ func (b *BrowserContext) Cookies(urls ...string) ([]*api.Cookie, error) { //noli
...
+	}
+
+	// remove all cookies after the last cookie that should be kept.
+	return cookies[:n], nil
+}
+
+// shouldFilterCookie determines whether a cookie should be kept,
+// based on its compatibility with a specific URL.
+// Returns true if the cookie should be kept, false otherwise.
+func shouldFilterCookie(c *api.Cookie, uri *url.URL) bool {
```
ankur22, line 623: `shouldFilterCookie` -> if `true` shouldn't it filter the cookie? I guess you could swap the logic around in the implementation or give it a different name such as `shouldKeepCookie` or `shouldNotFilterCookie`.

ankur22, line 623: Thanks for explaining how this works 🙇 I think the comments are very useful here!

common/browser_context.go
```diff
@@ -546,6 +547,131 @@ func (b *BrowserContext) Cookies(urls ...string) ([]*api.Cookie, error) { //noli
...
+	}
+	// if no cookies should be kept, return nil instead of
+	// an empty slice to conform with the API error behavior.
+	// also makes tests concise.
+	if n == 0 {
+		return nil, nil
+	}
+
+	// remove all cookies after the last cookie that should be kept.
+	return cookies[:n], nil
```
ankur22, line 617: Clever implementation 🙂 Thanks for the comment which gave a heads up before looking at the code, it was super useful 👍

common/browser_context.go
```diff
@@ -546,6 +547,131 @@ func (b *BrowserContext) Cookies(urls ...string) ([]*api.Cookie, error) { //noli
...
+	// dot turns it into .example.com, making the cookie valid
+	// for sub.example.com, another.example.com, etc.
+	domain := c.Domain
+	if !strings.HasPrefix(domain, ".") {
+		domain = "." + domain
+	}
+	// Confirm that the cookie's domain is a suffix of the URL's
+	// hostname, emulating how a browser would scope cookies to
+	// specific domains.
+	if !strings.HasSuffix("."+uri.Hostname(), domain) {
```
ankur22, line 636: Possibly not a big deal, but shouldn't this be the other way around:

```suggestion
	if !strings.HasSuffix(domain, "."+uri.Hostname()) {
```

So now I would read that as _if the cookie's domain doesn't match the expected filter uri hostname_... WDYT?

common/browser_context.go
```diff
@@ -546,6 +547,131 @@ func (b *BrowserContext) Cookies(urls ...string) ([]*api.Cookie, error) { //noli
...
+	// be treated as "/".
+	//
+	// See: https://datatracker.ietf.org/doc/html/rfc6265#section-5.1.4
+	if uri.Path == "" {
+		uri.Path = "/"
+	}
+	// Ensure that the cookie applies to the specific path of the
+	// URL, emulating how a browser would scope cookies to specific
+	// paths within a domain.
+	if !strings.HasPrefix(uri.Path, c.Path) {
```
ankur22, line 649: Same as above suggestion on the `HasSuffix` if statement.

common/browser_context_test.go
```diff
@@ -50,3 +51,240 @@ func TestNewBrowserContext(t *testing.T) {
 		assert.True(t, k6ObjScriptFound, "k6ObjScript was not initialized in the context")
 	})
 }
+
+func TestFilterCookies(t *testing.T) {
```
ankur22, line 55: Do we need a `t.Parallel` in the main body of the test and also in the sub tests?

common/browser_context_test.go
```diff
@@ -50,3 +51,240 @@ func TestNewBrowserContext(t *testing.T) {
...
+					Domain: "foo.com",
+				},
+				{
+					Domain: "baz.com",
+				},
+			},
+		},
+		"filter_by_subdomain": {
+			filterByURLs: []string{
+				"https://sub.foo.com",
```
ankur22, line 134: Is it possible to filter by subdomain?

common/browser_context_test.go
```diff
@@ -50,3 +51,240 @@ func TestNewBrowserContext(t *testing.T) {
...
+				},
+			},
+			wantCookies: []*api.Cookie{
+				{
+					Domain: "foo.com",
+					Path:   "/bar",
+				},
+			},
+		},
+		"disallow_secure_cookie_on_http": {
```
ankur22, line 217: Do we need another test for the `localhost` case?

common/browser_context_test.go
```diff
@@ -50,3 +51,240 @@ func TestNewBrowserContext(t *testing.T) {
...
+				tt.filterByURLs...,
+			)
+			if tt.wantErr {
+				assert.Nilf(t, cookies, "want no cookies after an error, but got %#v", cookies)
+				require.Errorf(t, err, "want an error, but got none")
+				return
+			}
+			require.NoError(t, err)
+
+			assert.Lenf(t,
```
ankur22, line 279: Do we need this length assertion? I would've thought the equal assertion below will deal with that and we will get a nice output of both the `want` and `got` to make debugging simpler.

tests/browser_context_test.go
```diff
@@ -442,14 +565,28 @@ func TestBrowserContextCookies(t *testing.T) {
...
+				require.Errorf(t,
+					err, "expected an error, but got none",
+				)
+				return
+			}
 			require.NoErrorf(t,
 				err, "failed to get cookies from the browser context",
 			)
-			require.Lenf(t,
+			assert.Lenf(t,
```
ankur22, line 589: Same here, i wonder if we can remove the length assertion.

ankur22 (approved):
Great bit of work! Thanks for adding this feature in. Very well put together. I left some comments for you to consider.

inancgumus:
> Great bit of work! Thanks for adding this feature in. Very well put together. I left some comments for you to consider.

Thanks for the great review! 🙇 I've applied all your suggestions.

inancgumus:
> Great bit of work! Thanks for adding this feature in. Very well put together. I left some comments for you to consider.

Thanks for the helpful review! 🙇 I've applied all your suggestions. cc: @ankur22

common/browser_context_test.go
```diff
@@ -134,13 +136,16 @@ func TestFilterCookies(t *testing.T) {
 				"https://sub.foo.com",
 			},
 			cookies: []*api.Cookie{
+				{
+					Domain: "sub.foo.com",
+				},
 				{
 					Domain: ".foo.com",
```
ankur22, line 143: Oh, now `.foo.com` is being filtered out, is that correct? It wasn't being filtered out in the previous iteration. I'm not 100% clued up on this so i'm mainly asking asking what's the expected result 🙂

inancgumus: I've noticed when the subdomain is explicitly given to filter cookies with subdomains, then it should do so :)

ankur22 (approved):
Thanks for taking into consideration the comments I made. LGTM 🙌 🚀

PR #1019, opened 2023-08-31
Optimize tests to run in parallel

### LINKED ISSUE #1018: Optimize tests run speed
While working on #1016, I realized that we forgot to parallelize tests throughout the codebase 🤦‍♂️

* The current tests take almost a minute to run.
* We can speed up tests tremendously by running them in parallel. Like, 2X 3X, _depending on the number of cores on the machine running the tests_.
* This also lets us detect races much faster and prevent unintended deadlocks while our users running their tests.

```[tasklist]
### Related issues
- [ ] #1019
- [ ] #1021
```

inancgumus:
## What?

1. Makes unit and integration tests run in parallel.
2. Makes some CI optimizations, considering that this is an optimization PR. Let me know if that part deserves another PR.

## Why?

1. The current tests take about 1 minute to run (locally), and there is no reason to run tests sequentially.
2. This also lets us detect race conditions much faster and prevent unintended deadlocks in code.
3. The E2E runner is hard to debug when running without the `-q` flag. With the flag, the result looks clearer to read and debug without the k6 banner and progress bar.

Here are the results after this change (On a 10-core machine):

```
Before: ~48s
After : ~24s

===Before===
ok      github.com/grafana/xk6-browser/browser  1.853s
ok      github.com/grafana/xk6-browser/common   1.669s
ok      github.com/grafana/xk6-browser/chromium 1.511s
ok      github.com/grafana/xk6-browser/tests    46.902s

===After===
ok      github.com/grafana/xk6-browser/browser  0.956s
ok      github.com/grafana/xk6-browser/common   0.689s
ok      github.com/grafana/xk6-browser/chromium 0.251s
ok      github.com/grafana/xk6-browser/tests    23.150s
```

On CI, it's the same because our runner runs on a 2-core. We can make this much faster if/when we run it on a more powerful runner. 

## Checklist

## Related PR(s)/Issue(s)

Updates #1018

### DIFF
--- a/common/browser_context_test.go
+++ b/common/browser_context_test.go
@@ -14,7 +14,11 @@ import (
 )
 
 func TestNewBrowserContext(t *testing.T) {
+	t.Parallel()
+
 	t.Run("add_web_vital_js_scripts_to_context", func(t *testing.T) {
+		t.Parallel()
+
 		ctx, cancel := context.WithCancel(context.Background())
 		logger := log.NewNullLogger()
 		b := newBrowser(ctx, cancel, nil, NewLocalBrowserOptions(), logger)
--- a/tests/browser_test.go
+++ b/tests/browser_test.go
@@ -21,6 +21,8 @@ import (
 )
 
 func TestBrowserNewPage(t *testing.T) {
+	t.Parallel()
+
 	b := newTestBrowser(t)
 	p1 := b.NewPage(nil)
 	c := b.Context()
@@ -47,6 +49,8 @@ func TestBrowserNewPage(t *testing.T) {
 }
 
 func TestBrowserNewContext(t *testing.T) {
+	t.Parallel()
+
 	b := newTestBrowser(t)
 	bc1, err := b.NewContext(nil)
 	assert.NoError(t, err)
@@ -186,6 +190,8 @@ func TestBrowserOn(t *testing.T) {
 
 // This only works for Chrome!
 func TestBrowserVersion(t *testing.T) {
+	t.Parallel()
+
 	const re = `^\d+\.\d+\.\d+\.\d+$`
 	r, _ := regexp.Compile(re)
 	ver := newTestBrowser(t).Version()
@@ -196,6 +202,8 @@ func TestBrowserVersion(t *testing.T) {
 // TODO: Improve this test, see:
 // https://github.com/grafana/xk6-browser/pull/51#discussion_r742696736
 func TestBrowserUserAgent(t *testing.T) {
+	t.Parallel()
+
 	b := newTestBrowser(t)
 
 	// testBrowserVersion() tests the version already
@@ -208,6 +216,8 @@ func TestBrowserUserAgent(t *testing.T) {
 }
 
 func TestBrowserCrashErr(t *testing.T) {
+	t.Parallel()
+
 	// create a new VU in an environment that requires a bad remote-debugging-port.
 	vu := k6test.NewVU(t, env.ConstLookup(env.BrowserArguments, "remote-debugging-port=99999"))
 
@@ -257,6 +267,9 @@ func TestBrowserLogIterationID(t *testing.T) {
 }
 
 func TestMultiBrowserPanic(t *testing.T) {
+	// this test should run sequentially.
+	// don't use t.Parallel() here.
+
 	var b1, b2 *testBrowser
 
 	// run it in a test to kick in the Cleanup() in testBrowser.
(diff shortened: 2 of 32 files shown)


tests/element_handle_test.go
```diff
@@ -279,6 +299,9 @@ func TestElementHandleIsChecked(t *testing.T) {
 }
 
 func TestElementHandleQueryAll(t *testing.T) {
+	// tests are faster than parallel execution.
+	// do not make this test parallel.
```
ka3de, line 303: Can you clarify this please? This does not seem correct to me :thinking: 
Also, even if we want subtests to be sequential, should at least `TestElementHandleQueryAll` be parallel?

tests/page_test.go
```diff
@@ -336,6 +350,10 @@ func TestPageInputSpecialCharacters(t *testing.T) {
 }
 
 func TestPageFill(t *testing.T) {
+	// these tests are not parallel by intention because
+	// they're testing the same page instance and they're
+	// faster when run sequentially.
```
ka3de, line 355: Can you clarify this please? How did you measure this? At least with my silly tests this does not seem to be the case :thinking:

ka3de (commented):
Thanks a lot for working on this @inancgumus ! :raised_hands: 
Just made a couple of comments. Also, I think there are a few tests missing to be executed in parallel or add a comment explaining why the can't not be run as such.

Unit tests:
- TestBrowserContextOptionsPermissions
- TestNewBrowserContext
- TestConnection
- TestConnectionClosureAbnormal
- TestConnectionSendRecv
- TestConnectionCreateSession
- TestContextWithDoneChan
- TestPageLocator

Integration tests:
- TestLocatorPress

inancgumus:
Thank you for the review 🙇

> Also, I think there are a few tests missing to be executed in parallel or add a comment explaining why the can't not be run as such.

> TestBrowserContextOptionsPermissions, TestConnection, TestConnectionClosureAbnormal, TestConnectionSendRecv, TestConnectionCreateSession, TestContextWithDoneChan, TestPageLocator, TestLocatorPress.

Ah, I missed these ones; thanks! I'll make them parallel.

> TestNewBrowserContext

I had sent this PR before our today's discussion in #1007. I was expecting #1007 to use globals, and it wouldn't make sense for this test to be parallel. After the discussion we had, we can make it parallel. WDYT?

tests/element_handle_test.go
inancgumus: Nice catch! I forgot why I jot it down there like that. I'll make this parallel, including subtests.

tests/page_test.go
inancgumus: Subtests race with each other since they change the same page content. I didn't bother making only the top-level test parallel, but not the subtests. But we would have to add a `nolinter` direction. Since we have many tests, running this sequentially didn't make a difference.

On the other hand, even if we make each subtest to create and run their own browsers (and not share them), the test run for this test will take 2X. On my machine, it takes 7.5s.

inancgumus: I've improved the comment: https://github.com/grafana/xk6-browser/pull/1019/commits/122818fc3d0bd754d8fb563e49b080406fb55d5f.

tests/element_handle_test.go
inancgumus: See: [9cd296e](https://github.com/grafana/xk6-browser/pull/1019/commits/9cd296ea60f7902d46146703d1b8248b0cccc233)

tests/page_test.go
ka3de: Right, I missed that there are two subests that use the same selector. Does not make much sense that the linter only complains if you run the top level test in parallel and not the subtests, but not if you just don't run the top level in parallel :man_shrugging: 
Anyway, we can leave it like this :+1:

ka3de (dismissed):
LGTM :clap:

tests/page_test.go
inancgumus: Linters :)

common/browser_context_test.go
```diff
@@ -14,7 +14,11 @@ import (
 )
 
 func TestNewBrowserContext(t *testing.T) {
+	t.Parallel()
+
 	t.Run("add_web_vital_js_scripts_to_context", func(t *testing.T) {
+		t.Parallel()
+
```
ankur22, line 21: Thank you 🙏 🙂

ankur22 (approved):
Nice! This is awesome! It;s take much less time on my machine to run through all the tests.

tests/browser_test.go
```diff
@@ -257,6 +267,9 @@ func TestBrowserLogIterationID(t *testing.T) {
 }
 
 func TestMultiBrowserPanic(t *testing.T) {
+	// this test should run sequentially.
+	// don't use t.Parallel() here.
```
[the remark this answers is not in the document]
ankur22: Ok, thanks for trying 👍

ka3de (approved):
LGTM.

PR #1022, opened 2023-08-31
Fix the check on examples/fillform.js

ankur22:
## What?

This change will fix the `check` in the [example/fillform.js](https://github.com/grafana/xk6-browser/blob/main/examples/fillform.js) test.

## Why?

`check` actually expects a function that returns a `boolean`, as is [documented](https://k6.io/docs/javascript-api/k6/check/). Although the `check` accepts the object as we've been using it, it's not correct. When users work with this example and the type definitions for k6, they will see a red line under the `'header'` in `check`, which will cause confusion and annoyance as to why our own examples have errors.

Screenshot 2023-08-31 at 14 48 04

## Checklist

### DIFF
--- a/examples/fillform.js
+++ b/examples/fillform.js
@@ -39,7 +39,7 @@ export default async function() {
       page.locator('input[type="submit"]').click(),
     ]);
     check(page, {
-      'header': page => page.locator('h2').textContent() == 'Welcome, admin!',
+      'header': p => p.locator('h2').textContent() == 'Welcome, admin!',
     });
 
     // Check whether we receive cookies from the logged site.
--- a/examples/fillform.js
+++ b/examples/fillform.js
@@ -39,7 +39,7 @@ export default async function() {
       page.locator('input[type="submit"]').click(),
     ]);
     check(page, {
-      'header': page.locator('h2').textContent() == 'Welcome, admin!',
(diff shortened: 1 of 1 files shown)


examples/fillform.js
```diff
@@ -39,7 +39,7 @@ export default async function() {
       page.locator('input[type="submit"]').click(),
     ]);
     check(page, {
-      'header': page.locator('h2').textContent() == 'Welcome, admin!',
+      'header': page => page.locator('h2').textContent() == 'Welcome, admin!',
```
inancgumus, line 42: WDYT about saying just `p` to distinguish from the actual `page` object?

```suggestion
      'header': p => p.locator('h2').textContent() == 'Welcome, admin!',
```

inancgumus (approved):
LGTM! I guess there is a superfluous commit: [Add some](https://github.com/grafana/xk6-browser/pull/1022/commits/f45856d0934fd413758dc55a13e6f62c3198cb18).

examples/fillform.js
ankur22: Nice idea, i've resoled this in 84ca43cc7eab12d88c3c17f9ed295ebea4fba51b. I've also changed it from `'header'` to `header` 👍

inancgumus (approved):

inancgumus (approved):

Issue #1025, opened 2023-09-01
Cannot add cookies with an expiration


inancgumus:
There is a problem while converting the `expires` property (in seconds) to `cdproto`'s `cdp.TimeSinceEpoch`.

1. Users provide seconds since Unix time epoch while passing the `expires` property.
2. CDP and `cdproto` require setting an exact date (time since epoch).
3. We directly pass the property to `cdproto` while it expects us a `time.Time` for setting an expiration date for cookies.

### xk6-browser version

v1.0.2

### Steps to reproduce the problem

Add a cookie with an expiration property (in seconds).

```js
context.addCookies([{
  name: 'foo',
  value: '42',
  domain: '127.0.0.1',
  path: '/',
  expires: 10,  // <---
}]);
```

### Expected behaviour

Cookies should be added with an expiration date.

### Actual behaviour

```
ERRO[0000] Uncaught (in promise) GoError:
cannot recognize cookie values:
could not convert array element [object Object] to []network.CookieParam at 0:
could not convert struct value 10 to *cdp.TimeSinceEpoch for field Expires:
could not convert 10 to cdp.TimeSinceEpoch
```

## Suggestion

The current `addCookies` API accepts an internal cookie object: `cdproto.network.Cookie`. This object contains superfluous information and can be dangerous because it lets users set internal CDP details from their scripts. We can use `api.Cookie` instead of `cdproto.network.Cookie` to be consistent with our `BrowserContext.cookies` API. And also prevent possible security problems.

```[tasklist]
### Tasks
- [ ] #1030
- [ ] #1031
```

PR #1030, opened 2023-09-05
Refactor add cookie and improve tests

inancgumus:
## What?

Refactors `BrowserContext.addCookie` to `[]*api.Cookie` and improve its tests.

## Why?

* Provides a consistent API with the `Cookies` method. `addCookies` and `Cookies` both now use `api.Cookie`.
* Separates Goja from the business logic.
* Increases security by only accepting a well-defined `api.Cookie` type instead of letting users set cookies by using `network.Cookie` (which carries CDP internal data).
* Makes code and tests more maintainable and robust.

## Checklist

## Related PR(s)/Issue(s)

Updates: #1025


ankur22 (approved):
This is a great refactor! LGTM 🙂

common/browser_context.go
```diff
@@ -443,21 +443,16 @@ func (b *BrowserContext) getSession(id target.SessionID) *Session {
 
 // AddCookies adds cookies into this browser context.
 // All pages within this context will have these cookies installed.
-func (b *BrowserContext) AddCookies(cookies goja.Value) error {
+func (b *BrowserContext) AddCookies(cookies []*api.Cookie) error {
```
ankur22, line 446: Does goja automatically try to convert from it's internal type into `[]*api.Cookie`? I don't see a change in the mapping layer to handle that.

inancgumus: Yes, it does. We don't need to change the mapping layer 🥳

PR #1031, opened 2023-09-07
Fix the expiration bug while adding cookies

### LINKED ISSUE #1025: Cannot add cookies with an expiration

inancgumus:
## What?

1. Fixes the expiration bug while adding cookies.
2. Add an expiring cookie example.
3. Fixes a bug in the test.

## Why?

Users won't set cookies with an expiration date. Now they can.

## Checklist

## Related PR(s)/Issue(s)

Closes #1025

### DIFF
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -464,13 +464,19 @@ func (b *BrowserContext) AddCookies(cookies []*api.Cookie) error {
 			const msg = "if cookie URL is not provided, both domain and path must be specified: %#v"
 			return fmt.Errorf(msg, c)
 		}
+		// calculate the cookie expiration date, session cookie if not set.
+		var ts *cdp.TimeSinceEpoch
+		if c.Expires > 0 {
+			t := cdp.TimeSinceEpoch(time.Unix(c.Expires, 0))
+			ts = &t
+		}
 		cookiesToSet = append(cookiesToSet, &network.CookieParam{
 			Name:     c.Name,
 			Value:    c.Value,
 			Domain:   c.Domain,
 			Path:     c.Path,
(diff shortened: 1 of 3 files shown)


ankur22 (approved):
LGTM! 🎉 Thanks for the fix! 🚀

Issue #940, opened 2023-06-19
Setting the timeout with `page.setDefaultTimeout` doesn't then get used anywhere


ankur22:
### Brief summary

When we set the default timeout of the `page` with `page.setDefaultTimeout`, the value is set for the `page`'s instance of `timeoutSettings`. The problem is that that value isn't used anywhere else or propagated to `frame`.

### xk6-browser version

v0.10.0

### OS

NA

### Chrome version

NA

### Docker version and image (if applicable)

NA

### Steps to reproduce the problem

When working with the following script, it timesout, which is expected since there is no such element in test.k6.io:

```js
import { chromium } from 'k6/experimental/browser';

export const options = {
  scenarios: {
    browser: {
      executor: 'per-vu-iterations',
      exec: 'browser',
      vus: 1,
      iterations: '1'
    }
  }
};

export async function browser() {
  const browser = chromium.launch({
    headless: false,
    timeout: '2m',
  });
  const context = browser.newContext();
  const page = context.newPage();
  try {
    page.setDefaultTimeout(90000);
    await page.goto('https://test.k6.io', { waitUntil: 'networkidle' });
    page.waitForSelector("div[class$='overlay-box']");
  } finally {
    page.close();
    browser.close();
  }
}
```

It results in the following error:

```bash
ERRO[0000] Failed to load resource: the server responded with a status of 404 ()  browser_source=network line_number=0 source=browser stacktrace="<nil>" url="https://k6.test.io/"
ERRO[0062] Uncaught (in promise) GoError: waiting for selector "div[class$='overlay-box']": timed out after 30000ms
running at github.com/grafana/xk6-browser/browser.mapPage.func11 (native)
browser at browser (file:///Users/ankuragarwal/go/src/github.com/grafana/xk6-browser/examples/com-test.js:24:25(37))  executor=per-vu-iterations scenario=browser
```

### Expected behaviour

The error should be since the timeout was set to 90seconds by `page.setDefaultTimeout(90000);`:

```bash
ERRO[0062] Uncaught (in promise) GoError: waiting for selector "div[class$='overlay-box']": timed out after 90000ms
```

### Actual behaviour

```bash
ERRO[0062] Uncaught (in promise) GoError: waiting for selector "div[class$='overlay-box']": timed out after 30000ms
```

PR #1033, opened 2023-09-14
Fix propagation of page timeout

ankur22:
## What?

When working with the following APIs, the timeout was being set, but it wasn't then being propagated to the underlying `frameManager` and `frame`s which meant that APIs the relied on the timeout were always working with the default 30 second timeout or the timeout set in the `browserContext`.

```js
page.setDefaultTimeout()
page.setDefaultNavigationTimeout()
```

## Why?

We want to be able to set the timeouts on the `page` and affect it's APIs, while also having a different timeout on its `browserContext`. This is the behaviour that we want, and at the moment the setting of timeouts through the `page` object does nothing.

## Checklist

## Related PR(s)/Issue(s)

Closes: https://github.com/grafana/xk6-browser/issues/940

### DIFF
--- a/common/page.go
+++ b/common/page.go
@@ -138,7 +138,7 @@ func NewPage(
 	}
 
 	var err error
-	p.frameManager = NewFrameManager(ctx, s, &p, bctx.timeoutSettings, p.logger)
+	p.frameManager = NewFrameManager(ctx, s, &p, p.timeoutSettings, p.logger)
 	p.mainFrameSession, err = NewFrameSession(ctx, s, &p, nil, tid, p.logger)
 	if err != nil {
 		p.logger.Debugf("Page:NewPage:NewFrameSession:return", "sid:%v tid:%v err:%v",
--- a/tests/browser_context_test.go
+++ b/tests/browser_context_test.go
@@ -1,6 +1,7 @@
 package tests
 
 import (
+	"fmt"
 	"net/http"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


inancgumus (approved):
I've tested this with the following script, and it worked 🥳 

Can you merge the (almost) duplicated test code for better maintainability? Suggestion: Run the same test with `page` and `browserContext` in the same test.

```js

export const options = {
  scenarios: {
    ui: {
      executor: 'per-vu-iterations',
      vus: 1,
      iterations: '1',
      options: {
        browser: {
            type: 'chromium',
        },
      },
    },
  }
};

  context.setDefaultTimeout(1);
  try {
    // page.setDefaultTimeout(1);
    page.waitForSelector("div[class$='overlay-box']");
  } finally {
    page.close();
  }
}
```

ka3de (approved):
LGTM.

ankur22:
> Suggestion 1: Can you merge the (almost) duplicated test code for better maintainability? Suggestion: Run the same test with `page` and `browserContext` in the same test.

I wanted to highlight that although they seem the same, they're in fact not the same. The tests for `browserContext` will have worked prior to the fix, whereas the tests for `page` only work after the fix. Happy to merge them though in a new PR once this and https://github.com/grafana/xk6-browser/pull/1035 are merged in 👍 

> Suggestion 2: Should we update the fillform example with this timeout setting and a comment on top of it?

Happy to do this, just wondering what the hope is by doing that?

inancgumus:
> I wanted to highlight that although they seem the same, they're in fact not the same.

Ah, OK, makes sense 👍 

> Happy to do this, just wondering what the hope is by doing that?

I thought it might be helpful, but it's fine.

@ankur22

PR #1034, opened 2023-09-14
Fix reload & setContent to use navigation timeout

ankur22:
## What?

The following page APIs were not using the correct timeout:

```js
page.reload()
page.setContent()
```

## Why?

These methods should work with the default navigation instead of the default timeout to match the behaviour in [Playwright](https://playwright.dev/docs/api/class-page#page-set-default-navigation-timeout).

## Checklist

### DIFF
--- a/common/page.go
+++ b/common/page.go
@@ -862,7 +862,7 @@ func (p *Page) Reload(opts goja.Value) api.Response {
 
 	parsedOpts := NewPageReloadOptions(
 		LifecycleEventLoad,
-		time.Duration(p.timeoutSettings.navigationTimeout())*time.Second,
+		p.timeoutSettings.navigationTimeout(),
 	)
 	if err := parsedOpts.Parse(p.ctx, opts); err != nil {
 		k6ext.Panic(p.ctx, "parsing reload options: %w", err)
--- a/common/page.go
+++ b/common/page.go
@@ -860,7 +860,10 @@ func (p *Page) QueryAll(selector string) ([]api.ElementHandle, error) {
 func (p *Page) Reload(opts goja.Value) api.Response {
 	p.logger.Debugf("Page:Reload", "sid:%v", p.sessionID())
 
-	parsedOpts := NewPageReloadOptions(LifecycleEventLoad, p.defaultTimeout())
+	parsedOpts := NewPageReloadOptions(
+		LifecycleEventLoad,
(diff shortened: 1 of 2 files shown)


inancgumus (approved):
I've tested this with the following script, and it worked 🥳 Nice catch!

I've also checked the other methods we have by exploring the link you shared, and they all seem fine and use the correct timeout 👍 

```js

export const options = {
  scenarios: {
    ui: {
      executor: 'per-vu-iterations',
      vus: 1,
      iterations: '1',
      options: {
        browser: {
            type: 'chromium',
        },
      },
    },
  }
};

  try {
    context.setDefaultNavigationTimeout(0.1); // <---
    page.reload();
  } finally {
    page.close();
  }
}
```

I wish we had a better error message (concise and without the first execution context changed error) when the navigation timeout happens :) Current one:

```
WARN[0002] failed to hide page: evaluating JS: execution context changed; most likely because of a navigation  category="Page:Close" elapsed="0 ms" source=browser
ERRO[0002] Uncaught (in promise) GoError: reloading page: timed out
        at github.com/grafana/xk6-browser/browser.mapPage.func9 (native)
        at file:///Users/inanc/grafana/k6browser/tests/pr1034.js:24:4(27)  executor=per-vu-iterations scenario=ui
```

common/page.go
```diff
@@ -860,7 +860,10 @@ func (p *Page) QueryAll(selector string) ([]api.ElementHandle, error) {
 func (p *Page) Reload(opts goja.Value) api.Response {
 	p.logger.Debugf("Page:Reload", "sid:%v", p.sessionID())
 
-	parsedOpts := NewPageReloadOptions(LifecycleEventLoad, p.defaultTimeout())
+	parsedOpts := NewPageReloadOptions(
+		LifecycleEventLoad,
+		time.Duration(p.timeoutSettings.navigationTimeout())*time.Second,
```
ankur22, line 865: Reminder: This needs to be updated once https://github.com/grafana/xk6-browser/pull/1035 has been merged in.

inancgumus: Read for additional context: https://github.com/grafana/xk6-browser/pull/1035#pullrequestreview-1630425853 :)

ka3de (approved):
LGTM.

inancgumus (approved):
Nice! Remember to squash as there is a fix-up commit :)

PR #1035, opened 2023-09-14
Fix/use time duration timeout

ankur22:
## What?

This PR fixes a problem where the timeout was an int64 making it difficult to workout whether that was in seconds or milliseconds. This change enforces us to work with `time.Duration` when it comes to the default and navigation timeouts.

## Why?

This prevents confusion as to whether timeout (which was a int64) was in milliseconds or seconds.

## Checklist


inancgumus (approved):
Nice catch, thanks for this! WDYT about the following?

Should we update this part of the code (WaitForEvent), too? And make it millisecond-based for consistency throughout our API?

https://github.com/grafana/xk6-browser/blob/22ca96326a385c0c9a1eacf3632c4d4f06198ee4/common/browser_context.go#L351

Maybe not related to this PR, but I guess we have a bug in `WaitForEvent`:

https://github.com/grafana/xk6-browser/blob/22ca96326a385c0c9a1eacf3632c4d4f06198ee4/common/browser_context.go#L351

It multiplies `30 * time.Second` with `time.Second` if that option is the default:

https://github.com/grafana/xk6-browser/blob/22ca96326a385c0c9a1eacf3632c4d4f06198ee4/common/consts.go#L11

ankur22:
> Should we update this part of the code (WaitForEvent), too? And make it millisecond-based for consistency throughout our API?

I like that idea, and yeah, maybe best for another PR. Happy to take that on after i've got these few timeout PRs merged in 👍

inancgumus (approved):

ka3de (approved):
LGTM.

PR #1039, opened 2023-09-19
Cookies: Fix Goja conversions

inancgumus:
## What?

Adds `js` struct tags to `api.Cookie` to allow Goja to convert `api.Cookie` values.

## Why?

Goja cannot convert `httpOnly` and `sameSite` fields while sending and retrieving cookies from the JavaScript side.

## Checklist

## Related PR(s)/Issue(s)

Updates: #6

### DIFF
--- a/api/browser_context.go
+++ b/api/browser_context.go
@@ -41,16 +41,16 @@ type BrowserContext interface {
 //
 // https://datatracker.ietf.org/doc/html/rfc6265.
 type Cookie struct {
-	Name     string         `json:"name"`                   // Cookie name.
-	Value    string         `json:"value"`                  // Cookie value.
-	Domain   string         `json:"domain"`                 // Cookie domain.
-	Path     string         `json:"path"`                   // Cookie path.
+	Name     string         `js:"name" json:"name"`         // Cookie name.
+	Value    string         `js:"value" json:"value"`       // Cookie value.
+	Domain   string         `js:"domain" json:"domain"`     // Cookie domain.
+	Path     string         `js:"path" json:"path"`         // Cookie path.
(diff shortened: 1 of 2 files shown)


api/browser_context.go
```diff
@@ -41,15 +41,16 @@ type BrowserContext interface {
...
-	Path     string         `json:"path"`          // Cookie path.
-	Expires  int64          `json:"expires"`       // Cookie expiration date as the number of seconds since the UNIX epoch.
-	HTTPOnly bool           `json:"httpOnly"`      // True if cookie is http-only.
-	Secure   bool           `json:"secure"`        // True if cookie is secure.
-	SameSite CookieSameSite `json:"sameSite"`      // Cookie SameSite type.
-	URL      string         `json:"url,omitempty"` // Cookie URL.
+	Name     string         `json:"name"`                   // Cookie name.
+	Value    string         `json:"value"`                  // Cookie value.
+	Domain   string         `json:"domain"`                 // Cookie domain.
+	Path     string         `json:"path"`                   // Cookie path.
```
ankur22, line 47: Just a suggestion, but also happy to leave it as is.

Just for consistency, would it be a good idea to add the `js` struct tags to all the fields?

ankur22 (dismissed):
Good catch! LGTM 🙂

api/browser_context.go
inancgumus: Sure thing, ac67ec1 👍

ka3de (approved):
LGTM.

api/browser_context.go
ankur22: Thanks! 🎉

Issue #447, opened 2022-07-07
`waitForEvent` on `BrowserContext` is waiting for async support


ankur22:
Tested against: dbede120c63df43995813a847a25b0e66e289592

`waitForEvent` on `BrowserContext` is waiting for async support.
```[tasklist]
### Tasks
- [x] PRs: https://github.com/grafana/xk6-browser/pull/1042, https://github.com/grafana/xk6-browser/pull/1069
- [x] Release notes: https://github.com/grafana/k6/pull/3413
- [x] Docs: https://github.com/grafana/k6-docs/pull/1384
- [x] Type definitions: https://github.com/grafana/k6-DefinitelyTyped/pull/18
```

PR #1042, opened 2023-09-20
Fix browserContext.waitForEvent

ankur22:
## What?

This refactors, fixes various issues, promisifies the API, and adds tests for the `browserContext.waitForEvent` API.

## Why?

This initial [motivation](https://github.com/grafana/xk6-browser/pull/1035#pullrequestreview-1630425853) for tackling this was to solely rectify a confusion with the timeout which was sometimes in milliseconds and sometimes in seconds.

While taking a look at the timeout issue, it was clear that there were various issues with this API:

1. It needed a refactor to align with how we parse options elsewhere.
2. The fix on the timeout to work with the default browserContext time.
3. It was unusable unless we promisified it.
4. Parsing of a callable didn't quite work as expected when the options object was passed in.
5. It needed integration and an example test.

There is an argument to be made against this change and API since we've not had any feedback on this API, and so it might not bear any real value still.

## Checklist

## Related PR(s)/Issue(s)

Initial motivation: [comment](https://github.com/grafana/xk6-browser/pull/1035#pullrequestreview-1630425853).
Closes: https://github.com/grafana/xk6-browser/issues/447

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -652,6 +652,7 @@ func mapBrowserContext(vu moduleVU, bc api.BrowserContext) mapping {
 			ctx := vu.Context()
 			return k6ext.Promise(ctx, func() (result any, reason error) {
 				resp, err := bc.WaitForEvent(event, optsOrPredicate)
+				panicIfFatalError(ctx, err)
 				if err != nil {
 					return nil, err //nolint:wrapcheck
 				}
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -648,7 +648,22 @@ func mapBrowserContext(vu moduleVU, bc api.BrowserContext) mapping {
 		"setOffline":         bc.SetOffline,
 		"storageState":       bc.StorageState,
 		"unroute":            bc.Unroute,
-		"waitForEvent":       bc.WaitForEvent,
+		"waitForEvent": func(event string, optsOrPredicate goja.Value) *goja.Promise {
+			ctx := vu.Context()
+			return k6ext.Promise(ctx, func() (result any, reason error) {
+				resp, err := bc.WaitForEvent(event, optsOrPredicate)
+				if err != nil {
+					return nil, err //nolint:wrapcheck
+				}
+
+				p, ok := resp.(api.Page)
+				if !ok {
+					panicIfFatalError(ctx, fmt.Errorf("response object is not a page: %w", k6error.ErrFatal))
+				}
+
+				return mapPage(vu, p), nil
+			})
+		},
 		"pages": func() *goja.Object {
 			var (
 				mpages []mapping
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -345,7 +345,7 @@ func (b *BrowserContext) WaitForEvent(event string, optsOrPredicate goja.Value) 
 	b.logger.Debugf("BrowserContext:WaitForEvent", "bctxid:%v event:%q", b.id, event)
 
 	parsedOpts := NewWaitForEventOptions(
-		b.browser.browserOpts.Timeout * time.Second,
+		b.timeoutSettings.timeout(),
 	)
 	if err := parsedOpts.Parse(b.ctx, optsOrPredicate); err != nil {
 		k6ext.Panic(b.ctx, "parsing waitForEvent options: %w", err)
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -381,6 +381,8 @@ func (b *BrowserContext) waitForEvent(event string, predicateFn goja.Callable, t
 	}
 }
 
+// runWaitForEventHandler can work with a nil predicateFn.
+// If predicateFn is nil, it will return the response straight away.
 func (b *BrowserContext) runWaitForEventHandler(
 	ctx context.Context, evCancelFn func(), chEvHandler chan Event, out chan any, predicateFn goja.Callable,
 ) {
@@ -405,8 +407,20 @@ func (b *BrowserContext) runWaitForEventHandler(
 			if ev.typ == EventBrowserContextPage {
 				b.logger.Debugf("BrowserContext:WaitForEvent:go():EventBrowserContextPage", "bctxid:%v", b.id)
 				p, _ := ev.data.(*Page)
-				if retVal, err := predicateFn(b.vu.Runtime().ToValue(p)); err == nil && retVal.ToBoolean() {
+
+				if predicateFn == nil {
 					b.logger.Debugf("BrowserContext:WaitForEvent:go():EventBrowserContextPage:return", "bctxid:%v", b.id)
+					out <- p
+					close(out)
+
+					// We wait for one matching event only,
+					// then remove event handler by cancelling context and stopping goroutine.
+					evCancelFn()
+					return
+				}
+
+				if retVal, err := predicateFn(b.vu.Runtime().ToValue(p)); err == nil && retVal.ToBoolean() {
+					b.logger.Debugf("BrowserContext:WaitForEvent:go():EventBrowserContextPage:predicateFn:return", "bctxid:%v", b.id)
 					out <- p
 					close(out)
 
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -342,7 +342,6 @@ func (b *BrowserContext) Unroute(url goja.Value, handler goja.Callable) {
 
 // WaitForEvent waits for event.
 func (b *BrowserContext) WaitForEvent(event string, optsOrPredicate goja.Value) (any, error) {
-	// TODO: This public API needs Promise support (as return value) to be useful in JS!
 	b.logger.Debugf("BrowserContext:WaitForEvent", "bctxid:%v event:%q", b.id, event)
 
 	parsedOpts := NewWaitForEventOptions(
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -370,7 +370,7 @@ func (b *BrowserContext) waitForEvent(
 	timeout time.Duration,
 ) (any, error) {
 	if event != waitForEventTypePage {
-		return nil, fmt.Errorf("%q is the only event that is supported, you passed in %q", waitForEventTypePage, event)
+		return nil, fmt.Errorf("incorrect event %q, %q is the only event supported", event, waitForEventTypePage)
 	}
 
 	evCancelCtx, evCancelFn := context.WithCancel(b.ctx)
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -422,21 +422,24 @@ func (b *BrowserContext) runWaitForEventHandler(
 				b.logger.Debugf("BrowserContext:runWaitForEventHandler:go():EventBrowserContextClose:return", "bctxid:%v", b.id)
 				return
 			}
-			if ev.typ == EventBrowserContextPage {
-				b.logger.Debugf("BrowserContext:runWaitForEventHandler:go():EventBrowserContextPage", "bctxid:%v", b.id)
-				p, _ = ev.data.(*Page)
 
-				if predicateFn == nil {
-					b.logger.Debugf("BrowserContext:runWaitForEventHandler:go():EventBrowserContextPage:return", "bctxid:%v", b.id)
-					return
-				}
+			if ev.typ != EventBrowserContextPage {
+				continue
+			}
 
-				if retVal, err := predicateFn(b.vu.Runtime().ToValue(p)); err == nil && retVal.ToBoolean() {
-					b.logger.Debugf(
-						"BrowserContext:runWaitForEventHandler:go():EventBrowserContextPage:predicateFn:return",
-						"bctxid:%v", b.id,
-					)
-				}
+			b.logger.Debugf("BrowserContext:runWaitForEventHandler:go():EventBrowserContextPage", "bctxid:%v", b.id)
+			p, _ = ev.data.(*Page)
+
+			if predicateFn == nil {
+				b.logger.Debugf("BrowserContext:runWaitForEventHandler:go():EventBrowserContextPage:return", "bctxid:%v", b.id)
+				return
+			}
+
+			if retVal, err := predicateFn(b.vu.Runtime().ToValue(p)); err == nil && retVal.ToBoolean() {
+				b.logger.Debugf(
+					"BrowserContext:runWaitForEventHandler:go():EventBrowserContextPage:predicateFn:return",
+					"bctxid:%v", b.id,
+				)
 			}
 		}
 	}
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -344,36 +344,14 @@ func (b *BrowserContext) WaitForEvent(event string, optsOrPredicate goja.Value) 
 	// TODO: This public API needs Promise support (as return value) to be useful in JS!
 	b.logger.Debugf("BrowserContext:WaitForEvent", "bctxid:%v event:%q", b.id, event)
 
-	var (
-		isCallable  bool
-		predicateFn goja.Callable
-		// TODO: Find out whether * time.Second is necessary.
-		timeout = b.browser.browserOpts.Timeout * time.Second //nolint:durationcheck
+	parsedOpts := NewWaitForEventOptions(
+		b.browser.browserOpts.Timeout * time.Second,
 	)
-	if gojaValueExists(optsOrPredicate) {
-		switch optsOrPredicate.ExportType() {
-		case reflect.TypeOf(goja.Object{}):
-			opts := optsOrPredicate.ToObject(b.vu.Runtime())
-			for _, k := range opts.Keys() {
-				switch k {
-				case "predicate":
-					predicateFn, isCallable = goja.AssertFunction(opts.Get(k))
-					if !isCallable {
-						k6ext.Panic(b.ctx, "predicate function is not callable")
-					}
-				case "timeout":
-					timeout = time.Duration(opts.Get(k).ToInteger()) * time.Millisecond
-				}
-			}
-		default:
-			predicateFn, isCallable = goja.AssertFunction(optsOrPredicate)
-			if !isCallable {
-				k6ext.Panic(b.ctx, "predicate function is not callable")
-			}
-		}
+	if err := parsedOpts.Parse(b.ctx, optsOrPredicate); err != nil {
+		k6ext.Panic(b.ctx, "parsing waitForEvent options: %w", err)
 	}
 
-	return b.waitForEvent(event, predicateFn, timeout)
+	return b.waitForEvent(event, parsedOpts.PredicateFn, parsedOpts.Timeout)
 }
 
 func (b *BrowserContext) waitForEvent(event string, predicateFn goja.Callable, timeout time.Duration) any {
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/url"
 	"reflect"
@@ -375,16 +376,14 @@ func (b *BrowserContext) waitForEvent(event string, predicateFn goja.Callable, t
 
 	select {
 	case <-b.ctx.Done():
-		b.logger.Debugf("BrowserContext:WaitForEvent:ctx.Done", "bctxid:%v event:%q", b.id, event)
+		return nil, errors.New("test iteration ended")
 	case <-time.After(timeout):
 		b.logger.Debugf("BrowserContext:WaitForEvent:timeout", "bctxid:%v event:%q", b.id, event)
+		return nil, fmt.Errorf("waitForEvent timed out after %v", timeout)
 	case evData := <-ch:
 		b.logger.Debugf("BrowserContext:WaitForEvent:evData", "bctxid:%v event:%q", b.id, event)
 		return evData, nil
 	}
-	b.logger.Debugf("BrowserContext:WaitForEvent:return nil", "bctxid:%v event:%q", b.id, event)
-
-	return nil, nil
 }
 
 func (b *BrowserContext) runWaitForEventHandler(
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -378,8 +378,9 @@ func (b *BrowserContext) waitForEvent(
 
 	chEvHandler := make(chan Event)
 	ch := make(chan any)
+	errCh := make(chan error)
 
-	go b.runWaitForEventHandler(evCancelCtx, chEvHandler, ch, predicateFn)
+	go b.runWaitForEventHandler(evCancelCtx, chEvHandler, ch, errCh, predicateFn)
 
 	b.on(evCancelCtx, []string{EventBrowserContextPage}, chEvHandler)
 
@@ -392,19 +393,23 @@ func (b *BrowserContext) waitForEvent(
 	case evData := <-ch:
 		b.logger.Debugf("BrowserContext:WaitForEvent:evData", "bctxid:%v event:%q", b.id, event)
 		return evData, nil
+	case err := <-errCh:
+		b.logger.Debugf("BrowserContext:WaitForEvent:err", "bctxid:%v event:%q, err:%v", b.id, event, err)
+		return nil, err
 	}
 }
 
 // runWaitForEventHandler can work with a nil predicateFn. If predicateFn is
 // nil it will return the response straight away.
 func (b *BrowserContext) runWaitForEventHandler(
-	ctx context.Context, chEvHandler chan Event, out chan<- any, predicateFn goja.Callable,
+	ctx context.Context, chEvHandler chan Event, out chan<- any, errOut chan<- error, predicateFn goja.Callable,
 ) {
 	b.logger.Debugf("BrowserContext:runWaitForEventHandler:go():starts", "bctxid:%v", b.id)
 	defer b.logger.Debugf("BrowserContext:runWaitForEventHandler:go():returns", "bctxid:%v", b.id)
 
 	defer func() {
 		close(out)
+		close(errOut)
 	}()
 
 	for {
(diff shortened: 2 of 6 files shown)


browser/mapping.go
```diff
@@ -648,7 +648,22 @@ func mapBrowserContext(vu moduleVU, bc api.BrowserContext) mapping {
...
+			return k6ext.Promise(ctx, func() (result any, reason error) {
+				resp, err := bc.WaitForEvent(event, optsOrPredicate)
+				if err != nil {
+					return nil, err //nolint:wrapcheck
+				}
+
+				p, ok := resp.(api.Page)
+				if !ok {
+					panicIfFatalError(ctx, fmt.Errorf("response object is not a page: %w", k6error.ErrFatal))
+				}
```
inancgumus, line 662: What's the difference between returning an error and panicking here? We discussed this before, but please remind me. Thanks!

inancgumus, line 662: What's the difference between returning an error and panicking here? We discussed this before, but please remind me. Thanks!

common/browser_context.go
```diff
@@ -340,43 +341,24 @@ func (b *BrowserContext) Unroute(url goja.Value, handler goja.Callable) {
 }
 
 // WaitForEvent waits for event.
-func (b *BrowserContext) WaitForEvent(event string, optsOrPredicate goja.Value) any {
-	// TODO: This public API needs Promise support (as return value) to be useful in JS!
```
inancgumus, line 344: Yay! 🥳

common/browser_context.go
```diff
@@ -340,43 +341,24 @@ func (b *BrowserContext) Unroute(url goja.Value, handler goja.Callable) {
...
 	}
 
-	return b.waitForEvent(event, predicateFn, timeout)
+	return b.waitForEvent(event, parsedOpts.PredicateFn, parsedOpts.Timeout)
 }
 
-func (b *BrowserContext) waitForEvent(event string, predicateFn goja.Callable, timeout time.Duration) any {
+func (b *BrowserContext) waitForEvent(event string, predicateFn goja.Callable, timeout time.Duration) (any, error) {
+	if event != "page" {
+		return nil, fmt.Errorf("\"page\" is the only event that is supported, you passed in %q", event)
```
inancgumus, line 359:
```suggestion
		return nil, fmt.Errorf(`incorrect event: %q, "page" is the only event supported", event)		
```

It's okay, but should this `"page"` be a constant somewhere, as it is used in multiple places.

common/browser_context.go
```diff
@@ -388,23 +370,34 @@ func (b *BrowserContext) waitForEvent(event string, predicateFn goja.Callable, t
 
 	select {
 	case <-b.ctx.Done():
-		b.logger.Debugf("BrowserContext:WaitForEvent:ctx.Done", "bctxid:%v event:%q", b.id, event)
+		return nil, errors.New("test iteration ended")
```
inancgumus, line 373: Could context be canceled by something else, too? I'm wondering if this error message's would be accurate then 🤔

common/browser_context.go
```diff
@@ -388,23 +370,34 @@ func (b *BrowserContext) waitForEvent(event string, predicateFn goja.Callable, t
...
 		b.logger.Debugf("BrowserContext:WaitForEvent:timeout", "bctxid:%v event:%q", b.id, event)
+		return nil, fmt.Errorf("waitForEvent timed out after %v", timeout)
 	case evData := <-ch:
 		b.logger.Debugf("BrowserContext:WaitForEvent:evData", "bctxid:%v event:%q", b.id, event)
-		return evData
+		return evData, nil
 	}
-	b.logger.Debugf("BrowserContext:WaitForEvent:return nil", "bctxid:%v event:%q", b.id, event)
-
-	return nil
```
inancgumus, line 400: 👍

common/browser_context.go
```diff
@@ -388,23 +370,34 @@ func (b *BrowserContext) waitForEvent(event string, predicateFn goja.Callable, t
...
 }
 
+// runWaitForEventHandler can work with a nil predicateFn.
+// If predicateFn is nil, it will return the response straight away.
 func (b *BrowserContext) runWaitForEventHandler(
 	ctx context.Context, evCancelFn func(), chEvHandler chan Event, out chan any, predicateFn goja.Callable,
 ) {
 	b.logger.Debugf("BrowserContext:WaitForEvent:go():starts", "bctxid:%v", b.id)
 	defer b.logger.Debugf("BrowserContext:WaitForEvent:go():returns", "bctxid:%v", b.id)
```
inancgumus, line 389:
```suggestion
	b.logger.Debugf("BrowserContext:runWaitForEventHandler():starts", "bctxid:%v", b.id)
	defer b.logger.Debugf("BrowserContext:runWaitForEventHandler():returns", "bctxid:%v", b.id)
```

common/browser_context.go
```diff
@@ -413,25 +406,19 @@ func (b *BrowserContext) runWaitForEventHandler(
 		case ev := <-chEvHandler:
 			if ev.typ == EventBrowserContextClose {
 				b.logger.Debugf("BrowserContext:WaitForEvent:go():EventBrowserContextClose:return", "bctxid:%v", b.id)
```
inancgumus, line 408: We might want to update these log messages as well. Like: `"BrowserContext:runWaitForEventHandler():`.

common/browser_context.go
```diff
@@ -413,25 +406,19 @@ func (b *BrowserContext) runWaitForEventHandler(
...
-				out <- nil
-				close(out)
-
-				// We wait for one matching event only,
-				// then remove event handler by cancelling context and stopping goroutine.
-				evCancelFn()
 				return
 			}
 			if ev.typ == EventBrowserContextPage {
```
inancgumus, line 411: Early return from the function here?

```go
if ev.typ != EventBrowserContextPage {
   // log and return
}

File: common/browser_context_options.go [hunk 9]
```diff
@@ -132,3 +134,48 @@ func (b *BrowserContextOptions) Parse(ctx context.Context, opts goja.Value) erro
...
+		Timeout: defaultTimeout,
+	}
+}
+
+// Parse will parse the options or a callable predicate function.
+// It can parse only a callable predicate function or an object
+// which contains a callable predicate function and a timeout.
+func (w *WaitForEventOptions) Parse(ctx context.Context, optsOrPredicate goja.Value) error {
+	var isCallable bool
+	if gojaValueExists(optsOrPredicate) {
```
inancgumus, line 158: Early return?

```go
	if !gojaValueExists(optsOrPredicate) {
		return nil
	}
	// actual code is here
```

common/browser_context_options.go
```diff
@@ -132,3 +134,48 @@ func (b *BrowserContextOptions) Parse(ctx context.Context, opts goja.Value) erro
...
+
+		opts := optsOrPredicate.ToObject(rt)
+		for _, k := range opts.Keys() {
+			switch k {
+			case "predicate":
+				w.PredicateFn, isCallable = goja.AssertFunction(opts.Get(k))
+				if !isCallable {
+					return errors.New("predicate function is not callable")
+				}
+			case "timeout": //nolint:goconst
```
inancgumus, line 174: I remember a dozen of bugs because of using a wrong string value. So what the linter suggests here makes sense :)

common/browser_context_options.go
```diff
@@ -132,3 +134,48 @@ func (b *BrowserContextOptions) Parse(ctx context.Context, opts goja.Value) erro
 	}
 	return nil
 }
+
+// WaitForEventOptions are the options used by
+// the browserContext.waitForEvent API.
+type WaitForEventOptions struct {
```
inancgumus, line 140: Nice! A new type to manage this!

examples/waitForEvent.js
```diff
@@ -0,0 +1,41 @@
...
+    return false
+  } })
+  
+  // Now we create two pages.
+  const page = context.newPage()
+  const page2 = context.newPage()
+
+  // We await for the page creation
+  // events to be processed and the
+  // predicate to pass.
```
inancgumus, line 35: Is there a reason to split these comment lines extremely short? I noticed a few of them in the Go code in this PR, too. I wouldn't mind it if this weren't a refactoring PR :) It's OK if it's your personal style :)

tests/browser_context_test.go
```diff
@@ -688,3 +689,156 @@ func TestBrowserContextTimeout(t *testing.T) {
...
+	t.Parallel()
+
+	testCases := []struct {
+		name            string
+		event           string
+		optsOrPredicate *optsOrPredicate
+		wantErr         string
+	}{
+		{
+			name:  "successfully wait for page creation",
```
inancgumus, line 703: No strong opinion on this one as I'll share what I find useful for me.

Should we keep these test names concise? So the test run "glancability" would be better. I also prefer using underscores in test names because it makes it easier to copy fast and find-search when they fail.

We can elaborate on what's going on with a comment.

```go
		{
			// successfully wait for page creation with only predicate
			name:            "with_predicate",
			event:           "page",
			optsOrPredicate: &optsOrPredicate{justPredicate: stringPtr("() => true;")},
		},
```

I omitted "page" from the name because the event is `"page"` is there. And I didn't say "wait" because the test is about wait for event.

tests/browser_context_test.go
```diff
@@ -688,3 +689,156 @@ func TestBrowserContextTimeout(t *testing.T) {
...
+			tb := newTestBrowser(t)
+
+			bc, err := tb.NewContext(nil)
+			require.NoError(t, err)
+
+			ctx, cancel := context.WithTimeout(tb.context(), 5*time.Second)
+			defer cancel()
+
+			var p1, p2 api.Page
+			err = tb.run(
```
inancgumus, line 748: It would be great if you could comment on what this code does. A simple comment, like: "Wait until the page is created". By the way, what would happen if the page is created before `WaitForEvent` is called? Is it expected?

tests/browser_context_test.go
```diff
@@ -688,3 +689,156 @@ func TestBrowserContextTimeout(t *testing.T) {
...
+				assert.Equal(t, p1.MainFrame().ID(), p2.MainFrame().ID())
+				return
+			}
+
+			assert.ErrorContains(t, err, tc.wantErr)
+		})
+	}
+}
+
+type optsOrPredicate struct {
```
inancgumus, line 778: Can you help me understand what this type and functions do? Commenting inside the code would be great!

inancgumus (commented):
It's much better than before 🥳  And now `WaitForEvent` works! Finally 👏 

I have some weakly-held suggestions.

Nit: I wished the refactoring in one PR and the fix in another 🙇

browser/mapping.go
ankur22: The reason i didn't want to use `k6ext.Abort` from here is to use `panicIfFatalError` which could have other cleanup functionality in the future. At the moment `panicIfFatalError` doesn't do anything but call `k6ext.Abort`.

In general it tries to help isolate where an `k6ext.Abort` can be called from, and so it encourages us to bubble up errors from downstream back up. If we wrap an error as `FatelError` then it will abort at the mapping layer instead of us having to call abort/panic in the business logic area of the codebase.

In this case, maybe it's ok to call `k6ext.Abort` from here instead of first wrapping it in `FatelError` and then calling `panicIfFatalError`.

The reason we wanted something like this in the first place was due to https://github.com/grafana/xk6-browser/issues/688

Let me know how you feel about this.

inancgumus (approved):
Ah, forgot to approve 😅

common/browser_context.go
ankur22: Yeah, that makes sense. Resolved in 4c096de323dcee4e6b88cc80cd8ea227fd8aef14 and d741d451341620082cb433f50a75f30710c0afbb.

common/browser_context.go
ankur22: Good spot. I was meant to leave a WDYT note. So when the context is closed, it definitely means that iteration is ending, but it could also mean the whole test run is ending. Would you prefer the original error from the context itself?

common/browser_context.go
ankur22: Resolved in af81e9a81d4e02f3501a755fc3edb495f6cf3c3f

common/browser_context.go
ankur22: Good catch on this. Resolved altogether in af81e9a81d4e02f3501a755fc3edb495f6cf3c3f.

common/browser_context.go
```diff
@@ -388,52 +380,60 @@ func (b *BrowserContext) waitForEvent(event string, predicateFn goja.Callable, t
...
+	}()
+
 	for {
 		select {
 		case <-ctx.Done():
-			b.logger.Debugf("BrowserContext:WaitForEvent:go():ctx:done", "bctxid:%v", b.id)
+			b.logger.Debugf("BrowserContext:runWaitForEventHandler:go():ctx:done", "bctxid:%v", b.id)
 			return
 		case ev := <-chEvHandler:
 			if ev.typ == EventBrowserContextClose {
```
ankur22, line 417: Reminder: This isn't being emitted, so best to remove from codebase.

common/browser_context.go
ankur22: Yep, thanks! Resolved in bbfe8df5481b1d3b0616c4699b8adb8be85bcb6e.

common/browser_context_options.go
ankur22: 👍 resolved in 3d1f55840554dd1fda8ce87161e1e5092b2ac8b6.

common/browser_context_options.go
ankur22: Since there's 14 places for this change. I might do them in a new PR, WDYT?

examples/waitForEvent.js
ankur22: Yeah, personal preference but it changes 😅 

I'll change them though so that they're consistent. I think around 80 characters max line length should be good enough as a guide.

ankur22: Resolved in 197378ca4f9662c5cdc6f49441f621e47b84cfec

tests/browser_context_test.go
ankur22: Sure, I can give it a go. I personally prefer to be a bit more explicit but i think being consistent is better. Resolved in b2d2503d7fa48f4800e38cf37308a9148bf36caa

browser/mapping.go
ankur22: Ah, right, so the reason I prefer an abort here over a plain error is because if we get something other than `api.page` then something has gone very wrong in the implementation, and requires us to look into. Happy change it to a error though.

tests/browser_context_test.go
ankur22: Sure, i've added some comments in 8fd96c5507cd0d0f1d86f00b498123f08475d56d. Let me know if they help or not.

tests/browser_context_test.go
ankur22: Sure, i've added some comments in 8fd96c5507cd0d0f1d86f00b498123f08475d56d. Let me know if they help or not.

common/browser_context_options.go
ankur22: https://github.com/grafana/xk6-browser/pull/1044

common/browser_context.go
ankur22: PR https://github.com/grafana/xk6-browser/pull/1045

common/browser_context.go
ka3de: I agree, I don't think we should mention "iteration end". Maybe something more generic?

> Would you prefer the original error from the context itself?

What would that be?

browser/mapping.go
inancgumus: > Happy to change it to an error though.

Yep, please 👍

I understand your reasoning better now, thanks. Returning something other than `api.Page` deserves a panic in the business logic code, which will be a programmer error (that's what panics are for—truly exceptional situations). For usual error handling, I think returning an error is fine here.

tests/browser_context_test.go
```diff
@@ -688,3 +689,184 @@ func TestBrowserContextTimeout(t *testing.T) {
...
+			// With a predicate function and a new timeout in an option object.
+			name:            "success_with_option_predicate_timeout",
+			event:           "page",
+			optsOrPredicate: &optsOrPredicate{predicate: stringPtr("() => true;"), timeout: int64Ptr(1000)},
+		},
+		{
+			// Fails when an event other than "page" is passed in.
+			name:    "fails_incorrect_event",
+			event:   "browser",
+			wantErr: "incorrect event \"browser\", \"page\" is the only event supported",
```
inancgumus, line 729:
```suggestion
			wantErr: `incorrect event "browser", "page" is the only event supported`,
```

common/browser_context.go
inancgumus: You create [this type](https://github.com/grafana/xk6-browser/commit/4c096de323dcee4e6b88cc80cd8ea227fd8aef14#diff-688b96a8694c1d64dd8e32b212658c9953b9045c1225b0cea4ec2796bf6f2f26R36) because we'll add more event types, right? Makes sense 👍

A small suggestion for [this](https://github.com/grafana/xk6-browser/commit/d741d451341620082cb433f50a75f30710c0afbb#diff-143d56aaa3500065e9f275f9d202a6182d95a4887685bd54f9861356f0e38230R724):

```diff
- wantErr: "incorrect event \"browser\", \"page\" is the only event supported",
+ wantErr: `incorrect event "browser", "page" is the only event supported`,
```

common/browser_context.go
inancgumus: @ankur22, yes, the context error itself would be better 👍 So we can better diagnose when it ends because of something else.

common/browser_context.go
inancgumus: Thanks!

common/browser_context.go
inancgumus: Perfect, thanks!

common/browser_context.go
ka3de: With the current implementation, my understanding is that if this `EventBrowserContextClose` is fired, we would be returning a `null/undefined` page to the JS context, as we do not do any verification in the `case evData := <-ch:` case in `BrowserContext.waitForEvent` select. But if this event is never fired and we will remove it, then this function will only end either in a TO or with a page being returned and passed to the given predicate function (if any). Is that right?

common/browser_context_options.go
inancgumus: Thanks!

common/browser_context_options.go
inancgumus: Nice! Thanks.

examples/waitForEvent.js
inancgumus: Perfect, thanks!

common/browser_context.go
```diff
@@ -388,52 +384,63 @@ func (b *BrowserContext) waitForEvent(event string, predicateFn goja.Callable, t
...
+				b.logger.Debugf("BrowserContext:runWaitForEventHandler:go():EventBrowserContextClose:return", "bctxid:%v", b.id)
+				return
+			}
+
+			if ev.typ != EventBrowserContextPage {
+				continue
+			}
+
+			b.logger.Debugf("BrowserContext:runWaitForEventHandler:go():EventBrowserContextPage", "bctxid:%v", b.id)
+			p, _ = ev.data.(*Page)
```
ka3de, line 431: Is there a reason to not verifying that the conversion is correct?

tests/browser_context_test.go
inancgumus: Explicitness is suitable for other cases, yeah, I agree. On the other hand, naming in Go is usually concise, including test names :) Thanks for giving it a shot ❤️

tests/browser_context_test.go
inancgumus: Thanks! It looks good.

I would remove the self-explanatory parts like:
- "// Call waitForEvent."
- "// Call newPage."
- "// For the happy paths."
- "// For the failure cases."

It's fine to keep only the ones that include whys. Also, the func comments you added are nice 👍

browser/mapping.go
```diff
@@ -648,7 +648,22 @@ func mapBrowserContext(vu moduleVU, bc api.BrowserContext) mapping {
...
-		"waitForEvent":       bc.WaitForEvent,
+		"waitForEvent": func(event string, optsOrPredicate goja.Value) *goja.Promise {
+			ctx := vu.Context()
+			return k6ext.Promise(ctx, func() (result any, reason error) {
+				resp, err := bc.WaitForEvent(event, optsOrPredicate)
+				if err != nil {
+					return nil, err //nolint:wrapcheck
+				}
+
+				p, ok := resp.(api.Page)
```
ka3de, line 659: I wonder if this verification should be moved to the `BrowserContext.waitForEvent` implementation, and in case it fails there, we would already receive an error here from the `bc.waitForEvent` call, that then can be propagated for the promise.

Did we move this here because we want to panic if the data received is not a page and we want to avoid panicking away from the mapping layer?

_Related with the discussion [here](https://github.com/grafana/xk6-browser/pull/1042/commits/2e2e0637c43c5ea5f7cd3838bdfec987e2e24866#r1332839382)._

tests/browser_context_test.go
inancgumus: Thanks, see: https://github.com/grafana/xk6-browser/pull/1042#discussion_r1334015121

ka3de (commented):
Great work @ankur22 ! :clap: These event handling features through users defined code are tricky. Just made some small comments.

browser/mapping.go
ka3de: I have mixed thoughts here (see my [comment](https://github.com/grafana/xk6-browser/pull/1042#discussion_r1334016282) kind of related to this).
I guess the first thing that we should decide is if receiving an event data that does not comply with a `Page` should produce a panic or not.

Many times we used panic because there was not a "direct channel" to communicate the error to the user, but in this case we do have it as we are in the context of a promise, so would it make sense to return every error, even if it is an unexpected one, through this promise?

inancgumus: I believe unexpected programming mistakes should cause a panic (i.e., [1](https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/crypto/tls/conn.go;l=252-274), [2](https://github.com/golang/go/blob/master/src/cmd/compile/internal/base/flag.go#L357-L433)), not an error. Errors are for expected and usual working conditions, like a file is not found, or a network is down. That's how I see it, and I'm happy to leave the final decision to you 👍

common/browser_context.go
ankur22: Nice catch! 🤦 No idea how I overlooked this. If a `FatelError` is returned from here, then we can panic on the mapping layer, which is fine since that will panic on the "main" goroutine.

Resolved in: ee6a951ced1eabc035e6d1d0efcdb4431e3c2d65

browser/mapping.go
ankur22: Yeah, i agree with both of what you are saying. Here's an update:

I somehow missed this type cast check that @ka3de pointed out [here](https://github.com/grafana/xk6-browser/pull/1042#discussion_r1334010235) 🤦 So that has now been resolved.

At the moment, `resp` could either be a nil `page` or a valid `page`. I feel that it's probably still valid to panic here still though even though we're now correctly checking the casting downstream from here. WDYT?

common/browser_context.go
ankur22: Yes correct, there will be (might be) more types that we handle later on.

Suggestion added in c07abdf5286789f69e0308a53d4bc83fbc48510b.

common/browser_context.go
ankur22: Resolved in dc62a318fcb0d2ccb9110fb933e30cbed02e103d

tests/browser_context_test.go
ankur22: Resolved in a fixup: ee71432cc301e6b3a0f36a809fc0e0f09316375a

common/browser_context.go
ankur22: > if this `EventBrowserContextClose` is fired, we would be returning a `null/undefined` page to the JS context, as we do not do any verification in the `case evData := <-ch:` case in `BrowserContext.waitForEvent` select

Yes, that's the same understanding I have of this.

> But if this event is never fired and we will remove it, then this function will only end either in a TO or with a page being returned and passed to the given predicate function (if any). Is that right?

Yes correct, or if the iteration ends.

Since `EventBrowserContextClose` isn't being emitted, feels like a good opportunity to clean things up. My main concern is that `EventBrowserContextClose`'s value is `close` and other events which are emitted for other reasons also use that same value `close` (e.g. [1](https://github.com/grafana/xk6-browser/blob/main/common/event_emitter.go#L23C39-L23C39) and [2](https://github.com/grafana/xk6-browser/blob/main/common/event_emitter.go#L32)). So I wanted to avoid any unexpected unblocks of `waitForEvent` because a `close` was emitted for an undocumented situation. I've not tested this so I could be wrong and it's safe to leave it in or complete the implementation of emitting `EventBrowserContextClose` when the `browserContext` is closed.

WDYT?

browser/mapping.go
ankur22: > I wonder if this verification should be moved to the BrowserContext.waitForEvent implementation

browser/mapping.go
inancgumus: I'm leaning towards the idea that type mistakes likely come from programming errors, and it might be best to _use the built-in `panic`_. However, I'm open to other approaches and happy to go with what the team thinks is best!

inancgumus (approved):
Thanks for taking my suggestions on board and working on this 👍 🙌

browser/mapping.go
ankur22: You mean panic in the `BrowserContext.waitForEvent` implementation? The reason not to do that would be to avoid panicking on goroutine other than the "main" goroutine.

If you're wondering why use `panicIfFatalError` instead of just the `k6ext.Abort` method directly, i'd be happy to do that too here too if that's the consensus 🙂

ka3de (approved):
LGTM :rocket:

inancgumus:
@ankur22 Please squash the fixup commits before merging 🙇 :)

Issue #1048, opened 2023-09-26
Concurrent write error with waitForEvent.js example script


ankur22:
### Brief summary

When running the `waitForEvent.js` example script in the CI it is [failing](https://github.com/grafana/xk6-browser/actions/runs/6311217342/job/17134781659?pr=1047) (not on all runs though) on `fatal error: concurrent map writes`. I'm unable to reproduce this locally though. My initial guess is that the [predicateFn](https://github.com/grafana/xk6-browser/blob/main/common/browser_context.go#L442) call needs to be made on the [taskqueue](https://github.com/grafana/xk6-browser/blob/main/common/page.go#L1152-L1157) like we do for `page.on('console')`.

### xk6-browser version

a7598e135deecb7ac1f64bea605feb547fda345d

### OS

NA

### Chrome version

NA

### Docker version and image (if applicable)

NA

### Steps to reproduce the problem

Run the examples/waitForEvent.js test script in the CI and it might result in an error. It doesn't seem to occur on all runs. I've failed to reproduce it locally.

### Expected behaviour

The test to run without any issues.

### Actual behaviour

The test fails with the following output:

```bash
fatal error: concurrent map writes

goroutine 58 [running]:
github.com/dop251/goja.(*Runtime).fieldsInfo(0xc002428800, {0x19eef48, 0x168f480})
	github.com/dop251/goja@v0.0.0-20230919151941-fc55792775de/object_goreflect.go:627 +0xbc
github.com/dop251/goja.(*objectGoReflect).init(0xc002a702a0)
	github.com/dop251/goja@v0.0.0-20230919151941-fc55792775de/object_goreflect.go:150 +0x50b
github.com/dop251/goja.(*Runtime).toValue(0xc002428800, {0x16cb400?, 0xc001bca340?}, {0x0?, 0x0?, 0x1?})
	github.com/dop251/goja@v0.0.0-20230919151941-fc55792775de/runtime.go:1919 +0x1885
github.com/dop251/goja.(*Runtime).ToValue(...)
	github.com/dop251/goja@v0.0.0-20230919151941-fc55792775de/runtime.go:1735
github.com/grafana/xk6-browser/common.(*BrowserContext).runWaitForEventHandler(0xc0023822c0, {0x19da488, 0xc000faa910}, 0xc0023cbf20, 0xc0023cbf80, 0xc00004e420, 0xc001ceb260)
[...]

PR #1050, opened 2023-09-26
POC: Fix for waitForEvent predicate 

ankur22:
## What?

NOTE: This is a POC and what I wanted was for us to discuss where would be a better place for placing the taskqueue.

This fixes an issue with `waitForEvent` and forces the predicate function to be called on the taskqueue, just like how `page.on('console')` works.

## Why?

If you run the following test with the existing implementation of `waitForEvent` in `main` (a7598e135deecb7ac1f64bea605feb547fda345d), it will cause unexpected `nil` errors.

<details><summary>Test script</summary>

```js

export const options = {
  scenarios: {
    browser: {
      options: {
        browser: {
            type: 'chromium',
        },
      },
    },
  },
}

  for (var i = 0; i < 10; i++) {
    
    try {
      // We want to wait for two page creations before carrying on.
      var counter = 0
      const promise = context.waitForEvent("page", { predicate: page => {
        if (++counter >= 5) {
          console.log('predicate passed')
          return true
        }
        return false
      } })
      
      // Now we create two pages.
      const page2 = context.newPage()
      const page3 = context.newPage()
      const page4 = context.newPage()
      const page5 = context.newPage()

      // We await for the page creation events to be processed and the predicate
      // to pass.
      console.log('wait for predicate')
      await promise

      page.close()
      page2.close()
      page3.close()
      page4.close()
      page5.close()
    } catch (e) {
      console.log('test failed: ' + e)
    } finally {
      context.close()
      console.log(`iteration ${i} completed`)
    }
  }
};
```

</details>

The reason is that js calls that are ran in the gojs context need to run on the taskqueue otherwise there will be race conditions and unexpected syntonisation issues, since goja seems to not be thread safe. I don't understand the full details of why but something similar was done for `page.on('console')`. After replicating the creation and queueing up of the predicate on the taskqueue the nil issue resolved themselves.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/xk6-browser/issues/1048

### DIFF
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -439,10 +439,6 @@ func (b *BrowserContext) runWaitForEventHandler(
 				return
 			}
 
-			// TODO: Find a better place to instantiate a page/context/iteration
-			//       scoped taskqueue.
-			p.createTaskQueue()
-
 			tempPassed := make(chan bool)
 			tempErrOut := make(chan error)
 			// TODO: Is this is a suitable place to add the predicate function to
--- a/common/browser_context.go
+++ b/common/browser_context.go
@@ -439,14 +439,59 @@ func (b *BrowserContext) runWaitForEventHandler(
 				return
 			}
 
-			if retVal, err := predicateFn(b.vu.Runtime().ToValue(p)); err == nil && retVal.ToBoolean() {
+			// TODO: Find a better place to instantiate a page/context/iteration
+			//       scoped taskqueue.
+			p.createTaskQueue()
+
+			tempPassed := make(chan bool)
+			tempErrOut := make(chan error)
+			// TODO: Is this is a suitable place to add the predicate function to
+			//       the taskqueue?
+			p.queueTask(func() error {
+				retVal, err := predicateFn(b.vu.Runtime().ToValue(p))
+				if err != nil {
+					tempErrOut <- fmt.Errorf("executing predicate function: %w", err)
+					return nil
+				}
+
+				tempPassed <- retVal.ToBoolean()
+				return nil
+			})
+
+			select {
+			case <-ctx.Done():
 				b.logger.Debugf(
-					"BrowserContext:runWaitForEventHandler:go():EventBrowserContextPage:predicateFn:return",
+					"BrowserContext:runWaitForEventHandler:go():EventBrowserContextPage:predicateFn:contextCanceled",
 					"bctxid:%v", b.id,
 				)
-				out <- p
+				return
+			case v := <-tempPassed:
+				if !v {
+					b.logger.Debugf(
+						"BrowserContext:runWaitForEventHandler:go():EventBrowserContextPage:predicateFn:false",
+						"bctxid:%v", b.id,
+					)
+
+					// Predicate didn't pass so wait and try again on the next
+					// EventBrowserContextPage event.
+					continue
+				}
+			case err := <-tempErrOut:
+				b.logger.Debugf(
+					"BrowserContext:runWaitForEventHandler:go():EventBrowserContextPage:predicateFn:error",
+					"bctxid:%v", b.id,
+				)
+				errOut <- err
 				return
 			}
+
+			b.logger.Debugf(
+				"BrowserContext:runWaitForEventHandler:go():EventBrowserContextPage:predicateFn:passed",
+				"bctxid:%v", b.id,
+			)
+			out <- p
+
+			return
 		}
 	}
 }
--- a/common/page.go
+++ b/common/page.go
@@ -89,7 +89,8 @@ type Page struct {
 	routes           []api.Route
 	vu               k6modules.VU
 
-	tq *taskqueue.TaskQueue
+	tqOnce sync.Once
+	tq     *taskqueue.TaskQueue
 
 	logger *log.Logger
 }
@@ -794,6 +795,12 @@ func (p *Page) MainFrame() api.Frame {
 	return mf
 }
 
+func (p *Page) createTaskQueue() {
+	p.tqOnce.Do(func() {
+		p.tq = taskqueue.New(p.vu.RegisterCallback)
+	})
+}
+
 // On subscribes to a page event for which the given handler will be executed
 // passing in the ConsoleMessage associated with the event.
 // The only accepted event value is 'console'.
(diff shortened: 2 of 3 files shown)


common/browser_context.go
```diff
@@ -439,14 +439,58 @@ func (b *BrowserContext) runWaitForEventHandler(
 				return
 			}
 
-			if retVal, err := predicateFn(b.vu.Runtime().ToValue(p)); err == nil && retVal.ToBoolean() {
+			// TODO: Find a better place to instantiate a page/context/iteration
```
ankur22, line 442: How should we scope the taskqueue? Should it be scoped at the page, browserContext, browser, or iteration? Or should there be a taskqueue scoped at all these levels? For now `page` will be the best place for it since all the APIs that need to work with it require the `page` in some form either directly or indirectly (`page.on` and `browserContext.waitForEvent`).

common/browser_context.go
```diff
@@ -439,14 +439,58 @@ func (b *BrowserContext) runWaitForEventHandler(
...
+			//       scoped taskqueue.
+			p.createTaskQueue()
+
+			tempPassed := make(chan bool)
+			tempErrOut := make(chan error)
+			// TODO: Is this is a suitable place to add the predicate function to
```
ankur22, line 448: Is it ok to add things to the taskqueue within the business logic or should we try to add it in the mapping layer?

common/browser_context.go
ka3de: When I was working in the `page.on` feature I tried exploring on abstracting the `TaskQueue` to a global level, such as the [registry](https://github.com/grafana/xk6-browser/blob/89bb1fbb705d7d511f4f344684d0c9147dc6c3d6/browser/registry.go) implementations that we have right now, and pass it encapsulated in the `VU` or its context. The problem is that we need some event that allows us to close the `TaskQueue` at the proper time so the iteration can finish (e.g.: the `page.close` action that we are currently using because the `TaskQueue` is scoped at the `page` level). The most intuitive event to do this in a global level is the `IterEnd` event, but this will not happen unless the `TaskQueue` is closed and then the iteration can end, so it's not a valid use case.

I think this is the biggest hurdle to move the `TaskQueue` to a "global" level. In any case, if we realize that "all"/most of the use cases for the `TaskQueue` are in the `page` implementation, for me it's OK to keep it at this level polishing a little bit the APIs so it can be used easily for different methods.

inancgumus: Even if we don't have the `IterEnd` event support, we can still move it to the mapping layer. We can call `p.tq.Queue` and `p.tq.Close` in the mapping layer's `mapMsgAndHandleEvent` and `mapPage.close`, and there shouldn't be any behavioral differences. The thing that we should be careful is to find a way to share `TaskQueue` between the mappers and pages.

common/browser_context.go
inancgumus: Blending `TaskQueue` into the business logic binds us more to Goja and k6 instead of native Go channel constructs. This makes tests complicated and codes error-prone if we continue to let it infiltrate more. The [tests look more like event loop event handling](https://github.com/grafana/xk6-browser/blob/main/tests/page_test.go#L923-L983) instead of focusing on the actual logic. This makes them harder to craft and understand and makes them error-prone. 

I have a strong opinion but can hold it weakly, depending on our discussions :) It's fine to continue blending it to make progress until we find a better way, though 👍

common/browser_context.go
ankur22: Thanks for the input!

>  if we realize that "all"/most of the use cases for the TaskQueue are in the page implementation, for me it's OK to keep it at this level polishing a little bit the APIs so it can be used easily for different methods.

Yeah, that's what it currently seems like 👍 I also think in the short term this is good enough.

> The most intuitive event to do this in a global level is the IterEnd event, but this will not happen unless the TaskQueue is closed and then the iteration can end, so it's not a valid use case.

Ah yes, that was a difficulty with the `TaskQueue`. Makes sense for it to be scoped at a lower level in that case. COuld it be scoped at the `Browser` level? That way we could rely on ourselves to ensure that the `TaskQueue` is closed when the `Browser` is no longer needed. I wonder though if that wouldn't work with the event loop.

> We can call p.tq.Queue and p.tq.Close in the mapping layer's mapMsgAndHandleEvent and mapPage.Close

I like this idea. Just as a note, we would still need to call `p.tq.Close` from the `page.close` though since `page.on` works with logs which can be logged at any point in the test while a website is being tested.

>  find a way to share TaskQueue between the mappers and pages.

Could we still work with the registry? When a page is created, could we register a new `TaskQueue` and when a page is closed, close the `TaskQueue` associated to the page that is closing.

inancgumus: Sure :)

> the short term this is good enough

👍 

> Could it be scoped at the Browser level?

I haven't checked this out. But sounds like a good idea because it makes sense `Browser` to have a single taskqueue—shared by all pages. The thing is, will handlers attached with `page.On` be called before the `Browser` closes? So that the handlers run. Since we have [a possible race condition](https://github.com/grafana/xk6-browser/issues/971) near a browser close.

> we would still need to call p.tq.Close from the page.close

Yep, for sure. This is also related to my previous comment above. Is it possible to close it once from the browser side or not? This might take some investigation.

> Could we still work with the registry?

It's worth giving a try 👍 Or, we need to find another way of sharing 🤔

ka3de: I've given a try to move the `TaskQueue` to the mapping layer. I've done this from `main` ([84deb83](https://github.com/grafana/xk6-browser/commit/84deb8332a89d3c43cfec6d233d4677ac572c79e)) and only based in the `page.on` method by now. You can see my attempt in [this branch](https://github.com/grafana/xk6-browser/compare/poc/tq-to-mapping).

It is very experimental. I have doubts whether there is an issue on creating multiple local `TasQueue` instances, one per handler, but with this approach, even if the handler is not executed, the test finishes correctly, and we do not require users to close anything explicitly. Actually I believe this approach (the local `TaskQueue`s) is something that we could probably implement with our current main implementation.

I also tried moving the `TaskQueue` to be a VU scoped variable, by including it in the `moduleVU` struct, but in that case I was hitting again the issue that we don't know when to close the `TaskQueue` so there is a deadlock between the `TaskQueue` and the iteration execution.

This needs refinement and better understanding of implications, but I think there are interesting learning points.

ankur22:
Final solution in https://github.com/grafana/xk6-browser/pull/1069

PR #975, opened 2023-07-14
Update to the latest master version of k6

ankur22:
### Description of changes

Updating to the latest version of `master` k6. The main reason for this is to pick up a cloud output fix from k6.

### Checklist

- [ ] Written tests for the changes
- [ ] Update k6 documentation (if relevant) -- PR link: ?
- [ ] Update [k6 browser get started](https://k6.io/blog/get-started-with-k6-browser/) blog (if relevant) -- PR link: ?
- [ ] Generate and update [TypeScript definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/k6/experimental/browser) (if relevant)

PR #830, opened 2023-03-16
Update to k6-core master and fix BPool

inancgumus:
We [detected an incompatibility](https://github.com/grafana/xk6-browser/actions/runs/4434049483/attempts/2) with k6-core:

- `BPool` becomes `BufferPool` in this [PR](https://github.com/grafana/k6/pull/2879).

This PR fixes the issue.

PR #739, opened 2023-01-30
Update/k6 master

ankur22:

PR #1051, opened 2023-09-27
Add dependabot

inancgumus:
## What?

Adds Github dependabot for our repository.

## Why?

To keep our dependencies up to date, like k6-core, CDP proto, etc.

We usually do this manually, which can be a hassle and causes our dependencies go stale:
- #1047, #975, #907, #830, #740, #739, ...

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,8 +10,6 @@ updates:
         dependency-type: "all"
       - dependency-name: "github.com/chromedp/cdproto"
         dependency-type: "all"
-      - dependency-name: "github.com/mailru/easyjson"
-        dependency-type: "all"
       - dependency-name: "github.com/mstoykov/k6-taskqueue-lib"
         dependency-type: "all"
       - dependency-name: "github.com/stretchr/testify"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,8 +18,6 @@ updates:
         dependency-type: "all"
       - dependency-name: "github.com/mccutchen/go-httpbin"
         dependency-type: "all"
-      - dependency-name: "github.com/dop251/goja"
-        dependency-type: "all"
     target-branch: "main"
     commit-message:
       prefix: "Bump "
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,6 +20,8 @@ updates:
         dependency-type: "all"
       - dependency-name: "github.com/mccutchen/go-httpbin"
         dependency-type: "all"
+      - dependency-name: "github.com/dop251/goja"
+        dependency-type: "all"
     target-branch: "main"
     commit-message:
       prefix: "Bump "
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,32 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 10
+    allow:
+      - dependency-name: "go.k6.io/k6"
+        dependency-type: "all"
+      - dependency-name: "github.com/chromedp/cdproto"
+        dependency-type: "all"
+      - dependency-name: "github.com/mailru/easyjson"
+        dependency-type: "all"
+      - dependency-name: "github.com/mstoykov/k6-taskqueue-lib"
+        dependency-type: "all"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
+version: 2
+updates:
+
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 10
+    allow:
```
ankur22, line 9: Should we also add `goja`? What was the criteria for add these in here?

ankur22 (approved):
Seems like a good idea.

My only thought around this is that we shouldn't update before k6 unless there's a good reason to do so (blocking us due to a bug in a dependency etc).

.github/dependabot.yml
inancgumus: Yes, I was indecisive about Goja 😟 Although we have a direct dependency, k6 also has its own Gona. Any suggestions?

ankur22: I'd be happy top add it in and see how the bot behaves.

inancgumus:
> Seems like a good idea.

Thanks, @ankur22 😊

> My only thought around this is that we shouldn't update before k6 unless there's a good reason to do so (blocking us due to a bug in a dependency, etc.).

No worries. The bot only makes PR requests, and we can decide whether to merge them. It has a UI that we can talk to using Github comments. So, we can include or exclude some of the steps.

inancgumus:
@ankur22 I made some changes because the configuration syntax was incorrect.

ankur22 (approved):
LGTM 🚀

.github/dependabot.yml
```diff
@@ -0,0 +1,34 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 10
+    allow:
```
ka3de, line 8: I used dependabot in the past and was working well, but I think in an "extension" repo this can be tricky. Similar to @ankur22 comment [here](https://github.com/grafana/xk6-browser/pull/1051#pullrequestreview-1646158314), I think we should add to the allowed list only the dependencies that are directly related with xk6-browser, for the rest, the ones that are used in k6 and transitively in k6 browser, I would not add them, as we will always be "restricted" by k6 core updates pace.

What I would make sure to include is every security related update, so at least we are aware. Not sure if this is already included in this definition :thinking:

inancgumus: I agree; that's why I [was hesitant](https://github.com/grafana/xk6-browser/pull/1051#discussion_r1338421455) to add Goja. Yes, the security updates should be included. And, as I [explained earlier](https://github.com/grafana/xk6-browser/pull/1051#issuecomment-1737160204), we can manage the bot by [its UI](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/managing-pull-requests-for-dependency-updates#managing-dependabot-pull-requests-with-comment-commands). So it won't be like a fully automated progress, and we don't have to include everything the bot says. We can fine-tune this config in time. Let's first see how it will work for us.

ka3de: So should we remove the goja dependency then? I think it will add just noise for us to review if we are just going to do whatever k6 does, and also that repo is pretty active in development.

inancgumus: As I said, we can tune the bot in time :) But, sure, we can do that 👍 @ankur22, WDYT?

ankur22: Yeah, sounds good 👍

Should the other shared dependencies also be removed such as `github.com/stretchr/testify` and `github.com/mailru/easyjson`?

inancgumus: I think we can keep `testify` and remove `easyjson`. `testify` is fine because we never merge tests into k6-core.

inancgumus: See [b2c72c4](https://github.com/grafana/xk6-browser/pull/1051/commits/b2c72c44d583154027e4444f9943adc65deb9f69) and [8e28d45](https://github.com/grafana/xk6-browser/pull/1051/commits/8e28d4525801d524098f783d4d6b79b0730aabd1).

ankur22 (approved):
LGTM 🚀

ka3de (approved):
LGTM :clap:

PR #1055, opened 2023-09-29
Remove the api package

inancgumus:
## What?

- Suffixes the `api` pkg types with `API` and fixes the linter errors (commits that start with "Rename".)
- Moves the `API` types into the `common` pkg (commits that start with "Move" and "Use".)
- Removes the `api` pkg for good (commits that start with "Remove".)

## Why?

- We postfix the `api` types with `API` to distinguish them from the `common` types. So, we can move them into the `common` pkg.
- Moving them into the `common` pkg is necessary because there are many dependencies to these types throughout the module, and we can only sanely refactor these into concrete types within the `common` pkg.
- In another PR, we can turn these interfaces into concrete types now that we don't have an `api` pkg dependency.

See the other reasonings in #898.

## Checklist

## Related PR(s)/Issue(s)

Updates: #898 (Actually, it finishes the work, but it's better to move the interfaces to concrete types to realize the issue's goal fully).


inancgumus:
I'll squash this PR when I finish.

PR #1057, opened 2023-10-06
Refactor the API interfaces to concrete types

### LINKED ISSUE #898: Remove the `api` package to use concrete `common` types

inancgumus:
## What?

* Uses the concrete types in `common` instead of the API interface types.
* Unexports and moves the API interface declarations into the mapping layer test.
* Breaks down `types.go` and moves types to where they are closely used.

## Why?

* Using concrete types makes sense as they make understanding and navigation in the code straightforward. Also, we only support one browser. And, we can support more thanks to the mapping layer.
* Removes unnecessary type assertions.
* Removes unnecessary interface redirects.

## Checklist

## Related PR(s)/Issue(s)

Closes grafana/xk6-browser#898

## Note 👀

Although this PR is relatively large, reviewing it should be relatively easy. It's because most of the changes are done by the editor auto-renaming and moving. So, there should be few issues present. Each commit is organized to move a single interface type to a concrete type for easy reviewing.


browser/mapping_test.go
```diff
@@ -239,3 +239,17 @@ func isCustomMapping(customMappings map[string]string, typ, method string) (stri
 
 	return "", false
 }
+
+// ----------------------------------------------------------------------------
+// JavaScript API definitions.
+// ----------------------------------------------------------------------------
+
+// keyboardAPI is the interface of a keyboard input device.
+// TODO: map this to page.GetKeyboard().
```
ka3de, line 593: Can you clarify what does this TODO mean?

browser/mapping_test.go
```diff
@@ -265,3 +265,14 @@ type cdpSessionAPI interface { //nolint: unused
 	Detach()
 	Send(method string, params goja.Value) goja.Value
 }
+
+// mouseAPI is the interface of a mouse input device.
+// TODO: map this to page.GetMouse().
```
ka3de, line 615: Can you clarify what does this TODO mean?

common/element_handle.go
```diff
@@ -22,8 +22,8 @@ const resultDone = "done"
 
 // Ensure ElementHandle implements the api.ElementHandle and api.JSHandle interfaces.
 var (
-	_ ElementHandleAPI = &ElementHandle{}
-	_ JSHandleAPI      = &ElementHandle{}
+	_ *ElementHandle = &ElementHandle{}
```
ka3de, line 25: What is the purpose of this?

browser/mapping_test.go
```diff
@@ -511,6 +511,81 @@ type responseAPI interface {
 	URL() string
 }
 
+// Strict mode:
+// All operations on locators throw an exception if more
+// than one element matches the locator's selector.
+//
+// See Issue #100 for more details.
```
ka3de, line 521: As a general comment, to me it seems odd that we have moved all these comments for all interfaces to a test file. Would it make more sense to move them to the type implementation instead?

common/types.go
```diff
@@ -574,3 +574,20 @@ func (c *Credentials) Parse(ctx context.Context, credentials goja.Value) error {
...
+// HTTPMessageSize are the sizes in bytes of the HTTP message header and body.
+type HTTPMessageSize struct {
+	Headers int64 `json:"headers"`
+	Body    int64 `json:"body"`
+}
+
+// Total returns the total size in bytes of the HTTP message.
+func (s HTTPMessageSize) Total() int64 {
+	return s.Headers + s.Body
+}
```
ka3de, line 589: This `types` file seems too generic, and probably all elements defined in it could be moved somewhere else. But focusing on these modifications, would it make sense to move these types to a separate file? Such as `http.go` maybe?

common/browser_context.go
```diff
@@ -27,6 +27,42 @@ import (
 // with the browserContext.waitForEvent API.
 type waitForEventType string
 
+// Cookie represents a browser cookie.
+//
+// https://datatracker.ietf.org/doc/html/rfc6265.
+type Cookie struct {
```
ka3de, line 33: I wonder if types such as `Cookie` and `ConsoleMessage` should be in their own files for better organization. At the same time, considering the "small associated implementation" and that they are pretty much tight to `BrowserContext`/`Page`, I can see them placed along with these types (as they are), so no strong opinion.

common/api.go
```diff
@@ -6,6 +6,10 @@ import (
 )
 
 // JSHandleAPI is the interface of an in-page JS object.
+//
+// TODO: Find a way to move this to a concrete type. It's too difficult to
+// do that right now because of the tests and the way we're using the
+// JSHandleAPI interface.
```
ka3de, line 12: If the complexity here is because of tests, could we define specific `JSHandle` interfaces in the tests and in the rest use the concrete type? So at least we can get rid of this `common/api.go` file? It's a pity to leave just one type as interface everywhere.

ka3de (commented):
Thanks for working on this @inancgumus !
I still think interfaces provide a few benefits that we were not taking advantage of, but it's true that our previous implementation was not right. Having too long interfaces definition and in a dedicated package was wrong, so I think this is an improvement and we can iterate introducing more specific interfaces that actually improve the codebase.

Made a few small comments. Also, I believe these comments should be updated:
https://github.com/grafana/xk6-browser/blob/5528008fe9ba4b5c73d2b259eee60fc419d19b76/browser/mapping_test.go#L20
https://github.com/grafana/xk6-browser/blob/5528008fe9ba4b5c73d2b259eee60fc419d19b76/browser/mapping_test.go#L51

common/element_handle.go
inancgumus: Nice catch. We can remove it 👍

browser/mapping_test.go
inancgumus: Makes sense 👍

common/types.go
inancgumus: Sure 👍

common/browser_context.go
inancgumus: Keeping them close to where they are used makes more sense to me. I'd like to keep them as they are. WDYT?

common/api.go
inancgumus: `JSHandle` complexity is dispersed over the whole module. I've tried to move this to a concrete type but couldn't. Not only because of the tests but also because of the code dependencies. We can discuss how to refactor this and send another PR, as it's quite complex to do in this PR.

inancgumus:
> Thanks for working on this @inancgumus! I still think interfaces provide a few benefits that we were not taking advantage of, but it's true that our previous implementation was not right. Having too long interfaces definition and in a dedicated package was wrong, so I think this is an improvement and we can iterate introducing more specific interfaces that actually improve the codebase.

Agreed, thanks for your review!

browser/mapping_test.go
```diff
@@ -253,3 +253,9 @@ type keyboardAPI interface { //nolint: unused
 	Type(text string, opts goja.Value)
 	Up(key string)
 }
+
+// touchscreenAPI is the interface of a touchscreen.
+// TODO: map this to page.GetTouchscreen().
```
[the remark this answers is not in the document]
inancgumus: See: 5528008

browser/mapping_test.go
inancgumus: See: 5528008

common/element_handle.go
inancgumus: See: 4c3e2c2

browser/mapping_test.go
inancgumus: See: af05e07

common/types.go
inancgumus: See: 24ed386 and 62255c2

browser/mapping_test.go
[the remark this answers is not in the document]
ka3de: I see you updated the [comment](https://github.com/grafana/xk6-browser/blob/f0c037e4fd528ab79e6be2b53134212ea3b3067c/browser/mapping_test.go#L537-L540), but I'm not sure I'm understanding this properly.

With these changes for `page.[GetKeyboard()|GetMouse()|GetTouchscreen()`, because now we return the concrete type instead of the interfaces, aren't we exposing properties and methods to the JS layer that we do not want to expose?

browser/mapping_test.go
```diff
@@ -239,3 +235,360 @@ func isCustomMapping(customMappings map[string]string, typ, method string) (stri
...
+// touchscreenAPI is the interface of a touchscreen.
+// TODO: map this to page.GetTouchscreen(). Currently, the common.TouchscreenAPI type
+// mapping is not tested using this interface. We use the concrete type
+// without testing its exported methods.
+type touchscreenAPI interface { //nolint: unused
+	Tap(x float64, y float64)
+}
+
+// cdpSessionAPI is the interface of a raw CDP session.
+type cdpSessionAPI interface { //nolint: unused
```
ka3de, line 558: A comment for `cdpSessionAPI`, `videoAPI` and `routeAPI`. If we are not using them, why should we keep them?

browser/mapping_test.go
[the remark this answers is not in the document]
inancgumus: Yes, we expose their properties, but we haven't been testing what methods they return. For that, we need to check what methods they return using their API interfaces defined in the mapping layer. For example:

https://github.com/grafana/xk6-browser/blob/f0c037e4fd528ab79e6be2b53134212ea3b3067c/browser/mapping.go#L495

`keyboard` returns `*Keyboard`, but `page` mapping test doesn't test `*Keyboard` methods:

https://github.com/grafana/xk6-browser/blob/f0c037e4fd528ab79e6be2b53134212ea3b3067c/browser/mapping_test.go#L138-L147

The test should use the `keyboardAPI` interface. Even if you change the methods in `keyboardAPI`, the `page` mapping test would still pass.

browser/mapping_test.go
inancgumus: Sure, we can remove them: 43dfcab

browser/mapping_test.go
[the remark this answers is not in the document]
inancgumus: I created an issue for this https://github.com/grafana/xk6-browser/issues/1061.

ka3de (approved):
LGTM.

PR #1069, opened 2023-10-16
Move taskqueue to mapping layer

ankur22:
## What?

This PR moves the taskqueue to the mapping layer. The change will affect `page.on` and `browserContext.waitForEvent`.

## Why?

As we work more with goja and it's runtime, it is starting to find its way into our business logic code. This make maintaining the project more difficult due to having to work with raw `goja.Value` and (more recently) the event loop and `taskqueue`. One of these areas that we do not want in the business logic is most definitely the `taskqueue` which was introduced when we added `page.on`. `page.on` requires a predicate method on the goja runtime since the function is written in JS. Since the goja runtime is not thread safe we need to work with the `taskqueue`. Working with the `taskqueue` in the business logic makes reasoning, maintaining and writing tests very tricky since various components need to be orchestrated in multiple goroutines.

This change forces the creation and deletion of the `taskqueue` in the mapping layer. This change also forces us to convert `goja.Value`s to native go structs, which has enabled us to move some goja code out of the business logic (specifically for `browsercontext.waitForEvent`). We can verify that the changes work with JS test scripts, whereas the business logic can be tested with native go code using either unit or integration tests with less go routine orchestration.

## Checklist

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -433,7 +433,7 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
 			})
 		},
 		"close": func(opts goja.Value) error {
-			vu.taskQueueRegistry.close(p.GetTargetID())
+			vu.taskQueueRegistry.close(p.TargetID())
 
 			return p.Close(opts) //nolint:wrapcheck
 		},
@@ -507,7 +507,7 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
 		},
 		"mouse": rt.ToValue(p.GetMouse()).ToObject(rt),
 		"on": func(event string, handler goja.Callable) error {
-			tq := vu.taskQueueRegistry.get(p.GetTargetID())
+			tq := vu.taskQueueRegistry.get(p.TargetID())
 
 			mapMsgAndHandleEvent := func(m *common.ConsoleMessage) error {
 				mapping := mapConsoleMessage(vu, m)
@@ -671,7 +671,7 @@ func mapBrowserContext(vu moduleVU, bc *common.BrowserContext) mapping { //nolin
 				var runInTaskQueue func(p *common.Page) (bool, error)
 				if parsedOpts.PredicateFn != nil {
 					runInTaskQueue = func(p *common.Page) (bool, error) {
-						tq := vu.taskQueueRegistry.get(p.GetTargetID())
+						tq := vu.taskQueueRegistry.get(p.TargetID())
 
 						var rtn bool
 						var err error
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -435,8 +435,10 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
 		},
 		"close": func(opts goja.Value) error {
 			vu.tqMu.Lock()
-			if vu.tq != nil {
-				vu.tq.Close()
+			tq := vu.tq[p.GetTargetID()]
+			if tq != nil {
+				tq.Close()
+				delete(vu.tq, p.GetTargetID())
 			}
 			vu.tqMu.Unlock()
 
@@ -513,8 +515,10 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
 		"mouse": rt.ToValue(p.GetMouse()).ToObject(rt),
 		"on": func(event string, handler goja.Callable) error {
 			vu.tqMu.Lock()
-			if vu.tq == nil {
-				vu.tq = taskqueue.New(vu.RegisterCallback)
+			tq := vu.tq[p.GetTargetID()]
+			if tq == nil {
+				tq = taskqueue.New(vu.RegisterCallback)
+				vu.tq[p.GetTargetID()] = tq
 			}
 			vu.tqMu.Unlock()
 
@@ -525,7 +529,7 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
 			}
 			runInTaskQueue := func(m *common.ConsoleMessage) {
 				vu.tqMu.RLock()
-				vu.tq.Queue(func() error {
+				tq.Queue(func() error {
 					if err := mapMsgAndHandleEvent(m); err != nil {
 						return fmt.Errorf("executing page.on handler: %w", err)
 					}
@@ -672,12 +676,6 @@ func mapBrowserContext(vu moduleVU, bc *common.BrowserContext) mapping { //nolin
 		"waitForEvent": func(event string, optsOrPredicate goja.Value) *goja.Promise {
 			ctx := vu.Context()
 			return k6ext.Promise(ctx, func() (result any, reason error) {
-				vu.tqMu.Lock()
-				if vu.tq == nil {
-					vu.tq = taskqueue.New(vu.RegisterCallback)
-				}
-				vu.tqMu.Unlock()
-
 				parsedOpts := common.NewWaitForEventOptions(
 					bc.Timeout(),
 				)
@@ -688,6 +686,14 @@ func mapBrowserContext(vu moduleVU, bc *common.BrowserContext) mapping { //nolin
 				var runInTaskQueue func(p *common.Page) (bool, error)
 				if parsedOpts.PredicateFn != nil {
 					runInTaskQueue = func(p *common.Page) (bool, error) {
+						vu.tqMu.Lock()
+						tq := vu.tq[p.GetTargetID()]
+						if tq == nil {
+							tq = taskqueue.New(vu.RegisterCallback)
+							vu.tq[p.GetTargetID()] = tq
+						}
+						vu.tqMu.Unlock()
+
 						var rtn bool
 						var err error
 						// The function on the taskqueue runs in its own goroutine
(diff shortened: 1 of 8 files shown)


browser/mapping.go
```diff
@@ -434,9 +434,16 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
 			})
 		},
 		"close": func(opts goja.Value) error {
-			if vu.tq != nil {
+			// There could be multiple pages and they could all need the taskqueue
+			// which means that we should only close the taskqueue once we only
+			// have one or less pages in the browserContext.
+			pp := p.Context().Pages()
+
+			if len(pp) <= 1 && vu.tq != nil {
```
ankur22: This fix will never work since the page close is asynchronous, which means we would have to work with an event loop or wait here for some time until the page count decreases by one. Neither of these are good solution. Instead i've opted to work with a `taskqueue` per page, which you will find in a subsequent commit.

tests/browser_context_test.go
```diff
@@ -750,119 +745,50 @@ func TestBrowserContextWaitForEvent(t *testing.T) {
...
-	return strPointer
-}
-
-func int64Ptr(value int64) *int64 {
-	int64Pointer := new(int64)
-	*int64Pointer = value
-	return int64Pointer
-}
+				var resp any
+				close(aboutToCallWait)
```
ka3de, line 758: I understand we use this channel to guarantee that `bc.NewPage()` is called after `bc.WaitForEvent`, is that right? If that is the case, should we close it after actually calling `waitForEvent`?

tests/browser_context_test.go
```diff
@@ -750,119 +745,50 @@ func TestBrowserContextWaitForEvent(t *testing.T) {
...
 
-		opts := tb.toGojaValue(struct {
-			Predicate goja.Value
-		}{
-			Predicate: predicate,
-		})
+			select {
+			case <-waitDone:
+			case <-ctx.Done():
+				err = ctx.Err()
```
ka3de, line 780: Should we assert here that this error is `nil` which would mean that the context was not canceled due to TO? Or maybe use another error, so we don't overlap `err` which is used to store the error resulting from `bc.WaitForEvent` and assert it separately?

browser/mapping.go
```diff
@@ -513,8 +515,10 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
...
 		"on": func(event string, handler goja.Callable) error {
 			vu.tqMu.Lock()
-			if vu.tq == nil {
-				vu.tq = taskqueue.New(vu.RegisterCallback)
+			tq := vu.tq[p.GetTargetID()]
+			if tq == nil {
+				tq = taskqueue.New(vu.RegisterCallback)
+				vu.tq[p.GetTargetID()] = tq
 			}
 			vu.tqMu.Unlock()
```
ka3de, line 523: I wonder if we should implement a higher level API in `moduleVU` to encapsulate this logic and similar ones that are in the mapping layer now so it is a little cleaner?

browser/mapping.go
```diff
@@ -688,14 +686,22 @@ func mapBrowserContext(vu moduleVU, bc *common.BrowserContext) mapping { //nolin
 				var runInTaskQueue func(p *common.Page) (bool, error)
 				if parsedOpts.PredicateFn != nil {
 					runInTaskQueue = func(p *common.Page) (bool, error) {
+						vu.tqMu.Lock()
```
ka3de, line 689: If I understand correctly, we use this `vu.tqMu` to sync access for all task queues used for every page independently. So here for example we use it as the same task queue for a given page might be used concurrently, for example in a handler registered through `page.on` and a predicate in `waitForEvent`. Is this correct?

ka3de (commented):
Thank you for working on this @ankur22 !
Overall I think it looks good, it's true that we are adding complexity to the mapping layer, which ideally it's not desired, but it's a tradeoff in order to simplify our core logic, so I think it's probably a good step forward.

tests/browser_context_test.go
ankur22: I see where you are getting at, i'm wondering whether it might be clearer to assert the context canceled error separately.

The idea is that the `err` is checked in the lines below. If `wantErr` is an empty string, then `err` should be `nil`. If we expect an error then it should match `wantErr`.

browser/mapping.go
ankur22: Yep, that was something I was considering. Let me give it a go 👍

browser/mapping.go
ankur22: One more thing to add is that once a taskqueue is retrieved from the taskqueues, we're not protecting the access to that since it look thread safe. So i would expect that `tq.Queue(func() error {...` can be called from multiple threads on the same `tq` and that's ok. So `tqMu` is protecting the taskqueues data structure itself.

ankur22: Oh, but i think i know what you're getting at now though, there's no need to use `tqMu` when working with `tq.Queue`. That'll be fixed in the next commit with the taskqueue higher level API refactor.

browser/mapping.go
ankur22: Moved the implementation to a higher level (`taskqueueRegistry`): 8ddf4494b121cd895c155ecfcd128d8c6b72870e, fb900e1b26a7ee9f75cc668c9503ac5bdd597521, d422abaf4531c5e490fa876e62d8804287601580.

browser/mapping.go
ankur22: Fixed in 1e7d7b75cfcd5dc3405faf5ad7e3354ccad561fc

tests/browser_context_test.go
ka3de: Yes, I understood how it works, but it was a little bit tricky. That's why I thought it could make sense to be more explicit on asserting both possible errors. But no strong opinion, we can leave it as it is also.

browser/mapping.go
ka3de: Yep, so that was my doubt basically, whether `taskQueue` itself was actually thread safe, as we could be queuing tasks concurrently between different executions of `waitForEvent` and `on`. But [it is actually thread](https://github.com/mstoykov/k6-taskqueue-lib/blob/111459812ad27077dfadf4cd01c131019d24b7ae/taskqueue/queuer.go#L62-L63) safe within its implementation, so I think this change is correct.

common/page.go
```diff
@@ -1286,6 +1270,12 @@ func (p *Page) Workers() []*Worker {
 	return workers
 }
 
+// GetTargetID retrieve the unique id that is associated to this page.
+// For internal use only.
+func (p *Page) GetTargetID() string {
```
inancgumus, line 1275: Should we just say `TargetID` as per https://go.dev/doc/effective_go#Getters.
> it's neither idiomatic nor necessary to put Get into the getter's name. If you have a field called owner (lower case, unexported), the getter method should be called Owner (upper case, exported), not GetOwner

tests/browser_context_test.go
ankur22: Sure, i've added an assert to make it clearer: e18558adba1a3cbef4eaa6b05abdf36323f1f420

inancgumus (dismissed):
It's great that we're moving toward a more straightforward, more maintainable code base 👍 Thanks!

tests/browser_context_test.go
```diff
@@ -750,119 +745,50 @@ func TestBrowserContextWaitForEvent(t *testing.T) {
...
-	int64Pointer := new(int64)
-	*int64Pointer = value
-	return int64Pointer
-}
+				var resp any
+				close(aboutToCallWait)
+				resp, err = bc.WaitForEvent(tc.event, tc.predicate, tc.timeout)
+				if resp != nil {
+					p, ok := resp.(*common.Page)
+					require.True(t, ok)
```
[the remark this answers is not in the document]
ankur22: Good spot! Didn't know that 🙂 Resolved in 04895b2af62610d603a1392a412ac0ced71e04c4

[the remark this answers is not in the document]
inancgumus: Perfect 🙇 :)

tests/browser_context_test.go
```diff
@@ -750,119 +746,58 @@ func TestBrowserContextWaitForEvent(t *testing.T) {
...
+				defer close(waitDone)
+				defer close(errChan)
+
+				var resp any
+				close(aboutToCallWait)
+				resp, err = bc.WaitForEvent(tc.event, tc.predicate, tc.timeout)
+				if resp != nil {
+					p, ok := resp.(*common.Page)
+					if !ok {
+						errChan <- errors.New("response from waitForEvent is not a page")
```
inancgumus, line 766: One more point :) This goroutine won't return if the test goroutine finishes earlier without reading the `errChan` (maybe due to the `Context` timeout). Should we make the `errChan` with a single-item buffered chan to be on the safe side?

ankur22: Good point, what about reusing the context like so:

```go
						select {
						case errChan <- errors.New("response from waitForEvent is not a page"):
						case <-ctx.Done():
						}
```

inancgumus: Makes sense, but it's safer to accept a one-buffered channel if we ever update this code later. These kinds of bugs are subtle and can easily go unnoticed. Have you tried using `tb.run` here? IDK if it makes sense, but it can make things easier.

ankur22: Ah, nice, i like `tb.run`, i was thinking errgroup might be a good fit 🙂  Resolved in 6855e30afdc593b67318ec34e3ad6578dae39a25

inancgumus: Perfect 👏 🙇

inancgumus (approved):
🚀

Issue #1064, opened 2023-10-12
Move option parsing to the mapping layer


inancgumus:
### What?

Move the option parsing of the types (like `common`) to the mapping layer.

### Why?

* Better type-safety.
* Developer productivity increase.
* Separation of concerns.
* Moving Goja handling to the mapping layer will make the code easy to reason about, reducing the business logic's Goja dependency and simplifying the business logic code and tests.

### How?

For example, `BrowserContextOptions` parses the Goja option values and converts them into a `BrowserContextOptions` value (business language). This translation can be made in the mapping layer instead:

```go
func mapBrowser(vu moduleVU) mapping {
	...
	"newContext": func(gopts goja.Value) (*goja.Object, error) {
		b, err := vu.browser()
		if err != nil {
			return nil, err
		}
		opts, err := parseBrowserContextOptions(gopts)
		if err != nil {
			return nil, fmt.Errorf("parsing browser context options: %w", err)
		}
		bctx, err := b.NewContext(opts)
		if err != nil {
			return nil, err //nolint:wrapcheck
		}
		m := mapBrowserContext(vu, bctx)
		return rt.ToValue(m).ToObject(rt), nil
	},
	...
}
```

Then the `Browser.NewContext` can be simplified into:

```go
func (b *Browser) NewContext(opts *BrowserContextOptions) (*BrowserContext, error) {
	// ...
	browserCtx, err := NewBrowserContext(b.ctx, b, browserContextID, opts, b.logger)
	// ...
}
```

Instead of this:

```go
// NewContext creates a new incognito-like browser context.
func (b *Browser) NewContext(opts goja.Value) (*BrowserContext, error) {
	// ...
	browserCtxOpts := NewBrowserContextOptions()
	if err := browserCtxOpts.Parse(b.ctx, opts); err != nil {
		k6ext.Panic(b.ctx, "parsing newContext options: %w", err)
	}
	browserCtx, err := NewBrowserContext(b.ctx, b, browserContextID, browserCtxOpts, b.logger)
	// ...
}
```

The tests can be simplified, too:

```go
func TestBasicAuth(t *testing.T) {
	// ...
	browser := newTestBrowser(t, withHTTPServer())
	bc, err := browser.NewContext(&common.BrowserContextOptions{
		HttpCredentials: &common.Credentials{
			Username: user,
			Password: pass,
		},
	})
	// ...
}
```

Instead of this:

```go
func TestBasicAuth(t *testing.T) {
	// ...
[...]

PR #1075, opened 2023-10-18
Refactor browserContext.grantPermissions

ankur22:
## What?

This refactors the `browserContext.grantPermissions` implementation.

## Why?

This helps bring it inline with current ways of implementing the APIs:
1. Parsing the options in a new parsed type.
2. Returning an `error` instead of panicking.
3. Explicit errors when a permission is invalid/unrecognised.
4. Working with goja and in the mapping layer and not in the business logic.

## Checklist

## Related PR(s)/Issue(s)

Prerequisite to working on https://github.com/grafana/xk6-browser/issues/443


ka3de (dismissed):
LGTM :clap:

browser/mapping.go
```diff
@@ -616,16 +616,21 @@ func mapWorker(vu moduleVU, w *common.Worker) mapping {
...
+		"addInitScript":    bc.AddInitScript,
+		"browser":          bc.Browser,
+		"clearCookies":     bc.ClearCookies,
+		"clearPermissions": bc.ClearPermissions,
+		"close":            bc.Close,
+		"cookies":          bc.Cookies,
+		"exposeBinding":    bc.ExposeBinding,
+		"exposeFunction":   bc.ExposeFunction,
+		"grantPermissions": func(permissions []string, opts goja.Value) error {
+			parsedOpts := common.NewGrantPermissionsOptions()
```
inancgumus, line 629: I don't mind, and this is bikeshedding (😅), but it's more idiomatic to use short variable names within short contexts. Instead of `parsedOpts`, we can just say something like `popts`. Also, before parsing, semantically, `parsedOpts` are not parsed yet.

common/browser_context.go
```diff
@@ -212,7 +215,7 @@ func (b *BrowserContext) ExposeFunction(name string, callback goja.Callable) {
 }
 
 // GrantPermissions enables the specified permissions, all others will be disabled.
-func (b *BrowserContext) GrantPermissions(permissions []string, opts goja.Value) {
+func (b *BrowserContext) GrantPermissions(permissions []string, parsedOpts *GrantPermissionsOptions) error {
```
inancgumus, line 218: Should we just say `opts` here as `BrowserContext` now doesn't care about options parsing?

```suggestion
func (b *BrowserContext) GrantPermissions(permissions []string, opts *GrantPermissionsOptions) error {
```

common/browser_context.go
```diff
@@ -232,28 +235,22 @@ func (b *BrowserContext) GrantPermissions(permissions []string, opts goja.Value)
...
+			return fmt.Errorf("%q is an invalid permission", p)
+		}
+		perms = append(perms, proto)
 	}
 
-	action := cdpbrowser.GrantPermissions(perms).WithOrigin(origin).WithBrowserContextID(b.id)
+	action := cdpbrowser.GrantPermissions(perms).WithOrigin(parsedOpts.Origin).WithBrowserContextID(b.id)
 	if err := action.Do(cdp.WithExecutor(b.ctx, b.browser.conn)); err != nil {
-		k6ext.Panic(b.ctx, "internal error while granting browser permissions: %w", err)
+		return fmt.Errorf("error granting browser permissions: %w", err)
```
inancgumus, line 250: Should we follow our typical error text also here without saying "error"?

```suggestion
		return fmt.Errorf("granting browser permissions: %w", err)
```

common/browser_context_options.go
```diff
@@ -232,3 +232,29 @@ func (w *WaitForEventOptions) Parse(ctx context.Context, optsOrPredicate goja.Va
 
 	return nil
 }
+
+// GrantPermissionsOptions are the options used by the browserContext.grantPermissions
+// API.
```
inancgumus, line 237: Since this comment appears on pkg.go.dev. Is it better to refer to Go-specific types instead of the JavaScript API? Maybe like: `// GrantPermissionsOptions are used by BrowserContext.GrantPermissions`? Also, there's work going on on the Go side to make these references clickable on the documentation (soon).

common/browser_context_options.go
```diff
@@ -232,3 +232,29 @@ func (w *WaitForEventOptions) Parse(ctx context.Context, optsOrPredicate goja.Va
...
+type GrantPermissionsOptions struct {
+	Origin string
+}
+
+// NewGrantPermissionsOptions creates a new instance of GrantPermissionsOptions.
```
inancgumus, line 242: I don't mind, but just a minor suggestion to have a more idiomatic comment on pkg.go.dev:

```suggestion
// NewGrantPermissionsOptions returns a new GrantPermissionsOptions.
```

common/browser_context_options.go
```diff
@@ -232,3 +232,29 @@ func (w *WaitForEventOptions) Parse(ctx context.Context, optsOrPredicate goja.Va
...
+func NewGrantPermissionsOptions() *GrantPermissionsOptions {
+	return &GrantPermissionsOptions{}
+}
+
+// Parse will parse the options. It can parse origin out of the supplied object.
```
inancgumus, line 247: Also, here since Go doesn't have objects and instances:

```suggestion
// Parse parses the options from opts if opts exists in the Goja runtime.
```

tests/browser_context_test.go
```diff
@@ -866,3 +866,50 @@ func optsOrPredicateToGojaValue(t *testing.T, tb *testBrowser, op *optsOrPredica
...
 }
+
+func TestBrowserContextGrantPermissions(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name       string
+		permission string
+		wantErr    string
+	}{
```
inancgumus, line 877: Perfect tests and naming 👏

inancgumus (approved):
Nice work 👍 It's a good step towards #271 and #1064. Thanks! Some minor suggestions. Feel free to skip.

browser/mapping.go
ankur22: Yeah, makes sense 👍 I've renamed it to `pOpts`, since they will be parsed on the next line and we can't change the name of the variable after it's been parsed 🙂 

Resolved in 74822b1ddc3524ca37fe80cafeef623bcd0b8b75

common/browser_context_options.go
ankur22: Yeah, good point, i'll change that.

Resolved in cf5a95cf9ac1f4fd02374ae29b86b4fa8357fc2b

browser/mapping.go
inancgumus: To be honest, I also say `popts` everytime 😄

ka3de (approved):
LGTM.

inancgumus (approved):

PR #1090, opened 2023-11-01
Update to use os & io instead ioutil

ankur22:
## What?

Replacing `ioutil` with `os` and `io` where needed.

## Why?

ioutil is deprecated, and instead we should work with `os` and `io`.

## Checklist


inancgumus (approved):
Nice!

ka3de (approved):
LGTM. Thanks!

Issue #887, opened 2023-05-17
Support CPU and Network throttling


ka3de:
### Feature Description

CPU and network throttling are useful in order to emulate specific network conditions, for example in order to emulate a mobile connection speed when actually running in a faster connection.

See Lighthouse project documentation on throttling: https://github.com/GoogleChrome/lighthouse/blob/main/docs/throttling.md
See CDP APIs in order to throttle [CPU](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setCPUThrottlingRate) and [network](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-emulateNetworkConditions).

```[tasklist]
### Tasks
- [x] Implement Network throttling -- https://github.com/grafana/xk6-browser/pull/1094
- [x] Implement CPU throttling -- https://github.com/grafana/xk6-browser/pull/1095
- [x] Implement predefined network profiles -- https://github.com/grafana/xk6-browser/pull/1098
- [x] Update k6-docs -- https://github.com/grafana/k6-docs/pull/1415
- [x] Update release notes -- https://github.com/grafana/k6/pull/3453
- [x] Update Type definitions -- https://github.com/grafana/k6-DefinitelyTyped/pull/19
```

mdcruz:
I support this feature, and it would be great if we could prioritise this somehow 👍🏼

One strength of using lab data is the ability to run tests in a controlled environment. Currently, we can do this in k6 browser by setting the screen size or removing certain Chrome arguments, but not the ability to configure network throttling (e.g. from 5g to 4g). I want to be able to change my network connection to see how this impacts my web vital metrics.

mdcruz:
Example on how Playwright can do it: https://github.com/microsoft/playwright/issues/6038#issuecomment-812521882

ankur22:
We should look at existing tools and see how they've implemented the same feature. Playwright doesn't seem to support it as a dedicated API yet, but puppeteer does have dedicated APIs for both [CPU](https://pptr.dev/api/puppeteer.page.emulatecputhrottling) and [network](https://pptr.dev/api/puppeteer.page.emulatenetworkconditions). It's probably worth looking through any discussions for both tools to see whether there's anything we can do to improve it in the first iteration.

inancgumus:
Here's some info why Playwright lacks the feature: https://github.com/microsoft/playwright/issues/8622#issuecomment-1284845091

[...]

PR #1094, opened 2023-11-09
Add throttleNetwork on page

ankur22:
## What?

This adds a new `throttleNetwork` API on `page` which throttles the network from chrome's perspective.

## Why?

A good use case for throttling the network is to see (within a lab environment) how the frontend reacts when the network is slow, e.g. on a slow 3G connection. With new measurements under a constraint network, users will be able to fine tune their website so that the most important information is returned in an acceptable timeframe.

## Checklist

## Related PR(s)/Issue(s)

Updates https://github.com/grafana/xk6-browser/issues/887

### DIFF
--- a/browser/mapping_test.go
+++ b/browser/mapping_test.go
@@ -343,6 +343,7 @@ type pageAPI interface {
 	SetViewportSize(viewportSize goja.Value)
 	Tap(selector string, opts goja.Value)
 	TextContent(selector string, opts goja.Value) string
+	ThrottleNetwork(common.NetworkProfile) error
 	Title() string
 	Type(selector string, text string, opts goja.Value)
 	Uncheck(selector string, opts goja.Value)
--- a/common/network_manager.go
+++ b/common/network_manager.go
@@ -28,6 +28,18 @@ import (
 	"github.com/dop251/goja"
 )
 
+// NetworkProfile is used in ThrottleNetwork.
+type NetworkProfile struct {
+	// Minimum latency from request sent to response headers received (ms).
+	Latency float64
+
+	// Maximal aggregated download throughput (bytes/sec). -1 disables download throttling.
+	DownloadThroughput float64
+
+	// Maximal aggregated upload throughput (bytes/sec). -1 disables upload throttling.
+	UploadThroughput float64
+}
+
 // Credentials holds HTTP authentication credentials.
 type Credentials struct {
 	Username string `js:"username"`
--- a/common/network_manager.go
+++ b/common/network_manager.go
@@ -742,6 +742,22 @@ func (m *NetworkManager) SetOfflineMode(offline bool) {
 	}
 }
 
+// ThrottleNetwork changes the network attributes in chrome to simulate slower
+// networks e.g. a slow 3G connection.
+func (m *NetworkManager) ThrottleNetwork(networkProfile NetworkProfile) error {
+	action := network.EmulateNetworkConditions(
+		m.offline,
+		networkProfile.Latency,
+		networkProfile.DownloadThroughput,
+		networkProfile.UploadThroughput,
+	)
+	if err := action.Do(cdp.WithExecutor(m.ctx, m.session)); err != nil {
+		return fmt.Errorf("throttling network: %w", err)
+	}
+
+	return nil
+}
+
 // SetUserAgent overrides the browser user agent string.
 func (m *NetworkManager) SetUserAgent(userAgent string) {
 	action := emulation.SetUserAgentOverride(userAgent)
--- a/common/network_manager.go
+++ b/common/network_manager.go
@@ -40,6 +40,15 @@ type NetworkProfile struct {
 	UploadThroughput float64
 }
 
+// NewNetworkProfile creates a non-throttled network profile.
+func NewNetworkProfile() NetworkProfile {
+	return NetworkProfile{
+		Latency:            0,
+		DownloadThroughput: -1,
+		UploadThroughput:   -1,
+	}
+}
+
 // Credentials holds HTTP authentication credentials.
 type Credentials struct {
 	Username string `js:"username"`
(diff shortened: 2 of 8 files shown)


common/network_manager.go
```diff
@@ -28,6 +28,18 @@ import (
 	"github.com/dop251/goja"
 )
 
+// NetworkProfile is used in ThrottleNetwork.
+type NetworkProfile struct {
+	// Minimum latency from request sent to response headers received (ms).
+	Latency float64
+
+	// Maximal aggregated download throughput (bytes/sec). -1 disables download throttling.
+	DownloadThroughput float64
```
ankur22, line 37: Any thoughts on what this can be renamed to make it shorter?

common/network_manager.go
```diff
@@ -742,6 +742,22 @@ func (m *NetworkManager) SetOfflineMode(offline bool) {
 	}
 }
 
+// ThrottleNetwork changes the network attributes in chrome to simulate slower
+// networks e.g. a slow 3G connection.
+func (m *NetworkManager) ThrottleNetwork(networkProfile NetworkProfile) error {
+	action := network.EmulateNetworkConditions(
+		m.offline,
```
ka3de, line 749: If I understand this correctly, we are using here the `NetworkManager.offline` attribute, which is set when calling `BrowserContext.SetOfflineMode`. This implies that, if there is a previous call to `SetOfflineMode`, a call to `ThrottleNetwork` won't apply, as the page is being emulated as being "offline".  Is that correct?

I wonder if we should automatically overwrite the `offline` attribute to "false" when calling `ThrottleNetwork`, as the user is expressing an intention on applying this emulation to this page. Instead of requiring our users to call `BrowserContext.SetOffline(false)`, which will actually apply to all `BrowserContext` pages, not just this one.

browser/mapping_test.go
```diff
@@ -343,6 +343,7 @@ type pageAPI interface {
 	SetViewportSize(viewportSize goja.Value)
 	Tap(selector string, opts goja.Value)
 	TextContent(selector string, opts goja.Value) string
+	throttleNetwork(common.NetworkProfile) error
```
ka3de, line 346: There is actually no need for this to be capitalized, but considering that the other methods follow this format, should this one follow that pattern too?

ka3de (commented):
Great work @ankur22 :clap: 
Just a couple of considerations.

common/network_manager.go
ka3de: For me it's OK, I can't find a good shorter alternative, and I don't think it's so long that it must be changed.

inancgumus: I think just `Download` is clear as it’s a field of the network profile.

common/network_manager.go
ankur22: > This implies that, if there is a previous call to SetOfflineMode, a call to ThrottleNetwork won't apply, as the page is being emulated as being "offline". Is that correct?

Yes, that's correct.

> I wonder if we should automatically overwrite the offline attribute to "false" when calling ThrottleNetwork, as the user is expressing an intention on applying this emulation to this page. Instead of requiring our users to call BrowserContext.SetOffline(false), which will actually apply to all BrowserContext pages, not just this one.

That could work. My hesitation towards this idea is that the API is quite explicit at the moment to enable/disable offline mode (as well as the other APIs we support when it comes to enabling/disabling features such as `grantPermissions`). If we allow the offline mode to be automatically disabled on `ThrottleNetwork`, then this breaks the explicit nature of our APIs. I feel that being explicit (at least in the short term) is the best way forward until we see how it is being used.

browser/mapping_test.go
ankur22: Nice spot, i think it's best to be consistent, i've changed it in 20e9ca5cbda25f7844aab89636bbf78c7b0276b2

common/network_manager.go
ka3de: I see your point. Still, I'm not sure this is the better approach, maybe documenting the overwrite when calling these page level methods would make sense? Otherwise I feel we should inform the user somehow that the network throttling call will not apply, if `offline` is `true` when calling `ThrottleNetwork`, but I guess for that the only option is to return an error.

ankur22: > Otherwise I feel we should inform the user somehow that the network throttling call will not apply, if `offline` is `true` when calling `ThrottleNetwork`, but I guess for that the only option is to return an error.

I'm happy with this idea since it makes things more transparent and quickly highlights an issue preventing the `ThrottleNetwork` from working.

ka3de: Ok, to be honest I don't quite like the idea of returning an error from `ThrottleNetwork` call anyway, so let's follow your proposal. Still, I feel this might get some confusion, althought I don't think `setOfflineMode` is used frequently, so let's see what is user's feedback. 
Will we then need to store the current page network profile?

ankur22: > Will we then need to store the current page network profile?

Yes, correct, i'll do that next.

ankur22: Done in 24b9b97678dd30366b2b4b03e56cd1d71a8c4ef8. I'll refactor this so that `SetOfflineMode` will work with `ThrottleNetwork` in another PR. This change now works with the following example:

<details>
  <summary>test script</summary>

```js

export const options = {
  scenarios: {
    throttled: {
      options: {
        browser: {
            type: 'chromium',
        },
      },
      exec: 'throttled',
    },
  },
}

export async function throttled() {

  try {
    context.setOffline(true);

    page.throttleNetwork({
      latency: 1000,
      downloadThroughput: 50,
      uploadThroughput: 50,
    });

    context.setOffline(false); // Throttled network profile is kept to what was set in the line above.

    await page.goto('https://test.k6.io/', { waitUntil: 'networkidle' });
  } finally {
    page.close();
  }
}
```

</details>

common/network_manager.go
ankur22: `Download` and `Upload` were in the original spec i shared in the issue. I also think they should be enough since it makes sense contextually. Changed in 98693e507d9aa66a55d2d5569110b9b0022ba446

ka3de (approved):
LGTM :clap:

inancgumus (approved):
Nice work. Thanks 👏

PR #1095, opened 2023-11-09
Add throttleCPU on page

ankur22:
## What?

This adds a new `throttleCPU` API on `page` which throttles the cpu from chrome's perspective.

## Why?

A good use case for throttling the cpu is to see (within a lab environment) how the frontend reacts when the test is ran on a mobile device. With new measurements under a constraint CPU, users will be able to fine tune their website so that the most important information is returned in an acceptable timeframe.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/887


examples/throttle.js
```diff
@@ -20,10 +20,21 @@ export const options = {
...
+        },
+      },
+      exec: 'cpuThrottled',
+    },
   },
   thresholds: {
     'browser_http_req_duration{scenario:normal}': ['p(99)<500'],
     'browser_http_req_duration{scenario:throttled}': ['p(99)<1500'],
+    'iteration_duration{scenario:normal}': ['p(99)<3000'],
+    'iteration_duration{scenario:cpuThrottled}': ['p(99)<5000'],
```
ankur22, line 37: Is this good enough for now? It's difficult to find the right sweet spot where this threshold will hold true on all platforms. It seems to fail on the Windows CI runs, so i need to fine tune it still.

tests/page_test.go
```diff
@@ -1177,3 +1177,61 @@ func TestPageThrottleNetwork(t *testing.T) {
...
+// average time it takes to run the test. Next it will repeat the steps but
+// first apply CPU throttling. The average duration with CPU throttling
+// enabled should be longer than without it.
+func TestPageThrottleCPU(t *testing.T) {
+	t.Parallel()
+
+	tb := newTestBrowser(t, withFileServer())
+
+	tb.withHandler("/ping", func(w http.ResponseWriter, req *http.Request) {
+		defer func() {
```
inancgumus, line 1191: Is there any reason this is in a defer?

tests/page_test.go
```diff
@@ -1177,3 +1177,61 @@ func TestPageThrottleNetwork(t *testing.T) {
...
+			require.NoError(t, err)
+		}()
+		bb, err := io.ReadAll(req.Body)
+		require.NoError(t, err)
+
+		fmt.Fprint(w, string(bb))
+	})
+
+	page := tb.NewPage(nil)
+	iterations := 5
```
inancgumus, line 1202: const instead of a variable?

tests/page_test.go
```diff
@@ -1177,3 +1177,61 @@ func TestPageThrottleNetwork(t *testing.T) {
...
+		Rate: 50,
+	})
+	require.NoError(t, err)
+
+	withCPUThrottle := performPingTest(t, tb, page, iterations)
+
+	assert.Greater(t, withCPUThrottle, noCPUThrottle)
+}
+
+func performPingTest(t *testing.T, tb *testBrowser, page *common.Page, iterations int) int64 {
```
inancgumus, line 1216: This can be a part of the same test if there is no other reason to keep this as a separate function.

inancgumus (approved):
Nice work! LGTM. Added some weakly held suggestions. Feel free to skip.

browser/mapping_test.go
```diff
@@ -343,6 +343,7 @@ type pageAPI interface {
 	SetViewportSize(viewportSize goja.Value)
 	Tap(selector string, opts goja.Value)
 	TextContent(selector string, opts goja.Value) string
+	throttleCPU(common.CPUProfile) error
```
ka3de, line 346: Should this be capitalized?

tests/page_test.go
ankur22: It was due to the linter, which wanted a check on `err` from `req.Body.Close()`. Happy to consider other options.

tests/page_test.go
ankur22: This gets called twice by the `TestPageThrottleCPU`, the first to time the duration without throttling and the second with throttling. I think it's best left as a function.

browser/mapping_test.go
ankur22: Fixed in 51207e4c98548658bc1baa7d2f32dc5f79230586

ka3de (approved):
LGTM.

inancgumus (approved):

PR #1098, opened 2023-11-10
Add predefined network profiles

ankur22:
## What?

Adding predefined network profiles `No Throttling`, `Slow 3G` and `Fast 3G`.

## Why?

These are the most commonly used during frontend testing when throttling network connections.

These could have been defined in the type definitions repo. To make it simpler for our users to work with these network profiles, it's defined in the browser module repo which will make it available for browser module users.

## Checklist

## Related PR(s)/Issue(s)

Updates https://github.com/grafana/xk6-browser/issues/887

### DIFF
--- a/browser/module.go
+++ b/browser/module.go
@@ -27,8 +27,9 @@ type (
 
 	// JSModule exposes the properties available to the JS script.
 	JSModule struct {
-		Browser *goja.Object
-		Devices map[string]common.Device
+		Browser         *goja.Object
+		Devices         map[string]common.Device
+		NetworkProfiles map[string]common.NetworkProfile `js:"networkProfiles"`
 	}
 
 	// ModuleInstance represents an instance of the JS module.
@@ -69,7 +70,8 @@ func (m *RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
 				browserRegistry:   newBrowserRegistry(vu, m.remoteRegistry, m.PidRegistry),
 				taskQueueRegistry: newTaskQueueRegistry(vu),
 			}),
-			Devices: common.GetDevices(),
+			Devices:         common.GetDevices(),
+			NetworkProfiles: common.GetNetworkProfiles(),
 		},
 	}
 }
--- a/browser/module_test.go
+++ b/browser/module_test.go
@@ -20,4 +20,5 @@ func TestModuleNew(t *testing.T) {
 	require.NotNil(t, m.mod, "Module should be set")
 	require.NotNil(t, m.mod.Browser, "Browser should be set")
 	require.NotNil(t, m.mod.Devices, "Devices should be set")
+	require.NotNil(t, m.mod.NetworkProfiles, "Profiles should be set")
 }
--- a/common/network_manager.go
+++ b/common/network_manager.go
@@ -28,27 +28,6 @@ import (
 	"github.com/dop251/goja"
 )
 
-// NetworkProfile is used in ThrottleNetwork.
-type NetworkProfile struct {
-	// Minimum latency from request sent to response headers received (ms).
-	Latency float64
-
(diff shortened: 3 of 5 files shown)


common/networkProfile.go
```diff
@@ -0,0 +1,30 @@
+package common
+
```
ka3de, line 2: Should the file be named `network_profile.go`?

common/networkProfile.go
```diff
@@ -0,0 +1,30 @@
...
+// NetworkProfile is used in ThrottleNetwork.
+type NetworkProfile struct {
+	// Minimum latency from request sent to response headers received (ms).
+	Latency float64 `js:"latency"`
+
+	// Maximal aggregated download throughput (bytes/sec). -1 disables download throttling.
+	DownloadThroughput float64 `js:"downloadThroughput"`
```
inancgumus, line 9: This is a bit verbose. Should we just say `Download` as the comment makes it clear? And it's in a `NetwrokProfile` already-also clear from the context. No strong opinion.

common/networkProfile.go
```diff
@@ -0,0 +1,30 @@
...
+	// Maximal aggregated upload throughput (bytes/sec). -1 disables upload throttling.
+	UploadThroughput float64 `js:"uploadThroughput"`
+}
+
+// GetNetworkProfiles returns NetworkProfiles which are ready to be used to
+// throttle the network with page.throttleNetwork.
+func GetNetworkProfiles() map[string]NetworkProfile {
+	return map[string]NetworkProfile{
+		"Slow 3G": {
+			DownloadThroughput: ((500 * 1000) / 8) * 0.8,
```
inancgumus, line 20: Should we add a comment to clarify what these magic numbers mean?

inancgumus (approved):
LGTM after renaming the file as @ka3de suggests. Thanks!

common/networkProfile.go
ankur22: Thanks, resolved. See [comment](https://github.com/grafana/xk6-browser/pull/1094#discussion_r1388140886) for details.

common/networkProfile.go
ankur22: Good spot, fixed in 389b0f1328ffc9a8e2b31a5d2a19e2a1a809cc08

common/networkProfile.go
ankur22: Tbh, i'm not sure what these numbers mean. They were copied from [here](https://github.com/microsoft/playwright/issues/6038#issuecomment-812521882) and [here](https://github.com/puppeteer/puppeteer/blob/0a41f8d01e85ff732fdd2e50468bc746d7bc6475/packages/puppeteer-core/src/cdp/PredefinedNetworkConditions.ts#L41-L52). What about:

```js
		"Slow 3G": {
			Download: 50000,
			Upload:   50000,
			Latency:  2000,
		},
		"Fast 3G": {
			Download: 180000
			Upload:   84375,
			Latency:  562.5,
		},
```

inancgumus: LGTM. Here is more info on what the numbers mean: https://stackoverflow.com/questions/48367042/in-chrome-dev-tools-what-is-the-speed-of-each-preset-option-for-network-throttl.

ka3de (dismissed):
LGTM.

common/networkProfile.go
ankur22: Thanks for the link. I quite the explanation, so i'll copy it over into here instead of converting them to the final value as I suggested in the comment [above](https://github.com/grafana/xk6-browser/pull/1098#discussion_r1391225957). Having them broken down into smaller steps might help us to create new predefined profiles later on.

ankur22: Added the notes on what the numbers mean in 9118e7764a6895c5fec126720e071ccf171e5cd1

ka3de (approved):
LGTM.

PR #1100, opened 2023-11-14
Add tracing

ka3de:
## What?

This PR uses the tracing instrumentation provided by k6 through the `VU.State` in order to generate traces that represent the inner workings of k6 browser.

## Why?

Tracing has been chosen as the best suited method in order to build a timeline view of a k6 browser test, which can provide a more visual representation of a browser test along with the metrics associated for each action, such as specific WebVitals measurements per navigation, errors produced, etc.

## Checklist

## Related PR(s)/Issue(s)

Related: https://github.com/grafana/k6/pull/3445

### DIFF
--- a/browser/registry.go
+++ b/browser/registry.go
@@ -186,7 +186,9 @@ type browserRegistry struct {
 
 type browserBuildFunc func(ctx context.Context) (*common.Browser, error)
 
-func newBrowserRegistry(vu k6modules.VU, remote *remoteRegistry, pids *pidRegistry) *browserRegistry {
+func newBrowserRegistry(
+	ctx context.Context, vu k6modules.VU, remote *remoteRegistry, pids *pidRegistry,
+) *browserRegistry {
 	bt := chromium.NewBrowserType(vu)
 	builder := func(ctx context.Context) (*common.Browser, error) {
 		var (
@@ -231,12 +233,14 @@ func newBrowserRegistry(vu k6modules.VU, remote *remoteRegistry, pids *pidRegist
 	}
 
 	go r.handleExitEvent(exitCh, unsubscribe)
-	go r.handleIterEvents(eventsCh, unsubscribe)
+	go r.handleIterEvents(ctx, eventsCh, unsubscribe)
 
 	return r
 }
 
-func (r *browserRegistry) handleIterEvents(eventsCh <-chan *k6event.Event, unsubscribeFn func()) {
+func (r *browserRegistry) handleIterEvents( //nolint:funlen
+	ctx context.Context, eventsCh <-chan *k6event.Event, unsubscribeFn func(),
+) {
 	var (
 		ok   bool
 		data k6event.IterData
@@ -281,7 +285,7 @@ func (r *browserRegistry) handleIterEvents(eventsCh <-chan *k6event.Event, unsub
 
 			// Wrap the tracer into the browser context to make it accessible for the other
 			// components that inherit the context so these can use it to trace their actions.
-			tracerCtx := common.WithTracer(context.Background(), r.tr.tracer)
+			tracerCtx := common.WithTracer(ctx, r.tr.tracer)
 			tracedCtx := r.tr.startIterationTrace(tracerCtx, data)
 
 			b, err := r.buildFn(tracedCtx)
--- a/browser/registry.go
+++ b/browser/registry.go
@@ -13,11 +13,14 @@ import (
 	"sync/atomic"
 
 	"github.com/mstoykov/k6-taskqueue-lib/taskqueue"
+	"go.opentelemetry.io/otel/attribute"
+	oteltrace "go.opentelemetry.io/otel/trace"
 
 	"github.com/grafana/xk6-browser/chromium"
 	"github.com/grafana/xk6-browser/common"
 	"github.com/grafana/xk6-browser/env"
 	"github.com/grafana/xk6-browser/k6ext"
+	browsertrace "github.com/grafana/xk6-browser/trace"
 
 	k6event "go.k6.io/k6/event"
 	k6modules "go.k6.io/k6/js/modules"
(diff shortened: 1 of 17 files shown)


common/trace.go
```diff
@@ -0,0 +1,58 @@
...
+
+// TraceAPICall is a helper method that retrieves the Tracer from the given ctx and
+// calls its TraceAPICall implementation. If the Tracer is not present in the given
+// ctx, it returns a noopSpan and the given context.
+func TraceAPICall(
+	ctx context.Context, targetID string, spanName string, opts ...trace.SpanStartOption,
+) (context.Context, trace.Span) {
+	if tracer := GetTracer(ctx); tracer != nil {
+		return tracer.TraceAPICall(ctx, targetID, spanName, opts...)
+	}
```
ka3de, line 32: I considered panicking here, as the `tracer` should always be present in the context, whether it is a _noop_ or a "valid implementation", but that would require us to mock it in all of our tests. Also, we already have a test to verify that the tracer is correctly created from the `VU.State.TracerProvider` and propagated to all the components.

browser/registry.go
```diff
@@ -279,7 +278,12 @@ func (r *browserRegistry) handleIterEvents(eventsCh <-chan *k6event.Event, unsub
 			// we have to initialize traces registry on the first VU iteration
 			// so we can get access to the k6 TracerProvider.
 			r.initTracesRegistry()
-			tracedCtx := r.tr.startIterationTrace(ctx, data)
+
+			// Wrap the tracer into the browser context to make it accessible for the other
+			// components that inherit the context so these can use it to trace their actions.
+			tracerCtx := common.WithTracer(context.Background(), r.tr.tracer)
```
inancgumus, line 284: I'm wondering if this should use `vuCtx`?

common/frame_session.go
```diff
@@ -749,6 +749,33 @@ func (fs *FrameSession) onFrameNavigated(frame *cdp.Frame, initial bool) {
...
 		)
+
+		var (
+			spanID   = fs.mainFrameSpan.SpanContext().SpanID().String()
+			newFrame = fs.manager.getFrameByID(frame.ID)
+		)
+		// Set k6SpanId property in the page so it can be retrieved when pushing
+		// the Web Vitals events from the page execution context and used to
+		// correlate them with the navigation span to which they belong to.
+		if newFrame != nil {
```
inancgumus, line 766: This `if isMainFrame := frame.ParentID...-> if newFrame -> go func -> if err != nil` is too nested, making it harder to follow. Can we find a way to make it flat?

common/js/web_vital_init.js
```diff
@@ -8,6 +8,7 @@ function print(metric) {
     numEntries: metric.entries.length,
     navigationType: metric.navigationType,
     url: window.location.href,
+    spanID: window.k6SpanId,
```
inancgumus, line 11: I would add this commit's comment as a comment in the code here.

common/element_handle.go
```diff
@@ -1170,13 +1170,20 @@ func (h *ElementHandle) setChecked(apiCtx context.Context, checked bool, p *Posi
 }
 
 func (h *ElementHandle) Screenshot(opts goja.Value) goja.ArrayBuffer {
+	sCtx, span := GetTracer(h.ctx).TraceAPICall(
```
inancgumus: Should we say `spanCtx` here to make it a little bit more explicit?

common/screenshotter.go
```diff
@@ -229,6 +232,8 @@ func (s *screenshotter) screenshot(
 		if err := os.WriteFile(path, buf, 0o644); err != nil {
 			return nil, fmt.Errorf("saving screenshot to %q: %w", path, err)
 		}
+
+		s.span.SetAttributes(attribute.String("path", path))
```
inancgumus, line 236: Should we call this statement in `Page.Screenshot` instead of a `span` field into `screenshotter`? To see the overall flow and keep it simple. We can add this field back if needed to add more granularity later on.

common/trace.go
```diff
@@ -0,0 +1,20 @@
+package common
+
+import (
+	"context"
+
+	"go.opentelemetry.io/otel/trace"
+)
+
+// Tracer defines the interface with the tracing methods used in the common package.
+type Tracer interface {
```
inancgumus, line 12: Nice idea 👍

inancgumus (dismissed):
Colossal work and nicely done 👏 I have some weak suggestions. Thanks!

common/trace.go
ankur22: I wonder if it's worth adding a warning log when we do return a `NoopSpan`? Is there a valid case for tracer to ever be `nil`?

ankur22 (dismissed):
This is great! Excited to see the results, and the next steps.

One question I have is why were only a subset of APIs instrumented?

browser/registry.go
ka3de: We can't use `vuCtx` as its lifecycle is shorter than the iteration. Instead, we use background context for the browser, and then stop it based on the `iterEnd` event.

common/frame_session.go
ka3de: It is a little bit nested, but there are different set of actions depending on the different conditions. In this case I think being nested makes it more explicit and easier to understand, I believe making it flatter would make it more difficult to understand. Can you see a way on how to improve it?

common/frame_session.go
```diff
@@ -749,6 +749,33 @@ func (fs *FrameSession) onFrameNavigated(frame *cdp.Frame, initial bool) {
...
+			// Executing a CDP command in the event parsing goroutine might deadlock in some cases.
+			// For example a deadlock happens if the content loaded in the frame that has navigated
+			// includes a JavaScript initiated dialog which we have to explicitly accept or dismiss
+			// (see onEventJavascriptDialogOpening). In that case our EvaluateGlobal call can't be
+			// executed, as the browser is waiting for us to accept/dismiss the JS dialog, but we
+			// can't act on that because the event parsing goroutine is stuck in onFrameNavigated.
+			// Because in this case the action is to set an attribute to the global object (window)
+			// it should be safe to just execute this in a separate goroutine.
+			go func() {
+				js := fmt.Sprintf("window.k6SpanId = '%s';", spanID)
```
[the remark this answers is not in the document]
ka3de: Initially I was thinking that considering that this is set in the context of the web page, I don't think it's gonna collide with other tracing related functionalities that could be implemented in k6, and also that it's an internal implementation to k6 browser, we can change it at any time, so I would prefer to keep it simple and shorter as of now. But I'm thinking this might be referenced by the Faro integration, so... maybe we could add the `browser` naming there, eve though I don't see having an collisions in the page context. What are your thoughts also @inancgumus ?

common/frame_session.go
inancgumus: I think nesting creates cognitive pressure, but no strong opinions. Sure, there's a similar code [here](https://github.com/grafana/xk6-browser/blob/86329fbb36f04c962073298c25233553a952e4f9/common/network_manager.go#L377-L411).

browser/registry.go
ka3de: That makes sense, see [a355a21](https://github.com/grafana/xk6-browser/pull/1100/commits/a355a21b2cf0e2c9e6c681bb97f394e9ff61dd31).

inancgumus: 🙇

common/frame_session.go
```diff
@@ -727,6 +741,15 @@ func (fs *FrameSession) onFrameNavigated(frame *cdp.Frame, initial bool) {
 		k6ext.Panic(fs.ctx, "handling frameNavigated event to %q: %w",
 			frame.URL+frame.URLFragment, err)
 	}
+
+	// Trace navigation only for the main frame.
+	// TODO: How will this affect sub frames such as iframes?
```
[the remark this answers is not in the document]
ka3de: See https://github.com/grafana/xk6-browser/issues/1119.

common/frame_session.go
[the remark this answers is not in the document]
ka3de: Rethinking this, this is a completely internal mechanism for k6 browser in order to correlate traces, so no integration should take this into account as a reference, instead, the _test run id_ should be used. So I think we can keep this simple and short by now.

common/trace.go
ka3de: Reviewing this, adding a warning log when the tracer is not present in the context should be easy, but we'll have to accept the logger as input for the common package tracing functions. Take into account also that we have an integration test that already verifies that the tracer is present and propagated to all components that currently generate traces.

common/screenshotter.go
ka3de: Changed in [533a9fa](https://github.com/grafana/xk6-browser/pull/1100/commits/533a9fadce6358ca6c70d447cd8b977e4223b7a6).

ka3de:
> One question I have is why were only a subset of APIs instrumented?

I knew I have read this question somewhere but missed to answer it @ankur22.
The methods instrumented are the ones that were implemented in the PoC which was accepted by the working group, so these are the ones that I focused on. Instrumenting more exposed API methods should be straight forward now, so I would say let's merge this implementation first, and then we can evaluate if more methods are required for the next release. What do you think?

ankur22 (dismissed):
LGTM 🚀

inancgumus (dismissed):
🚀

ka3de:
Squashed fixup commits, rebased on top of main to resolve conflict on `go.mod` and fixed a couple of lint error that appeared after applying the fixup commits.

inancgumus (approved):

ankur22 (approved):
LGTM 🚀

Issue #982, opened 2023-07-26
page.$ returns an error when selector doesn't match


ankur22:
### Brief summary

When we use `page.$` giving it a selector that we know will not match on the navigated site, we expect the return value to be `null`, but instead we see an error. We've [documented](https://k6.io/docs/javascript-api/k6-experimental/browser/page/page-$/) this behaviour so that a `null` should be returned when the selector doesn't match.

### xk6-browser version

0fcdd3618f29b1ed8459c3b491e22817e9cc726e

### OS

NA

### Chrome version

NA

### Docker version and image (if applicable)

NA

### Steps to reproduce the problem

Run the following test script:

```js
import { browser } from 'k6/experimental/browser';

export const options = {
  scenarios: {
    browser: {
      executor: 'shared-iterations',
      options: {
        browser: {
            type: 'chromium',
        },
      },
    },
  },
}

export default async function() {
  const page = browser.newPage();
 
  try {
    await page.goto('https://test.k6.io');

    const missing = page.$('//h2[text()="Public page"]')
  } finally {
    page.close();
  }
}
```

### Expected behaviour

The test should complete with no errors.

### Actual behaviour

The test fails with the following error:

```bash
ERRO[0001] Uncaught (in promise) GoError: querying selector "//h2[text()=\"Public page\"]"
running at github.com/grafana/xk6-browser/browser.mapPage.func13 (native)
browser at file:///Users/ankuragarwal/go/src/github.com/grafana/xk6-browser/e2e/asd.js:22:36(17)  executor=shared-iterations scenario=browser
```
```[tasklist]
### Tasks
- [x] Implement fix -- https://github.com/grafana/xk6-browser/pull/1101
- [x] Add to release notes -- https://github.com/grafana/k6/pull/3465
```

PR #1101, opened 2023-11-15
Fix error on page.query when no matches with given selector

ankur22:
## What?

This fixes the issue where `Page.Query` (`page.$`) returned an error when no elements match the given selector.

## Why?

The actual expectation in this scenario, is for the API to return `nil`.

This was tested with the following script:

<details>
  <summary>Click me</summary>

  ```js
  import { browser } from 'k6/experimental/browser';
  
  export const options = {
    scenarios: {
      ui: {
        iterations: 1,
        options: {
          browser: {
            type: 'chromium'
          },
        },
      },
    },
  }
  
    await page.goto("https://test.k6.io");
  
    const result = page.$('input[name="login"]');
    if (!result) {
      console.log("It's null, hooray!");
    }
  
    page.close();
  }
  ```

</details>

## Checklist

## Related PR(s)/Issue(s)

Updates: [#982](https://github.com/grafana/xk6-browser/issues/982)

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -253,6 +253,12 @@ func mapElementHandle(vu moduleVU, eh *common.ElementHandle) mapping {
 		if err != nil {
 			return nil, err //nolint:wrapcheck
 		}
+		// ElementHandle can be null when the selector does not match any elements.
+		// We do not want to map nil elementHandles since the expectation is a
+		// null result in the test script for this case.
+		if eh == nil {
+			return nil, nil //nolint:nilnil
+		}
 		ehm := mapElementHandle(vu, eh)
 		return ehm, nil
 	}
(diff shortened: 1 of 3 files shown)


common/element_handle.go
```diff
@@ -1042,7 +1042,8 @@ func (h *ElementHandle) Query(selector string) (*ElementHandle, error) {
 		return nil, fmt.Errorf("querying selector %q: %w", selector, err)
 	}
 	if result == nil {
-		return nil, fmt.Errorf("querying selector %q", selector)
+		h.logger.Debugf("ElementHandle:Query:result:nil", "sel:%q", selector)
```
ankur22, line 1045: Left a debug log here which might be useful. Happy to remove if you think there's no value in it.

browser/mapping.go
```diff
@@ -253,6 +253,9 @@ func mapElementHandle(vu moduleVU, eh *common.ElementHandle) mapping {
 		if err != nil {
 			return nil, err //nolint:wrapcheck
 		}
+		if eh == nil {
```
ka3de, line 256: nit: We might want to add a comment here (and in the other cases) similar to: _"ElementHandle can be nil if selector does not match any elements"_.

ka3de (approved):
LGTM.

common/element_handle.go
ka3de: Not sure if it will add much value, but no strong opinion against it, if you think it can help us in some cases, let's keep it.

browser/mapping.go
ankur22: Sounds good, done in a2dc2585edf7b3ab11254faafd82b50f6e563af9

common/element_handle.go
ankur22: Removed in 54a09aeba83c7c2c376e617e3df328af059038f3

ka3de (approved):
LGTM.

Issue #981, opened 2023-07-26
elementHandle.isVisible should return straight away even if the selector was not found


ankur22:
### Brief summary

Currently when we work with `elementHandle.isVisible` it will wait and try to match on the given selector before returning `true` or `false` on whether it is visible or not. To match Playwrights behaviour, and the expectations of our users, we should change the behaviour to return straight away even if the selector is not present in the DOM.

```js
  await page.goto(`url-redacted`)

  const alreadyInstalled = await page.locator('//h3[text()="Resource consumption"]').isVisible()

  if (alreadyInstalled) {
    console.log('Dashboards and alert rules already installed on the page')
    return
  }
```

### xk6-browser version

v0.45.0

### OS

NA

### Chrome version

NA

### Docker version and image (if applicable)

NA

### Steps to reproduce the problem

1. Work with `elementHandle.isVisible` and set the selector to something that will not be present on the page that the test navigates to.

### Expected behaviour

It should return `false` straight away.

### Actual behaviour

It waits for the selector to match, but eventually times out.
```[tasklist]
### Tasks
- [x] Use `force` and `noWaitAfter` in `frame.newAction` -- https://github.com/grafana/xk6-browser/pull/1102
- [x] Add 1102 to release notes -- https://github.com/grafana/k6/pull/3465
- [x] Refactor query to allow strict mode -- https://github.com/grafana/xk6-browser/pull/1107
- [x] Implement the fix -- https://github.com/grafana/xk6-browser/pull/1108
- [x] Fix the checks for other APIs that shouldn't wait -- https://github.com/grafana/xk6-browser/pull/1109 & https://github.com/grafana/xk6-browser/pull/1110
- [x] Remove timeout option since the API will not wait for the selector -- https://github.com/grafana/xk6-browser/pull/1111
- [x] Add it to the release notes -- https://github.com/grafana/k6/pull/3506
- [x] Update the type definitions -- https://github.com/grafana/k6-DefinitelyTyped/pull/23
- [x] Update the documentation -- https://github.com/grafana/k6-docs/pull/1457
```

PR #1102, opened 2023-11-15
Use force and noWaitAfter in frame.newAction

ankur22:
## What?

The two input arguments `force` and `noWaitAfter` were not being used by `frame.newAction`, and instead were being overriden to always be `false` for both flags.

## Why?

Some of the APIs that call through to `frame.newAction` (such as `isVisible`) do not need to:

1. wait for a navigation, which is what `noWaitAfter` controls;
2. wait for actionability checks (attached to dom, visible, stable and enabled) and instead should be actioned upon straight away, which is what `force` controls.

## Checklist

## Related PR(s)/Issue(s)

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -2001,7 +2001,7 @@ func (f *Frame) newAction(
 			}
 			return
 		}
-		f := handle.newAction(states, fn, false, false, timeout)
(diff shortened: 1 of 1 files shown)


ka3de (approved):
LGTM.

inancgumus (approved):
👍

PR #1107, opened 2023-11-22
Refactor query

ankur22:
## What?

It refactors and updates the query API for `elementHandle` , `frame` and `page` so that we can enable `strict` mode. The change is internal only and doesn't affect the public API.

## Why?

The change is required so that in the future when we work on an API (such as `locator`) that needs to work with `query` but allow `strict` mode on.

## Checklist

## Related PR(s)/Issue(s)

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -249,7 +249,7 @@ func mapElementHandle(vu moduleVU, eh *common.ElementHandle) mapping {
 		},
 	}
 	maps["$"] = func(selector string) (mapping, error) {
-		eh, err := eh.Query(selector)
+		eh, err := eh.Query(selector, false)
 		if err != nil {
 			return nil, err //nolint:wrapcheck
 		}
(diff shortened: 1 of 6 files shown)


browser/mapping.go
```diff
@@ -249,7 +249,7 @@ func mapElementHandle(vu moduleVU, eh *common.ElementHandle) mapping {
 		},
 	}
 	maps["$"] = func(selector string) (mapping, error) {
-		eh, err := eh.Query(selector)
+		eh, err := eh.Query(selector, false)
```
inancgumus, line 252: WDYT about putting `false` in a package-level constant like `strictModeOff` to be more explicit?

inancgumus (approved):

browser/mapping.go
ankur22: Resolved in d7f3f532c35291924fc0f78f1c9985a7a61d5565

PR #1108, opened 2023-11-22
Fix isVisible

ankur22:
## What?

This fixes `isVisible` so that it does not wait for an element to match with the given selector, and returns straight away. This makes the `timeout` option obsolete.

## Why?

There are two reason to make this change:
1. It doesn't make sense to wait for an element to match before checking whether it is visible;
2. It will match Playwrights behaviour which is what some users will expect.

## Checklist

## Related PR(s)/Issue(s)

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -1306,20 +1306,12 @@ func (f *Frame) isVisible(selector string, opts *FrameIsVisibleOptions) (bool, e
 		}
 		return v, err
 	}
-	act := f.newAction(
-		selector, DOMElementStateAttached, opts.Strict, isVisible, []string{}, false, true, opts.Timeout,
-	)
-	v, err := call(f.ctx, act, opts.Timeout)
+	v, err := f.runActionOnSelector(f.ctx, selector, opts.Strict, isVisible)
 	if err != nil {
-		return false, errorFromDOMError(err)
+		return false, fmt.Errorf("checking is %q visible: %w", selector, err)
 	}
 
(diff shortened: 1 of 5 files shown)


inancgumus:
@ankur22 Does this conform to auto-waiting actionability checks? https://playwright.dev/docs/actionability

common/frame.go
```diff
@@ -1973,6 +1965,31 @@ type frameExecutionContext interface {
...
+		return false, fmt.Errorf("query: %w", err)
+	}
+	if handle == nil {
+		f.log.Debugf("Frame:runActionOnSelector:nilHandler", "fid:%s furl:%q selector:%s", f.ID(), f.URL(), selector)
+		return false, nil
+	}
+
+	v, err := fn(ctx, handle)
+	if err != nil {
+		return false, fmt.Errorf("function called: %w", err)
```
inancgumus, line 1982: `calling function` instead of `function called` to be consistent with the rest of our errors?

inancgumus (approved):
👍 Some suggestions.

ankur22:
> @ankur22 Does this conform to auto-waiting actionability checks? https://playwright.dev/docs/actionability

No, this isn't a prerequisite for the `isVisiblity` check. In the link you posted, visible is an actionability check for other APIs.

common/frame.go
ankur22: Resolved in 9e1d3ed837aa954889a44780bd302c431e4b7b11

PR #1110, opened 2023-11-22
Fix isHidden

ankur22:
## What?

This fixes `isHidden` so that it does not wait for an element to match with the given selector, and returns straight away. This makes the `timeout` option obsolete.

## Why?

There are two reason to make this change:
1. It doesn't make sense to wait for an element to match before checking whether it is hidden;
2. It will match Playwrights behaviour which is what some users will expect.

## Checklist

## Related PR(s)/Issue(s)

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -1251,7 +1251,7 @@ func (f *Frame) IsHidden(selector string, opts goja.Value) (bool, error) {
 	}
 	hidden, err := f.isHidden(selector, popts)
 	if err != nil {
-		return false, fmt.Errorf("checking is %q hidden: %w", selector, err)
+		return false, err
 	}
 
 	return hidden, nil
--- a/common/frame.go
+++ b/common/frame.go
@@ -1306,7 +1306,7 @@ func (f *Frame) isVisible(selector string, opts *FrameIsVisibleOptions) (bool, e
 		}
 		return v, err
 	}
-	v, err := f.runActionOnSelector(f.ctx, selector, opts.Strict, isVisible)
+	v, err := f.runActionOnSelector(f.ctx, selector, opts.Strict, isVisible, func() bool { return false })
 	if err != nil {
 		return false, fmt.Errorf("checking is %q visible: %w", selector, err)
 	}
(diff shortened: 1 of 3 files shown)


common/frame.go
```diff
@@ -1265,20 +1265,12 @@ func (f *Frame) isHidden(selector string, opts *FrameIsHiddenOptions) (bool, err
...
-	v, err := call(f.ctx, act, opts.Timeout)
+	v, err := f.runActionOnSelector(f.ctx, selector, opts.Strict, isHidden, func() bool { return true })
 	if err != nil {
-		return false, errorFromDOMError(err)
-	}
-
-	bv, ok := v.(bool)
-	if !ok {
-		return false, fmt.Errorf("checking is %q hidden: unexpected type %T", selector, v)
+		return false, fmt.Errorf("checking is %q visible: %w", selector, err)
```
inancgumus, line 1270: Shouldn't this be `hidden` instead of `visible`?

common/frame.go
```diff
@@ -1966,15 +1958,15 @@ type frameExecutionContext interface {
 }
 
 func (f *Frame) runActionOnSelector(
-	ctx context.Context, selector string, strict bool, fn elementHandleActionFunc,
+	ctx context.Context, selector string, strict bool, fn elementHandleActionFunc, nfn func() bool,
```
inancgumus, line 1961: What does `nfn` mean here? Maybe we can make the parameter name more explicit.

tests/page_test.go
```diff
@@ -1302,3 +1302,70 @@ func TestPageIsVisible(t *testing.T) {
 		})
 	}
 }
+
+func TestPageIsHidden(t *testing.T) {
```
inancgumus, line 1306: This test is very similar to `TestPageIsVisible`. What about merging both? They both are about visibility.

inancgumus (approved):

tests/page_test.go
ankur22: Yeah, true, they are very similar, but they're testing two different APIs, so i think it's best to keep them separate.

common/frame.go
ankur22: Resolved in 869847bcd1e43e4516eee6e836fa3249e54f4e11

common/frame.go
ankur22: Resolved in 4eb45e1d6a23b22079645ec2f065ad3fee2a8b80

PR #1121, opened 2023-12-05
Fix dblClick

ankur22:
## What?

Fixes the behaviour of the `dblClick` API calls from `locator`, `page`, `elementHandle` and `frame` (they all share the same logic which is implemented in `mouse`).

## Why?

When a `dblClick` API call is made, it needs to perform two things:
1. Ensure a `onDblClick` event is initiated and any listeners receive the event. This was already part of the implementation.
2. Ensure that double the mouse actions are performed so that an element is clicked on twice, e.g. a counter increments from 0 to 2. This was missing.

## Checklist

## Related PR(s)/Issue(s)

Playwright reference: https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/input.ts#L199-L219

Updates: https://github.com/grafana/xk6-browser/issues/469

### DIFF
--- a/common/mouse.go
+++ b/common/mouse.go
@@ -59,30 +59,6 @@ func (m *Mouse) click(x float64, y float64, opts *MouseClickOptions) error {
 	return nil
 }
 
-func (m *Mouse) dblClick(x float64, y float64, opts *MouseDblClickOptions) error {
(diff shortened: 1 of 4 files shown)


inancgumus (approved):
Nice thinking of removing the double click and converting it into a click count option.

Issue #987, opened 2023-07-31
Panic: Unexpected end of JSON input


inancgumus:
### Brief summary

This problem was handled before in #793. But it seems the problem persists in some cases. Test run ID: 155041.

### xk6-browser version

0fcdd36

### OS

Linux

### Steps to reproduce the problem

Run the same script in #986 several times on a restricted platform.

### Expected behaviour

Not panics.

### Actual behaviour

Panics.

```go
panic: GoError: parsing remote object value: unexpected end of JSON input

goroutine 214897 [running]:
go.k6.io/k6/js/common.Throw(0x174df7b?, {0x19f6860?, 0xc0006b1740?})
	go.k6.io/k6@v0.45.1-0.20230725090219-9fb9b886ed4f/js/common/util.go:20 +0x4d
github.com/grafana/xk6-browser/k6ext.Panic.func1(0x1504b00?, {0xc0007dcb90?, 0x1594501?, 0xc00127b000?})
	github.com/grafana/xk6-browser@v1.0.0-rc0.0.20230726091449-0fcdd3618f29/k6ext/panic.go:35 +0x4a
github.com/grafana/xk6-browser/k6ext.sharedPanic({0x1a08080, 0xc01f6ec5f0}, 0xc0035877b8, {0xc0007dcb90?, 0x1, 0x1})
	github.com/grafana/xk6-browser@v1.0.0-rc0.0.20230726091449-0fcdd3618f29/k6ext/panic.go:64 +0x265
github.com/grafana/xk6-browser/k6ext.Panic({0x1a08080?, 0xc01f6ec5f0?}, {0x174df7b?, 0xc0000131d8?}, {0xc0007dcb90?, 0x18?, 0x18?})
	github.com/grafana/xk6-browser@v1.0.0-rc0.0.20230726091449-0fcdd3618f29/k6ext/panic.go:37 +0x65
github.com/grafana/xk6-browser/common.handleParseRemoteObjectErr({0x1a08080, 0xc01f6ec5f0}, {0x19f6760?, 0xc0220f91a0}, 0x2735d00?)
	github.com/grafana/xk6-browser@v1.0.0-rc0.0.20230726091449-0fcdd3618f29/common/remote_object.go:200 +0x132
github.com/grafana/xk6-browser/common.(*FrameSession).onConsoleAPICalled(0xc0192fae70, 0xc011710000)
[...]

PR #1129, opened 2023-12-11
Mitigate risk of panic on parse of remote object during console.log

ankur22:
## What?

This change reduces the risk of a panic occurring when the website under test uses `console.log` to print logs.

## Why?

Currently there is an unknown case where a panic can occur when an object from console.log is unmarshaled and fails with an invalid JSON error. So this isn't fixing the issue but mitigating the risk. This is done by not needing to parse the console.log text at all. The reason behind this choice is that when a user prints something to the console, they don't tend to need the object to be parsed.

Other areas of the module still rely on the existing parse logic which does still attempt to unmarshal the object which is returned when working with APIs such as `JSONValue` and `Eval`.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/987

### DIFF
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -616,16 +616,13 @@ func (fs *FrameSession) onConsoleAPICalled(event *cdpruntime.EventConsoleAPICall
 		}
 		*/
 
-	parsedObjects := make([]any, 0, len(event.Args))
+	parsedObjects := make([]string, 0, len(event.Args))
 	for _, robj := range event.Args {
-		i, err := parseRemoteObject(robj)
-		if err != nil {
-			handleParseRemoteObjectErr(fs.ctx, err, l)
-		}
+		i := parseConsoleRemoteObject(fs.logger, robj)
 		parsedObjects = append(parsedObjects, i)
 	}
 
-	l = l.WithField("objects", parsedObjects)
+	l = l.WithField("stringObjects", parsedObjects)
 
 	switch event.Type {
 	case "log", "info":
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -619,8 +619,8 @@ func (fs *FrameSession) onConsoleAPICalled(event *cdpruntime.EventConsoleAPICall
 
 	parsedObjects := make([]string, 0, len(event.Args))
 	for _, robj := range event.Args {
-		i := parseConsoleRemoteObject(fs.logger, robj)
-		parsedObjects = append(parsedObjects, i)
+		s := parseConsoleRemoteObject(fs.logger, robj)
+		parsedObjects = append(parsedObjects, s)
 	}
 
 	l = l.WithField("stringObjects", parsedObjects)
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -10,10 +10,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/grafana/xk6-browser/k6ext"
-	"github.com/grafana/xk6-browser/log"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
+
+	"github.com/grafana/xk6-browser/k6ext"
+	"github.com/grafana/xk6-browser/log"
 
 	k6modules "go.k6.io/k6/js/modules"
 	k6metrics "go.k6.io/k6/metrics"
--- a/common/page.go
+++ b/common/page.go
@@ -1374,9 +1374,9 @@ func (p *Page) consoleMsgFromConsoleEvent(e *cdpruntime.EventConsoleAPICalled) (
 	)
 
 	for _, robj := range e.Args {
-		i := parseConsoleRemoteObject(p.logger, robj)
+		s := parseConsoleRemoteObject(p.logger, robj)
 
-		objects = append(objects, i)
+		objects = append(objects, s)
 		objectHandles = append(objectHandles, NewJSHandle(
 			p.ctx, p.session, execCtx, execCtx.Frame(), robj, p.logger,
 		))
--- a/common/page.go
+++ b/common/page.go
@@ -1369,19 +1369,12 @@ func (p *Page) consoleMsgFromConsoleEvent(e *cdpruntime.EventConsoleAPICalled) (
 	}
 
 	var (
-		l = p.logger.WithTime(e.Timestamp.Time()).
-			WithField("source", "browser").
-			WithField("browser_source", "console-api")
-
-		objects       = make([]any, 0, len(e.Args))
+		objects       = make([]string, 0, len(e.Args))
 		objectHandles = make([]JSHandleAPI, 0, len(e.Args))
 	)
 
 	for _, robj := range e.Args {
-		i, err := parseRemoteObject(robj)
-		if err != nil {
-			handleParseRemoteObjectErr(p.ctx, err, l)
-		}
+		i := parseConsoleRemoteObject(p.logger, robj)
 
 		objects = append(objects, i)
 		objectHandles = append(objectHandles, NewJSHandle(
@@ -1424,7 +1417,7 @@ func (p *Page) sessionID() (sid target.SessionID) {
 
 // textForConsoleEvent generates the text representation for a consoleAPICalled event
 // mimicking Playwright's behavior.
-func textForConsoleEvent(e *cdpruntime.EventConsoleAPICalled, args []any) string {
+func textForConsoleEvent(e *cdpruntime.EventConsoleAPICalled, args []string) string {
 	if e.Type.String() == "dir" || e.Type.String() == "dirxml" ||
 		e.Type.String() == "table" {
 		if len(e.Args) > 0 {
@@ -1434,17 +1427,5 @@ func textForConsoleEvent(e *cdpruntime.EventConsoleAPICalled, args []any) string
 		return ""
 	}
 
-	// args is a mix of string and non strings, so using fmt.Sprint(args...)
-	// might not add spaces between all elements, therefore use a strings.Builder
-	// and handle format and concatenation
-	var b strings.Builder
-	for i, a := range args {
-		format := " %v"
-		if i == 0 {
-			format = "%v"
-		}
-		b.WriteString(fmt.Sprintf(format, a))
-	}
-
-	return b.String()
+	return strings.Join(args, " ")
 }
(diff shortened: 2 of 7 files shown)


tests/console_log_test.go
```diff
@@ -0,0 +1,115 @@
...
+			name: "number", log: "1", want: "1",
+		},
+		{
+			name: "string", log: `"some string"`, want: "some string",
+		},
+		{
+			name: "bool", log: "true", want: "true",
+		},
+		{
+			name: "empty_array", log: "[]", want: "{}", // TODO: Improve this output
```
ankur22, line 32: I will create a new PR for this soon.

tests/console_log_test.go
```diff
@@ -0,0 +1,115 @@
...
+			name: "empty", log: "", want: "",
+		},
+		{
+			name: "null", log: "null", want: "null",
+		},
+		{
+			name: "undefined", log: "undefined", want: "undefined",
+		},
+		{
+			name: "bigint", log: `BigInt("2")`, want: "2n",
```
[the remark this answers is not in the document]
ankur22: Fixed in https://github.com/grafana/xk6-browser/pull/1133

tests/console_log_test.go
ankur22: Done in https://github.com/grafana/xk6-browser/pull/1135

common/frame_session.go
```diff
@@ -616,16 +617,13 @@ func (fs *FrameSession) onConsoleAPICalled(event *cdpruntime.EventConsoleAPICall
...
 
-	parsedObjects := make([]any, 0, len(event.Args))
+	parsedObjects := make([]string, 0, len(event.Args))
 	for _, robj := range event.Args {
-		i, err := parseRemoteObject(robj)
-		if err != nil {
-			handleParseRemoteObjectErr(fs.ctx, err, l)
-		}
+		i := parseConsoleRemoteObject(fs.logger, robj)
 		parsedObjects = append(parsedObjects, i)
```
inancgumus, line 623: IDK why it's been done so with `i` before (probably to say it's an interface)—it feels like we're getting an index. I believe it's better to name this `s` because the returned value is a `string`.

```suggestion
		s := parseConsoleRemoteObject(fs.logger, robj)
		parsedObjects = append(parsedObjects, s)		
```

common/page.go
```diff
@@ -1369,19 +1369,12 @@ func (p *Page) consoleMsgFromConsoleEvent(e *cdpruntime.EventConsoleAPICalled) (
...
 	)
 
 	for _, robj := range e.Args {
-		i, err := parseRemoteObject(robj)
-		if err != nil {
-			handleParseRemoteObjectErr(p.ctx, err, l)
-		}
+		i := parseConsoleRemoteObject(p.logger, robj)
 
 		objects = append(objects, i)
```
inancgumus, line 1379: Similar here.

```suggestion
		s := parseConsoleRemoteObject(p.logger, robj)

		objects = append(objects, s)
```

common/remote_object.go
```diff
@@ -162,6 +161,8 @@ func parseExceptionDetails(exc *cdpruntime.ExceptionDetails) string {
 	return errMsg
 }
 
+// parseRemoteObject is to be used by callers that require the string value
+// to be parsed to a go type.
```
inancgumus, line 165:
```suggestion
// to be parsed to a Go type.
```

common/remote_object.go
```diff
@@ -189,25 +193,71 @@ func valueFromRemoteObject(ctx context.Context, robj *cdpruntime.RemoteObject) (
...
-			logger.Warn(ooe)
-		case errors.As(e, &ope):
-			logger.WithError(ope).Error()
-		default:
-			// If this panics it's a bug :)
-			k6ext.Panic(ctx, "parsing remote object value: %w", e)
+func parseConsoleRemoteObjectPreview(logger *log.Logger, op *cdpruntime.ObjectPreview) string {
+	obj := make(map[string]string)
+	if op.Overflow {
+		logger.Warnf("parseConsoleRemoteObjectPreview", "object is too large and will be parsed partially")
```
inancgumus, line 199: Shouldn't this return an error and let the callers decide whether to log or what to do with that error?

common/remote_object.go
```diff
@@ -189,25 +193,71 @@ func valueFromRemoteObject(ctx context.Context, robj *cdpruntime.RemoteObject) (
...
+	}
+
+	for _, p := range op.Properties {
+		val := parseConsoleRemoteObjectValue(logger, p.Type, p.Subtype, p.Value, p.ValuePreview)
+		obj[p.Name] = val
+	}
+
+	bb, err := json.Marshal(obj)
+	if err != nil {
+		logger.Errorf("parseConsoleRemoteObjectPreview", "failed to marshal object to string: %v", err)
```
inancgumus, line 209: Same here, return an error?

tests/remote_obj_test.go
```diff
@@ -0,0 +1,115 @@
...
+				testTO   = 2500 * time.Millisecond
+			)
+
+			select {
+			case <-done:
+			case <-time.After(testTO):
+				assertTO = true
+			}
+
+			assert.False(t, assertTO, "test timed out before event handlers were called")
```
inancgumus, line 112: No strong opinion, but to make the selected code here slightly simpler, wdyt about moving these variables and `assert.FailNow` into the `select`?

common/remote_object.go
```diff
@@ -189,25 +193,71 @@ func valueFromRemoteObject(ctx context.Context, robj *cdpruntime.RemoteObject) (
...
+	t cdpruntime.Type,
+	st cdpruntime.Subtype,
+	val string,
+	op *cdpruntime.ObjectPreview,
+) string {
+	switch t {
+	case cdpruntime.TypeNumber:
+	case cdpruntime.TypeBoolean:
+	case cdpruntime.TypeSymbol:
+	case cdpruntime.TypeBigint:
```
inancgumus, line 227: Do we need to handle these?

```suggestion
```

common/frame_session.go
```diff
@@ -616,16 +617,13 @@ func (fs *FrameSession) onConsoleAPICalled(event *cdpruntime.EventConsoleAPICall
 		}
 		*/
 
-	parsedObjects := make([]any, 0, len(event.Args))
+	parsedObjects := make([]string, 0, len(event.Args))
 	for _, robj := range event.Args {
-		i, err := parseRemoteObject(robj)
-		if err != nil {
-			handleParseRemoteObjectErr(fs.ctx, err, l)
```
inancgumus, line 623: I'm wondering about the reason for removing error handling.

inancgumus (approved):
👍 Nice work. Just some suggestions.

common/remote_object.go
ankur22: I added this fro two reasons:
1. It makes it clear that all the types have been considered.
2. The linter was having an issue with not checking all cases.

Happy to remove them and add a nolint directive. WDYT?

common/remote_object.go
ankur22: This is a good candidate for a warning over an error that is returned back to the caller. The warning notifies the user that the module was unable to print the whole output from `console.log`, but we don't want to end the process here since the module will still parse and print some of the object in the lines below.

common/remote_object.go
ankur22: I could return an error here. I feel that printing an error is good enough though. The new functions are solely for parsing remote objects from `console.log`, and if something fails here i think it's better to try to log something in stdout rather than give up completely.

tests/remote_obj_test.go
ankur22: Resolved in ac1862188b622f4eb2014dbed8fff594fd3bde06

common/remote_object.go
inancgumus: Sorry for my confusing suggestion. I wasn't suggesting to return an error to the user. I'm with you there 👍 I mean to return an error from this function to the caller function chains, and then they can decide whether to log it as a warning. This can make these methods more testable.

common/remote_object.go
inancgumus: Then, it makes sense 👍 Maybe we can make it a single-liner?

common/frame_session.go
ankur22: I don't think it needs it for this specific use case since there's nothing that can be done if we return an error. The best course of action is to warn or error on the two cases that you've already highlighted, and then carry on with the processing of the remote object. If we return an error while processing the remote object it will abruptly stop the parsing, but then there's no way to retry it or work with the error to try and solve the parsing issue.

common/remote_object.go
ankur22: If we returned an error at this point, then we wouldn't complete the parsing of the object. Thinking about it a bit more now, I think a warn is probably not quite right here, it should be an info. It's telling the user that the module didn't receive the whole object, just part of it, and so only the part that it received will be displayed.

inancgumus: Hmm, I see your point. That was why we had included multi-errors here. Yes, it could be info, too.

ankur22: Update log from warn to info in a13e10dbbc9da518059c3d3f6c150ec8b1d0a85e

common/remote_object.go
ka3de: Maybe add them to the end of the switch also, and with a comment? For me it makes it more intuitive to understand that we are doing nothing for these cases.

log/logger.go
```diff
@@ -195,5 +195,9 @@ func (f *consoleLogFormatter) Format(entry *logrus.Entry) ([]byte, error) {
 		entry.Message = strings.Join(msg, " ")
 		delete(entry.Data, "objects")
 	}
+	if stringObjects, ok := entry.Data["stringObjects"].([]string); ok {
```
ka3de, line 198: If `stringObjects` is set, we overwrite the `entry.Message`. Should `stringObjects` be exclusive from `objects`?
In any case, after the changes in this PR, `objects` field is unused in the codebase. Should we maintain its particular formatter implementation?

common/remote_object.go
ankur22: Resolved in 0139348b022c3d9b8bfaa9b32845fb32c1242c1b

log/logger.go
ankur22: Good spot! I've removed the original `objects` formatter from the logger in 2e9e38701f46676fb5593e9d963c40f7e117766f

ka3de (approved):
LGTM.

ka3de (approved):
LGTM.

inancgumus (approved):

PR #1135, opened 2023-12-12
Fix array console output

ankur22:
## What?

This changes the way the output of an array with `console.out` is parsed which affects the display from:

```
{"0":"foo","1":"bar"}
```

to:

```
["foo","bar"]
```

## Why?

To better represent the array type when it is printed to the console.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/987

### DIFF
--- a/common/remote_object.go
+++ b/common/remote_object.go
@@ -248,6 +248,25 @@ func parseConsoleRemoteObjectPreview(logger *log.Logger, op *cdpruntime.ObjectPr
 	return string(bb)
 }
 
+func parseConsoleRemoteArrayPreview(logger *log.Logger, op *cdpruntime.ObjectPreview) string {
+	arr := make([]any, 0)
+	if op.Overflow {
+		logger.Warnf("parseConsoleRemoteArrayPreview", "array is too large and will be parsed partially")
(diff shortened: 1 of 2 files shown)


inancgumus (approved):
👍 Nice

common/remote_object.go
```diff
@@ -246,6 +246,25 @@ func parseConsoleRemoteObjectPreview(logger *log.Logger, op *cdpruntime.ObjectPr
 	return string(bb)
 }
 
+func parseConsoleRemoteArrayPreview(logger *log.Logger, op *cdpruntime.ObjectPreview) string {
+	arr := make([]any, 0)
```
ka3de, line 250: Why specify len `0`? Shouldn't this be:
```suggestion
	arr := make([]any, 0, len(op.Properties))
```
To make the underlying array big enough for the properties we know in advance and avoid extra allocations?

ankur22: Resolved in 11be8b388289f450f9f52a2e065b5a59a993a496

ka3de (approved):
LGTM.

Issue #1136, opened 2023-12-13
Provide a better error message when Chromium isn't installed


inancgumus:
### What?

Provide a new error message telling users to install Chromium to run browser tests on their system.

### Why?

When users don't provide an executable path, or we can't find one, we say: `"exec: no command"`, which can confuse most users (including me!). For instance, the following error says: "no command at browserFunction", which is hard to guess what's happening.

```
error building browser on IterStart: launching browser: exec: no command at browserFunction (file:///github/workspace/test.js:28:7(0))
```

Instead, we can say:

```
error building browser on IterStart: launching browser: chromium is not installed on this system at browserFunction (file:///github/workspace/test.js:28:7(0))
```

### Tasks

```[tasklist]
### Tasks
- [ ] https://github.com/grafana/xk6-browser/pull/1137
- [ ] https://github.com/grafana/xk6-browser/pull/1145
- [ ] https://github.com/grafana/k6/pull/3516
```

### Related PR(s)/Issue(s)

https://community.grafana.com/t/k6-browser-github-action-error-building-browser-no-command/109937

PR #1137, opened 2023-12-13
Provide a better error message when Chromium isn't found

inancgumus:
## What?

1. Provides a new error message telling users they must install Chromium on their system to run browser tests.
2. Moves `BrowserContext.ExecutablePath` to `ExecutablePath`.

## Why?

1. The current error message is confusing and just says: _"exec: no command"_.
2. To make `ExecutablePath` testable to prevent future regressions.

## Checklist

## Related PR(s)/Issue(s)

Updates #1136

### DIFF
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -183,7 +183,12 @@ func (b *BrowserType) launch(
 	}
 	flags["user-data-dir"] = dataDir.Dir
 
-	browserProc, err := b.allocate(ctx, opts, flags, dataDir, logger)
+	path, err := executablePath(opts.ExecutablePath, b.envLookupper, exec.LookPath)
+	if err != nil {
+		return nil, 0, fmt.Errorf("finding browser executable: %w", err)
+	}
+
+	browserProc, err := b.allocate(ctx, path, flags, dataDir, logger)
 	if browserProc == nil {
 		return nil, 0, fmt.Errorf("launching browser: %w", err)
 	}
@@ -223,7 +228,7 @@ func (b *BrowserType) Name() string {
 
 // allocate starts a new Chromium browser process and returns it.
 func (b *BrowserType) allocate(
-	ctx context.Context, opts *common.BrowserOptions,
+	ctx context.Context, path string,
 	flags map[string]any, dataDir *storage.Dir,
 	logger *log.Logger,
 ) (_ *common.BrowserProcess, rerr error) {
@@ -237,11 +242,6 @@ func (b *BrowserType) allocate(
 	args, err := parseArgs(flags)
 	if err != nil {
 		return nil, err
-	}
-
-	path, err := executablePath(opts.ExecutablePath, b.envLookupper, exec.LookPath)
-	if err != nil {
-		return nil, fmt.Errorf("finding browser executable: %w", err)
 	}
 
 	return common.NewLocalBrowserProcess(bProcCtx, path, args, dataDir, bProcCtxCancel, logger) //nolint: wrapcheck
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -247,8 +247,13 @@ func (b *BrowserType) allocate(
 	return common.NewLocalBrowserProcess(bProcCtx, path, args, dataDir, bProcCtxCancel, logger) //nolint: wrapcheck
 }
 
-// ErrChromeNotInstalled is returned when the Chrome executable is not found.
-var ErrChromeNotInstalled = errors.New("neither chrome nor chromium is installed on this system")
+var (
+	// ErrChromeNotInstalled is returned when the Chrome executable is not found.
+	ErrChromeNotInstalled = errors.New("neither chrome nor chromium is installed on this system")
+
+	// ErrChromeNotFoundAtPath is returned when the Chrome executable is not found at the given path.
+	ErrChromeNotFoundAtPath = errors.New("neither chrome nor chromium found on the path")
+)
 
 // executablePath returns the path where the extension expects to find the browser executable.
 func executablePath(
@@ -261,7 +266,7 @@ func executablePath(
 		if _, err := lookPath(path); err == nil {
 			return path, nil
 		}
-		return "", ErrChromeNotInstalled
+		return "", fmt.Errorf("%w: %s", ErrChromeNotFoundAtPath, path)
 	}
 
 	// find the browser executable in the default paths below
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -239,7 +239,7 @@ func (b *BrowserType) allocate(
 		return nil, err
 	}
 
-	path, ok := ExecutablePath(opts.ExecutablePath, b.envLookupper, exec.LookPath)
+	path, ok := executablePath(opts.ExecutablePath, b.envLookupper, exec.LookPath)
 	if !ok {
 		return nil, ErrChromeNotInstalled
 	}
@@ -250,8 +250,8 @@ func (b *BrowserType) allocate(
 // ErrChromeNotInstalled is returned when the Chrome executable is not found.
 var ErrChromeNotInstalled = errors.New("chromium is not installed on this system")
 
-// ExecutablePath returns the path where the extension expects to find the browser executable.
-func ExecutablePath(
+// executablePath returns the path where the extension expects to find the browser executable.
+func executablePath(
 	path string,
 	env env.LookupFunc,
 	lookPath func(file string) (string, error), // os.LookPath
--- a/chromium/browser_type.go
+++ b/chromium/browser_type.go
@@ -239,9 +239,9 @@ func (b *BrowserType) allocate(
 		return nil, err
 	}
 
-	path, ok := executablePath(opts.ExecutablePath, b.envLookupper, exec.LookPath)
-	if !ok {
-		return nil, ErrChromeNotInstalled
+	path, err := executablePath(opts.ExecutablePath, b.envLookupper, exec.LookPath)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


chromium/browser_type.go
```diff
@@ -240,22 +239,26 @@ func (b *BrowserType) allocate(
...
+	path, ok := ExecutablePath(opts.ExecutablePath, b.envLookupper, exec.LookPath)
+	if !ok {
+		return nil, ErrChromeNotInstalled
 	}
 
 	return common.NewLocalBrowserProcess(bProcCtx, path, args, dataDir, bProcCtxCancel, logger) //nolint: wrapcheck
 }
 
+// ErrChromeNotInstalled is returned when the Chrome executable is not found.
+var ErrChromeNotInstalled = errors.New("chromium is not installed on this system")
```
inancgumus, line 251: I've exported this in case we'll want to test against it later on.

chromium/browser_type.go
```diff
@@ -34,7 +34,6 @@ type BrowserType struct {
 	vu           k6modules.VU
 	hooks        *common.Hooks
 	k6Metrics    *k6ext.CustomMetrics
-	execPath     string // path to the Chromium executable
```
inancgumus, line 37: This wasn't used other than setting it.

chromium/browser_type.go
```diff
@@ -240,22 +239,26 @@ func (b *BrowserType) allocate(
...
+var ErrChromeNotInstalled = errors.New("chromium is not installed on this system")
+
 // ExecutablePath returns the path where the extension expects to find the browser executable.
-func (b *BrowserType) ExecutablePath() (execPath string) {
-	if b.execPath != "" {
-		return b.execPath
+func ExecutablePath(
+	path string,
+	env env.LookupFunc,
+	lookPath func(file string) (string, error), // os.LookPath
```
inancgumus, line 257: This is a little bit mouthful to type, but it's a good first step making stuff testable.

chromium/browser_type_test.go
```diff
@@ -150,3 +153,87 @@ func TestBrowserTypePrepareFlags(t *testing.T) {
 		})
 	}
 }
+
+func TestExecutablePath(t *testing.T) {
```
inancgumus, line 157: This test doesn't test the `ErrChromeNotInstalled`, but at least, it verifies the `ExecutablePath`'s behavior.

chromium/browser_type.go
ankur22: Maybe it needs to be `chrome/chromium` otherwise people might wonder whether chrome is supported or not. I think i see questions about that in the community sometimes.

chromium/browser_type.go
```diff
@@ -240,22 +239,26 @@ func (b *BrowserType) allocate(
...
 }
 
+// ErrChromeNotInstalled is returned when the Chrome executable is not found.
+var ErrChromeNotInstalled = errors.New("chromium is not installed on this system")
+
 // ExecutablePath returns the path where the extension expects to find the browser executable.
-func (b *BrowserType) ExecutablePath() (execPath string) {
-	if b.execPath != "" {
-		return b.execPath
+func ExecutablePath(
```
ankur22, line 254: Why is this exported?

chromium/browser_type.go
ankur22: I think it's ok here for now.

chromium/browser_type_test.go
```diff
@@ -150,3 +153,86 @@ func TestBrowserTypePrepareFlags(t *testing.T) {
...
+
+	// we pick a random file name to look for in our tests
+	// this doesn't matter as long as it's in the paths we look for
+	// in ExecutablePath function.
+	const chromiumExecutable = "google-chrome"
+
+	tests := map[string]struct {
+		path        string                            // user provided path
+		lookPath    func(file string) (string, error) // determines if a file exists
+		userProfile env.LookupFunc
```
ankur22, line 168: I'm not sure `userProfile` is a good name for this, why not `lookup` or something similar?

chromium/browser_type_test.go
ankur22: Why doesn't it test for that?

ankur22 (changes requested):
Nice bit of refactoring and the error message definitely looks better when chrome or chromium aren't installed. Left a few things to consider and for clarification.

chromium/browser_type_test.go
ankur22: I get it now, happy for you to ignore this.

chromium/browser_type.go
```diff
@@ -240,22 +239,26 @@ func (b *BrowserType) allocate(
 		return nil, err
 	}
 
-	path := opts.ExecutablePath
-	if path == "" {
-		path = b.ExecutablePath()
+	path, ok := ExecutablePath(opts.ExecutablePath, b.envLookupper, exec.LookPath)
```
inancgumus, line 242: Should this also check whether the user-provided path exists? Currently, `NewLocalBrowserProcess` does that. But it might make more sense to encapsulate it in `ExecutablePath`—even before launching a process.

chromium/browser_type.go
inancgumus: It was exported from `BrowserType` before. I'm fine unexporting it.

chromium/browser_type.go
ankur22: I agree, i found it confusing that it didn't do that.

chromium/browser_type.go
ankur22: I think unexporting would be good for now.

chromium/browser_type.go
ankur22: If you make this change, then the `ErrChromeNotInstalled` could be returned from `ExecutablePath` instead of a `bool`.

chromium/browser_type_test.go
ankur22: I think if you follow your [comment](https://github.com/grafana/xk6-browser/pull/1137/files#r1425442535) then that error could be returned from `ExecutablePath` instead of a `bool`, which would allow a test for that error in here.

ankur22 (commented):
I like the idea you came up with to check the path in `ExecutablePath` 👍

chromium/browser_type.go
inancgumus: 👍

chromium/browser_type.go
inancgumus: See 72d3b03b.

chromium/browser_type.go
inancgumus: See 6c46e7c8.

chromium/browser_type_test.go
inancgumus: See 5499e10.

chromium/browser_type.go
inancgumus: See 5499e10.

chromium/browser_type.go
```diff
@@ -240,23 +244,27 @@ func (b *BrowserType) allocate(
...
-
 	return common.NewLocalBrowserProcess(bProcCtx, path, args, dataDir, bProcCtxCancel, logger) //nolint: wrapcheck
 }
 
-// ExecutablePath returns the path where the extension expects to find the browser executable.
-func (b *BrowserType) ExecutablePath() (execPath string) {
-	if b.execPath != "" {
-		return b.execPath
+// ErrChromeNotInstalled is returned when the Chrome executable is not found.
+var ErrChromeNotInstalled = errors.New("chrome or chromium is not installed on this system")
```
ka3de, line 251: Strictly speaking this message seems to say that one or the other is not installed and therefore there's an error. Should it be something like:
```suggestion
var ErrChromeNotInstalled = errors.New("neither chrome nor chromium are installed in this system")
```

ka3de (approved):
Nice! Just a small comment on the error message, but LGTM.

chromium/browser_type.go
inancgumus: Ah, right, thanks 👍 Should we say "is": neither chrome nor chromium is installed on this system"?

ka3de: I would think "are" is OK, as we are speaking about two different subjects, but maybe I'm wrong :thinking:

inancgumus: @heitortsergent WDYT? :)

inancgumus: I'm wondering since the linked article suggests: _When neither and nor link singular terms, the verb is singular: Neither the car nor the truck is available._

inancgumus: @ka3de See: bbfea56f.

ka3de (approved):
LGTM.

chromium/browser_type.go
```diff
@@ -255,11 +255,16 @@ func executablePath(
...
-) (outPath string, ok bool) {
-	if strings.TrimSpace(path) != "" {
-		return path, true
+) (string, error) {
+	// find the browser executable in the user provided path
+	if path := strings.TrimSpace(path); path != "" {
+		if _, err := lookPath(path); err == nil {
+			return path, nil
+		}
+		return "", ErrChromeNotInstalled
```
ankur22, line 264: I wonder if it would make it clearer if there was a different error returned here, such as "chrome nor chromium not found at given path"

ankur22 (dismissed):
LGTM. Left one minor thing to consider.

chromium/browser_type.go
inancgumus: I was considering the same, TBH 👍 It might be not very clear for users to see Chrome/Chrome not installed while they're providing the path. These two are separate cases. See: 8227bb0.

ankur22 (approved):
LGTM

PR #1139, opened 2023-12-15
Refactor page.on integration tests

ankur22:
## What?

Refactors the `page.on` integration test.

## Why?

This change helps us to quickly debug any issues with the `page.on` integration tests by asserting on all individual expectations, which means that in the test output the expectation and actual values will be visible.

## Checklist

### DIFF
--- a/tests/page_test.go
+++ b/tests/page_test.go
@@ -9,7 +9,6 @@ import (
 	"io"
 	"net/http"
 	"strconv"
-	"strings"
 	"testing"
 	"time"
 
@@ -961,7 +960,7 @@ func TestPageOn(t *testing.T) { //nolint:gocognit
 			assertFn: func(t *testing.T, cm *common.ConsoleMessage) {
 				t.Helper()
 				assert.Equal(t, "timeEnd", cm.Type)
-				assert.True(t, strings.HasPrefix(cm.Text, "k6: 0."), `missing expected prefix "k6: 0."`)
+				assert.Regexp(t, `^k6: [0-9]+\.[0-9]+`, cm.Text, `expected prefix "k6: <a float>" but got %q`, cm.Text)
 				assert.True(t, cm.Page.URL() == blankPage, "url is not %s", blankPage)
 			},
 		},
--- a/tests/page_test.go
+++ b/tests/page_test.go
@@ -787,164 +787,182 @@ func TestPageOn(t *testing.T) { //nolint:gocognit
 	testCases := []struct {
 		name      string
 		consoleFn string
-		assertFn  func(*common.ConsoleMessage) bool
+		assertFn  func(*testing.T, *common.ConsoleMessage)
 	}{
 		{
 			name:      "on console.log",
 			consoleFn: "() => console.log('this is a log message')",
-			assertFn: func(cm *common.ConsoleMessage) bool {
-				return cm.Type == "log" &&
-					cm.Text == "this is a log message" &&
-					cm.Args[0].JSONValue().String() == "this is a log message" &&
-					cm.Page.URL() == blankPage
+			assertFn: func(t *testing.T, cm *common.ConsoleMessage) {
+				t.Helper()
+				assert.Equal(t, "log", cm.Type)
+				assert.Equal(t, "this is a log message", cm.Text)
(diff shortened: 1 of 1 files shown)


inancgumus (approved):

inancgumus:
Asserting for each case is indeed better for determining where the test/code fails 👍 We should follow this practice more often 👍

ka3de (approved):
Seems to be working consistently OK, at least locally. LGTM.

tests/page_test.go
```diff
@@ -961,7 +960,7 @@ func TestPageOn(t *testing.T) { //nolint:gocognit
 			assertFn: func(t *testing.T, cm *common.ConsoleMessage) {
 				t.Helper()
 				assert.Equal(t, "timeEnd", cm.Type)
-				assert.True(t, strings.HasPrefix(cm.Text, "k6: 0."), `missing expected prefix "k6: 0."`)
+				assert.Regexp(t, "^k6: [0-9]*.[0-9]*", cm.Text, `expected prefix "k6: 0." but got %q`, cm.Text)
```
inancgumus, line 963: Can you help me understand why we use a regex here?

ankur22: When working with `console.time` it calculates the time in between the two calls `console.time('k6');` and  `console.timeEnd('k6');` and prints the delta in the console output. Since it's time we can't be sure how long it will take. In CI it seems to be quite a bit slower than running it locally. Using a regex to verify that the prefix of the output is made up of `k6` and a number.

inancgumus: Thanks for explaining. So, why does the assert error message say `expected prefix "k6: 0."`? Since we try to match for k6: `0-9.0-9` instead of `k6: 0.0-9`.

ankur22: Good spot, i've changed it in f604d5807fe828169b67482d2e9c24ab9238fdd1

tests/page_test.go
```diff
@@ -960,7 +960,7 @@ func TestPageOn(t *testing.T) { //nolint:gocognit
 			assertFn: func(t *testing.T, cm *common.ConsoleMessage) {
 				t.Helper()
 				assert.Equal(t, "timeEnd", cm.Type)
-				assert.Regexp(t, "^k6: [0-9]*.[0-9]*", cm.Text, `expected prefix "k6: 0." but got %q`, cm.Text)
+				assert.Regexp(t, "^k6: [0-9]*.[0-9]*", cm.Text, `expected prefix "k6: <a float>" but got %q`, cm.Text)
```
inancgumus, line 963: I think this should be like:

```suggestion
				assert.Regexp(t, "^k6: [0-9]*\.[0-9]*", cm.Text, `expected prefix "k6: <a float>" but got %q`, cm.Text)
```

Otherwise `.` will match to any character.

inancgumus (commented):
LGTM, one [suggestion](https://github.com/grafana/xk6-browser/pull/1139#discussion_r1427994045).

tests/page_test.go
inancgumus: If we always expect a number `*`s should be `+`s instead. Like this.

```suggestion
				assert.Regexp(t, "^k6: [0-9]+\.[0-9]+", cm.Text, `expected prefix "k6: <a float>" but got %q`, cm.Text)
```

ankur22: Resolved in 138f3b8ac8cb3d844d357e6ac57248a7b9637340

inancgumus: Thanks, should we mind [this](https://github.com/grafana/xk6-browser/pull/1139#discussion_r1428000647)?

ankur22: `*` and `+` work for this use case.

inancgumus: OK, so, we expect n.m, .m, or n. then. Alright 👍

inancgumus (dismissed):

tests/page_test.go
ankur22: Ah, wait, i see resolved in 559a597015f757602f77eea4d79c9fd0566b497d

inancgumus: Regexes are always P*ITA :)

ka3de (approved):

PR #1141, opened 2023-12-15
Refactor TraceNavigation

ka3de:
## What?

Based on discussions in #1140 , a decision was made to, instead of using the URL as the span name for navigation spans, use the word 'navigation' to identify the action, following the same pattern as the rest of the spans. Due to this, and because a navigation span will always have that fixed span name, it is no longer necessary to accept the span name as input for the method.

## Why?

Even though the traces generated from k6 browser have some particularities to fit our needs, this change and the changes done in #1140 try to make our implementation more compliant with Open Telemetry conventions and recommendations. These changes will also improve traces searches.

## Checklist

## Related PR(s)/Issue(s)

#1100, #1140

### DIFF
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -756,8 +756,7 @@ func (fs *FrameSession) onFrameNavigated(frame *cdp.Frame, initial bool) {
 	}
 
 	_, fs.mainFrameSpan = TraceNavigation(
-		fs.ctx, fs.targetID.String(), frame.URL,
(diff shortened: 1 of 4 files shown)


ankur22 (approved):
LGMT 🚀

ka3de:
Because k6 has merged https://github.com/grafana/k6/pull/3503, our integration tests with `main` version of k6 fails. Therefore we have to merge #1142 first and then rebase this branch on top.

Issue #1114, opened 2023-11-27
Add `.clear()` to Locator API


mdcruz:
### Feature Description

Playwright's Locator API contains more methods that are useful for browser testing, such as:

-  `.clear()` for clearing text inputs

When typing or filling textboxes or input fields, it's useful to clear existing fields in case they already have values.

### Suggested Solution (optional)

https://playwright.dev/docs/api/class-locator#locator-clear

```[tasklist]
### Tasks
- [x] Implement new API -- https://github.com/grafana/xk6-browser/pull/1149
- [x] Add to release notes -- https://github.com/grafana/k6/pull/3524
- [x] Add documentation -- https://github.com/grafana/k6-docs/pull/1468
- [x] Add type definitions -- https://github.com/grafana/k6-DefinitelyTyped/commit/6d7f3241527d832e2302463a7fb9876902d3624e & https://github.com/grafana/k6-DefinitelyTyped/pull/24
```

PR #1149, opened 2024-01-04
Add Clear to locator

ankur22:
## What?

This adds a new method `clear` to the `locator` class. This works with the existing `Fill` method which is in the `locator` class. Resources that contributed to the implementation can be found by following the links below:
- https://playwright.dev/docs/api/class-locator
- https://github.com/microsoft/playwright/blob/e30c0a6ff7c48a3abc55eab63fa15d31f6241280/packages/playwright-core/src/client/locator.ts#L130

## Why?

When typing or filling textboxes or input fields, it's useful to clear existing fields in case they already have values.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1114

### DIFF
--- a/tests/locator_test.go
+++ b/tests/locator_test.go
@@ -2,6 +2,7 @@ package tests
 
 import (
 	"testing"
+	"time"
 
 	"github.com/dop251/goja"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


tests/locator_test.go
```diff
@@ -53,6 +54,19 @@ func TestLocator(t *testing.T) {
 				})
 			},
 		},
+		{
+			"Clear", func(tb *testBrowser, p *common.Page) {
+				const value = "fill me up"
```
inancgumus, line 59: 😂 "fill me up"

inancgumus (approved):
Nice work 👍

Issue #1148, opened 2024-01-04
panic: runtime error: invalid memory address or nil pointer dereference


rauldeheer:
### Brief summary

We're trying to run a test within our application using the k6x-browser. We're encountering a weird issue where the test sometimes crashes with the following error:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x98 pc=0x1def4c9]
register [--------------------------------------] 1 VUs 00m30.9s/16m40s 0/1 iters, 1 per VU
goroutine 12019 [running]:
github.com/grafana/xk6-browser/common.(*Frame).ID(0x213f5e0?)
github.com/grafana/xk6-browser@v1.2.1/common/frame.go:1327 +0x29
github.com/grafana/xk6-browser/common.(*FrameSession).initIsolatedWorld(0xc0002daa50, {0x23b0e8a, 0x1c})
github.com/grafana/xk6-browser@v1.2.1/common/frame_session.go:421 +0x45a
github.com/grafana/xk6-browser/common.NewFrameSession({0x2688968, 0xc0004ac0a0}, {0x268f738?, 0xc0013be320?}, 0xc000683ba0, 0xc000e1e6e0, {0xc0003927a0, 0x20}, 0xc00037f830)
github.com/grafana/xk6-browser@v1.2.1/common/frame_session.go:134 +0x938
github.com/grafana/xk6-browser/common.(*FrameSession).attachIFrameToTarget(0xc000e1e6e0, 0xc002b6ba70, {0xc000392780, 0x20})
github.com/grafana/xk6-browser@v1.2.1/common/frame_session.go:905 +0xe5
github.com/grafana/xk6-browser/common.(*FrameSession).onAttachedToTarget(0xc000e1e6e0, 0xc002663520)
github.com/grafana/xk6-browser@v1.2.1/common/frame_session.go:838 +0x271
github.com/grafana/xk6-browser/common.(*FrameSession).initEvents.func1()
github.com/grafana/xk6-browser@v1.2.1/common/frame_session.go:267 +0x3f0
created by github.com/grafana/xk6-browser/common.(*FrameSession).initEvents in goroutine 77
github.com/grafana/xk6-browser@v1.2.1/common/frame_session.go:216 +0x19b
```

### k6 version

0.48.0

### OS

macOS 14 (arm) - Sonoma 14.2.1 (23C71)

### Steps to reproduce the problem

[...]

PR #1152, opened 2024-01-10
Fix potential nil exception when getting frame

ankur22:
## What?

This change avoids the need to check for `nil` when getting the frame with an id, and reinforces a requirement to check with an addition of a `bool` return value to indicate whether the frame was found or not.

## Why?

When a part of the browser module needs to retrieve the frame given an id, it should check whether the frame is nil before using it. This is not safe, since it's easy to forget to check for nil, and so a new return value is also returned.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1148

### DIFF
--- a/common/element_handle.go
+++ b/common/element_handle.go
@@ -440,13 +440,13 @@ func (h *ElementHandle) ownerFrame(apiCtx context.Context) *Frame {
 	if frameId == "" {
 		return nil
 	}
-	frame := h.frame.page.frameManager.getFrameByID(frameId)
-	if frame != nil {
+	frame, found := h.frame.page.frameManager.getFrameByID(frameId)
+	if found {
 		return frame
 	}
 	for _, page := range h.frame.page.browserCtx.browser.pages {
-		frame = page.frameManager.getFrameByID(frameId)
-		if frame != nil {
+		frame, found = page.frameManager.getFrameByID(frameId)
+		if found {
 			return frame
 		}
 	}
@@ -766,7 +766,12 @@ func (h *ElementHandle) ContentFrame() (*Frame, error) {
 		return nil, fmt.Errorf("element is not an iframe")
 	}
 
-	return h.frame.manager.getFrameByID(node.FrameID), nil
+	frame, found := h.frame.manager.getFrameByID(node.FrameID)
+	if !found {
+		return nil, fmt.Errorf("frame not found for id")
+	}
+
+	return frame, nil
 }
 
 func (h *ElementHandle) Dblclick(opts goja.Value) {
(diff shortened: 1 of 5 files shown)


ka3de (dismissed):

common/element_handle.go
```diff
@@ -766,7 +766,12 @@ func (h *ElementHandle) ContentFrame() (*Frame, error) {
 		return nil, fmt.Errorf("element is not an iframe")
 	}
 
-	return h.frame.manager.getFrameByID(node.FrameID), nil
+	frame, found := h.frame.manager.getFrameByID(node.FrameID)
+	if !found {
+		return nil, fmt.Errorf("frame not found for id")
```
inancgumus, line 771: How about?

```suggestion
		return nil, fmt.Errorf("frame not found for id %d", node.FrameID)
```

common/frame_manager.go
```diff
@@ -390,10 +390,16 @@ func (m *FrameManager) frameRequestedNavigation(frameID cdp.FrameID, url string,
 	return nil
 }
 
-func (m *FrameManager) getFrameByID(id cdp.FrameID) *Frame {
+// getFrameByID will attempt to find a frame with the given id. If one is found
+// the frame and true will be returned. If one isn't found nil and false will
+// be returned.
```
inancgumus, line 395: What about using the present tense as the rest of the code?

```suggestion
// getFrameByID finds a frame with id. If found, it returns the frame and true,
// otherwise, it returns nil and false.
```

common/frame_manager.go
```diff
@@ -390,10 +390,16 @@ func (m *FrameManager) frameRequestedNavigation(frameID cdp.FrameID, url string,
...
+func (m *FrameManager) getFrameByID(id cdp.FrameID) (*Frame, bool) {
 	m.framesMu.RLock()
 	defer m.framesMu.RUnlock()
-	return m.frames[id]
+
+	frame, ok := m.frames[id]
+
+	return frame, ok
```
inancgumus, line 402: We use `ok` here, and `found` in other parts of the code. It might be better to use `ok` in the rest, too, as it's also idiomatic to use `ok`.

inancgumus (commented):
It's an excellent idea to use the comma-ok idiom here 👍 LGTM and have some suggestions.

common/element_handle.go
ankur22: Nice spot! Thanks!

common/frame_manager.go
ankur22: Yeah, i was wondering whether `found` read a bit better. Happy to change it ok.

ankur22: Refactored in 5bb8c607033917d1ce243a921c8ca48451eedb6b

inancgumus (approved):

Issue #1162, opened 2024-01-18
Goja runtime shouldn't be accessed while parsing options in async APIs


ankur22:
# What

We must not work with the goja runtime when the browser module is running a promisified API call off the event loop i.e. on a new goroutine. All promisified APIs in the browser run in a new goroutine. A concrete example of such a case is with [page.goto](https://github.com/grafana/xk6-browser/blob/5f754c2c2252896016fd77d1fd3234a795049a0e/common/frame.go#L954-L956) where it is parsing the goja values in the new goroutine, which should be moved to the mapping layer.

This issue only deals with the refactoring of the option and argument parsing, https://github.com/grafana/xk6-browser/issues/1182 deals with the inner goja usage.

# Why

The goja runtime is not thread safe and should only be accessed when running on the event loop. If this is not adhered to, it can cause panics during runtime.

# How

The following is the list of all async/promisified APIs, check each one and ensure that the goja runtime is only accessed in the event loop thread and not in the new go routine.

```[tasklist]
### Tasks
- [x] https://github.com/grafana/xk6-browser/pull/1164
- [ ] https://github.com/grafana/xk6-browser/issues/1169
- [ ] https://github.com/grafana/xk6-browser/issues/1172
- [ ] https://github.com/grafana/xk6-browser/issues/1174
- [ ] https://github.com/grafana/xk6-browser/issues/1176
- [ ] https://github.com/grafana/xk6-browser/issues/1177
- [ ] https://github.com/grafana/xk6-browser/issues/1180
- [ ] https://github.com/grafana/xk6-browser/issues/1181
- [ ] https://github.com/grafana/xk6-browser/issues/1184
- [ ] https://github.com/grafana/xk6-browser/issues/1185
- [ ] https://github.com/grafana/xk6-browser/issues/1187
- [x] https://github.com/grafana/xk6-browser/issues/1178
- [x] Fix `browserContext.setExtraHTTPHeaders` (unimplemented method)
- [x] Add fixes to release notes: https://github.com/grafana/k6/pull/3578
```

# Note

inancgumus:
Should we close this one? @ankur22

PR #1164, opened 2024-01-18
Refactor locator.click parse

ankur22:
## What?

Refactor `locator.click` so that it parses the goja value using the goja runtime on the main goja thread and not in a new goroutine.

## Why?

This will avoid race issues since the goja runtime is not thread safe since it is a JS interpreter that is single threaded.

I've also confirmed that this API doesn't work with goja anywhere else, including when working with `eval`, which is does since it uses `newPointerAction`, which calls `waitForSelector`, that calls ,`evalWithScript` and eventually `eval`, but since `returnByValue` is `false` it doesn't work with goja to transform the return value to a goja value.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1162

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -52,17 +52,15 @@ func mapLocator(vu moduleVU, lo *common.Locator) mapping {
 
 			return lo.Clear(copts) //nolint:wrapcheck
 		},
-		"click": func(opts goja.Value) *goja.Promise {
+		"click": func(opts goja.Value) (*goja.Promise, error) {
 			popts, err := parseFrameClickOptions(vu.Context(), opts, lo.Timeout())
+			if err != nil {
+				return nil, err
+			}
 
 			return k6ext.Promise(vu.Context(), func() (any, error) {
-				if err != nil {
-					return nil, err
-				}
-
-				err = lo.Click(popts)
-				return nil, err //nolint:wrapcheck
-			})
+				return nil, lo.Click(popts) //nolint:wrapcheck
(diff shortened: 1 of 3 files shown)


browser/mapping.go
```diff
@@ -52,8 +53,14 @@ func mapLocator(vu moduleVU, lo *common.Locator) mapping {
...
+			popts, err := parseFrameClickOptions(vu.Context(), opts, lo.Timeout())
+
 			return k6ext.Promise(vu.Context(), func() (any, error) {
-				err := lo.Click(opts)
+				if err != nil {
+					return nil, err
+				}
+
+				err = lo.Click(popts)
 				return nil, err //nolint:wrapcheck
```
inancgumus, line 64: Is there a functional reason for assigning this operation's result to `err` here?

inancgumus (dismissed):
LGTM 👍

browser/mapping.go
ka3de: You mean, why it is not:
```suggestion
				return nil, lo.Click(popts) //nolint:wrapcheck
```
? :+1: (nit)

ka3de (approved):

browser/mapping.go
ankur22: Changed in eddc74e79501e35676b25af76f6193dffefd3a1d. I've also returned early from any parse errors in the same commit.

inancgumus (approved):

PR #1166, opened 2024-01-19
Refactor parseConsoleRemoteObject to return error

ankur22:
## What?

Returning an error from `parseConsoleRemoteObject` instead of logging an error.

## Why?

There is another use case for `parseConsoleRemoteObject`, which is when `JSONValue` is called. At the moment this API uses `valueFromRemoteObject` which returns a `goja.Value`.

Interestingly, the return value from `JSONValue` is always converted to string before being returned to the caller, so the extra step to convert to goja and then back to string is not needed.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1168

### DIFF
--- a/common/remote_object.go
+++ b/common/remote_object.go
@@ -227,42 +227,48 @@ func valueFromRemoteObject(_ context.Context, robj *cdpruntime.RemoteObject) (an
 	return val, err
 }
 
-func parseConsoleRemoteObjectPreview(logger *log.Logger, op *cdpruntime.ObjectPreview) string {
+func parseConsoleRemoteObjectPreview(logger *log.Logger, op *cdpruntime.ObjectPreview) (string, error) {
 	obj := make(map[string]string)
 	if op.Overflow {
 		logger.Infof("parseConsoleRemoteObjectPreview", "object is too large and will be parsed partially")
 	}
 
 	for _, p := range op.Properties {
-		val := parseConsoleRemoteObjectValue(logger, p.Type, p.Subtype, p.Value, p.ValuePreview)
+		val, err := parseConsoleRemoteObjectValue(logger, p.Type, p.Subtype, p.Value, p.ValuePreview)
+		if err != nil {
+			return "", err
+		}
 		obj[p.Name] = val
 	}
 
 	bb, err := json.Marshal(obj)
 	if err != nil {
-		logger.Errorf("parseConsoleRemoteObjectPreview", "failed to marshal object to string: %v", err)
+		return "", fmt.Errorf("marshaling object to string: %w", err)
 	}
 
-	return string(bb)
+	return string(bb), nil
 }
 
-func parseConsoleRemoteArrayPreview(logger *log.Logger, op *cdpruntime.ObjectPreview) string {
+func parseConsoleRemoteArrayPreview(logger *log.Logger, op *cdpruntime.ObjectPreview) (string, error) {
 	arr := make([]any, 0, len(op.Properties))
 	if op.Overflow {
-		logger.Warnf("parseConsoleRemoteArrayPreview", "array is too large and will be parsed partially")
+		logger.Infof("parseConsoleRemoteArrayPreview", "array is too large and will be parsed partially")
 	}
 
 	for _, p := range op.Properties {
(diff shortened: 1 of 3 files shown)


common/remote_object.go
```diff
@@ -229,42 +229,48 @@ func valueFromRemoteObject(ctx context.Context, robj *cdpruntime.RemoteObject) (
...
 	if op.Overflow {
-		logger.Warnf("parseConsoleRemoteArrayPreview", "array is too large and will be parsed partially")
+		logger.Infof("parseConsoleRemoteArrayPreview", "array is too large and will be parsed partially")
 	}
 
 	for _, p := range op.Properties {
-		val := parseConsoleRemoteObjectValue(logger, p.Type, p.Subtype, p.Value, p.ValuePreview)
+		val, err := parseConsoleRemoteObjectValue(logger, p.Type, p.Subtype, p.Value, p.ValuePreview)
+		if err != nil {
+			return "", err
```
inancgumus, line 263: Wrapping this error with contextual data will help while debugging. It's harder to track all these calls otherwise.

ka3de: Maybe also for https://github.com/grafana/xk6-browser/blob/8d48cf690355217cb9338c6d0868b4a9f21da167/common/remote_object.go#L241Which is also introduced in this PR.

ankur22: Yes, that might be a good idea. The reason I haven't is because:

1. `parseConsoleRemoteObjectValue` called
2. `parseConsoleRemoteObjectValue` calls `parseConsoleRemoteArrayPreview`
3. If there are `op.Properties` in `parseConsoleRemoteArrayPreview`, then it will call back to `parseConsoleRemoteObjectValue`.
4. `parseConsoleRemoteObjectValue` maybe now call back to `parseConsoleRemoteArrayPreview`.
5. If there's a marshalling error [here](https://github.com/grafana/xk6-browser/blob/fc47e2edd4bc3ab10365896f9e59e40a5e6c9fe3/common/remote_object.go#L266-L269) an error is returned which should only be `marshaling array to string: <some-error>`.

If more contextual detail was added then I think it would only be noise, which might be something like: `failed to parse console remote object: marshaling array to string: <some-error>`, and depending on how many levels deep this recursion goes then it could contain lots of occurrences of `failed to parse console remote object`.

I think that's correct. Let me know if i missed something or if there are strong opinions on this.

inancgumus: The noise is valuable when debugging 🥲 Train of errors like you listed won't be helpful, though. What about trimming them for this special case to get a useful error. 

Something like:

```
"parsing console remote object x 5: marshaling array to string: <error>"
```

`x 5` is added just by `strings.Count(err.Error(), "parsing console remote object")`.

inancgumus: If this is not viable or makes sense, we might want to know which object failed while parsing here: `marshaling object to a string: %w.` This might be especially important for remote-run failures.

ankur22: I like the simpler option of adding the string representation of the object that can't be marshalled. I guess something like: `fmt.Errorf("marshaling array %v to string: %w", arr, err)`?

inancgumus: `%v` print the bytes, maybe `%s`?

common/remote_object.go
```diff
@@ -227,42 +227,48 @@ func valueFromRemoteObject(_ context.Context, robj *cdpruntime.RemoteObject) (an
...
+		}
 		obj[p.Name] = val
 	}
 
 	bb, err := json.Marshal(obj)
 	if err != nil {
-		logger.Errorf("parseConsoleRemoteObjectPreview", "failed to marshal object to string: %v", err)
+		return "", fmt.Errorf("marshaling object to string: %w", err)
```
inancgumus, line 246: Here is one more marshaling error that we might want to enrich with the marshaled object.

ankur22: Changed in b0f41e70076a21a5bb9a2007e5b213785a80bd95

common/remote_object.go
ankur22: Changed in b0f41e70076a21a5bb9a2007e5b213785a80bd95

inancgumus (dismissed):

inancgumus (approved):

Issue #1170, opened 2024-01-19
Remove the call to goja runtime from `elementHandle.newPointerAction`


ankur22:
# What

Can [elementHandle.newPointerAction](https://github.com/grafana/xk6-browser/blob/5f754c2c2252896016fd77d1fd3234a795049a0e/common/element_handle.go#L1407-L1492) be de-gojafied? All `click` APIs use it.

# How

After #1189, there's only one method to de-gojafy: `ElementHandle.offsetPosition`.

```[tasklist]
### Tasks
- [ ] #1217
- [ ] grafana/k6#3603
```

inancgumus:
Closed by #1170.

Issue #1169, opened 2024-01-19
Refactor goja runtime value parsing out of `elementHandle.click`


ankur22:
# What

Refactor the parsing of the option out of the promise and into the mapping layer for `elementHandle.click`.

# Why

This will help avoid a scenario where more than one off main goja thread goroutine (i.e. promise) is ran concurrently which could try to access the goja runtime concurrently and cause a panic.

# How

Move [this](https://github.com/grafana/xk6-browser/blob/5f754c2c2252896016fd77d1fd3234a795049a0e/common/element_handle.go#L737-L740) to mapping layer. Should be a similar change to https://github.com/grafana/xk6-browser/pull/1164.
```[tasklist]
### Tasks
- [ ] https://github.com/grafana/xk6-browser/pull/1171
- [ ] https://github.com/grafana/k6/pull/3578
```

PR #1171, opened 2024-01-19
Refactor elementHandle.click parse

ankur22:
## What?

Move the parse logic for `elementHandle.click` into the mapping layer to reduce the call to the goja runtime by one less call.

## Why?

This will mitigate the risk of a panic occurring due to using the goja runtime in a new goroutine off the main goja thread. This doesn't solve the complete problem and it still relie on using the goja runtime, but that issue can be tracked separately (https://github.com/grafana/xk6-browser/issues/1170).

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1169

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -219,11 +219,18 @@ func mapElementHandle(vu moduleVU, eh *common.ElementHandle) mapping {
 	maps := mapping{
 		"boundingBox": eh.BoundingBox,
 		"check":       eh.Check,
-		"click": func(opts goja.Value) *goja.Promise {
+		"click": func(opts goja.Value) (*goja.Promise, error) {
+			ctx := vu.Context()
+
+			popts := common.NewElementHandleClickOptions(eh.Timeout())
+			if err := popts.Parse(ctx, opts); err != nil {
+				return nil, fmt.Errorf("parsing element click options: %w", err)
+			}
+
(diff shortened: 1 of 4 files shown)


browser/mapping.go
```diff
@@ -219,11 +219,18 @@ func mapElementHandle(vu moduleVU, eh *common.ElementHandle) mapping {
 	maps := mapping{
 		"boundingBox": eh.BoundingBox,
 		"check":       eh.Check,
-		"click": func(opts goja.Value) *goja.Promise {
+		"click": func(opts goja.Value) (*goja.Promise, error) {
+			ctx := vu.Context()
+
+			popts := common.NewElementHandleClickOptions(eh.Timeout())
```
ka3de, line 225: nit: What does `popts` stand for? Should it be `clickOpts`?

ka3de (approved):

browser/mapping.go
inancgumus: I think `copts` would be fine for "click options".

inancgumus (approved):

browser/mapping.go
ankur22: `popts` -> `parsed options`. Aren't `opts` before being parsed also click options? I prefer `popts`.

Issue #1172, opened 2024-01-19
Refactor goja runtime value parsing out of `frame.click` promise


ankur22:
# What

Refactor the parsing of the option out of the promise and into the mapping layer for `frame.click`.

# Why

This will help avoid a scenario where more than one off main goja thread goroutine (i.e. promise) is ran concurrently which could try to access the goja runtime concurrently and cause a panic.

# How

```[tasklist]
### Tasks
- [ ] https://github.com/grafana/xk6-browser/pull/1173
- [ ] https://github.com/grafana/k6/pull/3578
```

PR #1173, opened 2024-01-19
Refactor frame.click option parsing

ankur22:
## What?

This refactors the option parsing out of the promise and into the mapping layer for `frame.click`.

## Why?

This will help mitigate the risk of a panic occurring due to multiple goroutines accessing the goja runtime (which is not thread safe) concurrently.

More works needs to be done to totally remove the goja runtime usage from within the `frame.click` promise which can be tracked in https://github.com/grafana/xk6-browser/issues/1170.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1172

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -334,11 +334,16 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
 			}
 			return rt.ToValue(mcfs).ToObject(rt)
 		},
-		"click": func(selector string, opts goja.Value) *goja.Promise {
+		"click": func(selector string, opts goja.Value) (*goja.Promise, error) {
+			popts, err := parseFrameClickOptions(vu.Context(), opts, f.Timeout())
+			if err != nil {
+				return nil, err
+			}
+
 			return k6ext.Promise(vu.Context(), func() (any, error) {
-				err := f.Click(selector, opts)
+				err := f.Click(selector, popts)
 				return nil, err //nolint:wrapcheck
-			})
+			}), nil
 		},
 		"content":       f.Content,
 		"dblclick":      f.Dblclick,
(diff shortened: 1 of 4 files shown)


browser/mapping.go
```diff
@@ -327,11 +327,16 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
 			}
 			return rt.ToValue(mcfs).ToObject(rt)
 		},
-		"click": func(selector string, opts goja.Value) *goja.Promise {
+		"click": func(selector string, opts goja.Value) (*goja.Promise, error) {
+			popts, err := parseFrameClickOptions(vu.Context(), opts, f.Timeout())
```
ka3de, line 331: nit: I guess `popts` stands for `parsed opts`? Would `clickOpts` be more representative?

common/page.go
```diff
@@ -652,7 +652,12 @@ func (p *Page) IsChecked(selector string, opts goja.Value) bool {
 func (p *Page) Click(selector string, opts goja.Value) error {
 	p.logger.Debugf("Page:Click", "sid:%v selector:%s", p.sessionID(), selector)
 
-	return p.MainFrame().Click(selector, opts) //nolint:wrapcheck
+	popts := NewFrameClickOptions(p.defaultTimeout())
+	if err := popts.Parse(p.ctx, opts); err != nil {
```
ka3de, line 656: nit: s/popts/clickOpts

ka3de (approved):
Thanks you for working on this immediately :clap:

browser/mapping.go
inancgumus: `copts` since the code is very short and this can be understood from the context.

inancgumus (approved):

browser/mapping.go
ankur22: Going with popts for now. Happy to do a mass renaming of all options later if we need.

PR #1175, opened 2024-01-19
Refactor page.click option parsing

ankur22:
## What?

This refactors the option parsing out of the promise and into the mapping layer for `page.click`.

## Why?

This will help mitigate the risk of a panic occurring due to multiple goroutines accessing the goja runtime (which is not thread safe) concurrently.

More works needs to be done to totally remove the goja runtime usage from within the `page.click` promise which can be tracked in https://github.com/grafana/xk6-browser/issues/1170.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1174

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -475,11 +475,16 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
 		"addStyleTag":   p.AddStyleTag,
 		"bringToFront":  p.BringToFront,
 		"check":         p.Check,
-		"click": func(selector string, opts goja.Value) *goja.Promise {
+		"click": func(selector string, opts goja.Value) (*goja.Promise, error) {
+			popts, err := parseFrameClickOptions(vu.Context(), opts, p.Timeout())
+			if err != nil {
+				return nil, err
+			}
+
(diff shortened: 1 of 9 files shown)


browser/mapping.go
```diff
@@ -468,11 +468,16 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
 		"addStyleTag":   p.AddStyleTag,
 		"bringToFront":  p.BringToFront,
 		"check":         p.Check,
-		"click": func(selector string, opts goja.Value) *goja.Promise {
+		"click": func(selector string, opts goja.Value) (*goja.Promise, error) {
+			popts, err := parseFrameClickOptions(vu.Context(), opts, p.Timeout())
```
ka3de, line 472: nit: s/popts/clickOpts

ka3de (commented):
Overall LGTM, just left a nit comment. Also, there is a linting error for `tests/webvital_test.go` not being "_gci_-ed"

inancgumus (approved):

browser/mapping.go
ankur22: I'll do a mass renaming later if we need. Sticking with popts for now.

Issue #1176, opened 2024-01-22
Refactor goja runtime value parsing out of frame and page goto


inancgumus:
# What

Refactor the parsing of the option out of the promise and into the mapping layer for `page.goto` and `frame.goto`.

# Why

This will help avoid a scenario where more than one off main goja thread goroutine (i.e., promise) is run concurrently, which could try to access the goji runtime concurrently and cause a panic.

# How

```[tasklist]
### Tasks
- [ ] #1179
- [ ] https://github.com/grafana/k6/pull/3578
```

PR #1179, opened 2024-01-22
Move goja option parsing out of frame and page goto opts

inancgumus:
## What?

- Moves goja out of option parsing for `frame` and `page` `goto` to mapping.
- Updates tests accordingly.
- Adds two helpers: `Referrer` and `NavigationTimeout` for building options while mapping.

## Why?

See #1162 for the main reason. Also, the reason for doing both `frame` and `page` `goto` options in the same PR is that the code doesn't compile without the other. Or, it would require a lot of intermediary steps which would be confusing for the commit log.

## Checklist

## Related PR(s)/Issue(s)

Updates: #1176

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -538,8 +538,8 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
 		"goForward": p.GoForward,
 		"goto": func(url string, opts goja.Value) (*goja.Promise, error) {
 			gopts := common.NewFrameGotoOptions(
-				p.MainFrame().Referrer(),
-				p.MainFrame().NavigationTimeout(),
+				p.Referrer(),
+				p.NavigationTimeout(),
 			)
 			if err := gopts.Parse(vu.Context(), opts); err != nil {
 				return nil, fmt.Errorf("parsing page navigation options to %q: %w", url, err)
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -366,15 +366,22 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
 			return mapElementHandle(vu, fe), nil
 		},
 		"getAttribute": f.GetAttribute,
-		"goto": func(url string, opts goja.Value) *goja.Promise {
+		"goto": func(url string, opts goja.Value) (*goja.Promise, error) {
+			gopts := common.NewFrameGotoOptions(
+				f.Referrer(),
+				f.NavigationTimeout(),
+			)
+			if err := gopts.Parse(vu.Context(), opts); err != nil {
+				return nil, fmt.Errorf("parsing frame navigation options to %q: %w", url, err)
+			}
 			return k6ext.Promise(vu.Context(), func() (any, error) {
-				resp, err := f.Goto(url, opts)
+				resp, err := f.Goto(url, gopts)
 				if err != nil {
 					return nil, err //nolint:wrapcheck
 				}
 
 				return mapResponse(vu, resp), nil
-			})
+			}), nil
 		},
 		"hover":      f.Hover,
 		"innerHTML":  f.InnerHTML,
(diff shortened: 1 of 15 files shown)


browser/mapping.go
```diff
@@ -497,15 +504,22 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
...
 		},
 		"goForward": p.GoForward,
-		"goto": func(url string, opts goja.Value) *goja.Promise {
+		"goto": func(url string, opts goja.Value) (*goja.Promise, error) {
+			gopts := common.NewFrameGotoOptions(
+				p.MainFrame().Referrer(),
+				p.MainFrame().NavigationTimeout(),
+			)
+			if err := gopts.Parse(vu.Context(), opts); err != nil {
+				return nil, fmt.Errorf("parsing frame navigation options to %q: %w", url, err)
```
inancgumus, line 513: I forgot saying page here 🤦

```suggestion
				return nil, fmt.Errorf("parsing page navigation options to %q: %w", url, err)
```

tests/browser_context_options_test.go
```diff
@@ -86,7 +86,13 @@ func TestBrowserContextOptionsExtraHTTPHeaders(t *testing.T) {
 	require.NoError(t, err)
 
 	err = tb.awaitWithTimeout(time.Second*5, func() error {
-		resp, err := p.Goto(tb.url("/get"), nil)
+		opts := &common.FrameGotoOptions{
+			Timeout: common.DefaultTimeout,
+		}
```
inancgumus, line 91: `page.Goto` could implicitly set a default timeout. But I think this makes it explicit and less error-prone. WDYT?

browser/mapping.go
```diff
@@ -339,15 +339,22 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
 			return mapElementHandle(vu, fe), nil
 		},
 		"getAttribute": f.GetAttribute,
-		"goto": func(url string, opts goja.Value) *goja.Promise {
+		"goto": func(url string, opts goja.Value) (*goja.Promise, error) {
+			gopts := common.NewFrameGotoOptions(
```
inancgumus, line 343: Should we rename `FrameGotoOptions` to `GotoOptions` as both `Frame` and `Page` share the same type? And saying `FrameGotoOptions` in the `Page` signature feels a bit off.

browser/mapping.go
inancgumus: Or, just? for both?

```suggestion
				return nil, fmt.Errorf("parsing goto options to %q: %w", url, err)
```

browser/mapping.go
ankur22: I'm fine with having `FrameGotoOptions` in `page.goto` since `page.goto` uses `frame`'s implementation. If we refactored it out of `frame`, then I'd go down the route of renaming `FrameGotoOptions`.

tests/browser_context_options_test.go
ankur22: Yep, I was wondering the same. I went with the option of being explicit for now, and ensuring it matched the current behaviour. I think in a future refactor we can make it implicit.

browser/mapping.go
```diff
@@ -497,15 +504,22 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
 			})
 		},
 		"goForward": p.GoForward,
-		"goto": func(url string, opts goja.Value) *goja.Promise {
+		"goto": func(url string, opts goja.Value) (*goja.Promise, error) {
+			gopts := common.NewFrameGotoOptions(
+				p.MainFrame().Referrer(),
+				p.MainFrame().NavigationTimeout(),
```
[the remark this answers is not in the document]
ka3de: I think that makes sense, and it will probably make it easier for people not so familiar with the code?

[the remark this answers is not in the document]
inancgumus: @ka3de see: [e04f7be](https://github.com/grafana/xk6-browser/pull/1179/commits/e04f7bed4c2c11a0d7776282145b2cce96937052).

ka3de (approved):

browser/mapping.go
ankur22: If the `goto` implementation was moved out of `frame` and into a new file (e.g. `navigation`) then i think it would make sense to rename the the option. I think that the scope of that work is outside of this PR though.

ankur22 (approved):
LGTM 🚀

browser/mapping.go
inancgumus: Thanks for the clarification 👍

PR #1183, opened 2024-01-23
Refactor parsing of arguments for waitForFunction

ankur22:
## What?

Refactor the parsing of the arguments of `waitForFunction` APIs so that they are done outside the promise and therefore on the main goja thread.

## Why?

This will help mitigate issues which can occur when trying to access the goja runtime from multiple goroutines, which can causes panics since the goja runtime is not thread safe.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1180 & https://github.com/grafana/xk6-browser/issues/1181

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -411,21 +411,11 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
 		"uncheck":       f.Uncheck,
 		"url":           f.URL,
 		"waitForFunction": func(pageFunc, opts goja.Value, args ...goja.Value) (*goja.Promise, error) {
-			popts := common.NewFrameWaitForFunctionOptions(f.Timeout())
-			err := popts.Parse(vu.Context(), opts)
+			js, popts, pargs, err := parseWaitForFunctionArgs(
+				vu.Context(), f.Timeout(), pageFunc, opts, args...,
+			)
 			if err != nil {
-				return nil, fmt.Errorf("parsing waitForFunction options: %w", err)
-			}
-
-			js := pageFunc.ToString().String()
-			_, isCallable := goja.AssertFunction(pageFunc)
-			if !isCallable {
-				js = fmt.Sprintf("() => (%s)", js)
-			}
-
-			pargs := make([]any, 0, len(args))
-			for _, arg := range args {
-				pargs = append(pargs, arg.Export())
+				return nil, fmt.Errorf("frame waitForFunction: %w", err)
 			}
 
 			return k6ext.Promise(vu.Context(), func() (result any, reason error) {
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -410,10 +410,16 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
 		"type":          f.Type,
 		"uncheck":       f.Uncheck,
 		"url":           f.URL,
-		"waitForFunction": func(pageFunc, opts goja.Value, args ...goja.Value) *goja.Promise {
+		"waitForFunction": func(pageFunc, opts goja.Value, args ...goja.Value) (*goja.Promise, error) {
+			popts := common.NewFrameWaitForFunctionOptions(f.Timeout())
+			err := popts.Parse(vu.Context(), opts)
+			if err != nil {
+				return nil, fmt.Errorf("parsing waitForFunction options: %w", err)
+			}
+
 			return k6ext.Promise(vu.Context(), func() (result any, reason error) {
-				return f.WaitForFunction(pageFunc, opts, args...) //nolint:wrapcheck
-			})
+				return f.WaitForFunction(pageFunc, popts, args...) //nolint:wrapcheck
+			}), nil
 		},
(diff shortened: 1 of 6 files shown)


browser/mapping.go
```diff
@@ -457,6 +465,39 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
 	return maps
 }
 
+func parseWaitForFunctionArgs(
```
ankur22, line 468: Not sure about this. Should i just copy and paste the parsing instead of trying to abstract them out into its own function?

browser/mapping.go
```diff
@@ -457,6 +465,39 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping {
...
+	for _, arg := range args {
+		pargs = append(pargs, arg.Export())
+	}
+
+	return js, popts, pargs, nil
+}
+
+// MapPage will map a page for an integration test. It is only to be used for
+// internal testing purposes.
+func MapPage(vu k6modules.VU, p *common.Page) mapping {
```
ankur22, line 493: Not sure about this either. It's only meant to for integration tests. WDYT?

browser/mapping.go
ka3de: I prefer to keep it this way, grouped in a separate function, instead of having it scattered around the mappings implementation. Also considering that this is called more than once, for `mapPage` and `mapFrame`.

browser/mapping.go
ankur22: A new method of mapping the desired types was used, which was already being used by `TestTracing`. See 80598a1f7fd47e91fc340fbdb0a589f68618e4b0, 97420ece4f36426df6932d509d7ace38d8a516e4 and 6c957279b9f08c986b9554d4e242af714b6ccfb3.

tests/page_test.go
```diff
@@ -636,63 +630,117 @@ func TestPageWaitForFunction(t *testing.T) {
...
+}
+
+func runJSInEventLoop(t *testing.T, vu *k6test.VU, js string) error {
+	t.Helper()
+
+	rt := vu.Runtime()
+	_, err := rt.RunString(js)
+	require.NoError(t, err)
+
+	test, ok := goja.AssertFunction(rt.Get("test"))
```
ka3de, line 733: It seems confusing that we are asserting on a `test` function that is not created inside the `runJSInEventLoop` function. We seem to rely on the caller for this, but no comment is written either.

Would it make sense to do it in a similar way as it's done in the [tracing test](https://github.com/grafana/xk6-browser/blob/5f754c2c2252896016fd77d1fd3234a795049a0e/tests/tracing_test.go#L182-L191), where the `test` variable is created in the function that sets that up in the runtime? I took a quick look and I think it is possible. Otherwise I would consider declaring this `runJSInEventLoop` as a local function inside the test and use it there.

tests/page_test.go
```diff
@@ -636,63 +630,117 @@ func TestPageWaitForFunction(t *testing.T) {
...
+	var log []string
+	require.NoError(t, rt.Set("log", func(s string) { log = append(log, s) }))
+
+	vu.ActivateVU()
+	vu.StartIteration(t)
+
+	return vu, rt, &log, func() { t.Helper(); vu.EndIteration(t) }
+}
+
+func runJSInEventLoop(t *testing.T, vu *k6test.VU, js string) error {
```
mstoykov, line 726: I would recommend using [modulestest](https://pkg.go.dev/go.k6.io/k6@v0.48.0/js/modulestest)

ankur22: We're using `modulestest`, what specifically from that package?

ankur22: Got it to work, thanks for the heads up. b244cf026fc7dea00505ac185829b90d21133db9

tests/page_test.go
ankur22: Good point, I was wondering the same, but for some reason didn't change it. I've refactored the js running on the eventloop to the suggestion in this [comment](https://github.com/grafana/xk6-browser/pull/1183#discussion_r1464922157).

tests/page_test.go
```diff
@@ -636,63 +631,100 @@ func TestPageWaitForFunction(t *testing.T) {
...
-		assert.Contains(t, log, "ok: null")
+		assert.Contains(t, *log, "ok: null")
 	})
 }
 
+// startIteration will work with the event system to start chrome and
+// (more importantly) set browser as the mapped browser instance which will
+// force all tests that work with this to go through the mapping layer.
+// This returns a cleanup function which should be deferred.
+func startIteration(t *testing.T) (*k6test.VU, *goja.Runtime, *[]string, func()) {
```
inancgumus, line 704: What about enforcing this for all tests in test browser? This might enable us to have more comprehensive integration tests. Or, maybe with or with this, both.

ka3de (approved):

tests/page_test.go
inancgumus: Sure, this idea justifies a separate PR definitely.

Issue #1168, opened 2024-01-19
Move goja runtime usage out of remote_object.go


ankur22:
# What

Move the usage of goja runtime out of [remote_object.go](https://github.com/grafana/xk6-browser/blob/5e38a3ffd9157974b300ab53b26d8ae3b899a71e/common/remote_object.go#L224-L230).

# Why

We should avoid working with the goja runtime from within the business logic (i.e. outside the mapping layer), because we could be running such code from off the main goja thread (a new goroutine or promise) and when multiple goroutines try to work with the goja runtime at the same time it can cause (avoidable) panics.
```[tasklist]
### Tasks
- [ ] https://github.com/grafana/xk6-browser/pull/1166
- [ ] https://github.com/grafana/xk6-browser/pull/1190
- [ ] https://github.com/grafana/k6/pull/3578
```

PR #1190, opened 2024-01-24
Refactor JSONValue to use parseConsoleRemoteObject

ankur22:
## What?

Refactor `JSONValue` to use `parseConsoleRemoteObject`.

## Why?

This will help avoid working with the goja runtime outside of the mapping layer and therefore avoid possible race conditions with the unsafe goja runtime. `JSONValue` wasn't ever working with the goja values and so it made sense to always work with string values.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1168

### DIFF
--- a/common/js_handle.go
+++ b/common/js_handle.go
@@ -165,16 +165,14 @@ func (h *BaseJSHandle) GetProperty(_ string) JSHandleAPI {
 func (h *BaseJSHandle) JSONValue() (string, error) {
 	remoteObject := h.remoteObject
 	if remoteObject.ObjectID != "" {
-		var result *runtime.RemoteObject
 		var err error
 		action := runtime.CallFunctionOn("function() { return this; }").
 			WithReturnByValue(true).
 			WithAwaitPromise(true).
 			WithObjectID(h.remoteObject.ObjectID)
-		if result, _, err = action.Do(cdp.WithExecutor(h.ctx, h.session)); err != nil {
+		if remoteObject, _, err = action.Do(cdp.WithExecutor(h.ctx, h.session)); err != nil {
 			return "", fmt.Errorf("retrieving json value: %w", err)
 		}
-		remoteObject = result
 	}
 
 	res, err := parseConsoleRemoteObject(h.logger, remoteObject)
--- a/common/js_handle.go
+++ b/common/js_handle.go
@@ -177,7 +177,7 @@ func (h *BaseJSHandle) JSONValue() (string, error) {
 
 	res, err := parseConsoleRemoteObject(h.logger, remoteObject)
 	if err != nil {
-		return "", fmt.Errorf("extracting json value: %w", err)
+		return "", fmt.Errorf("extracting json value (remote object id: %v): %w", remoteObject.ObjectID, err)
 	}
 
 	return res, nil
--- a/common/js_handle.go
+++ b/common/js_handle.go
@@ -2,7 +2,6 @@ package common
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 
 	"github.com/grafana/xk6-browser/k6ext"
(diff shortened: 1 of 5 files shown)


common/js_handle.go
```diff
@@ -191,10 +173,19 @@ func (h *BaseJSHandle) JSONValue() string { // TODO: return error
 		if result, _, err = action.Do(cdp.WithExecutor(h.ctx, h.session)); err != nil {
 			k6ext.Panic(h.ctx, "getting properties for JS handle: %w", err)
 		}
-		remoteObject = result
+		res, err := parseConsoleRemoteObject(h.logger, result)
+		if err != nil {
+			return "", fmt.Errorf("extracting value from result: %w", err)
+		}
+
+		return res, nil
```
inancgumus, line 181: This part was refactored for clarity. Can we save `result` to `remoteObject` and call `parseConsoleRemoteObject` once?

inancgumus (commented):
LGTM but one last thing.

common/js_handle.go
ka3de: I would think that makes sense. Not sure why that was not being done before, but I don't think that has any issues.
Also I noticed we are returning two different error messages (`value from result` | `value from remote object`), shouldn't these be the same as we are executing the same action?

inancgumus: Yes, I agree. We don't have to differentiate between these errors. If there was an error retrieving the data via `ObjectID`, there will already be a clarifying error.

common/js_handle.go
```diff
@@ -191,10 +173,19 @@ func (h *BaseJSHandle) JSONValue() string { // TODO: return error
 		if result, _, err = action.Do(cdp.WithExecutor(h.ctx, h.session)); err != nil {
 			k6ext.Panic(h.ctx, "getting properties for JS handle: %w", err)
 		}
```
inancgumus, line 175: Why do we panic here? Can we return an error?

ankur22: Good spot, that was missed, changed in 8815b4198a6883c06f3147e389f18d2316a61bab

common/js_handle.go
ankur22: I guess the two error messages made sense for debug purposes to help clarify which conditional branch the method was working on. The [refactor](https://github.com/grafana/xk6-browser/blob/f83f34fe646f8cb0f94733a1657c8e7e3da04b19/common/js_handle.go#L183-L197) makes more sense still. It has been added to b77858857d1accc0509bddbb6e6efe6d816f7ad3

common/js_handle.go
```diff
@@ -163,23 +162,7 @@ func (h *BaseJSHandle) GetProperty(_ string) JSHandleAPI {
...
-			if err != nil {
-				k6ext.Panic(h.ctx, "marshaling value: %w", err)
-			}
-			return string(buf)
-		}
-	}
+func (h *BaseJSHandle) JSONValue() (string, error) {
 	remoteObject := h.remoteObject
 	if remoteObject.ObjectID != "" {
 		var result *runtime.RemoteObject
```
inancgumus: Do we still need `result` here? What about setting the action's result to `remoteObject` directly?

common/js_handle.go
```diff
@@ -189,12 +172,17 @@ func (h *BaseJSHandle) JSONValue() string { // TODO: return error
...
-			k6ext.Panic(h.ctx, "getting properties for JS handle: %w", err)
+			return "", fmt.Errorf("retrieving json value: %w", err)
 		}
 		remoteObject = result
 	}
 
-	return convertToString(remoteObject)
+	res, err := parseConsoleRemoteObject(h.logger, remoteObject)
+	if err != nil {
+		return "", fmt.Errorf("extracting json value: %w", err)
```
inancgumus, line 180: The comment [here](https://github.com/grafana/xk6-browser/pull/1190#discussion_r1469291804) explains:
> I guess the two error messages made sense for debug purposes to help clarify which conditional branch the method was working on.

What about adding the object ID in the error so that we can see which path is taken?

```suggestion
		return "", fmt.Errorf("extracting json value (remote object id: %v): %w", remoteObject.ObjectID, err)
```

common/js_handle.go
ankur22: Refactored in dc0abe90fb32958389a4e143fbbb725140f60ad1

inancgumus: Closing per https://github.com/grafana/xk6-browser/pull/1190/commits/dc0abe90fb32958389a4e143fbbb725140f60ad1.

inancgumus (approved):
🚀

inancgumus (approved):

PR #1191, opened 2024-01-25
Add goja escape assertions

inancgumus:
## What?

- Adds assertions for goja value escapes to `ExecutionContext`.
- Marks the failing tests as skipped until we fix them all.

Note that this will be added to the `fix/goja-execctx-eval` branch. Then, we'll fix the failing tests in other PRs (See #1182). So it's safe to approve this.

## Why?

- To ensure that no Goja value escapes to `ExecutionContext`.

## Checklist

- [x] I have performed a self-review of my code
- [x] I have added tests for my changes
- [x] I have commented on my code, particularly in hard-to-understand areas

## Related PR(s)/Issue(s)

Updates: #1182

Issue #1182, opened 2024-01-22
Remove Goja from Execution Context Eval


ankur22:
# What

`ExecutionContext.Eval` is at the heart of the module. Removing its Goja dependency will tremendously detach goja from most of the business logic. [executionContext.eval](https://github.com/grafana/xk6-browser/blob/main/common/execution_context.go#L155-L157) takes a slice argument, which can be of any type, including `goja.Value`. Find a way to enforce a change where the slice must be anything BUT a `goja.Value`.

# Why

- Improved stability: Enable avoiding working with the non-thread safe Goja from multiple goroutines.
- Productivity: Business logic can use native Go constructs. Code is simpler and easy to reason about.
- Testability: Testing the business logic becomes straightforward.

# How

- Remove `goja` from `Eval` and all related methods.
- Ensure that none of the methods in the module passes goja values to `Eval` and friends.
- Add assertions and keep them there for a few cycles.

# Related

#1162

```[tasklist]
### Tasks
- [ ] https://github.com/grafana/xk6-browser/issues/1170
- [ ] https://github.com/grafana/xk6-browser/pull/1189
- [ ] https://github.com/grafana/xk6-browser/pull/1191
- [ ] https://github.com/grafana/xk6-browser/pull/1193
- [ ] https://github.com/grafana/xk6-browser/issues/1168
- [ ] https://github.com/grafana/xk6-browser/pull/1190
- [x] What about the other APIs that still use goja? None of them escape Goja as it seems. (see https://github.com/grafana/xk6-browser/issues/1064)
- [x] Release notes: https://github.com/grafana/k6/pull/3578
```

inancgumus:
Wrapping up this one 🎉

PR #1193, opened 2024-01-25
Fix goja escape from DispatchEvents

inancgumus:
## What?

All `DispatchEvent` methods now:
- Return errors.
- Moves their option parsing to the mapping layer.
- Prevent passing down Goja values.

Fixes tests:
- TestLocator
- TestBrowserOptionsSlowMo
- The goja escape was related to `DispatchEvent`s.

## Why?

- This is a necessary fix to remove passing Goja values to `ExecutionContext`.
- See https://github.com/grafana/xk6-browser/pull/1191.
- See also #1182.

## Checklist

## Related PR(s)/Issue(s)

Updates: #1182

### DIFF
--- a/tests/locator_test.go
+++ b/tests/locator_test.go
@@ -26,7 +26,6 @@ type jsFrameWaitForSelectorOpts struct {
 
 func TestLocator(t *testing.T) {
 	t.Parallel()
-	t.Skip("TODO: fix goja escape")
 
 	tests := []struct {
 		name string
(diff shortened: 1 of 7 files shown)


tests/locator_test.go
```diff
@@ -91,7 +90,11 @@ func TestLocator(t *testing.T) {
 					return asBool(t, v)
 				}
 				require.False(t, result(), "should not be clicked first")
-				p.Locator("#link", nil).DispatchEvent("click", tb.toGojaValue("mouseevent"), nil)
+				opts := &common.FrameDispatchEventOptions{
```
ankur22, line 93: Nit: Could you have used `common.NewFrameDispatchEventOptions()` instead of declaring the options like this?

ankur22 (approved):
LGTM.

I left a nit for you to consider, but not important.

tests/locator_test.go
inancgumus: Sure, I thought about that. But it requires setting a timeout by the constructor. Since other tests don't use a timeout (and do it in the timeout tests below), I wanted to keep empty. WDYT?

ankur22: No strong opinion. My concern is whether it will work for everyone if the timeout is `0`.

inancgumus: You mean setting a `0` timeout using the constructor? It's been working so far that way. And the current code also does that.

ankur22: Yeah, it'll be defaulting to `0`. I'm happy with how it is now, it's not crucial to use the New method.

inancgumus: Done: [eca3338](https://github.com/grafana/xk6-browser/pull/1193/commits/eca33389a91ddaac3407df43745f5587eb3e3fc9).

Issue #1155, opened 2024-01-16
Define an interface for local and remote storage


ankur22:
### What?

Define an interface that exports a persist method, abstracting away how or where the file is being persisted.

### Why?

The browser module can take screenshots. These screenshots are currently either:

1. Persisted to the local disk relative to the k6 binary;
2. Returned to the user in the test script for them to do something with the binary data.

When a user works with the first option, we want the browser module to either:

1. persist the file to the local disk;
2. or upload it to remote location.

The interface needs to be generic enough to allow for both these cases, and it's only exported method should be to persist a blob of data.

### How?

Define an interface that takes:

1. the `context`;
2. the file path;
3. the stream of data.

which will satisfy the requirement of abstracting away the how and where the file is being saved.

It is a good idea to generalise the problem (use `io.writer`) to a blob of data even though the main use case is to store the screenshots so that we can stream the data either to disk or a remote location.

### Tasks

```[tasklist]
### Tasks
- [ ] PR
- [ ] Update the k6 release notes
```

### Related PR(s)/Issue(s)

Take a look at https://github.com/grafana/k6/issues/3017 for some more guidance.

ankur22:
The interface implementation work has been moved to https://github.com/grafana/xk6-browser/issues/1160.

PR #1194, opened 2024-01-25
Add file persister interface

ankur22:
## What?

This adds a file persister interface.

## Why?

It is a way to hide the details of where and how the files that the browser module saves to disk (and soon remotely).

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1155

### DIFF
--- a/storage/file_persister.go
+++ b/storage/file_persister.go
@@ -1,0 +1,12 @@
+package storage
+
+import (
+	"context"
+	"io"
+)
+
+// FilePersister will persist files. It abstracts away the where and how of
+// writing files to the source destination.
+type FilePersister interface {
+	Persist(ctx context.Context, path string, data io.Reader) error
+}
--- a/browser/modulevu.go
+++ b/browser/modulevu.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/grafana/xk6-browser/common"
 	"github.com/grafana/xk6-browser/k6ext"
+	"github.com/grafana/xk6-browser/storage"
 
 	k6modules "go.k6.io/k6/js/modules"
 )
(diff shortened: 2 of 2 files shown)


storage/file_persister.go
```diff
@@ -0,0 +1,14 @@
...
+	"context"
+	"io"
+
+	"github.com/grafana/xk6-browser/log"
+)
+
+// FilePersister will persist files. It abstracts away the where and how of
+// writing files to the source destination.
+type FilePersister interface {
+	Persist(ctx context.Context, logger *log.Logger, path string, data io.Reader) error
```
inancgumus, line 13: Why do we need an interface here? Since Go interfaces are explicit, we could declare it on the call side.

Besides, the signature is too fat. A value implementing this interface can use its field logger. The rest seems fine. Will we have a `path` on the cloud? Let's first define the concrete types and, finally, the interface.

Check [this](https://go.dev/wiki/CodeReviewComments#interfaces) for more information.

ankur22: Yep, I agree that this is the wrong way around in go. I set up the sub issues in such a way so that the local and remote persister could be done in parallel hence the initial PR to surface the interface that could be used by both. In hindsight that won't be how i do it next time since it's tricky to that without the concrete type.

> Besides, the signature is too fat. A value implementing this interface can use its field logger
I agree, and I don't like the addition of the `logger` as part of the method signiture. Take a look at this (unfinished) [PR](https://github.com/grafana/xk6-browser/pull/1197) on how it is instantiated and used. I don't think there's a way to instantiate it with a logger when the `ModuleInstance` is created. I could be wrong though.

> Will we have a path on the remote?
Yes, we will.

ankur22: The logger has been removed in de5f6bec23fe0c693db6f430981e2dd95a0d476a

ankur22:
Closing this and instead working off of https://github.com/grafana/xk6-browser/pull/1197

Issue #1156, opened 2024-01-16
Implement a local disk persister


ankur22:
### What?

Implement a concrete type that will satisfy the interface defined by https://github.com/grafana/xk6-browser/issues/1155, and will match the current behaviour of the browser module when a screenshot is saved to the local disk.

### Why?

To enable us to be be able to either persist the screenshot to the local disk or to upload to a remote location, we need to abstract away the details from the current [implementation](https://github.com/grafana/xk6-browser/blob/c0584af38c6ecaa29aa0c8287cd2daa4de3aeb5b/common/screenshotter.go#L222-L232) so that a concrete type can be used which could persist in either one of those two locations.

### How?

There currently is no API in k6 that allows files to be persisted to disk, so we should implement it the exact same way as we are doing now. Don't worry about making async as suggested in the [current](https://github.com/grafana/xk6-browser/blob/c0584af38c6ecaa29aa0c8287cd2daa4de3aeb5b/common/screenshotter.go#L223) comment, but leave the comment in so that we can work on it later if we need.

### Tasks

```[tasklist]
### Tasks
- [ ] https://github.com/grafana/xk6-browser/pull/1197
- [ ] https://github.com/grafana/k6/pull/3600
```

### Related PR(s)/Issue(s)

Prerequisite issue: https://github.com/grafana/xk6-browser/issues/1155

PR #1195, opened 2024-01-25
Add local file persister

ankur22:
## What?

This adds a local file persister which will save files to the local disk. It implements the interface defined in https://github.com/grafana/xk6-browser/issues/1155.

## Why?

So that we can replace how we're currently writing screenshots to disk. This will help hide the details of where and how the files are saved, allowing us to easily work with different file persisters.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1156

### DIFF
--- a/storage/file_persister.go
+++ b/storage/file_persister.go
@@ -2,11 +2,60 @@ package storage
 
 import (
 	"context"
+	"fmt"
 	"io"
+	"os"
+	"path/filepath"
 )
 
 // FilePersister will persist files. It abstracts away the where and how of
 // writing files to the source destination.
 type FilePersister interface {
 	Persist(ctx context.Context, path string, data io.Reader) error
+}
+
+// LocalFilePersister will persist files to the local disk.
+type LocalFilePersister struct{}
+
+// Persist will write the contents of data to the local disk on the specified path.
+// TODO: we should not write to disk here but put it on some queue for async disk writes.
+func (l *LocalFilePersister) Persist(_ context.Context, path string, data io.Reader) error {
+	cp := filepath.Clean(path)
+
+	dir := filepath.Dir(cp)
+	if err := os.MkdirAll(dir, 0o755); err != nil {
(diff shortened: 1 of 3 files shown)


storage/file_persister.go
```diff
@@ -12,3 +15,49 @@ import (
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return fmt.Errorf("reading from the reader: %w", err)
+		}
+	}
+
+	return nil
+}
```
inancgumus, line 63: Why don't we just use `io.Copy` here?

storage/file_persister.go
```diff
@@ -12,3 +15,49 @@ import (
...
+		return fmt.Errorf("creating a local directory %q: %w", dir, err)
+	}
+
+	f, err := os.OpenFile(cp, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600)
+	if err != nil {
+		return fmt.Errorf("creating a local file %q: %w", cp, err)
+	}
+	defer func() {
+		if err := f.Close(); err != nil {
+			logger.Errorf("LocalFilePersister:Persist", "closing the local file %q: %v", cp, err)
```
inancgumus, line 38: I think we can handle this also from the call site if we return an error here. So we don't have to pass a logger here. The call site can log the error if it sees it's a file close error. WDYT?

storage/file_persister.go
ankur22: I have no idea how i missed that 😆. Good suggestion, added to be155a55e95efa05b7195b81d99c2cd717d54c91

storage/file_persister.go
ankur22: Yeah, that's a nice idea! Added it to cfa2726d577b1a26a0233ba5a10dd540ded70ffe

In which case the `logger` can be removed 🎉

ankur22:
Closing this and moving this to https://github.com/grafana/xk6-browser/pull/1197

Issue #1157, opened 2024-01-16
Refactor screenshot to use local persister


ankur22:
### What?

Refactor the [current implementation](https://github.com/grafana/xk6-browser/blob/c0584af38c6ecaa29aa0c8287cd2daa4de3aeb5b/common/screenshotter.go#L222-L232) to persist screenshots to the local disk.

### Why?

This will help in two ways:

1. Ensure that the concrete type that was implemented in https://github.com/grafana/xk6-browser/issues/1156 behaves as it should;
2. Move us closer to being able to work with any type that satisfies the interface in https://github.com/grafana/xk6-browser/issues/1155.

### How?

Do the refactor by working with the interface defined by https://github.com/grafana/xk6-browser/issues/1155 and the concrete type defined by https://github.com/grafana/xk6-browser/issues/1156.

### Tasks

```[tasklist]
### Tasks
- [ ] https://github.com/grafana/xk6-browser/pull/1197
- [ ] https://github.com/grafana/k6/pull/3600
```

### Related PR(s)/Issue(s)

Prerequisite issue: https://github.com/grafana/xk6-browser/issues/1156

PR #1197, opened 2024-01-29
Add a localPersister

ankur22:
## What?

This change adds a new `LocalPersister` which will replace how the `screenshotter` writes to the local disk.

## Why?

This is the first step in removing the details of how and where the screenshot is being saved, allowing us to work with either a `LocalPersister` or (yet to be implemented) `RemotePersister`.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1155, https://github.com/grafana/xk6-browser/issues/1156 and https://github.com/grafana/xk6-browser/issues/1157.

### DIFF
--- a/common/element_handle.go
+++ b/common/element_handle.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/grafana/xk6-browser/common/js"
 	"github.com/grafana/xk6-browser/k6ext"
+	"github.com/grafana/xk6-browser/storage"
 )
 
 const resultDone = "done"
@@ -1176,7 +1177,8 @@ func (h *ElementHandle) setChecked(apiCtx context.Context, checked bool, p *Posi
 	return nil
 }
 
-func (h *ElementHandle) Screenshot(opts goja.Value) goja.ArrayBuffer {
+// Screenshot will instruct Chrome to save a screenshot of the current element and save it to specified file.
+func (h *ElementHandle) Screenshot(opts goja.Value, fp *storage.LocalFilePersister) goja.ArrayBuffer {
 	spanCtx, span := TraceAPICall(
 		h.ctx,
 		h.frame.page.targetID.String(),
@@ -1191,7 +1193,7 @@ func (h *ElementHandle) Screenshot(opts goja.Value) goja.ArrayBuffer {
 	}
 	span.SetAttributes(attribute.String("screenshot.path", parsedOpts.Path))
 
-	s := newScreenshotter(spanCtx)
+	s := newScreenshotter(spanCtx, fp)
 	buf, err := s.screenshotElement(h, parsedOpts)
 	if err != nil {
 		k6ext.Panic(h.ctx, "taking screenshot: %w", err)
--- a/common/page.go
+++ b/common/page.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/grafana/xk6-browser/k6ext"
 	"github.com/grafana/xk6-browser/log"
+	"github.com/grafana/xk6-browser/storage"
 
 	k6modules "go.k6.io/k6/js/modules"
 )
@@ -1120,7 +1121,7 @@ func (p *Page) Route(url goja.Value, handler goja.Callable) {
 }
 
 // Screenshot will instruct Chrome to save a screenshot of the current page and save it to specified file.
-func (p *Page) Screenshot(opts goja.Value) goja.ArrayBuffer {
+func (p *Page) Screenshot(opts goja.Value, fp *storage.LocalFilePersister) goja.ArrayBuffer {
 	spanCtx, span := TraceAPICall(p.ctx, p.targetID.String(), "page.screenshot")
 	defer span.End()
 
@@ -1130,7 +1131,7 @@ func (p *Page) Screenshot(opts goja.Value) goja.ArrayBuffer {
 	}
 	span.SetAttributes(attribute.String("screenshot.path", parsedOpts.Path))
 
-	s := newScreenshotter(spanCtx)
+	s := newScreenshotter(spanCtx, fp)
 	buf, err := s.screenshotPage(p, parsedOpts)
 	if err != nil {
 		k6ext.Panic(p.ctx, "capturing screenshot: %w", err)
--- a/common/screenshotter.go
+++ b/common/screenshotter.go
@@ -7,14 +7,14 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"os"
-	"path/filepath"
 	"strings"
 
 	"github.com/chromedp/cdproto/cdp"
 	"github.com/chromedp/cdproto/emulation"
 	cdppage "github.com/chromedp/cdproto/page"
 	"github.com/dop251/goja"
+
+	"github.com/grafana/xk6-browser/storage"
 )
 
 // ImageFormat represents an image file format.
@@ -61,11 +61,12 @@ func (f *ImageFormat) UnmarshalJSON(b []byte) error {
 }
 
 type screenshotter struct {
-	ctx context.Context
+	ctx       context.Context
+	persister *storage.LocalFilePersister
 }
 
-func newScreenshotter(ctx context.Context) *screenshotter {
-	return &screenshotter{ctx}
+func newScreenshotter(ctx context.Context, fp *storage.LocalFilePersister) *screenshotter {
+	return &screenshotter{ctx, fp}
 }
 
 func (s *screenshotter) fullPageSize(p *Page) (*Size, error) {
@@ -217,12 +218,8 @@ func (s *screenshotter) screenshot(
 	// Save screenshot capture to file
 	// TODO: we should not write to disk here but put it on some queue for async disk writes
 	if path != "" {
-		dir := filepath.Dir(path)
-		if err := os.MkdirAll(dir, 0o755); err != nil {
-			return nil, fmt.Errorf("creating screenshot directory %q: %w", dir, err)
-		}
-		if err := os.WriteFile(path, buf, 0o644); err != nil {
-			return nil, fmt.Errorf("saving screenshot to %q: %w", path, err)
+		if err := s.persister.Persist(path, bytes.NewBuffer(buf)); err != nil {
+			return nil, fmt.Errorf("persisting screenshot: %w", err)
 		}
 	}
 
--- a/common/screenshotter.go
+++ b/common/screenshotter.go
@@ -216,7 +216,6 @@ func (s *screenshotter) screenshot(
 	}
 
 	// Save screenshot capture to file
-	// TODO: we should not write to disk here but put it on some queue for async disk writes
 	if path != "" {
 		if err := s.persister.Persist(path, bytes.NewBuffer(buf)); err != nil {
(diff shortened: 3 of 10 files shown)


common/screenshotter.go
```diff
@@ -61,11 +61,12 @@ func (f *ImageFormat) UnmarshalJSON(b []byte) error {
 }
 
 type screenshotter struct {
-	ctx context.Context
+	ctx       context.Context
+	persister storage.FilePersister
```
inancgumus, line 65: Great work on the PR! Just a thought – how about we kick off by implementing our local and remote persisters before declaring the `persister` interface and adjusting the `screenshotter`? This way, we can avoid redoing stuff and streamline our PR process.

If that plan doesn't fit, what if we introduce the interface right here instead of using `FilePersister`? Since `screenshotter` will be switching between different implementations, placing the interface in the `common` package seems like a smart move. This would also make swapping implementations in our mapping layer and tests easier.

What do you think?

ankur22: I've made the change so that it is a stand alone type. The interface will come later when the remotePersister is finalised.

storage/file_persister.go
```diff
@@ -0,0 +1,38 @@
...
+		tempErr := f.Close()
+		// Only return the close error if there isn't already an existing error.
+		if tempErr != nil && err == nil {
+			err = fmt.Errorf("closing the local file %q: %w", cp, tempErr)
+		}
+	}()
+
+	_, err = io.Copy(f, data)
+
+	return
```
inancgumus, line 37: It's better to use `err` here for clarity, even though we don't need to do it.

storage/file_persister_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			dir, err := os.MkdirTemp("", "*")
+			require.NoError(t, err)
+			t.Cleanup(func() { _ = os.RemoveAll(dir) })
```
inancgumus, line 50: We can use [t.TempDir](https://pkg.go.dev/testing#T.TempDir) here so we don't have to clean it up ourselves.

```suggestion
			dir := t.TempDir()
```

storage/file_persister_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+			// We want to make sure that the persister truncates the existing
+			// data and therefore overwrites existing data. This sets up a file
+			// with some existing data that should be overwritten.
+			if tt.truncates {
+				err = os.WriteFile(p, []byte(tt.existingData), 0o600)
+				require.NoError(t, err)
+			}
+
+			l := &LocalFilePersister{}
```
inancgumus, line 61: Is this clearer?

```suggestion
			var l LocalFilePersister
```

storage/file_persister_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+			f, err := os.Open(filepath.Clean(p))
+			require.NoError(t, err)
+			defer func() {
+				err = f.Close()
+				require.NoError(t, err)
+			}()
+
+			bb, err := io.ReadAll(f)
+			require.NoError(t, err)
```
inancgumus, line 77: Why not just?

```suggestion
			bb, err := os.ReadFile(filepath.Clean(p))
			require.NoError(t, err)
```

storage/file_persister.go
```diff
@@ -0,0 +1,38 @@
...
+	f, err := os.OpenFile(cp, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600)
+	if err != nil {
+		return fmt.Errorf("creating a local file %q: %w", cp, err)
+	}
+	defer func() {
+		tempErr := f.Close()
+		// Only return the close error if there isn't already an existing error.
+		if tempErr != nil && err == nil {
+			err = fmt.Errorf("closing the local file %q: %w", cp, tempErr)
+		}
```
inancgumus, line 32: A suggestion, but no strong opinions.

```suggestion
		if cerr := f.Close(); cerr != nil && err == nil {
			err = fmt.Errorf("closing the local file %q: %w", cp, cerr)
		}
```

common/screenshotter.go
```diff
@@ -215,14 +216,9 @@ func (s *screenshotter) screenshot(
...
 	// Save screenshot capture to file
-	// TODO: we should not write to disk here but put it on some queue for async disk writes
 	if path != "" {
-		dir := filepath.Dir(path)
-		if err := os.MkdirAll(dir, 0o755); err != nil {
-			return nil, fmt.Errorf("creating screenshot directory %q: %w", dir, err)
-		}
-		if err := os.WriteFile(path, buf, 0o644); err != nil {
-			return nil, fmt.Errorf("saving screenshot to %q: %w", path, err)
+		if err := s.persister.Persist(path, bytes.NewBuffer(buf)); err != nil {
```
inancgumus, line 220: I suggest wrapping this within a `bufio.Reader` for better performance, especially important for large files.

common/page.go
```diff
@@ -1120,7 +1121,7 @@ func (p *Page) Route(url goja.Value, handler goja.Callable) {
 }
 
 // Screenshot will instruct Chrome to save a screenshot of the current page and save it to specified file.
-func (p *Page) Screenshot(opts goja.Value) goja.ArrayBuffer {
+func (p *Page) Screenshot(opts goja.Value, persister *storage.LocalFilePersister) goja.ArrayBuffer {
```
inancgumus, line 1124: Should we just say `fp` or `lfp` since the type makes it clear that it's persister? No strong opinion.

common/element_handle.go
```diff
@@ -1176,7 +1177,8 @@ func (h *ElementHandle) setChecked(apiCtx context.Context, checked bool, p *Posi
 	return nil
 }
 
-func (h *ElementHandle) Screenshot(opts goja.Value) goja.ArrayBuffer {
+// Screenshot will instruct Chrome to save a screenshot of the current element and save it to specified file.
+func (h *ElementHandle) Screenshot(opts goja.Value, filePersister *storage.LocalFilePersister) goja.ArrayBuffer {
```
inancgumus, line 1181: Should we just say `fp` or `lfp` since the type makes it clear that it's persister? No strong opinion.

common/screenshotter.go
inancgumus: 👍

common/screenshotter.go
ankur22: Yeah, i was wondering the same, and I was going to add it to the `LocalPersister.Persist` method itself (it feels like that should decide the best course of action on how to buffer things). Actually maybe it's better to have a `bufio.Writer` since the screenshot bytes are already in memory, so we want to buffer the write to the disk.

common/element_handle.go
ankur22: Resolved in 6868b5d16deb867fd64483336a1b6541c617c3cf

common/page.go
ankur22: Resolved in 6868b5d16deb867fd64483336a1b6541c617c3cf

storage/file_persister.go
ankur22: resolved in 661d2494a21f1f7e0821cc044fba0e6bdae46a6b

storage/file_persister.go
ankur22: Resolved in 661d2494a21f1f7e0821cc044fba0e6bdae46a6b

storage/file_persister_test.go
ankur22: Resolved in 8d21a696a6dbba7f779a64ff0f5b8d9f2761dab0

storage/file_persister_test.go
ankur22: Resolved in 8d21a696a6dbba7f779a64ff0f5b8d9f2761dab0

common/screenshotter.go
ankur22: Added it to 897862ba11263a6edc1793f03c73892beaa2dbd1. Actually it probably doesn't matter if it's a `bufio.Reader` or `bufio.Writer` as long as it writes a small amount to disk to avoid slowing down the app with large IO.

inancgumus: 👍 Using `bufio.Writer` in the `Copy` makes sense.

storage/file_persister.go
```diff
@@ -31,10 +31,15 @@ func (l *LocalFilePersister) Persist(path string, data io.Reader) (err error) {
...
 
-	_, err = io.Copy(bufio.NewWriter(f), data)
-	if err != nil {
+	bf := bufio.NewWriter(f)
+
+	if _, err := io.Copy(bf, data); err != nil {
 		return fmt.Errorf("copying data to file: %w", err)
 	}
 
+	if err := bf.Flush(); err != nil {
```
ankur22, line 40: Test were failing since I forgot to flush the data.

inancgumus (dismissed):
🚀

common/screenshotter.go
```diff
@@ -61,11 +61,12 @@ func (f *ImageFormat) UnmarshalJSON(b []byte) error {
...
 
 type screenshotter struct {
-	ctx context.Context
+	ctx       context.Context
+	persister *storage.LocalFilePersister
 }
 
-func newScreenshotter(ctx context.Context) *screenshotter {
-	return &screenshotter{ctx}
+func newScreenshotter(ctx context.Context, persister *storage.LocalFilePersister) *screenshotter {
```
inancgumus, line 68: Ah, one more `persister` here. Just noticed.

common/screenshotter.go
```diff
@@ -215,14 +216,9 @@ func (s *screenshotter) screenshot(
...
-	// TODO: we should not write to disk here but put it on some queue for async disk writes
 	if path != "" {
-		dir := filepath.Dir(path)
-		if err := os.MkdirAll(dir, 0o755); err != nil {
-			return nil, fmt.Errorf("creating screenshot directory %q: %w", dir, err)
-		}
-		if err := os.WriteFile(path, buf, 0o644); err != nil {
-			return nil, fmt.Errorf("saving screenshot to %q: %w", path, err)
+		if err := s.persister.Persist(path, bytes.NewBuffer(buf)); err != nil {
+			return nil, fmt.Errorf("saving screenshot")
```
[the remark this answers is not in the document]
ankur22: Ah, very good catch! Resolved in 39d2db9dabffe683fd59362178761634f1eedd0e

common/screenshotter.go
ankur22: Resolved in 39d2db9dabffe683fd59362178761634f1eedd0e

common/screenshotter.go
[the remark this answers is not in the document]
inancgumus: `saving` or `persisting` since the method name is `Persist` :) Although "saving" is more user friendly.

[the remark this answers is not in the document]
ankur22: I'm going with `persist` for now, but we can refactor this later if we need.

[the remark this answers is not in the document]
ankur22: Resolved in 70cd67a985d480abadadc0939020812aa809873c

inancgumus (approved):

PR #1198, opened 2024-01-29
Refactor goja out of screenshots and byte pointer changes

ankur22:
## What?

This refactors the parsing of the goja options for `frame.screenshot` and `page.screenshot` to the mapping layer, as well as moving goja out of screenshotter. This means that the screenshot business logic is free of goja.

It also changes the `screenshotter` to return a `[]byte` instead of a pointer to it (`*[]byte`).

## Why?

The options should be parsed out of the `goja.Value` representation in the mapping layer to avoid goja escaping into the business logic which makes the business logic more complex and this will also help if we convert this method into async.

There's no reason to return a `*[]byte` as far as i can tell.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1064

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -232,6 +232,7 @@ func mapJSHandle(vu moduleVU, jsh common.JSHandleAPI) mapping {
 //
 //nolint:funlen
 func mapElementHandle(vu moduleVU, eh *common.ElementHandle) mapping {
+	rt := vu.Runtime()
 	maps := mapping{
 		"boundingBox": eh.BoundingBox,
 		"check":       eh.Check,
@@ -280,8 +281,22 @@ func mapElementHandle(vu moduleVU, eh *common.ElementHandle) mapping {
 			return mapFrame(vu, f), nil
 		},
 		"press": eh.Press,
-		"screenshot": func(opts goja.Value) goja.ArrayBuffer {
-			return eh.Screenshot(opts, vu.LocalFilePersister)
+		"screenshot": func(opts goja.Value) (*goja.ArrayBuffer, error) {
+			ctx := vu.Context()
+
+			popts := common.NewElementHandleScreenshotOptions(eh.Timeout())
+			if err := popts.Parse(ctx, opts); err != nil {
+				return nil, fmt.Errorf("parsing frame screenshot options: %w", err)
+			}
+
+			bb, err := eh.Screenshot(popts, vu.LocalFilePersister)
+			if err != nil {
+				return nil, err //nolint:wrapcheck
+			}
+
+			ab := rt.NewArrayBuffer(bb)
+
+			return &ab, nil
 		},
 		"scrollIntoViewIfNeeded": eh.ScrollIntoViewIfNeeded,
 		"selectOption":           eh.SelectOption,
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -704,8 +704,22 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
 			return rt.ToValue(r).ToObject(rt)
 		},
 		"route": p.Route,
-		"screenshot": func(opts goja.Value) goja.ArrayBuffer {
-			return p.Screenshot(opts, vu.LocalFilePersister)
+		"screenshot": func(opts goja.Value) (*goja.ArrayBuffer, error) {
+			ctx := vu.Context()
+
+			popts := common.NewPageScreenshotOptions()
+			if err := popts.Parse(ctx, opts); err != nil {
+				return nil, fmt.Errorf("parsing page screenshot options: %w", err)
+			}
+
+			bb, err := p.Screenshot(popts, vu.LocalFilePersister)
+			if err != nil {
+				return nil, err //nolint:wrapcheck
+			}
+
+			ab := rt.NewArrayBuffer(bb)
+
+			return &ab, nil
 		},
 		"selectOption":                p.SelectOption,
 		"setContent":                  p.SetContent,
--- a/common/screenshotter.go
+++ b/common/screenshotter.go
@@ -12,7 +12,6 @@ import (
 	"github.com/chromedp/cdproto/cdp"
 	"github.com/chromedp/cdproto/emulation"
 	cdppage "github.com/chromedp/cdproto/page"
-	"github.com/dop251/goja"
 
 	"github.com/grafana/xk6-browser/storage"
 )
@@ -121,12 +120,14 @@ func (s *screenshotter) originalViewportSize(p *Page) (*Size, *Size, error) {
 	if err != nil {
 		return nil, nil, fmt.Errorf("getting viewport dimensions: %w", err)
 	}
-	r, ok := result.(goja.Object)
-	if !ok {
-		return nil, nil, fmt.Errorf("cannot convert to goja object: %w", err)
+
+	var returnVal Size
+	if err := convert(result, &returnVal); err != nil {
+		return nil, nil, fmt.Errorf("unpacking window size: %w", err)
 	}
-	viewportSize.Width = r.Get("width").ToFloat()
-	viewportSize.Height = r.Get("height").ToFloat()
+
+	viewportSize.Width = returnVal.Width
+	viewportSize.Height = returnVal.Height
 
 	return &viewportSize, &originalViewportSize, nil
 }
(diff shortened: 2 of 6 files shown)


inancgumus (commented):
This is great! If it isn't too much work for you, could we detach the file persisting logic out of this PR and make this PR focus only on de-gojafying? I believe adding the file persisting logic should be straightforward after this. Also, we'll make the `Screenshot` accept an interface later. This way, we can avoid double work, and it'll also help us keep the commit history cleaner. Thanks.

ankur22:
Yep, this was done after the persister work. I believe this PR is separate from the Persister work, although it does depend on it unfortunately. I'd rather keep this as is and merge later once the persister PRs are merged into `main`.

browser/mapping.go
```diff
@@ -280,8 +280,15 @@ func mapElementHandle(vu moduleVU, eh *common.ElementHandle) mapping {
...
-			return eh.Screenshot(opts, vu.LocalFilePersister)
+		"screenshot": func(opts goja.Value) ([]byte, error) {
+			ctx := vu.Context()
+
+			popts := common.NewElementHandleScreenshotOptions(eh.Timeout())
+			if err := popts.Parse(ctx, opts); err != nil {
+				return nil, fmt.Errorf("parsing frame screenshot options: %w", err)
+			}
+
+			return eh.Screenshot(popts, vu.LocalFilePersister) //nolint:wrapcheck
```
inancgumus, line 291: Should we return an `ArrayBuffer` here? I remember that transferring a large number of bytes can be problematic. WDYT @mstoykov?

common/screenshotter.go
```diff
@@ -121,12 +120,23 @@ func (s *screenshotter) originalViewportSize(p *Page) (*Size, *Size, error) {
...
+	if !ok {
+		return nil, nil, fmt.Errorf("converting viewport size width %q: %w", r["width"], err)
+	}
+	viewportSize.Width = width
+
+	height, ok := r["height"].(float64)
+	if !ok {
+		return nil, nil, fmt.Errorf("converting viewport size height %q: %w", r["height"], err)
+	}
+	viewportSize.Height = height
```
inancgumus, line 139: We can simply use the `convert` helper here.

common/screenshotter.go
```diff
@@ -276,13 +287,24 @@ func (s *screenshotter) screenshotElement(h *ElementHandle, opts *ElementHandleS
...
+	if !ok {
+		return nil, fmt.Errorf("converting scroll offset x %q: %w", so["x"], err)
+	}
+	documentRect.X += x
+
+	y, ok := so["y"].(float64)
+	if !ok {
+		return nil, fmt.Errorf("converting scroll offset y %q: %w", so["y"], err)
 	}
+	documentRect.Y += y
```
inancgumus, line 307: Same. We can simply use the `convert` helper here.

inancgumus (commented):
This is nice 🎉 Some suggestions.

browser/mapping.go
ankur22: Right, good spot. I forgot to make a note here to highlight the change. I'm going to run some tests to see how this may have affect the buffer that is returned.

ankur22: I ran some code in PW and k6 with screenshot and they were printing different results (PW was printing unreadable values such as `u��I���*` and k6 browser was printing numbers such as `9,148,148,164,`). I'm not entirely sure how to work with the buffer either. I think it's safer to change it to work with `ArrayBuffer` and keep the existing behaviour for now and then come back to fully grasping the buffer at a later stage 👍

ankur22: Resolved in 59ce5ab508a2cf0ce3d40feb71e0152aeb3c898f and 0a093fdb5ca8f8e9398505f59172153bfb4ee5c1

browser/mapping.go
```diff
@@ -280,8 +281,22 @@ func mapElementHandle(vu moduleVU, eh *common.ElementHandle) mapping {
...
+			ctx := vu.Context()
+
+			popts := common.NewElementHandleScreenshotOptions(eh.Timeout())
+			if err := popts.Parse(ctx, opts); err != nil {
+				return nil, fmt.Errorf("parsing frame screenshot options: %w", err)
+			}
+
+			bb, err := eh.Screenshot(popts, vu.LocalFilePersister)
+			if err != nil {
+				return &goja.ArrayBuffer{}, err //nolint:wrapcheck
```
inancgumus, line 294: What happens when return `nil` here?

```suggestion
				return nil, err //nolint:wrapcheck
```

browser/mapping.go
```diff
@@ -689,8 +704,22 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
...
 		},
 		"route": p.Route,
-		"screenshot": func(opts goja.Value) goja.ArrayBuffer {
-			return p.Screenshot(opts, vu.LocalFilePersister)
+		"screenshot": func(opts goja.Value) (*goja.ArrayBuffer, error) {
+			ctx := vu.Context()
+
+			popts := common.NewPageScreenshotOptions()
+			if err := popts.Parse(ctx, opts); err != nil {
+				return &goja.ArrayBuffer{}, fmt.Errorf("parsing page screenshot options: %w", err)
```
inancgumus, line 712: Also, here: `nil`?

browser/mapping.go
```diff
@@ -689,8 +704,22 @@ func mapPage(vu moduleVU, p *common.Page) mapping {
...
+			}
+
+			bb, err := p.Screenshot(popts, vu.LocalFilePersister)
+			if err != nil {
+				return &goja.ArrayBuffer{}, err //nolint:wrapcheck
```
inancgumus, line 717: `nil`?

common/screenshotter.go
```diff
@@ -121,12 +121,14 @@ func (s *screenshotter) originalViewportSize(p *Page) (*Size, *Size, error) {
 	if err != nil {
 		return nil, nil, fmt.Errorf("getting viewport dimensions: %w", err)
 	}
-	r, ok := result.(goja.Object)
-	if !ok {
-		return nil, nil, fmt.Errorf("cannot convert to goja object: %w", err)
+
+	var returnVal Size
+	if err := convert(result, &returnVal); err != nil {
+		k6ext.Panic(s.ctx, "unpacking selected options: %w", err)
```
inancgumus, line 127: Return an error?

common/screenshotter.go
ankur22: Nice, resolved in 4228c7982f6cb32eeb52d6dfcfabadfd2ffd8eb0. My concern is how it might impact the runtime to  marshal and unmarshal if we use this everywhere, but an issue for later.

common/screenshotter.go
ankur22: 🤦 Yep, bad copy and paste.

common/screenshotter.go
ankur22: Resolved in 5e38ec56b04eabc4a02f05d8230d784946520ede

common/screenshotter.go
ankur22: Resolved in 74c783c4448cdaeb00523b75b1e2e162f10c6188

browser/mapping.go
ankur22: Resolved in b7dff9f3c73b0dd207539e56bcb01ab47cf0911e

browser/mapping.go
ankur22: Resolved in 11219b4572a8bcfd5eecd4acedfa130291de2979

browser/mapping.go
ankur22: Resolved in 11219b4572a8bcfd5eecd4acedfa130291de2979

common/screenshotter.go
inancgumus: Do you mean performance-wise? But since we use Goja everywhere, that will not significantly impact us. We can only know that for sure by measuring it. Developer experience-wise, `convert` brings excellent value.

inancgumus (approved):

browser/mapping.go
mstoykov: I don't think I have enough information to answer this question.

Is this still a problem or have you fixed it in the following PRs.

And if it is (or you want me to confirm) can you please explain more what is happening.

Otherwise - yes if you are working with binary data you should return ArrayBuffer or some TypedArray( for example Uint8Array) which usually is more involved.

inancgumus: From what I understand, we should continue to use `ArrayBuffer` since we're working with binary data.

ankur22: Yep, it was a binary of screenshots, so we're working with `ArrayBuffer` (again) 👍

common/screenshotter.go
```diff
@@ -300,7 +300,7 @@ func (s *screenshotter) screenshotElement(h *ElementHandle, opts *ElementHandleS
 	return buf, nil
 }
 
-//nolint:funlen,cyclop
+//nolint:funlen,cyclop,gocognit
```
inancgumus, line 303:
```suggestion
//nolint:linter
```

😄

inancgumus (approved):

common/screenshotter.go
ankur22: I'll leave it as that for now 😄

inancgumus: But, I have a strong opinion :trollface: :trollface: :trollface:

Issue #1159, opened 2024-01-16
Use env vars to upload screenshot


ankur22:
### What?

With the remote persister from https://github.com/grafana/xk6-browser/issues/1158, use the available configuration to upload the screenshots.

### Why?

To upload the screenshot data to the remote location that has been identified by the env vars.

### How?

There are two requests that need to be made:

1. First request (which is authenticated) to retrieve one or more pre-signed urls.
2. Use the given pre-signed urls for the uploading of the screenshot data.

### Tasks

```[tasklist]
### Tasks
- [ ] PR
- [ ] Update the k6 release notes
```

### Related PR(s)/Issue(s)

Prerequisite issue: https://github.com/grafana/xk6-browser/issues/1158

PR #1202, opened 2024-01-30
Add RemoteFilePersister

ankur22:
## What?

This adds a remote persister, which will upload files (screenshots) to the remote location. It uses a two request mechanism to:

1. Retrieve a presigned url for the given file with the name.
2. Uploads the file with the retrieved presigned url.

Although it is possible to batch retrieve the pre-signed url, there is no easy way to retrieve the filenames that are required to retrieve the presigned urls, so it needs to be done one at a time when a request to persist the screenshot comes in.

## Why?

To enable the browser module to upload screenshots to a remote location.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1159

### DIFF
--- a/storage/file_persister.go
+++ b/storage/file_persister.go
@@ -2,10 +2,17 @@ package storage
 
 import (
 	"bufio"
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
+	"time"
 )
 
 // LocalFilePersister will persist files to the local disk.
@@ -42,4 +49,145 @@ func (l *LocalFilePersister) Persist(path string, data io.Reader) (err error) {
 	}
 
 	return nil
+}
+
+// RemoteFilePersister is to be used when files created by the browser module need
+// to be uploaded to a remote location. This uses a preSignedURLGetterURL to
+// retrieve one pre-signed URL. The pre-signed url is used to upload the file
+// to the remote location.
+type RemoteFilePersister struct {
+	preSignedURLGetterURL string
+	headers               map[string]string
+	basePath              string
+
+	httpClient *http.Client
+}
+
+// NewRemoteFilePersister creates a new instance of RemoteFilePersister.
+func NewRemoteFilePersister(
+	preSignedURLGetterURL string,
+	headers map[string]string,
+	basePath string,
+) *RemoteFilePersister {
+	return &RemoteFilePersister{
+		preSignedURLGetterURL: preSignedURLGetterURL,
+		headers:               headers,
+		basePath:              basePath,
+		httpClient: &http.Client{
+			Timeout: time.Second * 10,
+		},
+	}
+}
+
+// Persist will upload the contents of data to a remote location.
+func (r *RemoteFilePersister) Persist(ctx context.Context, path string, data io.Reader) (err error) {
+	pURL, err := r.getPreSignedURL(ctx, path)
+	if err != nil {
+		return fmt.Errorf("getting presigned url: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPut, pURL, data)
+	if err != nil {
+		return fmt.Errorf("creating upload request: %w", err)
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("performing upload request: %w", err)
+	}
+	defer resp.Body.Close() //nolint:errcheck
+
+	if _, err := io.Copy(io.Discard, resp.Body); err != nil {
+		return fmt.Errorf("draining upload response body: %w", err)
+	}
+
+	if err := checkStatusCode(resp); err != nil {
+		return fmt.Errorf("uploading: %w", err)
+	}
+
+	return nil
+}
+
+func checkStatusCode(resp *http.Response) error {
+	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusMultipleChoices {
+		return fmt.Errorf("server returned %d (%s)", resp.StatusCode, strings.ToLower(http.StatusText(resp.StatusCode)))
+	}
+
+	return nil
+}
+
+// getPreSignedURL will retrieve the presigned url for the current file.
+func (r *RemoteFilePersister) getPreSignedURL(ctx context.Context, path string) (string, error) {
+	b, err := buildPresignedRequestBody(r.basePath, path)
+	if err != nil {
+		return "", fmt.Errorf("building request body: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, r.preSignedURLGetterURL, bytes.NewReader(b))
+	if err != nil {
+		return "", fmt.Errorf("creating request: %w", err)
+	}
+
+	for k, v := range r.headers {
+		req.Header.Add(k, v)
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("performing request: %w", err)
+	}
+	defer resp.Body.Close() //nolint:errcheck
+
+	if err := checkStatusCode(resp); err != nil {
+		return "", err
+	}
+
+	return readResponseBody(resp)
+}
+
+func buildPresignedRequestBody(basePath, path string) ([]byte, error) {
+	b := struct {
+		Service string `json:"service"`
+		Files   []struct {
+			Name string `json:"name"`
+		} `json:"files"`
+	}{
+		Service: "aws_s3",
+		Files: []struct {
+			Name string `json:"name"`
+		}{
+			{
+				Name: filepath.Join(basePath, path),
+			},
+		},
+	}
+
+	bb, err := json.Marshal(b)
+	if err != nil {
+		return nil, fmt.Errorf("marshaling request body: %w", err)
+	}
+
+	return bb, nil
+}
+
+func readResponseBody(resp *http.Response) (string, error) {
+	rb := struct {
+		Service string `json:"service"`
+		URLs    []struct {
+			Name         string `json:"name"`
+			PreSignedURL string `json:"pre_signed_url"` //nolint:tagliatelle
+		} `json:"urls"`
+	}{}
+
+	decoder := json.NewDecoder(resp.Body)
+	err := decoder.Decode(&rb)
+	if err != nil {
+		return "", fmt.Errorf("decoding response body: %w", err)
+	}
+
+	if len(rb.URLs) == 0 {
+		return "", errors.New("missing presigned url in response body")
+	}
+
+	return rb.URLs[0].PreSignedURL, nil
 }
--- a/storage/file_persister.go
+++ b/storage/file_persister.go
@@ -91,7 +91,7 @@ func (r *RemoteFilePersister) Persist(ctx context.Context, path string, data io.
 		return fmt.Errorf("creating upload request: %w", err)
 	}
 
-	resp, err := http.DefaultClient.Do(req)
+	resp, err := r.httpClient.Do(req)
 	if err != nil {
 		return fmt.Errorf("performing upload request: %w", err)
 	}
@@ -132,7 +132,7 @@ func (r *RemoteFilePersister) getPreSignedURL(ctx context.Context, path string) 
 		req.Header.Add(k, v)
 	}
 
-	resp, err := http.DefaultClient.Do(req)
+	resp, err := r.httpClient.Do(req)
 	if err != nil {
 		return "", fmt.Errorf("performing request: %w", err)
 	}
--- a/storage/file_persister_test.go
+++ b/storage/file_persister_test.go
@@ -1,6 +1,11 @@
 package storage
 
 import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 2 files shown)


storage/file_persister.go
```diff
@@ -43,3 +48,142 @@ func (l *LocalFilePersister) Persist(path string, data io.Reader) (err error) {
...
+		return fmt.Errorf("perform upload request: %w", err)
+	}
+	defer func() {
+		if _, cerr := io.Copy(io.Discard, resp.Body); cerr != nil && err == nil {
+			err = fmt.Errorf("drain upload response body: %w", cerr)
+		}
+
+		if cerr := resp.Body.Close(); cerr != nil && err == nil {
+			err = fmt.Errorf("close upload response body: %w", cerr)
+		}
```
inancgumus, line 85: I don't think we should check the error from `Close`. `nolint` here is fine.

storage/file_persister_test.go
```diff
@@ -74,3 +79,174 @@ func TestLocalFilePersister(t *testing.T) {
...
+			wantPresignedURLBody: `{
+					"service":"aws_s3",
+					"files":[{"name":"screenshots/some/path/file.png"}]
+				}`,
+			wantPresignedHeaders: map[string]string{
+				"Authorization": "token asd123",
+				"Run_id":        "123456",
+			},
+			getPresignedURLResponse: http.StatusTooManyRequests,
+			wantError:               "get presigned url: status code out of range: request is being rate limited",
```
inancgumus, line 126: I believe we shouldn't do exact matches for errors for refactorability later on. Do you think just checking "request is being rate limited" is fine here?

storage/file_persister_test.go
```diff
@@ -74,3 +79,174 @@ func TestLocalFilePersister(t *testing.T) {
...
+					defer func() {
+						require.NoError(t, r.Body.Close())
+					}()
+
+					bb, err := io.ReadAll(r.Body)
+					require.NoError(t, err)
+
+					// Ensures that the body of the request matches the
+					// expected format.
+					assert.JSONEq(t, tt.wantPresignedURLBody, string(bb))
```
inancgumus, line 214: What about just `assert.HTTPBody`?

storage/file_persister_test.go
```diff
@@ -74,3 +79,174 @@ func TestLocalFilePersister(t *testing.T) {
...
+						"service": "aws_s3",
+						"urls": [{
+							"name": "%s",
+							"pre_signed_url": "%s"
+						}]
+						}`, basePath, uts.URL)
+
+					w.WriteHeader(tt.getPresignedURLResponse)
+					_, err = w.Write([]byte(body))
+					require.NoError(t, err)
```
inancgumus, line 232: `fmt.Fprintf`?

storage/file_persister.go
```diff
@@ -43,3 +48,142 @@ func (l *LocalFilePersister) Persist(path string, data io.Reader) (err error) {
...
+		} `json:"files"`
+	}{
+		Service: "aws_s3",
+		Files: []struct {
+			Name string `json:"name"`
+		}{
+			{
+				Name: filepath.Join(basePath, path),
+			},
+		},
```
inancgumus, line 154: It might make sense to define the `Files` as a type here to reduce repetition.

inancgumus (commented):
The feature is almost around the corner! Nice. Some suggestions.

storage/file_persister.go
```diff
@@ -43,3 +48,142 @@ func (l *LocalFilePersister) Persist(path string, data io.Reader) (err error) {
...
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("perform upload request: %w", err)
+	}
+	defer func() {
+		if _, cerr := io.Copy(io.Discard, resp.Body); cerr != nil && err == nil {
+			err = fmt.Errorf("drain upload response body: %w", cerr)
+		}
```
inancgumus, line 81: This part might go into the `if err != nil` above. No particular reason I can see for this to be in a `defer`.

storage/file_persister.go
```diff
@@ -43,3 +48,142 @@ func (l *LocalFilePersister) Persist(path string, data io.Reader) (err error) {
...
+	preSignedURLGetterURL string
+	headers               map[string]string
+	basePath              string
+}
+
+// Persist will upload the contents of data to a remote location.
+func (r *RemoteFilePersister) Persist(ctx context.Context, path string, data io.Reader) (err error) {
+	pURL, err := r.getPreSignedURL(ctx, path)
+	if err != nil {
+		return fmt.Errorf("get presigned url: %w", err)
```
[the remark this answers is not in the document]
ankur22: Resolved in da0b39822c1aac8813b3a425d1e618f7d6c55133

storage/file_persister.go
```diff
@@ -43,3 +48,142 @@ func (l *LocalFilePersister) Persist(path string, data io.Reader) (err error) {
...
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPut, pURL, data)
+	if err != nil {
+		return fmt.Errorf("create upload request: %w", err)
+	}
+
+	resp, err := http.DefaultClient.Do(req)
```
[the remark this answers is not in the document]
ankur22: I was relying on the context to cancel any long running requests, but yeah a timeout in the client makes sense. What should it be 10 secs?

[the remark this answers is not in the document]
ankur22: Resolved in 19fa7f8bc0db2c1fe5b414a660dcd5760d89d432

storage/file_persister.go
```diff
@@ -43,3 +48,142 @@ func (l *LocalFilePersister) Persist(path string, data io.Reader) (err error) {
...
+
+	if err := checkStatusCode(resp); err != nil {
+		return fmt.Errorf("upload status code out of range: %w", err)
+	}
+
+	return err
+}
+
+func checkStatusCode(resp *http.Response) error {
+	if resp.StatusCode == 429 {
```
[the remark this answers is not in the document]
ankur22: Resolved in bf11ae4b8b9ade8f88846586bfdde95996f059a0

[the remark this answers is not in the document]
ankur22: Resolved in bf11ae4b8b9ade8f88846586bfdde95996f059a0

storage/file_persister.go
ankur22: If this was in a test, i would agree. What's the reason for not checking?

storage/file_persister.go
```diff
@@ -43,3 +48,142 @@ func (l *LocalFilePersister) Persist(path string, data io.Reader) (err error) {
...
+			},
+		},
+	}
+
+	bb, err := json.Marshal(b)
+	if err != nil {
+		return nil, fmt.Errorf("marshal request body: %w", err)
+	}
+
+	return bytes.NewReader(bb), nil
```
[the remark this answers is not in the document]
ankur22: Not sure i understand why it's unusual, do you have an example why? In this scenario the response is being directly used in the request which takes in an `io.Reader`.

storage/file_persister_test.go
ankur22: Agreed that this is tightly coupling the implementation and the test. I actually found this quite useful to test how the errors would look so I can shape them better. Happy to change them though if you have a strong opinion on it.

storage/file_persister.go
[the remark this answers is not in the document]
inancgumus: [Here](https://pkg.go.dev/io#LimitReader), for example, it makes sense because it takes a `Reader` and changes its behavior with a new one. Returning a `Reader` is a must in this case.

storage/file_persister.go
```diff
@@ -43,3 +48,142 @@ func (l *LocalFilePersister) Persist(path string, data io.Reader) (err error) {
...
+
+	bb, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", fmt.Errorf("reading response body: %w", err)
+	}
+
+	err = json.Unmarshal(bb, &rb)
+	if err != nil {
+		return "", fmt.Errorf("unmarshaling response body: %w", err)
+	}
```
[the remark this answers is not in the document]
inancgumus: It will stream instead of reading everything in memory. Also, less code 🥳

storage/file_persister_test.go
ankur22: Resolved in 7dd12ad0451b5f4f917814ded786174d78a9837d

storage/file_persister.go
ankur22: They're not exactly the same since the request file object only has the `name` field, whereas the response `file` has a `name` and a `url`. But, if you have a strong option then how is best to implement that? Just `file` or both a `service` and a `file` which work for both request and response bodies? 

```go
type service struct {
	Service string `json:"service"`
	Files   []file `json:"files"`
	URLs    []file `json:"urls"`
}

type file struct {
	Name         string `json:"name"`
	PreSignedURL string `json:"pre_signed_url"` //nolint:tagliatelle
}
```

storage/file_persister.go
ankur22: Yep, good spot, resolved in 3c3b6fb879480090dc573d80f89aff295a25ec64

inancgumus: Do we still need `cerr` there?

storage/file_persister.go
inancgumus: No, it just caught my eye. It's fine if you think it's not the best way to use another type.

storage/file_persister.go
ankur22: Yep, that makes sense since we can't do much if it fails 👍 Resolved in 4e35b0857d614e2ced20499457bc30cc3ebd21f3

storage/file_persister.go
```diff
@@ -43,3 +48,142 @@ func (l *LocalFilePersister) Persist(path string, data io.Reader) (err error) {
...
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("perform request: %w", err)
+	}
+	defer func() {
+		if cerr := resp.Body.Close(); cerr != nil && err == nil {
+			err = fmt.Errorf("close response body: %w", cerr)
+		}
+	}()
```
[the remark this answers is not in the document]
ankur22: Resolved in 4e35b0857d614e2ced20499457bc30cc3ebd21f3

storage/file_persister.go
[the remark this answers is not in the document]
ankur22: I can see that there's no point in this particular case for `buildPresignedRequestBody` since the body is tiny. Resolved in a8991fdfbc46e2b939dc756a26d4498091b87dd3.

storage/file_persister.go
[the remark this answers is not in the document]
ankur22: Right, i see now, I did a bit of research on it 👍 Resolved in 6c55bf64cb165fd3dfc1a2b971f438e3accceab2.

storage/file_persister.go
ankur22: Ok, i'm leaving it as it is since there's no real reason to move them out as their own types for now.

storage/file_persister.go
ankur22: Resolved in 66062104624c5f6267ec6dc986f5b9ce303f81bb

storage/file_persister_test.go
ankur22: I'm leaning on the side of them being helpful otherwise the error comparison for both the presigned and upload responses would read the same `request is being rate limited`, whereas the full error message ensures that the correct paths within the implementation have been taken.

storage/file_persister.go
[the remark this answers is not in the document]
inancgumus: Yep. A better way would be returning a concrete `bytes.Reader` if the body was large, or just joining those functions together.

storage/file_persister_test.go
```diff
@@ -74,3 +79,174 @@ func TestLocalFilePersister(t *testing.T) {
...
+					defer func() {
+						require.NoError(t, r.Body.Close())
+					}()
+
+					bb, err := io.ReadAll(r.Body)
+					require.NoError(t, err)
+
+					// Ensures that the data is uploaded to the server and matches
+					// what was sent.
+					assert.Equal(t, tt.data, string(bb))
```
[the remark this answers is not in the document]
ankur22: I've never used `assert.HTTPBody` and I can't see how it would be useful here. The `HTTPBody` looks like it could be useful to test existing handler implementations where we can provide it with a `method`, `url` and any `url params` (although no body). I may have missed something though.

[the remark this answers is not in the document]
inancgumus: Sorry, may bad. I confused this code without completely understanding it. Resolving.

storage/file_persister.go
```diff
@@ -43,3 +49,144 @@ func (l *LocalFilePersister) Persist(path string, data io.Reader) (err error) {
...
+
+	resp, err := r.httpClient.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("performing request: %w", err)
+	}
+	defer resp.Body.Close() //nolint:errcheck
+
+	if err := checkStatusCode(resp); err != nil {
+		return "", fmt.Errorf("status code out of range: %w", err)
+	}
```
inancgumus, line 142: I believe we can remove this error wrapping to reduce noise. In that case, it makes sense to return the second error in `checkStatusCode` like this:

```go
return fmt.Errorf("server returned %d (%s)", resp.StatusCode, http.StatusText(resp.StatusCode))
```

The error will be clearer and less noisy:
- getting presigned url: https://..: server returned 503 (Service Unavailable)
- persisting: server returned 429 (too many requests)

Instead of:
- getting presigned url: https://..: status code out of range: status code is 503
- uploading status code out of range: status code is 429

WDYT?

storage/file_persister_test.go
```diff
@@ -74,3 +79,164 @@ func TestLocalFilePersister(t *testing.T) {
...
+
+func TestRemoteFilePersister(t *testing.T) {
+	t.Parallel()
+
+	basePath := "screenshots"
+
+	tests := []struct {
+		name                    string
+		path                    string
+		data                    string
```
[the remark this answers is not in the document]
ankur22: `data` is purposefully generic since the persister can upload any type of data, not only screenshots. I think it's clear based on the fact that it's data within a test that is testing some sort of persisting behaviour.

storage/file_persister.go
```diff
@@ -43,3 +49,144 @@ func (l *LocalFilePersister) Persist(path string, data io.Reader) (err error) {
...
+type RemoteFilePersister struct {
+	preSignedURLGetterURL string
+	headers               map[string]string
+	basePath              string
+
+	httpClient *http.Client
+}
+
+// NewRemoteFilePersister creates a new instance of RemoteFilePersister.
+func NewRemoteFilePersister(pURL string, h map[string]string, bp string) *RemoteFilePersister {
```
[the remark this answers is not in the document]
ankur22: Sure resolved in f6943fe9b2277b8181c34dac2ae7019b83f5ff5f

storage/file_persister.go
ankur22: Yep, i think this makes sense. I'm not so sure about adding the text versions of the http status codes, and instead will just work with the http status codes.

ankur22: Resolved in ac20d5b3d532c72443b9d5648c925b77b56f3a4d

storage/file_persister_test.go
```diff
@@ -74,3 +79,164 @@ func TestLocalFilePersister(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			// This starts the upload server where the files will be uploaded
+			// with the presigned url.
+			uts := httptest.NewServer(http.HandlerFunc(
```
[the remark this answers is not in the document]
ankur22: Yep, nice idea.

storage/file_persister.go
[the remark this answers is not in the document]
inancgumus: Perfect.

storage/file_persister_test.go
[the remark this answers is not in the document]
inancgumus: I still think a comment like this can be there for clarity.

storage/file_persister.go
inancgumus: Nice. The text was about seeing what the status code means for users and us, so nobody needs to look it up.

inancgumus: Maybe it wasn't clear in my previous comment, but, I also suggested combining these status code into one. Instead of saying, "request is being rate limited", "too many requests" is already clear. That's also why I suggest using status texts.

ankur22: Sure resolved in b9b25537c7f2cddb5c375c5dfa0301113ead217a

storage/file_persister_test.go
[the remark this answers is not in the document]
ankur22: Resolved in ef9e44df2a5352b9f5ccf94d3dbb218bcbdb1f83

storage/file_persister_test.go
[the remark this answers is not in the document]
inancgumus: Should we make these constants?

https://github.com/grafana/xk6-browser/commit/7db97ec7bc38b4f567f3b622889a7cd6dc42de5e#diff-51cb6e5b93355eb6fbd5fbcd7461c17296941f9b63ea1506316d1a20beb24ae0R86-R88

inancgumus (approved):
Nice work!

Issue #1160, opened 2024-01-16
Wire up the local disk and remote persisters


ankur22:
### What?

Wire up the local disk persister (https://github.com/grafana/xk6-browser/issues/1156) and remote persister (https://github.com/grafana/xk6-browser/issues/1159) so that when the environment is setup correctly the remote persister is used otherwise the local persister is used.

### Why?

When the configuration for the remote persister is present (the env var that was introduced in https://github.com/grafana/xk6-browser/issues/1159), that implies that the screenshots are to be sent to a remote location. When that is not the case, then they need to be saved local to the disk.

### How?

When the test run starts, if the env var (from https://github.com/grafana/xk6-browser/issues/1159) is present and the remote persister is created, that is to be passed to the [screenshotter](https://github.com/grafana/xk6-browser/blob/c0584af38c6ecaa29aa0c8287cd2daa4de3aeb5b/common/screenshotter.go#L67). When this is not the case a local persister should be created and passed to the `screenshotter`.

1. Read and parse `K6_BROWSER_SCREENSHOTS_OUTPUT ` which should contain all the configuration required to upload the screenshot. Take a look at how it's done in k6 for [traces](https://github.com/grafana/k6/blob/master/lib/trace/otel.go#L161).
2. Create a new interface that works with both `LocalPersister` and `RemotePersister` which may require amendments to one of both of the concrete types.

### Tasks

```[tasklist]
### Tasks
- [x] PR: https://github.com/grafana/xk6-browser/pull/1203
- [ ] https://github.com/grafana/k6/pull/3600
```

### Related PR(s)/Issue(s)

Prerequisite issues: https://github.com/grafana/xk6-browser/issues/1156 & https://github.com/grafana/xk6-browser/issues/1159.

Issue #1065, opened 2023-10-12
Rename `browser` and `common` packages


inancgumus:
### What?

1. Rename the `browser` package to `module`
2. Rename the `common` package to `browser`

### Why?

1. The current `browser` package includes code for the browser module.
2. The `common` package name is too generic.

Making this refactoring can make our Go documentation and code more readable for us and external contributors. Since we've removed the `api` package and have a mapping layer that can manage different kinds of browsers, there is no longer a need for a common functionality that will be shared by browsers. Even if there is, the name `common` is still too generic.

---

Here are some comparisons of what it would look like if we make these renamings.

**Before**

```go
xk6-browser/browser.JSModule
xk6-browser/browser.ModuleInstance
xk6-browser/browser.RootModule
xk6-browser/browser.New
```

**After**

```go
xk6-browser/module.JS         // enables renaming: JSModule to just JS
xk6-browser/module.Instance   // enables renaming: ModuleInstance to Instance
xk6-browser/module.Root       // enables renaming: ModuleRoot to Root
xk6-browser/module.New
```

**Before**

```go
xk6-browser/common.Browser
xk6-browser/common.BrowserOptions
xk6-browser/common.BrowserContext
xk6-browser/common.BrowserContextOptions
xk6-browser/common.BrowserProcess
xk6-browser/common.Page
xk6-browser/common.Frame
xk6-browser/common.ElementHandle
```

**After**

```go
xk6-browser/browser.Browser  // this is an outlier, like url.URL
xk6-browser/browser.Context
xk6-browser/browser.Options
xk6-browser/browser.ContextOptions
xk6-browser/browser.Process
xk6-browser/browser.Page
xk6-browser/browser.Frame
xk6-browser/browser.ElementHandle
```

Once we finish this, we can continue splitting the browser package into more meaningful packages if needed.

inancgumus:
One nice addition after this renaming is to document the `module` package's role. It's an entry point and drives the downstream components by passing them the external concrete dependencies.

PR #1203, opened 2024-02-01
Wire up the local and remote persisters

ankur22:
## What?

Now that we have the [LocalFilePersister](https://github.com/grafana/xk6-browser/blob/main/storage/file_persister.go#L19) and [RemoteFilePersister](https://github.com/grafana/xk6-browser/blob/main/storage/file_persister.go#L58), the next step is to easily switch between the two depending on the environment configuration.

The key environmental change is the addition of `K6_BROWSER_SCREENSHOTS_OUTPUT` which allows us to configure the necessary settings which will enable the browser module to upload the screenshots to the cloud. If this isn't present then it will assume that the screenshots are to be persisted to the local disk.

## Why?

It will allow us to easily swap out the local persister with the remote one. The design of this closely follows how it was implemented for tracing.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1160


storage/file_persister.go
```diff
@@ -13,14 +13,91 @@ import (
...
 	"time"
+
+	"github.com/grafana/xk6-browser/env"
 )
 
+// FilePersister is the type that all file persisters must implement. It's job is
+// to persist a file somewhere, hiding the details of where and how from the caller.
+type FilePersister interface {
+	Persist(ctx context.Context, path string, data io.Reader) (err error)
+}
```
ankur22, line 24: I'm not sure about where this should live. It feels correct, since this is needed elsewhere in the codebase, but would it be better to define the interface multiple times closer to where it's needed?

I took the inspiration of this design from how `context` is implemented, and how the interface and concrete types seem to live in the same package, and so not having to continually write an interface over and over again.

storage/file_persister.go
```diff
@@ -13,14 +13,91 @@ import (
...
+		return nil, fmt.Errorf("parsing %s: %w", env.ScreenshotsOutput, err)
+	}
+
+	return NewRemoteFilePersister(u, h, b), nil
+}
+
+// parseEnvVar will parse a value such as:
+// url=https://127.0.0.1/,basePath=/screenshots,header.1=a,header.2=b
+// and return them.
+func parseEnvVar(envVarValue string) (string, string, map[string]string, error) {
```
ankur22, line 46: This way of parsing an env var seems super flakey, am i missing a trick?

storage/file_persister.go
inancgumus: > but would it be better to define the interface multiple times

Where do we need to define it multiple times? AFAICS, there is only one usage in the `common` package, no?

storage/file_persister.go
inancgumus: Not that I know of :( Maybe `strings.Cut` can help here.

storage/file_persister.go
ankur22: > Where do we need to define it multiple times? AFAICS, there is only one usage in the common package, no?

In the `browser` package in `module.go` and `moduelVU.go`

ankur22: > This is an anti-pattern

👍 

> The interface should live in the common package since its user is the common package

And in the `browser` package.

> This way, changing the methods on the interface won't break other packages and produce a cascading effect of maintenance burden.

👍 

> and concrete types are never exported, unlike we did

What if we unexported the concrete types? Would that be acceptable?

storage/file_persister.go
ankur22: > Not that I know of :( Maybe strings.Cut can help here

No worries 👍 

> BTW, it might be a better idea to return this in a struct instead of a lot of primitive values.

I was wondering whether it would be better to do that instead of returning a long list of variables. I'll change it to return a struct.

storage/file_persister.go
inancgumus: The `Context` interface is exported because it wants others to implement this interface. It's a standard library package, after all. In our case, there are only two implementations, and they're in the same package. So, this is more like the Java/C# way of doing things with a factory pattern. Go doesn't need this.

Since Go has a structural type system, it's always OK to copy an interface in both packages if we don't expect others to implement the interface in different ways. The interface doesn't have to have a definite name and place, it doesn't have to be singular. The concrete types already provide documentation.

Moreover, by putting the factory in the `storage` package, we hide the decision to pick a storage. Since our code starts from the mapping layer, this makes code navigation difficult. We must go to the storage package to understand how it works.

So, I suggest:
- Adding an unexported interface in the `browser` package to make switching easier.
- Moving `NewFilePersister` into the mapping layer to make it explicit what concrete types it's passing to the inner layers.
- Adding an exported `Persister` interface in the `common` package so that the `common` package methods can accept the interface.

storage/file_persister.go
inancgumus: I believe a struct will be much clearer. We might also want to make these field names descriptive or add some comments.

storage/file_persister.go
ankur22: Yep, the bit about the interfaces, concrete types, `context` and DRY  I agree with.

> Since our code starts from the mapping layer, this makes code navigation difficult

Why is this particularly an issue in this case? We have to navigate from file to file and package to package often, but it's not the greatest of burdens.

> Moving `NewFilePersister` into the mapping layer as an unexported function to make it explicit what concrete types it's passing to the inner layers

By doing this we're moving away from the factory pattern, and allowing the decision to be made by the mapping layer before it injects it into the appropriate APIs that require it, which would mean that the `NewFilePersister` would still need to return an interface. I'm not sure I understand the benefit of doing that. Can you explain why moving `NewFilePersister` to the mapping layer would benefit us?

> Adding an unexported interface in the browser package to make this package's switching between different types easier.

👍 

> Adding an exported Persister interface in the common package so that the common package methods can accept the interface.

👍

inancgumus: > Why is this particularly an issue in this case?

Increases the cognitive load, especially when debugging. It's not about navigation per se but understanding and following the call path. A -> B -> C.

> which would mean that the NewFilePersister would still need to return an interface

Not necessarily. A method/function can read the environment variables and save the necessary concrete type to the module's interface field.

ankur22: I've moved the interfaces out into their own packages and unexported the types so that they can't be used outside of the package that they're defined in (7125509982597028318c97f70728dad4d952ae0c). I'm still unsure of the next step, which is to call `NewFilePersister` from the mapping layer or refactor/move `NewFilePersister` into the mapping layer?

> 1- Increases the cognitive load, especially when debugging.

👍 

> 2- We're hiding how we're picking a persister.

This is just abstracting that bit of detail away from the caller. I don't see why the `browser` or `common` packages need to know about how the `storage` package makes the decision.

> 3- We create a dependency between multiple packages, which is fragile and unnecessary

The only way around this is if the logic to make the decision to pick either the `LocalFilePersister` or the `RemoteFilePersister` resided in the `browser` package. It could then create either of the persisters, store it in the `moduleVU` within a package defined interface -- at this point should the `common` package export its version of the `filePersister` interface or not? Seems like a smell not to export it.

inancgumus: The new code looks nicer but still needs improvement. So, it can be an example implementation to follow for the upcoming features other than screenshot persisting.

> Which is to call NewFilePersister from the mapping layer or refactor/move NewFilePersister into the mapping layer?

We're moving the driving logic that deals with externals in the `browser` package. I recommend moving the logic in `NewFilePersister` to the mapping layer and setting the `filePersister` field in the `browser` package for clarity and maintainability.

> This is just abstracting that bit of detail away from the caller. I don't see why the browser or common packages need to know about how the storage package makes the decision.

The caller is the `browser` package, which is the driver. It doesn't need this kind of abstraction since it's an abstraction already. It's not the best place to abstract the machinery. More cognitive load. For consistency and clarity, we should see what happens there and which concrete values it's passing to downstream components.

> The only way around this is if the logic to make the decision to pick either the LocalFilePersister or the RemoteFilePersister resided in the browser package. It could then create either of the persisters, store it in the moduleVU within a package defined interface. At this point the common package should export its own persister interface.

Yes, this will be great. An exported interface field in `filePersister`. It doesn't have to be package-defined. One of the methods in `RootModule`, or a function in the `browser` package can update this field.

> Or move all the persister code into the browser package

I would advise against this as we don't need to see the internal machinery related to persisting. We should only see the decision-making process of picking a persisted for clarity.

> if that's the case then I have the following questions:
> Why do we have a storage package?
> When should we create new packages?

Like I said above, `storage` package is perfectly fine to have. It should export those concrete persister types to let the `browser` package drive this package.

---

I have a few more notes about the tests.

Since we're abstracting the persistency logic in `common`, we should not call the `NewFilePersister` there. Instead, we can just pass `Screenshot`, a fake type that we implement in `tests`. This will remove the dependency of `storage` from `tests`.

ankur22: > We're moving the driving logic that deals with externals in the browser package. I recommend moving the logic in NewFilePersister to the mapping layer and setting the filePersister field in the browser package for clarity and maintainability.

I think i know what you are getting at now. The `browser` package's role is the entry point, initialiser of various global and vu level types, and a translation layer between goja and the internal business logic. We should rename and reword what the packages do, at the moment [it's not clear](https://github.com/grafana/xk6-browser/blob/205f1734dc72351c0ee9f609e46cfc494ac6f99e/browser/module.go#L1).

> The caller is the browser package, which is the driver. It doesn't need this kind of abstraction since it's an abstraction already. It's not the best place to abstract the machinery. More cognitive load. For consistency and clarity, we should see what happens there and which concrete values it's passing to downstream components.

Yep, I have a better understanding of what the `browser` package does now and why you are making the suggestion 👍 

> An unexported interface field in filePersister

Not sure what you mean here, is it not already that in the latest change? The only thing remaining is to move `NewFilePersister` out of `storage` and into `browser`. I'll make the change and we can discuss any issues in a new thread.

> I would advise against this as we don't need to see the internal machinery related to persisting. We should only see the decision-making process of picking a concrete persister for clarity.

The packages I see that are similar to `storage` are `chromium`, `trace`, `k6ext`, and `log`. I guess these aren't "drivers" but abstracted away implementation details, which makes them simpler to swap out for a new implementation. So these packages are less to do with reuse, and more to do with hiding the details? Maybe the docs for the packages need to be updated.

> Like I said above, storage package is perfectly fine to have. It should export those concrete persister types to let the browser package drive this package.

👍 

> Since we're abstracting the persistency logic in common, we should not call the NewFilePersister in the screenshot tests. Instead, we can just pass a fake type that we implement in tests to the screenshot tests. This will remove the dependency of storage from tests.

👍 I'll work on the tests once we've settled on the implementation.

ankur22: I'm moved the `newFilePersister` code in 33c426b08be03cbf3555f94fb56e033aaad0d047.

inancgumus: > The browser package's role is the entry point

💯 We have an open issue about clarifying the role of the `browser` package.

https://github.com/grafana/xk6-browser/issues/1065

It's about renaming but can be extended or supplemented with other issues. I tackled this renaming work last month. But I couldn't pull it off because k6-core refers to the `browser` package, and renaming this package breaks k6-core. I couldn't find a good solution. We can, at least, update the package comment.

> We should rename and reword what the packages do, at the moment [it's not clear](https://github.com/grafana/xk6-browser/blob/205f1734dc72351c0ee9f609e46cfc494ac6f99e/browser/module.go#L1).

I'll do that after discussing it internally with you.

> Not sure what you mean here, is it not already that in the latest change?

Yes, it's in the latest commits. I just reiterated it. It's all fine.

> The packages I see that are similar to storage are chromium, trace, k6ext, and log. I guess these aren't "drivers" but abstracted away implementation details, which makes them simpler to swap out for a new implementation. So these packages are less to do with reuse, and more to do with hiding the details? Maybe the docs for the packages need to be updated.

👍 Yes, they're about doing their specific machinery and hiding the details of doing so. They're driven by other packages. I agree that we should update their docs.

> 👍 I'll work on the tests once we've settled on the implementation.

👍

inancgumus: > I'm moved the newFilePersister code in https://github.com/grafana/xk6-browser/commit/33c426b08be03cbf3555f94fb56e033aaad0d047.

Thanks. This is excellent.

storage/file_persister.go
ankur22: I've added this change in 3389a93947d654e531f932431213a1ba80240c5e.

inancgumus: Better 👍 For this:

https://github.com/grafana/xk6-browser/blob/3389a93947d654e531f932431213a1ba80240c5e/browser/file_persister.go#L11-L15

WDYT about `presignedURLConfig` since there isn't another remote persisted config, and it's about pre-signed URLs. This also simplifies the field names. Plus, this is used in a very short scope—its context makes it already meaningful.

```go
type presignedURLConfig struct {
	getterURL string
	headers   map[string]string
	basePath  string
}
```

No strong opinions.

ankur22: Makes sense, changed in d864d6d8058d44e6da07a172014e950fcbbc21a2

inancgumus: Nice. Thanks.

browser/file_persister.go
```diff
@@ -0,0 +1,103 @@
...
+// url=https://127.0.0.1/,basePath=/screenshots,header.1=a,header.2=b
+// and return them.
+func parseEnvVar(envVarValue string) (presignedURLConfig, error) {
+	ss := strings.Split(envVarValue, ",")
+
+	var (
+		preSignedURLGetterURL string
+		basePath              string
+		headers               = make(map[string]string)
+	)
```
inancgumus, line 45: Using `presignedURLConfig` here can shorten this code to turn off the linter error 🤔 Adding a `nolint` is fine if it won't work.

ankur22: I feel that `preSignedURLGetterURL` explains what the url is for without any additional comments, so i'd prefer to keep that. I've added a nolint to `parseEnvVar` for now.

browser/file_persister_test.go
```diff
@@ -0,0 +1,173 @@
...
+			envLookup: func(key string) (string, bool) {
+				return "", true
+			},
+			wantType: &storage.LocalFilePersister{},
+		},
+		{
+			name: "remote",
+			envLookup: func(key string) (string, bool) {
+				return "url=https://127.0.0.1/,basePath=/screenshots,header.1=a", true
+			},
```
inancgumus, line 39: Using the environment variable key makes it more explicit.

```suggestion
			envLookup: ConstLookup(env.ScreenshotsOutput, "url=https://127.0.0.1/,basePath=/screenshots,header.1=a")
```

browser/file_persister_test.go
```diff
@@ -0,0 +1,173 @@
...
+			wantType: &storage.RemoteFilePersister{},
+		},
+		{
+			name: "remote_parse_failed",
+			envLookup: func(key string) (string, bool) {
+				return "basePath=/screenshots,header.1=a", true
+			},
```
inancgumus, line 46:
```suggestion
			envLookup: ConstLookup(env.ScreenshotsOutput, "basePath=/screenshots,header.1=a")
```

browser/file_persister_test.go
```diff
@@ -0,0 +1,173 @@
...
+					"2": "b",
+				},
+			},
+		},
+		{
+			name:        "url",
+			envVarValue: "url=https://127.0.0.1/",
+			want: presignedURLConfig{
+				getterURL: "https://127.0.0.1/",
+				headers:   map[string]string{},
```
inancgumus, line 105: Should we allocate a map if it's empty? This comment is for `parseEnvVar` actually.

browser/module.go
```diff
@@ -18,19 +19,25 @@ import (
 	"github.com/grafana/xk6-browser/common"
 	"github.com/grafana/xk6-browser/env"
 	"github.com/grafana/xk6-browser/k6ext"
-	"github.com/grafana/xk6-browser/storage"
 
 	k6modules "go.k6.io/k6/js/modules"
 )
 
 type (
+	// FilePersister is the type that all file persisters must implement. It's job is
```
inancgumus, line 27: `filePersister` instead of `F`?

tests/element_handle_test.go
```diff
@@ -375,6 +375,8 @@ func TestElementHandleScreenshot(t *testing.T) {
 	elem, err := p.Query("div")
 	require.NoError(t, err)
 
+	// TODO: Use a mock instead of a LocalFilePersister when there's no need to
+	//       persist files.
```
inancgumus, line 379: Do we have a TODO for this somewhere on Github? Maybe add a task to the issue.

browser/file_persister_test.go
```diff
@@ -0,0 +1,173 @@
...
+	}{
+		{
+			name:        "url_headers_basePath",
+			envVarValue: "url=https://127.0.0.1/,basePath=/screenshots,header.1=a,header.2=b",
+			want: presignedURLConfig{
+				getterURL: "https://127.0.0.1/",
+				basePath:  "/screenshots",
+				headers: map[string]string{
+					"1": "a",
+					"2": "b",
```
inancgumus, line 85: Since tests are documentation, instead of `"1": "a"` and `"2": "b"`, would it be better to show here the would-be usage? This might show a reader why these `header`s exist. The same suggestion also for `Test_newFilePersister`

browser/file_persister.go
```diff
@@ -0,0 +1,105 @@
...
+	for _, s := range ss {
+		// The key value pair should be of the form key=value, so split
+		// on '=' to retrieve the key and value separately.
+		kv := strings.Split(s, "=")
+		if len(kv) <= 1 || len(kv) > 2 {
```
inancgumus, line 52: ?

```suggestion
		if len(kv) != 2 {
```

browser/file_persister.go
```diff
@@ -0,0 +1,105 @@
...
+		// A key with "header." means that the header name is encoded in the
+		// key, separated by a ".". Split the header on "." to retrieve the
+		// header name. The header value should be present in v from the previous
+		// split.
+		var hv, hk string
+		if strings.Contains(k, "header.") {
+			hv = v
+
+			hh := strings.Split(k, ".")
+			if len(hh) <= 1 || len(hh) > 2 {
```
inancgumus, line 68:
```suggestion
			if len(hh) != 2 {
```

inancgumus (commented):
Mostly excellent. Some suggestions.

browser/file_persister.go
ankur22: Fixed in 1790d3e576fcba88ce1e2ff0a38c4c36c34133af

browser/file_persister.go
ankur22: 🤦 yep. Changed in 9d293a06ec75da0170a104c04b19198662155da7

browser/file_persister.go
ankur22: Changed in 9d293a06ec75da0170a104c04b19198662155da7

browser/file_persister.go
```diff
@@ -0,0 +1,105 @@
...
+			hk = hh[1]
+
+			if hk == "" {
+				return presignedURLConfig{}, fmt.Errorf("empty header key, received %q", s)
+			}
+		}
+
+		switch k {
+		case "url":
+			_, err := url.ParseRequestURI(v)
```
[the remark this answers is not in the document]
ankur22: Ah, right. What do you think of this https://go.dev/play/p/rHI6lJgpo1E?

From: https://stackoverflow.com/questions/31480710/validate-url-with-standard-package-in-go

browser/file_persister.go
```diff
@@ -0,0 +1,105 @@
...
+			basePath = v
+		case "header":
+			headers[hk] = hv
+		default:
+			return presignedURLConfig{}, fmt.Errorf("invalid option %q", k)
+		}
+	}
+
+	if preSignedURLGetterURL == "" {
+		return presignedURLConfig{}, fmt.Errorf("missing required url")
```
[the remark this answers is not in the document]
ankur22: Fixed in 4f7960cf50770be041a0a1958fc356988941de24

browser/file_persister_test.go
```diff
@@ -0,0 +1,173 @@
...
+		name      string
+		envLookup env.LookupFunc
+		wantType  filePersister
+		wantErr   bool
+	}{
+		{
+			name: "local_no_env_var",
+			envLookup: func(key string) (string, bool) {
+				return "", false
+			},
```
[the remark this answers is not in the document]
ankur22: Fixed in 54d5961592d1531867598ba3b75bd5e1ad392cb0

browser/file_persister_test.go
```diff
@@ -0,0 +1,173 @@
...
+			wantType: &storage.LocalFilePersister{},
+		},
+		{
+			name: "local_empty_env_var",
+			envLookup: func(key string) (string, bool) {
+				return "", true
+			},
```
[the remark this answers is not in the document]
ankur22: Fixed in 54d5961592d1531867598ba3b75bd5e1ad392cb0

browser/file_persister_test.go
ankur22: Fixed in 54d5961592d1531867598ba3b75bd5e1ad392cb0

browser/file_persister_test.go
ankur22: Fixed in 54d5961592d1531867598ba3b75bd5e1ad392cb0

browser/file_persister_test.go
ankur22: I don't think it's the biggest issue if we do. In most use cases the headers will need to be used.

browser/module.go
ankur22: Fixed in 0af81728c03aa50c0b398a03ef78ae6358b616b1

browser/file_persister_test.go
inancgumus: Sure, just wanted to point it out.

browser/file_persister.go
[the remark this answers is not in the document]
inancgumus: Yeah, it makes sense, but it doesn't have to be another function. A more robust implementation could be:

```go
u, err := url.Parse(str)
s, h := strings.TrimSpace(u.Scheme), strings.TrimSpace(u.Host)
return err == nil && s != "" && h != ""
}
```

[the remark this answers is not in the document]
ankur22: Yeah, that function was only a way to showcase how it works easily in the playground. Why is this more robust?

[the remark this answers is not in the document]
inancgumus: It's not, actually :) Just hacked it quickly. The behavior should have changed since I last looked at it. The Playground version seems OK.

browser/file_persister_test.go
inancgumus: 👍 The suggestion was for maintainability and readability.

browser/file_persister.go
inancgumus: Actually I was suggesting this :)

```go
presignedURL := presignedURLConfig{
    headers: make(map[string]string),
}

...

for _, s := range ss {
    ...
    presignedURL.getterURL = v
...
    case "basePath":
        presignedURL.basePath = v
    case "header":
        presignedURL.headers[..]
...
}
return presignedURL, nil
```

browser/file_persister.go
```diff
@@ -0,0 +1,106 @@
...
+type presignedURLConfig struct {
+	getterURL string
+	headers   map[string]string
+	basePath  string
+}
+
+// newFilePersister will return either a persister that persists file to the local
+// disk or uploads the files to a remote location. This decision depends on whether
+// the K6_BROWSER_SCREENSHOTS_OUTPUT env var is setup with the correct configs.
+func newFilePersister(envLookup env.LookupFunc) (filePersister, error) {
```
inancgumus, line 22:
```
// newScreenshotPersister will return either a persister that persists file to the local
// disk or uploads the files to a remote location. This decision depends on whether
// the K6_BROWSER_SCREENSHOTS_OUTPUT env var is setup with the correct configs.
func newScreenshotPersister(envLookup env.LookupFunc) (filePersister, error) {
```

I know you wanted to keep it generic, but it uses `ScreenshotsOutput`. So it makes better sense to me to rename it like this to reflect its purpose.

browser/file_persister.go
```diff
@@ -0,0 +1,106 @@
...
+	envVar, ok := envLookup(env.ScreenshotsOutput)
+	if !ok || envVar == "" {
+		return &storage.LocalFilePersister{}, nil
```
inancgumus, line 25: Wondering why this is not `nil`?

```suggestion
		return nil, nil
```

browser/file_persister.go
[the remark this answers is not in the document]
ankur22: Fixed in fe76642f50cad31c992545a106c4eecd9cf6cdf5

browser/file_persister.go
ankur22: Fixed in ab27742ff7b334e8db66668d1fb2fdaa531d5a10

browser/file_persister.go
ankur22: Good spot, fixed in 97e899678203f030fec6a3d4197d46d6a3744ffe

ankur22: Oh wait! This is correct, if the env var hasn't been found and if it has but the value was an empty string then we want to return a `LocalFilePersister` here.

ankur22: Dropper commit https://github.com/grafana/xk6-browser/commit/97e899678203f030fec6a3d4197d46d6a3744ffe

inancgumus: :) That was the reason! I think this deserves a comment.

ankur22: I think it should be clear enough from the comment above the function.

tests/element_handle_test.go
ankur22: Thanks for reminding me. I've added a `mockPersister` in 145648579f8c73fe46a8de4e329c52656f9fe271

browser/file_persister.go
ankur22: Right, this feels like a nit pick that doesn't add that much value but i've changed it in 95c63fe0d639f83112969aedf6a1c9c0412cd932

tests/element_handle_test.go
inancgumus: Nice ❤️

inancgumus (approved):
LGTM. Nice work 👏 💯

browser/file_persister_test.go
```diff
@@ -0,0 +1,174 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/grafana/xk6-browser/env"
+	"github.com/grafana/xk6-browser/storage"
+)
+
+func Test_newFilePersister(t *testing.T) {
```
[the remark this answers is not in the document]
ankur22: Good spot! I can't keep track now 😅 I also fore pushed the change 🤦 into 93cc48f34a4d4b3da1105bf9c4340b860509dd27 instead of a fixup.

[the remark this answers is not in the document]
inancgumus: No worries 😄 We're almost there.

inancgumus (approved):

PR #1205, opened 2024-02-02
Fix 1204: panic while exporting nil goja values

### LINKED ISSUE #1204: Fix panic while exporting a nil Goja value
### What?

When running the script in [#1089](https://github.com/grafana/xk6-browser/issues/1089) it panics with an NPD.

Last commit: https://github.com/grafana/xk6-browser/commit/12ab6e93fadb1fc1348e2e4055dd5b51cac9c877

### Why

To prevent panics while calling browser methods with `null` or `undefined` values.

### How?

- We need to check for `nil` [here](https://github.com/grafana/xk6-browser/blob/1947368dc42a56133d87ce1352e5e2ba1b760a09/browser/mapping.go#L91) before exporting `goja.Value` types.
- This should also be checked for the other options that export.

### Tasks

```[tasklist]
### Tasks
- [ ] #1205
- [ ] https://github.com/grafana/k6/pull/3584
```

### Related PR(s)/Issue(s)

https://github.com/grafana/xk6-browser/issues/1182

inancgumus:
## What?

- Fixes #1204 
- Adds two examples to test the behavior

## Why?

- To prevent panics while passing `null` values to methods.

## Checklist

## Related PR(s)/Issue(s)

Updates: #1204

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -1021,7 +1021,7 @@ func panicIfFatalError(ctx context.Context, err error) {
 // exportArg exports the value and returns it.
 // It returns nil if the value is undefined or null.
 func exportArg(gv goja.Value) any {
-	if gv == nil || gv == goja.Undefined() || gv == goja.Null() {
+	if !gojaValueExists(gv) {
 		return nil
 	}
 	return gv.Export()
@@ -1036,4 +1036,10 @@ func exportArgs(gargs []goja.Value) []any {
 		args = append(args, exportArg(garg))
 	}
 	return args
+}
+
+// gojaValueExists returns true if a given value is not nil and exists
+// (defined and not null) in the goja runtime.
+func gojaValueExists(v goja.Value) bool {
+	return v != nil && !goja.IsUndefined(v) && !goja.IsNull(v)
 }
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -2,7 +2,6 @@ package browser
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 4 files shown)


browser/mapping.go
```diff
@@ -1042,3 +1017,23 @@ func panicIfFatalError(ctx context.Context, err error) {
 		k6ext.Abort(ctx, err.Error())
 	}
 }
+
+// exportArg exports the value and returns it.
+// It returns nil if the value is undefined or null.
+func exportArg(gv goja.Value) any {
+	if gv == nil || gv == goja.Undefined() || gv == goja.Null() {
```
ankur22, line 1024: Not important, but wondering if exporting `gojaValueExists` could be useful? Possibly in another PR.

ankur22 (approved):
LGTM 🚀

browser/mapping.go
inancgumus: Copying it here would be more useful :) It's a simple function. I can do that now. WDYT?

ankur22: I understand that a little copying is sometimes better than reusing a method, but I don't understand why that must be the case for this, especially in the mapping layer which is going to be very long and difficult to navigate if we keep adding functions into this file.

ankur22: On the other hand having a function closer to the caller does indeed make things simpler.

Since the mapping layer will be focused on the goja parsing, would it eventually make sense for the parsing logic to live in the `browser` package instead of the `common` package? If that's the eventual goal then I can see how copying `gojaValueExists` into `mapping.go` would be a good step forward. That could also mean that we refactor goja stuff into the mapping layer, we should also be refactoring the parse code from the `common` package too. WDYT?

inancgumus: > I don't understand why that must be the case for this

Because the `gojaValueExists` is a little function. Reusing it doesn't add any value. However, exporting the function pollutes the API surface area.

> On the other hand having a function closer to the caller does indeed make things simpler.

+1

> would it eventually make sense for the parsing logic to live in the browser package instead of the common package?

Exactly: https://github.com/grafana/xk6-browser/issues/1064

> That could also mean that when we refactor goja stuff into the mapping layer, we could also be moving the parse code from the common package too.

I'm all in!

ankur22: > Because the gojaValueExists is a little function (more on [this here](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=568s)). Reusing it doesn't add any value. However, exporting the function pollutes the API surface area.

Sure, happy with the idea 👍

inancgumus: @ankur22 Done: https://github.com/grafana/xk6-browser/pull/1205/commits/43be12cbba98953c37e5b043238a80f177c3e412, https://github.com/grafana/xk6-browser/pull/1205/commits/f5f26e93eedf7702ca0a68aed6859b8b5e5bab38, https://github.com/grafana/xk6-browser/pull/1205/commits/36a214ecbf2e038fb2f106f436174c85662a02a0.

Issue #1066, opened 2023-10-13
Data race due to concurrent map access


ka3de:
### Brief summary

The following test script reproduces a data race for concurrent access on a map:
```js
import { check, sleep } from 'k6';
import { browser } from 'k6/experimental/browser';


export const options = {
  thresholds: {
    'browser_web_vital_inp': ['p(75) < 80'],
  },
  scenarios: {
    ui: {
      executor: 'ramping-vus',
      stages: [
        { duration: '10s', target: 5 },
        { duration: '20s', target: 5 },
        { duration: '10s', target: 0 },
      ],
      options: {
        browser: {
            type: 'chromium',
        },
      },
    },
  },
}


export default async function() {
  const context = browser.newContext();
  const page = context.newPage();

  try {
    await page.goto('https://test.k6.io/', { waitUntil: 'networkidle' });

    await Promise.all([
      page.waitForNavigation(),
      page.locator('a[href="/browser.php"]').click(),
    ]);
    page.locator('#text1').type('lorem ipsum');

    sleep(1);

    page.locator('#counter-button').click();

    sleep(0.5);
    
    await Promise.all([
      page.waitForNavigation(),
      page.locator('a[href="/"]').click(),
    ]);
    check(page, {
      'header': p => p.locator('h1').textContent() == 'test.k6.io',
    });


  } finally {
    page.close();
  }
}

```
Which produces the following [truncated] output:

<details>
<summary>output</summary>

```shell

panic_contents="fatal error: concurrent map iteration and map write
goroutine 7 [running]:
go.k6.io/k6/log.(*lokiHook).Listen(0xc000695ee0, {0x19d8eb8, 0xc00009a550})
go.k6.io/k6/log/loki.go:234 +0x71b
go.k6.io/k6/cmd.(*rootCommand).setLoggerHook.func1()
go.k6.io/k6/cmd/root.go:293 +0x2d
created by go.k6.io/k6/cmd.(*rootCommand).setLoggerHook in goroutine 1
go.k6.io/k6/cmd/root.go:292 +0xc9

goroutine 1 [chan receive]:
go.k6.io/k6/execution.(*Scheduler).Run(0xc000a7b110, {0x19d8eb8, 0xc00009a5a0}, {0x19d8e80, 0xc0006a7800}, 0xc001eff620)
go.k6.io/k6/execution/scheduler.go:472 +0xc6e
[...]

PR #1215, opened 2024-02-13
Fix data race with logging

ankur22:
## What?

Remove the use of the logrus formatter.

## Why?

1. When we use the loki output, it can cause a race condition since the loki hook works with the entry (the log message and metadata) asynchronously, which means that while the hook reads the entry, the formatter could be amending the same entry. The entry is not thread safe.
2. The formatter is only used in one place, which is when writing the output from console.log that is used in the website under test. This change could safely be moved to where it's used, which also makes it clear how the console message is formatted before being logged.

Full details can be found here: https://github.com/grafana/xk6-browser/issues/1066#issuecomment-1941946881

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1066


inancgumus (approved):
LGTM

PR #1232, opened 2024-03-04
Add test run id to RootModule

ankur22:
## What?

This adds the test run ID to the `RootModule` allowing any browser component to work with the test run ID.

## Why?

This is a requirement to allow better integration with external tools (such as faro)to be used to label sessions with a test run ID. The test run id will eventually be populated in the `window.k6` object which is injected into each and every `browserContext` and `page`.

The change localises the retrieval of the test run id in one place instead of having every component need to do a lookup for an env var.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/k6-cloud/issues/2066

### DIFF
--- a/browser/module.go
+++ b/browser/module.go
@@ -38,6 +38,7 @@ type (
 		initOnce       *sync.Once
 		tracesMetadata map[string]string
 		filePersister  filePersister
+		testRunID      string
 	}
 
 	// JSModule exposes the properties available to the JS script.
@@ -91,6 +92,7 @@ func (m *RootModule) NewModuleInstance(vu k6modules.VU) k6modules.Instance {
 				),
 				taskQueueRegistry: newTaskQueueRegistry(vu),
 				filePersister:     m.filePersister,
+				testRunID:         m.testRunID,
 			}),
 			Devices:         common.GetDevices(),
 			NetworkProfiles: common.GetNetworkProfiles(),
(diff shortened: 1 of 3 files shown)


env/env.go
```diff
@@ -81,6 +81,12 @@ const (
 	ScreenshotsOutput = "K6_BROWSER_SCREENSHOTS_OUTPUT"
 )
 
+const (
```
inancgumus, line 84: To be consistent?

```go
// Infrastructural.
const (
...
```

Something like that.

browser/module.go
```diff
@@ -126,6 +128,9 @@ func (m *RootModule) initialize(vu k6modules.VU) {
 	if err != nil {
 		k6ext.Abort(vu.Context(), "failed to create file persister: %v", err)
 	}
+	if e, ok := initEnv.LookupEnv(env.K6TestRunID); ok && e != "" {
+		m.testRunID = e
+	}
```
inancgumus, line 133: Can we test if we can get this environment variable when we set this?

inancgumus (commented):
LGTM so far.

env/env.go
ankur22: Added in e8987b56d662397bce3d0c4be7be5f991cf075c1

browser/module.go
ankur22: Yep, I will add an integration test in the next PR when the `windows.k6.testRunId` object is set.

inancgumus (approved):

PR #1233, opened 2024-03-05
Fix error for `evaluate` APIs

ankur22:
## What?

This fixes the error that is returned when a user works with the `evaluate` APIs when a navigation isn't occurring:

```js

  const blah = page.evaluate("1+1");
  console.log(blah);
  
  page.close();
```

This used to return `GoError: evaluating JS: execution context changed; most likely because of a navigation`, but now it returns `given expression does not evaluate to a function`.

## Why?

The new error better represents what the user can do to fix the problem instead of being left confused when no navigation has occurred. In the example above and with the new error, they should be able to work out that `evaluate` only works with callable functions.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6-cloud/issues/2066

### DIFF
--- a/common/execution_context.go
+++ b/common/execution_context.go
@@ -206,7 +206,9 @@ func (e *ExecutionContext) eval(
 	)
 	if remoteObject, exceptionDetails, err = action.Do(cdp.WithExecutor(apiCtx, e.session)); err != nil {
 		var cdpe *cdproto.Error
-		if errors.As(err, &cdpe) && cdpe.Code == -32000 {
+		if errors.As(err, &cdpe) && cdpe.Message == "Given expression does not evaluate to a function" {
+			err = errors.New("given expression does not evaluate to a function")
+		} else if errors.As(err, &cdpe) && cdpe.Code == -32000 {
 			err = errors.New("execution context changed; most likely because of a navigation")
 		}
 		return nil, err
--- a/common/execution_context.go
+++ b/common/execution_context.go
@@ -22,6 +22,9 @@ import (
 
 const evaluationScriptURL = "__xk6_browser_evaluation_script__"
 
+// This error code originates from chromium.
+const devToolsServerErrorCode = -32000
+
 var sourceURLRegex = regexp.MustCompile(`^(?s)[\040\t]*//[@#] sourceURL=\s*(\S*?)\s*$`)
 
 type executionWorld string
@@ -206,11 +209,13 @@ func (e *ExecutionContext) eval(
 	)
 	if remoteObject, exceptionDetails, err = action.Do(cdp.WithExecutor(apiCtx, e.session)); err != nil {
 		var cdpe *cdproto.Error
-		if errors.As(err, &cdpe) && cdpe.Message == "Given expression does not evaluate to a function" {
-			err = errors.New("given expression does not evaluate to a function")
-		} else if errors.As(err, &cdpe) && cdpe.Code == -32000 {
-			err = errors.New("execution context changed; most likely because of a navigation")
+		if errors.As(err, &cdpe) && cdpe.Code == devToolsServerErrorCode {
+			// By creating a new error instead of reusing it, we're removing the
+			// chromium specific error code.
+			return nil, errors.New(cdpe.Message)
 		}
+
+		e.logger.Warn("ExecutionContext:eval", "Unexpected DevTools server error: %v", err)
 		return nil, err
 	}
 	if exceptionDetails != nil {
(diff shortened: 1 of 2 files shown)


common/execution_context.go
```diff
@@ -206,7 +206,9 @@ func (e *ExecutionContext) eval(
 	)
 	if remoteObject, exceptionDetails, err = action.Do(cdp.WithExecutor(apiCtx, e.session)); err != nil {
 		var cdpe *cdproto.Error
-		if errors.As(err, &cdpe) && cdpe.Code == -32000 {
+		if errors.As(err, &cdpe) && cdpe.Message == "Given expression does not evaluate to a function" {
+			err = errors.New("given expression does not evaluate to a function")
+		} else if errors.As(err, &cdpe) && cdpe.Code == -32000 {
```
inancgumus, line 211: Is there a constant in the cdp package for this (-32000)?

tests/page_test.go
```diff
@@ -174,6 +174,84 @@ func TestPageEvaluate(t *testing.T) {
...
 
+func TestPageEvaluateMapping(t *testing.T) {
+	tests := []struct {
+		name    string
+		script  string
+		want    any
+		wantErr string
+	}{
+		{
+			name:   "()_=>_0",
```
inancgumus, line 185: Can we give these names descriptive names like `arrow_func`, `func_null_return`, `async_func`, `not_a_func`, etc.? It'll be harder to run them this way on the command-line and they don't tell us the goal of the test cases.

tests/page_test.go
```diff
@@ -174,6 +174,84 @@ func TestPageEvaluate(t *testing.T) {
...
+
+			_, rt, _, cleanUp := startIteration(t)
+			defer cleanUp()
+
+			// Test script as non string input
+			got, err := rt.RunString(fmt.Sprintf(`
+				const p = browser.newPage()
+				p.evaluate(%s)
+			`, tt.script))
+			if tt.wantErr != "" {
```
inancgumus, line 234: I would write two top-level tests to reduce the conditional boilerplate in this test.

```go
func TestPageEvaluateMapping(t *testing.T) {
   ... happy path ...
}

func TestPageEvaluateMappingError(t *testing.T) {
   ... sad path ...
}
```

This way, we can make the test more readable. No strong opinion.

inancgumus (approved):
LGTM 👍

Some weak opinion suggestions.

There is also a linter error:

```
TestPageEvaluateMapping should call t.Parallel on the top level as well as its subtests (parallel)
```

tests/page_test.go
ankur22: Amended in 972a8da8ade8e5e9cf667cc5ee865ab2e7099f5e

tests/page_test.go
ankur22: I like this approach. This is how i prefer to split happy and error paths too 👍 I've amended it in 972a8da8ade8e5e9cf667cc5ee865ab2e7099f5e

common/execution_context.go
ankur22: Unfortunately not, it seems to originate from chromium [here](https://chromium.googlesource.com/chromium/src/+/665cdea2f5f29e166ad455b5021154456944fcbc/headless/lib/browser/headless_devtools_manager_delegate.cc#46) and [here](https://chromium.googlesource.com/chromium/src/third_party/WebKit/Source/platform/v8_inspector/+/d94ef126e8fd904cedea7bca029216e42fe9eb6a/V8RuntimeAgentImpl.cpp#377) for the error message.


Take a look at this [file](https://chromium.googlesource.com/chromium/src/+/665cdea2f5f29e166ad455b5021154456944fcbc/headless/lib/browser/headless_devtools_manager_delegate.cc). After searching for `kErrorServerError`, which is the enum name of error code `-32000`, it looks like we're handling that error incorrectly. Maybe we should be returning the error message instead of `execution context changed; most likely because of a navigation`, WDYT?

inancgumus: Nice digging into the Chromium source code. Thanks for the links 👍 

I would, then, declare a constant to represent this in a `devToolsServerErrorCode` constant for clarity. And, for this part: `unexpected non server error: %`; what about the following: `Unexpected DevTools server error: %v`.  WDYT?

One last thing. We can return the first case without using an `else`.

ankur22: Resolved in 9277d6013373403399da56efea13461dde4ef39f

inancgumus (approved):

PR #1235, opened 2024-03-05
Refactor k6 object and add testRunID

ankur22:
## What?

The first few commits refactor how the `k6` object is assigned to the `window` object. Instead of it being done implicitly in the `NewBrowserContext` constructor, it is now done in the mapping layer after a new `browserContext` has been created, which is when `browser.newPage` and `browser.newContext` are called.

The last commit adds a new field in the `k6` object which is `testRunId`.

## Why?

* The refactor makes it more explicit when a init script is added, instead of it being hidden.
* It remove the difficult to expand embedded approach to adding the `k6` object to `window`.
* Adds `testRunId` to `k6` which enables external apps (such as Faro) to link back to the test run.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/k6-cloud/issues/2066


inancgumus (approved):
Nice!

PR #1238, opened 2024-03-07
Fix fill functionality for textarea

bandorko:
## What?

Fixes a bug, where calling fill() on textarea throws error. #1225 

## Why?

fill() should work for any contenteditable node. Currently it works only for ```<input>```

## Checklist

## Related PR(s)/Issue(s)

#1225

### DIFF
--- a/common/element_handle.go
+++ b/common/element_handle.go
@@ -270,7 +270,10 @@ func (h *ElementHandle) fill(_ context.Context, value string) error {
 	if !ok {
 		return fmt.Errorf("unexpected type %T", result)
 	}
-	if ok && s != resultDone {
+
+	if s == "needsinput" {
+		h.frame.page.Keyboard.InsertText(value)
+	} else if s != resultDone {
 		// Either we're done or an error happened (returned as "error:..." from JS)
 		return errorFromDOMError(s)
 	}
--- a/common/element_handle.go
+++ b/common/element_handle.go
@@ -24,6 +24,7 @@ import (
 )
 
 const resultDone = "done"
+const resultNeedsInput = "needsinput"
 
 type (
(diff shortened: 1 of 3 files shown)


iandesj:
I would test that this works for `input` elements, too.

bandorko:
@iandesj 

Of course you can try it, if you build a custom k6 binary from the code on the branch. Fill() for **<input>** elements has worked so far, and it already has a test at https://github.com/grafana/xk6-browser/blob/main/tests/locator_test.go#L99-L105
If you find any scenario that fails with this implementation, I would be happy to deal with it.

common/element_handle.go
```diff
@@ -270,7 +270,10 @@ func (h *ElementHandle) fill(_ context.Context, value string) error {
 	if !ok {
 		return fmt.Errorf("unexpected type %T", result)
 	}
-	if ok && s != resultDone {
+
+	if s == "needsinput" {
```
inancgumus, line 274: This is great. Can you put this in a constant (like `resultDone`)?

inancgumus (dismissed):
🥳 Thanks!!

ankur22 (approved):
LGTM 🚀

common/element_handle.go
bandorko: done

inancgumus (approved):

iandesj:
Thanks @ankur22, @inancgumus, and of course @bandorko 👏🏻 

What does the process look like for getting this into the next k6 release? So I know what to expect.

ankur22:
> What does the process look like for getting this into the next k6 release? So I know what to expect.

It will be in the next version of k6 (`v0.50.0`) which should be released in the next couple of weeks 🤞

PR #1242, opened 2024-03-12
Fix DocumentFragment XPath evaluate

ankur22:
## What?

This adds javascript code to convert a `DocumentFragment` into a `Document`.

## Why?

`DocumentFragment`s do not implement `evaluate` and so do not work with `XPath` selectors. When a `DocumentFragment` is encountered it will convert that to a `Document` so that `evalaute` with `XPath` selectors can be performed.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1243

### DIFF
--- a/common/element_handle.go
+++ b/common/element_handle.go
@@ -23,8 +23,10 @@ import (
 	"github.com/grafana/xk6-browser/k6ext"
 )
 
-const resultDone = "done"
-const resultNeedsInput = "needsinput"
+const (
+	resultDone       = "done"
+	resultNeedsInput = "needsinput"
+)
 
 type (
 	elementHandleActionFunc        func(context.Context, *ElementHandle) (any, error)
--- a/common/js/injected_script.js
+++ b/common/js/injected_script.js
@@ -124,6 +124,11 @@ class XPathQueryEngine {
       selector = "." + selector;
     }
     const result = [];
+
+    if (root instanceof DocumentFragment) {
+      root = convertToDocument(root);
+    }
+
     const document = root instanceof Document ? root : root.ownerDocument;
     if (!document) {
       return result;
--- a/common/js/injected_script.js
+++ b/common/js/injected_script.js
@@ -125,6 +125,13 @@ class XPathQueryEngine {
     }
     const result = [];
 
+    // DocumentFragments cannot be queried with XPath and they do not implement
+    // evaluate. It first needs to be converted to a Document before being able
+    // to run the evaluate against it.
+    //
+    // This avoids the following error:
+    // - Failed to execute 'evaluate' on 'Document': The node provided is
+    //   '#document-fragment', which is not a valid context node type.
     if (root instanceof DocumentFragment) {
       root = convertToDocument(root);
     }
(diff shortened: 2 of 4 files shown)


tests/page_test.go
```diff
@@ -1633,6 +1633,7 @@ func TestPageIsHidden(t *testing.T) {
 	}
 }
 
+//nolint:tparallel
```
ankur22, line 1636: Couldn't find a way of running the subtests in parallel. The test server seemed to close before any of the tests ran 🤔

common/element_handle.go
```diff
@@ -23,8 +23,10 @@ import (
 	"github.com/grafana/xk6-browser/k6ext"
 )
 
-const resultDone = "done"
-const resultNeedsInput = "needsinput"
+const (
+	resultDone       = "done"
+	resultNeedsInput = "needsinput"
+)
```
ankur22, line 29: Not sure why i'm having to do this in this PR, which is unrelated 🤷

tests/page_test.go
ankur22: `t.Cleanup` helped here.

common/element_handle.go
bandorko: I caused this issue. I apologize for that.

ankur22: No worries 😄

inancgumus (commented):
Nice fix! Could you explain (or add comments) the changes to the `injected_script.js`? Can you help me understand the goals of these changes and how they all work? (I don't have specific questions to ask because I need some general guidance to understand it. I understand that we're "somehow" converting DocumentFragment to Document).

ankur22:
> Nice fix! Could you explain (or add comments) the changes to the injected_script.js? Can you help me understand the goals of these changes and how they all work? (I don't have specific questions to ask because I need some general guidance to understand it. I understand that we're "somehow" converting DocumentFragment to Document).

I've tried to tidying up the comments a bit so that it makes a bit more sense, as well as adding some more context. The details on the errors can be found in the [issue](https://github.com/grafana/xk6-browser/issues/1243) itself (including the comments). Let me know if this helps or not.

common/js/injected_script.js
```diff
@@ -143,6 +155,43 @@ class XPathQueryEngine {
...
+
+  copyNodesToDocument(fragment, newDoc.body);
+
+  return newDoc;
+}
+
+// copyNodesToDocument manually copies nodes to a new document, excluding
+// ShadowRoot nodes -- ShadowRoot are not cloneable so we need to manually
+// clone them one element at a time.
+function copyNodesToDocument(sourceNode, targetNode) {
```
inancgumus, line 172: Should we add a "too deep" error here? Think about what would happen if the element is super hierarchical that might clog our system. That might lead to instability issues.

inancgumus (approved):
Thanks for the great explanations 🙇 Now, it's super clear for me.

common/js/injected_script.js
ankur22: Yeah, that's a good point. The issue is how far deep do we allow this to go? This is where my lack of frontend knowledge hinders me in working out a good valid depth. Maybe it's ok to allow it to go until an OOM error occurs, and work backwards from there? Any ideas?

ankur22: I'm going to merge this in, but happy to discuss and work on this if you think it's an issue. My guess is that we might not need to worry for majority of use cases. I would imagine that Chrome would detect layouts which could go too deep before the browser module gets a chance.

inancgumus: Sure, let's wait for it to happen first.

inancgumus: @ankur22 

Sure, let's wait for it to happen first. I hope we can detect this is the cause if it ever happens 🤞

Issue #986, opened 2023-07-31
Frame requestByID panics when initializing networking


inancgumus:
### Brief summary

There is a flaky bug in `(*Frame).requestByID` that happens while initializing `(*NetworkManager).initEvents`. The test run ID is 155045.

### xk6-browser version

0fcdd36

### OS

Linux

### Steps to reproduce the problem

Run the following script on a resource restricted platform _several times_ as the issue is flaky.

<details>
<summary>The script</summary>

```js
import { browser } from 'k6/experimental/browser'
import { check } from 'k6'

export const options = {
  scenarios: {
    browser: {
      executor: 'per-vu-iterations',
      maxDuration: '15m',
      vus: 5,
      iterations: 10,
      options: {
        browser: {
          type: 'chromium',
        },
      },
    },
  }
}
export default async function () {
  const page = browser.newPage();

  try {
    // User journy in SPA application
    await page.goto('https://.../'); // TODO: change it to the otel app

    const productCard = page.locator('(//div[@data-cy="product-card"])[1]')
    productCard.click()

    const quantityOption = page.locator('[data-cy="product-quantity"]')
    quantityOption.selectOption('3')
    const addToCardBtn = page.locator('[data-cy="product-add-to-cart"]')
    addToCardBtn.click()

    check(page, {
      'cart item name': page => page.locator('//p[text()="National Park Foundation Explorascope"]').isVisible() === true,
      'cart item quantity': page => page.locator('//p[text()="3"]').isVisible() === true
    })

    // Check GH link to force another navigation
    await Promise.all([
      page.waitForNavigation(),
      page.locator('footer a').click(),
    ]);
    check(page, {
      'gh-author': page.locator(
        'div#repository-container-header span.author a'
      ).textContent().includes('open-telemetry'),
    });
  } finally {
    page.close();
  }
}
```

</details>

### Expected behaviour

No panics.

### Actual behaviour

```go
goroutine 246996 [running]:
github.com/grafana/xk6-browser/common.(*Frame).requestByID(0xc0089428c0, {0xc009718454, 0x7})
	github.com/grafana/xk6-browser@v1.0.0-rc0.0.20230726091449-0fcdd3618f29/common/frame.go:365 +0x5a
github.com/grafana/xk6-browser/common.(*FrameManager).requestFailed(0xc001d700c0, 0xc0055d4640, 0x0)
[...]

PR #1246, opened 2024-03-18
Remove logging in-flight requests

inancgumus:
## What?

Remove logging in-flight requests when a request is failed.

## Why?

We noticed we shouldn't find an internally failed request, as it can cause race condition issues within the module's internal components. Also, we've decided that logging in-flight requests only adds noise without bringing any tangential value.

We keep logging the failed request, though.

## Checklist

## Related PR(s)/Issue(s)

Updated #986


ankur22 (approved):
LGTM 🚀

Issue #1249, opened 2024-03-21
Label spans that have errors attached


ankur22:
We already have spans for some of our APIs, such as `page.screenshot`. This allows us to visualise the test run and all the metrics that go along with a particular navigation. What we're currently missing are the errors on the spans. For example, when `page.screenshot` is called, and an error is received from a remote server (when uploading screenshots to a remote server) we should mark the span with the error.
```[tasklist]
### Tasks
- [ ] https://github.com/grafana/xk6-browser/pull/1259
- [ ] https://github.com/grafana/xk6-browser/pull/1260
- [ ] https://github.com/grafana/k6/pull/3709
```

ankur22:
I think this involves the following calls:

```go
	span.SetStatus(codes.Error, "rate limit reached")
	span.RecordError(errors.New("429: Too Many Requests"))
```

We need to check the best practices thought to make sure.

PR #1260, opened 2024-04-05
Add errors to spans

ankur22:
## What?

Adding errors to the API spans that currently exist.

## Why?

When a user views the traces for the test, it's useful to see whether an error occurred and in which API call. This will help highlight spans where errors occurred and therefore caused the test iteration to fail.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/xk6-browser/issues/1249

### DIFF
--- a/common/browser.go
+++ b/common/browser.go
@@ -559,21 +559,21 @@ func (b *Browser) NewContext(opts goja.Value) (*BrowserContext, error) {
 	browserContextID, err := action.Do(cdp.WithExecutor(b.ctx, b.conn))
 	b.logger.Debugf("Browser:NewContext", "bctxid:%v", browserContextID)
 	if err != nil {
-		err = fmt.Errorf("creating browser context ID %s: %w", browserContextID, err)
+		err := fmt.Errorf("creating browser context ID %s: %w", browserContextID, err)
 		SpanRecordError(span, "browserContext creation in Chrome failed", err)
 		return nil, err
 	}
 
 	browserCtxOpts := NewBrowserContextOptions()
 	if err := browserCtxOpts.Parse(b.ctx, opts); err != nil {
-		err = fmt.Errorf("parsing newContext options: %w", err)
+		err := fmt.Errorf("parsing newContext options: %w", err)
 		SpanRecordError(span, "new browserContext options parsing failed", err)
 		return nil, err
 	}
 
 	browserCtx, err := NewBrowserContext(b.ctx, b, browserContextID, browserCtxOpts, b.logger)
 	if err != nil {
-		err = fmt.Errorf("new context: %w", err)
+		err := fmt.Errorf("new context: %w", err)
 		SpanRecordError(span, "new browserContext creation failed", err)
 		return nil, err
 	}
@@ -592,7 +592,7 @@ func (b *Browser) NewPage(opts goja.Value) (*Page, error) {
 
 	browserCtx, err := b.NewContext(opts)
 	if err != nil {
-		err = fmt.Errorf("new page: %w", err)
+		err := fmt.Errorf("new page: %w", err)
 		SpanRecordError(span, "new browserContext creation failed", err)
 		return nil, err
 	}
--- a/common/browser.go
+++ b/common/browser.go
@@ -551,7 +551,7 @@ func (b *Browser) NewContext(opts goja.Value) (*BrowserContext, error) {
 
 	if b.context != nil {
 		err := errors.New("existing browser context must be closed before creating a new one")
-		spanRecordError(span, "browserContext already exists", err)
+		spanRecordError(span, err)
 		return nil, err
 	}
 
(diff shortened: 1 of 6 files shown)


common/browser.go
```diff
@@ -550,24 +550,32 @@ func (b *Browser) NewContext(opts goja.Value) (*BrowserContext, error) {
...
+		SpanRecordError(span, "browserContext already exists", err)
+		return nil, err
 	}
 
 	action := target.CreateBrowserContext().WithDisposeOnDetach(true)
 	browserContextID, err := action.Do(cdp.WithExecutor(b.ctx, b.conn))
 	b.logger.Debugf("Browser:NewContext", "bctxid:%v", browserContextID)
 	if err != nil {
-		return nil, fmt.Errorf("creating browser context ID %s: %w", browserContextID, err)
+		err = fmt.Errorf("creating browser context ID %s: %w", browserContextID, err)
```
andrewslotin, line 562:
```suggestion
		err := fmt.Errorf("creating browser context ID %s: %w", browserContextID, err)
```
Does this wrapped error have meaning outside of this block? If not, I'd suggest to shadow it here and in similar blocks below to avoid overriding the original value.

common/page.go
```diff
@@ -874,7 +878,13 @@ func (p *Page) Goto(url string, opts *FrameGotoOptions) (*Response, error) {
 	)
 	defer span.End()
 
-	return p.MainFrame().Goto(url, opts)
+	resp, err := p.MainFrame().Goto(url, opts)
+	if err != nil {
+		SpanRecordError(span, "goto failed", err)
```
andrewslotin, line 883: Would it make sense to add some context to the error here?

andrewslotin (commented):
Left a few minor comments. @inancgumus, WDYT?

common/browser.go
inancgumus: 👍 Shadowing it here can also prevent unintentional bugs if we never return here. The linter will catch that the shadowed variable is not used.

inancgumus (commented):
I'm happy with the overall PR 👍

I have some concerns:
1. Why is `SpanRecordError` exported?
2. Can we use the existing error? Is there a reason for repurposing a new error message? Should we repurpose the original error and directly use it instead?
3. Weak suggestion: Can we register span errors in the mapping layer (instead of intermingling or handling them into the core logic)?

common/page.go
inancgumus: I believe the inner code already adds the necessary context: https://github.com/grafana/xk6-browser/blob/e4c6ed3a612ffe5f395804e825b2010800d7bc24/common/frame.go#L973-L987.

common/browser.go
ankur22: Good point, fixed in 5b93c3a95c2761ff55e3f6e76a21b962744afee7

common/page.go
ankur22: Yeah, exactly, the called method already returns the full context of the issue. I can wrap it in something like `err := fmt.Errorf("navigating page to %q: %w", url, err)`, to help identify that it was `page` calling `frame`, but it feels repetitive. WDYT?

ankur22:
> I'm happy with the overall PR 👍
> 
> I have some wonderings:
> 
> 1. Why is `SpanRecordError` exported?
> 2. Can we use the existing error? Is there a reason for repurposing a new error message? Should we repurpose the original error and directly use it instead?
> 3. Weak suggestion: Can we register span errors in the mapping layer (instead of intermingling or handling them into the core logic)?

1. Good spot, i copied the format from the other APIs in the file. Fixed in c0c2a6eb7220084fc59170ec65e43305b33184cf
2. I'm not sure what you mean here, Could you give an example?
3. This is something I was thinking about too. I think it's worth experimenting but in another PR. It would involve changing the method signature to always expect a `context.Context`, which is generally good, but also confusing in our case since the context would only carry the trace information and not be used to terminate the iteration/method 🤔

common/page.go
inancgumus: I think the existing error is good enough.

inancgumus:
@ankur22 

> 1. Good spot, i copied the format from the other APIs in the file. Fixed in [c0c2a6e](https://github.com/grafana/xk6-browser/commit/c0c2a6eb7220084fc59170ec65e43305b33184cf)

👍

> 2. I'm not sure what you mean here, Could you give an example?

Sure. I mean this (using the existing error message):

```go
err := errors.New("existing browser context must be closed before creating a new one")
spanRecordError(span, err)
return nil, err
```

Instead of:

```go
err := errors.New("existing browser context must be closed before creating a new one")
spanRecordError(span, "browserContext already exists", err)
return nil, err
```

In the second version, there are two error messages:
1. "existing browser context must be closed before creating a new one"
2. "browserContext already exists"

I thought maybe we could use a single error message. If the existing error message is too verbose, we could repurpose the existing ones to be compatible with span error messages. Keeping the error messages consistent will prevent confusions in the future, and also lead to less code.


> 3. This is something I was thinking about too. I think it's worth experimenting but in another PR. It would involve changing the method signature to always expect a `context.Context`, which is generally good, but also confusing in our case since the context would only carry the trace information and not be used to terminate the iteration/method 🤔

`Context` is also used to carry information. So, it's fine to experiment with it.

ankur22:
> > 2. I'm not sure what you mean here, Could you give an example?
> 
> Sure. I mean this (using the existing error message):
> 
> ```go
> err := errors.New("existing browser context must be closed before creating a new one")
> spanRecordError(span, err)
> return nil, err
> ```
> 
> Instead of:
> 
> ```go
> err := errors.New("existing browser context must be closed before creating a new one")
> spanRecordError(span, "browserContext already exists", err)
> return nil, err
> ```
> 
> In the current version, there are two error messages:
> 
> 1. "existing browser context must be closed before creating a new one"
> 2. "browserContext already exists"
> 
> I thought maybe we could use a single error message. If the existing error messages are too verbose, we could repurpose them to be compatible with span error messages. Keeping the error messages consistent will prevent confusion in the future and also lead to less code.

I see now. Adding a description of the error seems to be a requirement when recording the error in the span:

```
func spanRecordError(span trace.Span, description string, err error) {
	span.SetStatus(codes.Error, description)
	span.RecordError(err)
}
```

Any other options?

inancgumus:
@ankur22 

> I see now. Adding a description of the error seems to be a requirement when recording the error in the span.

Yes, but the description doesn't have to be a different error. [For example](https://github.com/grafana/grafana/blob/b28e6bc5d8547eda9754dac94a3faa4bb340e360/pkg/tsdb/tempo/trace.go#L51-L57).

ankur22:
> Yes, but the description doesn't have to be a different error. [For example](https://github.com/grafana/grafana/blob/b28e6bc5d8547eda9754dac94a3faa4bb340e360/pkg/tsdb/tempo/trace.go#L51-L57).

Nice find, if that's what they're doing then I will do the same 👍  Changed in 2207e4b25cc0b9ae665cbb7f6b93fb8b6c39b792

inancgumus (approved):

Issue #1089, opened 2023-10-31
Dispose panic after clicking on an element inside another frame


inancgumus:
### Brief summary

We receive a panic after clicking on an element inside another frame.

#### Early investigation result

The issue seems to be happening because we dispose of the main document after closing the browser with `Runtime.releaseObject`.

### xk6-browser version

v1.1.0

### Steps to reproduce the problem

1. Turn the headful browser option on.
2. Run the following script.

```js
export const options = {
  scenarios: {
    ui: {
      // mandatory fields; executor, options-->browser-->type
      executor: "shared-iterations",
      vus: 1,
      iterations: 1,
      options: {
        browser: {
          type: "chromium",
        },
      },
    },
  },
};

export default async function () {
  const page = browser.newPage();

  try {
    await page.goto("https://www.directline.com/my-dashboard#/sign-in", {
      waitUntil: "networkidle",
    });

    await Promise.all([
      page.locator('button[id="onetrust-accept-btn-handler"]').click(),
    ]);

    let frames = await page.frames();

    for (const frame of frames) {
      if (frame.title() === "reCAPTCHA" && frame.name().startsWith("a")) {
        const spanElement = frame.locator("#recaptcha-anchor");
        spanElement.focus();
        spanElement.dispatchEvent("click");
      }
    }
  } finally {
    page.close();
  }
}
```

### Expected behaviour

No panic.

### Actual behaviour

Panics with:

```go
panic: GoError: dispose: canceled

goroutine 100 [running]:
go.k6.io/k6/js/common.Throw(...)
	go.k6.io/k6@v0.45.1-0.20230804150043-5b5c5ccbb869/js/common/util.go:20
github.com/grafana/xk6-browser/k6ext.Panic.func1(0x101441de0?, {0x14002b0a9a0?, 0x101426601?, 0x14001719300?})
	github.com/grafana/xk6-browser@v0.10.0/k6ext/panic.go:35 +0x7c
github.com/grafana/xk6-browser/k6ext.sharedPanic({0x10167dbf8, 0x14001109540}, 0x14002663d30, {0x14002b0a9a0?, 0x1, 0x1})
	github.com/grafana/xk6-browser@v0.10.0/k6ext/panic.go:64 +0x1fc
[...]

PR #1261, opened 2024-04-05
Fix dispose context canceled errors

ankur22:
## What?

Ignore context canceled errors when attempting to dispose of a `frame`.

## Why?

When testing websites which contain `iframe`s, they can cause a `panic: GoError: dispose: canceled` panic. This is confusing behaviour since after investigating the issue it turns out that the dispose doesn't need to occur since the `page` the `iframe` is associated to is already closed. More details can be found in this [comment](https://github.com/grafana/xk6-browser/issues/1089#issuecomment-2040220958).

It's worth noting that this is more of a work around. A better solution might be to not end the iteration until all the CDP messages between k6 and Chrome ended successfully, but this would require quite a bit of refactoring and possibly more control over the iteration.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/xk6-browser/issues/1089

### DIFF
--- a/common/js_handle.go
+++ b/common/js_handle.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/grafana/xk6-browser/k6ext"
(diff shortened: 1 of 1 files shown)


common/js_handle.go
```diff
@@ -79,7 +80,13 @@ func (h *BaseJSHandle) AsElement() *ElementHandle {
 
 // Dispose releases the remote object.
 func (h *BaseJSHandle) Dispose() {
+	fmt.Println(">>>>>>: BaseJSHandler Dispose")
 	if err := h.dispose(); err != nil {
+		if errors.Is(err, context.Canceled) {
+			fmt.Println(">>>>>>: Skipping context canceled error on Dispose", err)
+			return
+		}
```
inancgumus, line 88: I'm [happy](https://github.com/grafana/xk6-browser/issues/1089#issuecomment-2056669729) with this approach 👍

inancgumus (approved):

Issue #1150, opened 2024-01-10
Fix flaky tracing test that fails with web_vital span err


inancgumus:
### Brief summary

When the integration tests are ran in CI, the workflow often fails on one of the [TestTracing](https://github.com/grafana/xk6-browser/blob/af54d56920a7f52b1e647c09b8fc3de3d870793f/tests/tracing_test.go#L54), specifically the [web_vital](https://github.com/grafana/xk6-browser/blob/af54d56920a7f52b1e647c09b8fc3de3d870793f/tests/tracing_test.go#L111) sub test.

When this test fails we can usually rerun the failed workflow and the test will pass the second time round, so it's flakey. It would be good if this wasn't the case so that it gives us confidence in our tests and workflows, as well as making it easier to contribute to for external contributors.

### xk6-browser version

fcc7f9e5f45b099d8516cb7cc38662b9110b3c23

### OS

NA

### Chrome version

NA

### Docker version and image (if applicable)

NA

### Steps to reproduce the problem

Run the test in CI, e.g.: https://github.com/grafana/xk6-browser/actions/runs/8138217177/job/22238484745?pr=1228

### Expected behaviour

The integration test passes.

### Actual behaviour

It fails with:

```bash
--- FAIL: TestTracing (3.71s)
    tracing_test.go:161: 
        	Error Trace:	/home/runner/work/xk6-browser/xk6-browser/tests/tracing_test.go:161
        	Error:      	Received unexpected error:
        	            	"web_vital" span was not found
        	Test:       	TestTracing
```
```[tasklist]
### Tasks
- [ ] https://github.com/grafana/xk6-browser/pull/1262
- [ ] https://github.com/grafana/k6/pull/3709
```

inancgumus:
It happens every other day. Here's [the latest one](https://github.com/grafana/xk6-browser/actions/runs/8151736116/attempts/1):

```
time="2024-03-05T06:01:29Z" level=error msg="error on evaluating window.k6SpanId: evaluating JS in global context: context canceled" category="FrameSession:onFrameNavigated" elapsed="0 ms" source=browser
time="2024-03-05T06:01:29Z" level=error msg="Failed to load resource: the server responded with a status of 404 (Not Found)" browser_source=network line_number=0 source=browser stacktrace="<nil>" url="http://127.0.0.1:35093/static/empty.html"
[...]

PR #1262, opened 2024-04-08
Fix web vital trace test

ankur22:
## What?

Catch the web vital spans after the `page.close` API call is made. This should definitely cause the web vitals to emit and be received in time before the sub-test completes.

## Why?

Instead of a sleep to wait for a web vital, which is problematic on slower machines that can cause the sub test to fail, catch the web vital spans that are emitted/received when the page.close() API is called.

I've ran the integration tests 10 times and there were no failures with the tracing tests, whereas usually I would get 2-3 failures.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/xk6-browser/issues/1150

### DIFF
--- a/tests/tracing_test.go
+++ b/tests/tracing_test.go
@@ -108,13 +108,6 @@ func TestTracing(t *testing.T) {
 			},
 		},
 		{
-			name: "web_vital",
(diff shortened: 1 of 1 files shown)


inancgumus (approved):
Nice 👍

PR #1264, opened 2024-04-10
Remove unimplemented apis

ankur22:
## What?

Removing the unimplemented API function bodies.

## Why?

It's confusing when auditing APIs to have some that do not have any implementation. It's as if they are a priority, which is not the case. We will add the APIs that we think we need and when the community asks for them.

## Checklist

### DIFF
--- a/browser/mapping.go
+++ b/browser/mapping.go
@@ -832,17 +832,9 @@ func mapBrowserContext(vu moduleVU, bc *common.BrowserContext) mapping { //nolin
 		},
 		"setDefaultNavigationTimeout": bc.SetDefaultNavigationTimeout,
 		"setDefaultTimeout":           bc.SetDefaultTimeout,
-		"setExtraHTTPHeaders": func(headers map[string]string) *goja.Promise {
-			ctx := vu.Context()
-			return k6ext.Promise(ctx, func() (result any, reason error) {
-				err := bc.SetExtraHTTPHeaders(headers)
-				panicIfFatalError(ctx, err)
-				return nil, err //nolint:wrapcheck
-			})
-		},
-		"setGeolocation":     bc.SetGeolocation,
-		"setHTTPCredentials": bc.SetHTTPCredentials, //nolint:staticcheck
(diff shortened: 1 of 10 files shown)


allansson (changes requested):
I've built and run the extension and it works normally. 

It's hard to test that something isn't there, but looking at the code I don't see any methods removed that have an implementation and there would definitely be a build error if some reference was left.

I did however find three other methods that don't have an implementation:

- https://github.com/grafana/xk6-browser/blob/a53651b8f89867bcb7ed5354ceb587e3f9ac1fd5/common/browser_context.go#L261
- https://github.com/grafana/xk6-browser/blob/a53651b8f89867bcb7ed5354ceb587e3f9ac1fd5/chromium/browser_type.go#L220
- https://github.com/grafana/xk6-browser/blob/a53651b8f89867bcb7ed5354ceb587e3f9ac1fd5/common/page.go#L820

I'm sure if there's a reason why these should remain, but I just wanted to point it out.

inancgumus:
👍 

> setExtraHTTPHeaders and LaunchPersistentContext

It seems safe to remove these ones.

> goBack

@ankur22 probably hasn't removed `goBack` because [this test](https://github.com/grafana/xk6-browser/blob/a53651b8f89867bcb7ed5354ceb587e3f9ac1fd5/tests/browser_test.go#L270-L314) requires it. However, we can easily implement an _undocumented_ method that panics. And then we can remove `GoBack`.

ankur22:
> setExtraHTTPHeaders and LaunchPersistentContext

Yep, good spot, I've removed them in 4a2e5848318014e0f36c57cb83b73630ed22f1db.

> goBack

As @inancgumus mentioned it was indeed due to the `goback` method, but I've removed that now and changed the test in f9d21900e843fc233ba572f362b7cb51ddb0445c to still behave in the same way (apart from creating the browserContext and page first).

inancgumus (approved):

allansson (approved):
👍

Issue #1317, opened 2024-05-09
`Mouse` is buggy


inancgumus:
### What?

Fix the `Mouse` class. [The current version](https://github.com/grafana/xk6-browser/blob/991c6995b7c4e77205da15d935c42d7ebb34e916/common/mouse.go) is buggy.

### Why?

The `Mouse` class has several problems:
* The `up` and `down` methods should not take x and y coordinates. Instead, they should dispatch the events on the current mouse position.
* The `move` method doesn't work and instead dispatches a `down` event!

### Tasks

```[tasklist]
### Tasks
- [ ] #1318
- [x] Update the k6 release notes
- [x] Rebase `main-async`
```

PR #1318, opened 2024-05-09
Fix `mouse`

inancgumus:
## What?

See #1317

## Checklist


tests/mouse_test.go
```diff
@@ -5,30 +5,104 @@ import (
...
+
+		// Simulate a click at the button coordinates
+		box := button.BoundingBox()
+		m.Click(box.X, box.Y, nil)
+
+		// Verify the button's text changed
+		assert.Equal(t, "Clicked!", button.TextContent())
+	})
+
+	t.Run("double_click", func(t *testing.T) {
```
ankur22, line 36: I think this test should also ensure that a click has occurred twice. There are two things that need to occur:

1. `ondblclick` event is fired.
2. the element is clicked twice.

Before https://github.com/grafana/xk6-browser/pull/1121, the `ondblclick` event was fired, but the element wasn't being clicked on twice. That's why there was a counter in the test to ensure that the element was clicked on twice.

So we need to add something into the test to ensure that the element is in fact being clicked on twice too.

ankur22 (approved):
LGTM.

I think the `dblclick` test needs to be amended though.

tests/mouse_test.go
inancgumus: 👍 [8beba15](https://github.com/grafana/xk6-browser/pull/1318/commits/8beba159aaf0f124f72c45404bd3035a3d38cc8e)

ankur22 (approved):
LGTM 🚀

PR #1325, opened 2024-05-09
Add hack to reload page

ankur22:
## What?

This seems to work to get around the issue where the new window never loads.

The next step should be to find a CDP command that works that isn't reload.

We still need to keep looking for the real solution though.

## Checklist


ankur22:
We found the correct fix.

Issue #827, opened 2023-03-14
Clicking on a link that opens a new tab never navigates to the href link


ankur22:
### Brief summary

When a link that opens in a new tab is clicked on, the navigation never completes in the new tab and seems to hang on the `about:blank` page.

### xk6-browser version

v0.43.1

### OS

Version 13.2 (22D49)

### Chrome version

Version 110.0.5481.177 (Official Build) (arm64)

### Docker version and image (if applicable)

NA

### Steps to reproduce the problem

This is the script that I'm trying to test, which should navigate to the reviews section in a new tab when the link has been clicked on:

```js
import { browser } from "k6/x/browser";

export const options = {
  scenarios: {
    ui: {
      executor: "shared-iterations",
      options: {
        browser: {
          type: "chromium",
        },
      },
    },
  }
};

export default async function () {
  const page = browser.newPage();

  try {
    await page.goto("https://k6.io", { waitUntil: "networkidle" });

    await Promise.all([
      page.waitForNavigation(),
      page.locator('a[href="https://www.g2.com/products/k6/reviews"]').click(),
    ]);
  } finally {
    page.close();
  }
}
```

In the new tab it never navigates to the link.

<img width="742" alt="Screenshot 2023-03-14 at 11 23 18" src="https://user-images.githubusercontent.com/1112428/224986337-d778fdb8-a0b9-458b-8e5b-eb46980bfb3f.png">

The output of the test is:

```bash
ERRO[0030] communicating with browser: websocket: close 1006 (abnormal closure): unexpected EOF  category=cdp elapsed="0 ms" goroutine=51
panic: GoError: creating a new page: initializing networking *network.EnableParams: websocket: close 1006 (abnormal closure): unexpected EOF
running (00m30.0s), 1/1 VUs, 0 complete and 0 interrupted iterations
goroutine 54 [running]:--------------------------] 1 VUs  00m30.0s/10m0s  0/1 shared iters
go.k6.io/k6/js/common.Throw(...)
        go.k6.io/k6@v0.42.1-0.20230130080633-582ec4d3940c/js/common/util.go:20
github.com/grafana/xk6-browser/k6ext.Panic.func1(0x0?, {0x14000abcd70?, 0x14002464b54?, 0x140006719c0?})
        github.com/grafana/xk6-browser@v0.0.0-00010101000000-000000000000/k6ext/panic.go:35 +0x7c
[...]

PR #1334, opened 2024-05-15
Fix navigate to link in new page

ankur22:
## What?

Move the CDP request `RunIfWaitingForDebugger` to be called first when creating a `FrameSession`.

## Why?

We have identified that the call to `RunIfWaitingForDebugger` wasn't being performed at the correct time when a new page was created after a link was clicked on that opened the page.

This is currently the simplest and safest solution. I tried to place some of the CDP actions that are in `NewFrameSession` within their own goroutines but there were data races.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/827

### DIFF
--- a/tests/page_test.go
+++ b/tests/page_test.go
@@ -1707,7 +1707,7 @@ func TestPageTargetBlank(t *testing.T) {
 	tb.withHandler("/home", func(w http.ResponseWriter, _ *http.Request) {
 		_, err := w.Write([]byte(
 			`<!DOCTYPE html><html><head></head><body>
-				<h1>click please</h1><a href="/link" target="_blank">click me</a>
+				<a href="/link" target="_blank">click me</a>
 			</body></html>`,
 		))
 		require.NoError(t, err)
@@ -1737,21 +1737,19 @@ func TestPageTargetBlank(t *testing.T) {
 	err = p.Click("a[href='/link']", common.NewFrameClickOptions(p.Timeout()))
 	require.NoError(t, err)
 
-	// Page count should now be 2.
-	for i := 0; i < 5; i++ {
-		pp = p.Context().Pages()
-		if len(pp) == 2 {
-			break
-		}
-		time.Sleep(time.Millisecond * 500)
-	}
+	// Wait for the page to be created and for it to navigate to the link.
+	obj, err := p.Context().WaitForEvent("page", nil, common.DefaultTimeout)
+	require.NoError(t, err)
+	p2, ok := obj.(*common.Page)
+	require.True(t, ok, "return from WaitForEvent is not a Page")
+
+	p2.WaitForLoadState(common.LifecycleEventLoad.String(), nil)
+
+	// Now there should be 2 pages.
+	pp = p.Context().Pages()
 	assert.Equal(t, 2, len(pp))
 
-	// Make sure the new page is in focus and assert this.
-	// At the moment pages() will not always return the pages in the expected order
-	// (the order they were opened in -- https://github.com/grafana/xk6-browser/issues/444)
-	// so we must check both pages for the expected header.
-	p0 := pp[0].MainFrame().InnerHTML("h1", nil)
-	p1 := pp[1].MainFrame().InnerHTML("h1", nil)
-	assert.True(t, p0 == "you clicked!" || p1 == "you clicked!", "Neither page had the expected h1 text")
+	// Make sure the new page contains the correct page.
+	got := p2.InnerHTML("h1", nil)
+	assert.Equal(t, "you clicked!", got)
 }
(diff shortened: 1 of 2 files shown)


tests/page_test.go
```diff
@@ -1699,3 +1699,59 @@ func TestShadowDOMAndDocumentFragment(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Page count should now be 2.
+	for i := 0; i < 5; i++ {
+		pp = p.Context().Pages()
+		if len(pp) == 2 {
+			break
+		}
+		time.Sleep(time.Millisecond * 500)
+	}
```
inancgumus, line 1747: This test might become a possible flaky test.

Instead of using a loop and `time.Sleep`, what about using:
* `p.Context().WaitForEvent("page", ...)` and,
* `WaitForLoadState` for the newly loaded page?

tests/page_test.go
```diff
@@ -1699,3 +1699,59 @@ func TestShadowDOMAndDocumentFragment(t *testing.T) {
...
+	assert.Equal(t, 2, len(pp))
```
inancgumus, line 1748: Shouldn't this be `require`. Otherwise, the code below might panic.

tests/page_test.go
```diff
@@ -1699,3 +1699,59 @@ func TestShadowDOMAndDocumentFragment(t *testing.T) {
...
+
+	p := tb.NewPage(nil)
+
+	// Navigate to the page with a link that opens a new page.
+	opts := &common.FrameGotoOptions{
+		Timeout: common.DefaultTimeout,
+	}
+	resp, err := p.Goto(tb.url("/home"), opts)
+	require.NoError(t, err)
+	require.NotNil(t, resp)
```
inancgumus, line 1730: Why do we need `resp`?

tests/page_test.go
```diff
@@ -1699,3 +1699,59 @@ func TestShadowDOMAndDocumentFragment(t *testing.T) {
 		})
 	}
 }
+
+func TestPageTargetBlank(t *testing.T) {
```
inancgumus, line 1703: Thanks for adding a test 🙇 Should we also add a link as a comment to the "new tab never loads" issue?

tests/page_test.go
```diff
@@ -1699,3 +1699,59 @@ func TestShadowDOMAndDocumentFragment(t *testing.T) {
...
+	opts := &common.FrameGotoOptions{
+		Timeout: common.DefaultTimeout,
+	}
+	resp, err := p.Goto(tb.url("/home"), opts)
+	require.NoError(t, err)
+	require.NotNil(t, resp)
+
+	// Current page count should be 1.
+	pp := p.Context().Pages()
+	assert.Equal(t, 1, len(pp))
```
inancgumus, line 1734: 👍 It's a good idea to ensure there is only one page at the beginning.

inancgumus (commented):
Nice fix and test :) A few suggestions.

tests/page_test.go
ankur22: I don't think we need to do that. We haven't done that for any of our other fixes. People can browser the git history if they want to find that information.

What's your reasoning?

tests/page_test.go
ankur22: We don't. It's just to ensure that it is there. I would remove it, but it don't think it hurts to leave it in. Unless there's a reason i'm missing?

tests/page_test.go
ankur22: Yeah, good point. So I always saw `assert` as a way to inform readers that this is something that is key to the test, whereas `require` is something that isn't really being tested in this test but should pass still.

That's why I used `assert` here.

Happy to change this to `require` if you have a strong opinion.

inancgumus: No strong opinion. It's OK.

tests/page_test.go
inancgumus: Alright. No worries.

tests/page_test.go
ankur22: I was avoiding `waitForLoadState` because of https://github.com/grafana/xk6-browser/issues/880, but I'll give that a go 👍 

Nice idea on `WaitForEvent`.

I've added these changes in 8b102b802bc78c412bceb7ec9a3a7739f12502a1

inancgumus: 👍 Got it. Still, IMO, it's better to fix `waitForLoadState` instead of creating a potentially flaky test.

inancgumus (approved):

tests/page_test.go
ankur22: It's been changed now so I think the `assert` should still be ok since we're no longer working with `pp` beyond checking it's length. I'll keep this in mind though `require` vs `assert` 👍

Issue #1307, opened 2024-05-06
Migrate Page APIs to async


inancgumus:
```[tasklist]
### Tasks
- [ ] https://github.com/grafana/xk6-browser/pull/1343
- [ ] #1345
- [ ] #1352
- [ ] #1355
- [ ] https://github.com/grafana/xk6-browser/pull/1370
- [ ] #1371
- [ ] grafana/k6-docs#1602
- [ ] https://github.com/grafana/k6-DefinitelyTyped/pull/48
- [ ] Release Notes:
```

---

## What

Migrate the following APIs according to the How section in #428:

| Class | Method | Current k6 Return | PW return |
| ---- | ---- | ---- | ---- |
|Page|$|ElementHandle\|null|Promise<ElementHandle<SVGElement\|HTMLElement>\|null>|
|Page|$$|ElementHandle[]|Promise<ElementHandle<SVGElement\|HTMLElement>[]>|
|Page|bringToFront|void|Promise<void>|
|Page|check|void|Promise<void>|
|Page|close|void|Promise<void>|
|Page|content|string|Promise<string>|
|Page|dblclick|void|Promise<void>|
|Page|dispatchEvent|void|Promise<void>|
|Page|emulateMedia|void|Promise<void>|
|Page|emulateVisionDeficiency|void|not found in Playwright|
|Page|evaluate<R, Arg>|R|Promise<R>|
|Page|evaluateHandle<R, Arg>|JSHandle<R>|Promise<SmartHandle<R>>|
|Page|fill|void|Promise<void>|
|Page|focus|void|Promise<void>|
|Page|getAttribute|null\|string|Promise<null\|string>|
|Page|hover|void|Promise<void>|
|Page|innerHTML|string|Promise<string>|
|Page|innerText|string|Promise<string>|
|Page|inputValue|string|Promise<string>|
|Page|isChecked|boolean|Promise<boolean>|
|Page|isDisabled|boolean|Promise<boolean>|
|Page|isEditable|boolean|Promise<boolean>|
|Page|isEnabled|boolean|Promise<boolean>|
|Page|isHidden|boolean|Promise<boolean>|
|Page|isVisible|boolean|Promise<boolean>|
|Page|on|void|this|
|Page|opener|Page\|null|Promise<null\|Page>|
|Page|press|void|Promise<void>|
|Page|reload|null\|Response|Promise<null\|Response>|
|Page|screenshot|ArrayBuffer|Promise<Buffer>|
[...]

PR #1345, opened 2024-05-23
Prepare `Page` for async migration part #2

inancgumus:
## What?

Turns panics into errors only for `Page`'s first set of methods. The rest will be in another PR.

## Why?

To make the async migration more reliable.

## Checklist

## Related PR(s)/Issue(s)

#1307

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -948,21 +948,19 @@ func (f *Frame) FrameElement() (*ElementHandle, error) {
 }
 
 // GetAttribute of the first element found that matches the selector.
-func (f *Frame) GetAttribute(selector, name string, opts goja.Value) any {
+func (f *Frame) GetAttribute(selector, name string, opts goja.Value) (any, error) {
 	f.log.Debugf("Frame:GetAttribute", "fid:%s furl:%q sel:%q name:%s", f.ID(), f.URL(), selector, name)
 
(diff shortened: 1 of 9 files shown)


common/frame.go
```diff
@@ -948,21 +948,19 @@ func (f *Frame) FrameElement() (*ElementHandle, error) {
 }
 
 // GetAttribute of the first element found that matches the selector.
-func (f *Frame) GetAttribute(selector, name string, opts goja.Value) any {
+func (f *Frame) GetAttribute(selector, name string, opts goja.Value) (any, error) {
```
ankur22, line 951: `GetAttribute` should only returns a `string`. Having taken a quick look at the call stack it looks like we're using [element.getAttribute](https://github.com/grafana/xk6-browser/blob/5711416a992a522b549917eaf9bfffd771a81679/common/element_handle.go#L314), maybe a `string` instead of `any`?

ankur22 (approved):
LGTM! Love seeing those panics being replaced by errors ❤️

common/frame.go
inancgumus: How about returning a `bool` to detect that in the mapping layer?

```
func (f *Frame) GetAttribute(selector, name string, opts goja.Value) (string, bool, error)
```

Even so, this should better go to a new issue because it'd change the existing signatures. I've tried to keep them intact other than adding errors.

Issue #1341, opened 2024-05-22
Panic with `we either navigate top level or have old version of the navigated frame`


ankur22:
### Brief summary

When navigating through certain websites, in some cases an error `we either navigate top level or have old version of the navigated frame` is reported which stops the test iteration. At the moment it's not clear why this occurs, but in the investigation so far we have found that the `frame` in question that is causing this panic isn't being tracked by k6 browser.

There are two ways in which a new `frame` is tracked by k6 browser:
1. Chrome tells us that a [frame is attached](https://github.com/grafana/xk6-browser/blob/ae847603ca332e4809e6a2af3845308a8245f109/common/frame_session.go#L256) (this is where we expect a CDP message);
2. We attach to the frame(s) by getting the [existing frame tree.](https://github.com/grafana/xk6-browser/blob/ae847603ca332e4809e6a2af3845308a8245f109/common/frame_session.go#L409-L412)

We have explored the CDP requests that are made between k6 browser and chrome, but we do not see a `frameAttached` event being sent from Chrome, but we can see that the `frameTree` does contain a reference to the new frame as a child.

### xk6-browser version

v1.5.1

### OS

NA

### Chrome version

NA

### Docker version and image (if applicable)

NA

### Steps to reproduce the problem

Run the script that is causing this issue to replicate the issue.

### Expected behaviour

Script passes with no issues.

### Actual behaviour

Panics with `we either navigate top level or have old version of the navigated frame`.
```[tasklist]
### Tasks
- [ ] https://github.com/grafana/xk6-browser/pull/1346
- [ ] Release notes
- [ ] https://github.com/grafana/xk6-browser/issues/1347
```

ankur22:
I think I have a better idea of what is happening.

1. We have the page, which has a frame and a session.
2. The page has an iframe, which is first attached with [page.frameAttached](https://github.com/grafana/xk6-browser/blob/ae847603ca332e4809e6a2af3845308a8245f109/common/frame_session.go#L736) CDP from Chrome.
3. Eventually we get a [target.attachTarget](https://github.com/grafana/xk6-browser/blob/ae847603ca332e4809e6a2af3845308a8245f109/common/frame_session.go#L891) for the iframe, but a new frameSession is also created.
4. Now we start to receive CDP requests from both the original session and new session for the iframe.
[...]

PR #1346, opened 2024-05-24
Fix iframe panic on navigation

ankur22:
## What?

This fixes an issue where a page with an iframe panics during a navigation.

## Why?

The reason are two fold:

1. Firstly we didn't have a reference to the iframes child frame. This was resolved by adding handling the frame tree of the iframe. It's not a mainframe but it still needs to be handled correctly so that we have a reference to the iframe and its children.
2. Now that we have the child frame, we don't want it to be removed during a detach (of type swap). We protect against this by checking the session of the incoming detach event and the frame that it is trying to detach. If they don't match we ignore the request.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1341

### DIFF
--- a/common/frame_manager.go
+++ b/common/frame_manager.go
@@ -173,6 +173,24 @@ func (m *FrameManager) frameDetached(frameID cdp.FrameID, reason cdppage.FrameDe
 		return
 	}
 
+	// This helps prevent an iframe and its child frames from being removed
+	// when the type of detach is a swap. After this detach event usually
+	// the iframe navigates, which requires the frames to be present for the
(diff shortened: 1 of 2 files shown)


inancgumus (approved):
Nice fix and your explanations make total sense 🚀

Since we know the cause of the issue, can we add a test to reproduce this and prevent regressions?

ankur22:
> Since we know the cause of the issue, can we add a test to reproduce this and prevent regressions?

I'm working on the test, but it's difficult one.

PR #1358, opened 2024-05-29
Revert request.headers and request.size to sync

ankur22:
## What?

Revert `request.headers` and `request.size` to sync.

## Why?

Both these APIs are safe to stay as sync because:
1. In Playwright they are sync APIs.
2. They don't make any long running IO calls or long running processes.

## Checklist

### DIFF
--- a/browser/request_mapping.go
+++ b/browser/request_mapping.go
@@ -57,11 +57,7 @@ func mapRequest(vu moduleVU, r *common.Request) mapping {
 				return mapResponse(vu, resp), nil
 			})
 		},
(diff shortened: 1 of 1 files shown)


inancgumus (approved):
I've compared this to PW's signatures, and all the methods seem fine 👍 

BTW: Do not merge this yet before rebasing first and force push (to be on the safe side) as the tests fail with: "tests/page_test.go:1804:9: assignment mismatch: 1 variable but p2.InnerHTML returns 2 values"

PR #1369, opened 2024-06-05
Improve `valueFromRemoteObject` `null` detection

inancgumus:
## What?

Improve `valueFromRemoteObject` `null` detection by returning a Go `nil` instead of `"null"` as a string.

## Why?

Previously, we couldn't detect a null value:

```go
v, _ := page.GetAttribute("#el", "missing", nil)
assert.Nil(t, nil, v) // v is not nil. it's "null"
```

After this PR, `v` will be `nil` instead of `"null"` as a string. This enables us to detect `nil` in methods like `GetAttribute`, `TextNode`, etc., to tell the user that what they're looking for does not exist instead of returning `"null"` or an empty string.

## Checklist

### DIFF
--- a/common/remote_object.go
+++ b/common/remote_object.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/grafana/xk6-browser/log"
 
+	"github.com/chromedp/cdproto/runtime"
 	cdpruntime "github.com/chromedp/cdproto/runtime"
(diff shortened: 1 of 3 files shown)


ankur22 (approved):
I can't fault the logic. Thanks for fixing this!

Issue #428, opened 2022-06-30
Browser v2: Async APIs


sniku:
```[tasklist]
### Migrate to Async
```

```[tasklist]
### Nice to have
- [ ] grafana/jslib.k6.io#125
```

```[tasklist]
### Migration is complete - Document the changes
```

```[tasklist]
### Done Done
- [ ] #1362
- [ ] #1302
- [ ] #683
- [ ] #463
- [ ] #439
- [ ] #440
- [ ] #1251
- [ ] #583
- [ ] #1294
- [ ] #1295
- [ ] #1296
- [ ] #1303
- [ ] #1304
- [ ] #1305
- [ ] #1306
- [ ] #1307
- [ ] #1308
- [ ] #1309
- [ ] #1359
- [ ] #1321
- [ ] #1298
- [ ] grafana/k6-docs#1600
```

---

## What

This issue has been repurposed to initiate and follow the progress of migrating most of the browser APIs to async (promise based).

A recent example is this [issue](https://github.com/grafana/xk6-browser/issues/1251) where `touchscreen.tap` was migrated to async:

From:

```js
page.touchscreen.tap(...);
```

To:

```js
await page.touchscreen.tap(...);
```

## Why

1. Most browser APIs use some form of long-running IO operation (networking) to perform the requested action on the web browser against the website under test. We need to avoid blocking JavaScript's runtime event loop for such operations.
2. We're going to add more asynchronous event-based APIs (such as [page.on](https://github.com/grafana/xk6-browser/issues/1227)) that our current synchronous APIs would block.
3. To align with how developers expect to work with JavaScript APIs.
4. To have better compatibility with Playwright.

## How

We have detailed exactly which APIs will most likely need to be migrated across in the [comment](https://github.com/grafana/xk6-browser/issues/428#issuecomment-1964020837) within this issue. For most of the APIs we can take a look at the comment and migrate the APIs that are listed there. It's worth noting that it's still best to double check whether the API needs to be migrated especially ones which k6 browser implements but aren't provided by Playwright, for example `page.throttleCPU`.

### NOT Implemented In PW

If an API is not implemented in PW, then you should be able to answer yes to one of the questions in the list below to determine whether it should be async or not, for example `browser.context` is not in PW, and it doesn't do any IO or long running tasks, therefore it can stay as a synchronous API.

1. Does the API perform a network call or other IO operations, this includes a CDP request to Chrome?
2. Does the API perform long running task?

### Implemented in PW

[...]

PR #1370, opened 2024-06-05
Improve `GetAttribute` `null` attribute detection

inancgumus:
## What?

The `GetAttribute` methods now return `false` when the attribute is missing.

## Why?

Previously, these methods returned an empty string (`""`). That approach makes it impossible to distinguish between whether the attribute is missing or empty.

## Note

I'm happy to convert the tests into table tests if desired.

## Checklist

## Related PR(s)/Issue(s)

#428


ankur22 (approved):
LGTM

PR #1371, opened 2024-06-06
Improve `TextContent` `null` element detection

inancgumus:
## What?

The `TextContent` methods now return `false` when the element's text content cannot be grabbed (like a JS `document`). I've added tests only for `ElementHandle` as the other types have their `TextContent` tests.

## Why?

Previously, these methods returned an empty string (`""`). That approach makes it impossible to distinguish between whether the text content is missing or empty.

## Checklist

## Related PR(s)/Issue(s)

#428


ankur22 (approved):
LGTM 🚀

Issue #1379, opened 2024-06-11
JSHandle.evaluate doesn't pass itself as the first argument


ankur22:
### Brief summary

In Playwright, JSHandle.evaluate passes itself as the first argument: https://playwright.dev/docs/api/class-jshandle#js-handle-evaluate. [k6 browser](https://github.com/grafana/xk6-browser/blob/07f10f8a1066ca51460ac2842dc1cc09edd490d2/common/js_handle.go#L127) doesn't seem to do the same.

### xk6-browser version

NA

### OS

NA

### Chrome version

NA

### Docker version and image (if applicable)

NA

### Steps to reproduce the problem

When calling the `evaluate` API like so:

```js
  const page = await browser.newPage();

  await page.goto('https://test.k6.io/browser.php');
  const jsHandle = await page.evaluateHandle(() => document.body);

  const innerHTML = await jsHandle.evaluate((node) => node.innerHTML);

  console.log(innerHTML);
```

We get an error:

```bash
ERRO[0000] Uncaught (in promise) evaluating element: TypeError: Cannot read properties of undefined (reading 'innerHTML')
    at __xk6_browser_evaluation_script__:1:15  executor=shared-iterations scenario=browser
```

### Expected behaviour

The expected bahaviour is that the innerHTML of the page's body is returned which can then be logged to the console.

### Actual behaviour

```bash
ERRO[0000] Uncaught (in promise) evaluating element: TypeError: Cannot read properties of undefined (reading 'innerHTML')
    at __xk6_browser_evaluation_script__:1:15  executor=shared-iterations scenario=browser
```
```[tasklist]
### Tasks
- [ ] https://github.com/grafana/xk6-browser/pull/1380
- [ ] Release notes
```

PR #1380, opened 2024-06-11
Fix/jshandle evaluate

ankur22:
## What?

This ensures that `jshandle.evaluate` and `jshandle.evaluateHandle` both set themselves as the first argument.

## Why?

It's the expected behaviour which matches Playwright: https://playwright.dev/docs/api/class-jshandle#js-handle-evaluate

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/xk6-browser/issues/1379

### DIFF
--- a/common/js_handle.go
+++ b/common/js_handle.go
@@ -124,6 +124,7 @@ func (h *BaseJSHandle) dispose() error {
 
 // Evaluate will evaluate provided page function within an execution context.
 func (h *BaseJSHandle) Evaluate(pageFunc string, args ...any) (any, error) {
+	args = append([]any{h}, args...)
 	res, err := h.execCtx.Eval(h.ctx, pageFunc, args...)
 	if err != nil {
 		return nil, fmt.Errorf("evaluating element: %w", err)
--- a/common/js_handle.go
+++ b/common/js_handle.go
@@ -135,6 +135,7 @@ func (h *BaseJSHandle) Evaluate(pageFunc string, args ...any) (any, error) {
 
 // EvaluateHandle will evaluate provided page function within an execution context.
 func (h *BaseJSHandle) EvaluateHandle(pageFunc string, args ...any) (JSHandleAPI, error) {
+	args = append([]any{h}, args...)
 	eh, err := h.execCtx.EvalHandle(h.ctx, pageFunc, args...)
 	if err != nil {
 		return nil, fmt.Errorf("evaluating handle for element: %w", err)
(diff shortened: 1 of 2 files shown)


tests/js_handle_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+			err := p.SetContent(`<html><head><title>Some title</title></head></html>`, nil)
+			require.NoError(t, err)
+
+			result, err := p.EvaluateHandle(`() => document.head`)
+			require.NoError(t, err)
+			require.NotNil(t, result)
+
+			got, err := result.Evaluate(tt.pageFunc, tt.args...)
+			assert.NoError(t, err)
```
inancgumus, line 52: Shouldn't this be `require`? The execution should stop if the result is an error, otherwise, there might be a panic due to a `nil` pointer error.

tests/js_handle_test.go
```diff
@@ -54,3 +54,56 @@ func TestJSHandleEvaluate(t *testing.T) {
...
+
+			err := p.SetContent(`<html><head><title>Some title</title></head></html>`, nil)
+			require.NoError(t, err)
+
+			result, err := p.EvaluateHandle(`() => document.head`)
+			require.NoError(t, err)
+			require.NotNil(t, result)
+
+			got, err := result.EvaluateHandle(tt.pageFunc, tt.args...)
+			assert.NoError(t, err)
```
inancgumus, line 101: I'd make this `require`.

tests/js_handle_test.go
```diff
@@ -54,3 +54,56 @@ func TestJSHandleEvaluate(t *testing.T) {
...
+			assert.NotNil(t, got)
+
+			j, err := got.JSONValue()
+			assert.NoError(t, err)
```
inancgumus, line 105: This one `require` as well.

common/js_handle.go
```diff
@@ -124,6 +124,7 @@ func (h *BaseJSHandle) dispose() error {
 
 // Evaluate will evaluate provided page function within an execution context.
 func (h *BaseJSHandle) Evaluate(pageFunc string, args ...any) (any, error) {
+	args = append([]any{h}, args...)
```
inancgumus, line 127: A little but important fix 👍

inancgumus (approved):
Thanks for fixing this 👍

tests/js_handle_test.go
ankur22: Sure, updated in b5ff202df57e398bb4f40c7a0e875a6c547bdd54

tests/js_handle_test.go
ankur22: Sure, updated in b5ff202df57e398bb4f40c7a0e875a6c547bdd54

inancgumus (approved):

Issue #1385, opened 2024-06-13
Improve empty browser string flag handling


inancgumus:
### What?

In v1.5.1, passing a string flag without a value results in an invalid argument that ends with an equal sign. For example:

```
K6_BROWSER_ARGS=disable-site-isolation-trials
```

Produces the following argument that we pass while launching the browser:

```
disable-site-isolation-trials=
```

It should be:

```
disable-site-isolation-trials
```

### Why?

Having a superfluous suffix (`=`) prevents browsers (e.g., Chrome) from interpreting the argument, making the argument defunct.

### How?

Handle the argument correctly [here](https://github.com/grafana/xk6-browser/blob/c27f714ca46a5f21fed4f2f5a92654c26ba19158/chromium/browser_type.go#L301).

### Tasks

```[tasklist]
### Tasks
- [ ] #1386
- [x] Update the k6 release notes
```

PR #1386, opened 2024-06-13
Fix browser empty string flag parsing

inancgumus:
## What?

Correctly handles empty string flags that don't have a value.

## Why?

See #1385.

## Checklist

## Related PR(s)/Issue(s)

Updates: #1385


chromium/browser_type.go
```diff
@@ -304,7 +304,15 @@ func parseArgs(flags map[string]any) ([]string, error) {
...
 		case string:
-			args = append(args, fmt.Sprintf("--%s=%s", name, value))
+			var arg string
+			if strings.TrimSpace(value) != "" {
+				arg = fmt.Sprintf("--%s=%s", name, value)
+			} else {
+				// If the value is empty, we don't include it in the args list.
+				// Otherwise, it will produce "--name=" which is invalid.
+				arg = fmt.Sprintf("--%s", name)
+			}
```
ankur22, line 314: The logic is sound and I usually wouldn't comment on this but on initial inspection it felt like there was too much going on here. Could it be made simpler with something like:

```go
arg := fmt.Sprintf("--%s", name)
if strings.TrimSpace(value) != "" {
	arg += fmt.Sprintf("=%s", name, value)
}
```

ankur22 (approved):
LGTM.

The comment I made can be ignored, seems more like i'm nit-picking.

chromium/browser_type.go
inancgumus: `Sprintf` is inefficient. What about? [ec48706](https://github.com/grafana/xk6-browser/pull/1386/commits/ec48706321e811b599d873880184a8c76f079580)

ankur22 (approved):
LGTM. Thanks for the tests, TIL about the parsing behaviour without looking at the code 🙂

Issue #1269, opened 2024-04-18
De-sobekify `Browser` methods


inancgumus:
```[tasklist]
### Tasks
- [ ] https://github.com/grafana/xk6-browser/pull/1389
- [x] Add release notes
```

----

Methods to de-sobekify:
- `newContext`
- `newPage`

Option parsing to move to the mapping layer:
- `BrowserContextOptions`
- `Geolocation`
- `Credentials`
- `Screen`
- `Viewport`

PR #1389, opened 2024-06-21
Desobekify `Browser`

inancgumus:
## What?

- Moves `BrowserContextOptions` out of the business logic.
- Sets default options in `NewBrowserContext` to prevent nil pointer issues in the rest of the module. _Later, we can turn some of these pointer fields into value types to take advantage of their zero values._
- Updates tests to work with the new structure.

## Note

These changes should had to be made in the single commit to be atomic.

## Why?

See #1269.

## Checklist

### DIFF
--- a/common/browser.go
+++ b/common/browser.go
@@ -14,7 +14,6 @@ import (
 	"github.com/chromedp/cdproto/cdp"
 	"github.com/chromedp/cdproto/target"
 	"github.com/gorilla/websocket"
-	"github.com/grafana/sobek"
 
 	"github.com/grafana/xk6-browser/k6ext"
 	"github.com/grafana/xk6-browser/log"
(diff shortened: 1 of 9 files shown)


common/browser.go
```diff
@@ -588,14 +587,7 @@ func (b *Browser) NewContext(opts sobek.Value) (*BrowserContext, error) {
 		return nil, err
 	}
 
-	browserCtxOpts := NewBrowserContextOptions()
-	if err := browserCtxOpts.Parse(b.ctx, opts); err != nil {
-		err := fmt.Errorf("parsing newContext options: %w", err)
-		spanRecordError(span, err)
```
allansson, line 594: It seems like the call to `spanRecordError` is lost in the new code. Is this intentional?

allansson (approved):
This looks a lot cleaner. Removes a lot of the clutter of the business logic. Nice job! 😍

common/browser.go
inancgumus: Good catch :) That's one of the issues with de-sobek-ification. Since the option parsing now happens in the mapping layer, but the trace span starts in [the business logic](https://github.com/grafana/xk6-browser/blob/3f26d5cd37bac7a306a99e3e730ee093a7e8f12f/common/browser.go#L571-L573), it's not viable to add this error to the span. I've created #1393 to discuss what we can do for these cases.

PR #1391, opened 2024-06-25
Fix potential races in createWaitForEvent*

mstoykov:
## What?

Try to prevent blocking on channel send.

## Why?

It is possible (and noticed in the real world) that the goroutine might enter the out select and then by the time get to writing to the channel back for the event handling to have been canceled.

## Checklist


andrewslotin (approved):
Good catch 👍

inancgumus (approved):
Nice catch 👍

Issue #1224, opened 2024-02-26
Panic on attaching iframe when getting browser window ID


ankur22:
### Brief summary

Panic on attaching iframe when [getting browser window ID](https://github.com/grafana/xk6-browser/blob/b90c091d8765ed28f74f62f03fbdff4afce12984/common/frame_session.go#L119-L127).

```bash
panic: GoError: attaching iframe: attaching iframe target ID 153A0D279584840F49FB1E718A8297AB to session ID DA84F3DFA9B0CA220314557F25BC8B37: getting browser window ID: No web contents for the given target id (-32000)

goroutine 97 [running]:
go.k6.io/k6/js/common.Throw(...)
	go.k6.io/k6/js/common/util.go:20
github.com/grafana/xk6-browser/k6ext.Panic.func1(0x155d220?, {0xc0018611c0?, 0x15fa301?, 0xc001bae5c0?})
	github.com/grafana/xk6-browser@v1.3.0/k6ext/panic.go:35 +0x74
github.com/grafana/xk6-browser/k6ext.sharedPanic({0x1b58098, 0xc000e131d0}, 0xc0026bdc80, {0xc0018611c0, 0x2, 0x2})
	github.com/grafana/xk6-browser@v1.3.0/k6ext/panic.go:64 +0x254
github.com/grafana/xk6-browser/k6ext.Panic({0x1b58098?, 0xc000e131d0?}, {0x17adf8a?, 0x294f5d0?}, {0xc0018611c0?, 0x0?, 0xc0026bde10?})
	github.com/grafana/xk6-browser@v1.3.0/k6ext/panic.go:37 +0x5a
github.com/grafana/xk6-browser/common.(*FrameSession).onAttachedToTarget(0xc0007d8540, 0xc001861000)
	github.com/grafana/xk6-browser@v1.3.0/common/frame_session.go:953 +0x671
github.com/grafana/xk6-browser/common.(*FrameSession).initEvents.func1()
	github.com/grafana/xk6-browser@v1.3.0/common/frame_session.go:282 +0x368
created by github.com/grafana/xk6-browser/common.(*FrameSession).initEvents in goroutine 73
	github.com/grafana/xk6-browser@v1.3.0/common/frame_session.go:223 +0x19b
```

### xk6-browser version

v1.3.0 (browser)

### OS

NA

### Chrome version

NA

### Docker version and image (if applicable)

NA

### Steps to reproduce the problem

[...]

PR #1406, opened 2024-08-20
Fix panic on iframe attach

ankur22:
## What?

This PR fixes an issue where the browser module would panic when navigating to certain website with iframes on them. The error is `panic: GoError: attaching iframe: attaching iframe target ID 153A0D279584840F49FB1E718A8297AB to session ID DA84F3DFA9B0CA220314557F25BC8B37: getting browser window ID: No web contents for the given target id (-32000)
`

## Why?

The reason for this is when the browser comes to attach an iframe, some iframes do not have any UI elements, which sets a state within [chromium such that no windowID can be returned](https://chromium.googlesource.com/chromium/src/+/665cdea2f5f29e166ad455b5021154456944fcbc/headless/lib/browser/headless_devtools_manager_delegate.cc#633) when one is requested for. This causes a panic [here](https://github.com/grafana/xk6-browser/blob/main/common/frame_session.go#L123). To avoid this issue we have to avoid performing the CDP calls that could cause errors when attaching a new iframe (retrieving a windowID and set the window bounds). PW does the [same](https://github.com/microsoft/playwright/blob/e17d1c498b9dc365fdea47a58c4c35cbad952a5a/packages/playwright-core/src/server/chromium/crPage.ts#L730).

## Test

<details>
  <summary>Test procedure</summary>

You can test against the following page:

```html
<!DOCTYPE html>
<html>
<head>
    <script src="https://js.stripe.com/v3/"></script>
</head>
<body></body>
</html>
```

First we need to replicate the issue, so set the current branch to `main`, and run the following test script (multiple times) until you get the same error as mentioned above:

```js
import { browser } from 'k6/browser';

export const options = {
  scenarios: {
    ui: {
      iterations: 1,
      options: {
        browser: {
          type: 'chromium'
        },
      },
    },
  },
}

  // You can try against http://2022wfh.ddns.net/static/tmp/panic.html
  await page.goto("URL for test site", {waitUntil:'networkidle'});

  await page.close();
}
```

Now test the same script against the fix branch. The same error shouldn't appear again.

</details>

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1224

### DIFF
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -57,7 +57,8 @@ type FrameSession struct {
 	k6Metrics *k6ext.CustomMetrics
 
 	targetID target.ID
-	windowID browser.WindowID
+	// windowID can be nil when it is associated to an iframe.
+	windowID *browser.WindowID
 
 	// To understand the concepts of Isolated Worlds, Contexts and Frames and
 	// the relationship betwween them have a look at the following doc:
@@ -121,7 +122,8 @@ func NewFrameSession(
 	}
 
 	action := browser.GetWindowForTarget().WithTargetID(fs.targetID)
-	if fs.windowID, _, err = action.Do(cdp.WithExecutor(fs.ctx, fs.session)); err != nil {
+	var windowID browser.WindowID
+	if windowID, _, err = action.Do(cdp.WithExecutor(fs.ctx, fs.session)); err != nil {
 		l.Debugf(
 			"NewFrameSession:GetWindowForTarget",
 			"sid:%v tid:%v err:%v",
@@ -129,6 +131,7 @@ func NewFrameSession(
 
 		return nil, fmt.Errorf("getting browser window ID: %w", err)
 	}
+	fs.windowID = &windowID
 
 	fs.initEvents()
 	if err = fs.initFrameTree(); err != nil {
@@ -1193,7 +1196,7 @@ func (fs *FrameSession) updateViewport() error {
 		fs.page.browserCtx.browser.browserOpts.Headless,
 		runtime.GOOS,
 	)
-	action2 := browser.SetWindowBounds(fs.windowID, &browser.Bounds{
+	action2 := browser.SetWindowBounds(*fs.windowID, &browser.Bounds{
 		Width:  viewport.Width,
 		Height: viewport.Height,
 	})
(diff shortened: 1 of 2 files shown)


common/frame_session.go
```diff
@@ -1187,18 +1197,20 @@ func (fs *FrameSession) updateViewport() error {
...
-		fs.page.browserCtx.browser.browserOpts.Headless,
-		runtime.GOOS,
-	)
-	action2 := browser.SetWindowBounds(fs.windowID, &browser.Bounds{
-		Width:  viewport.Width,
-		Height: viewport.Height,
-	})
-	if err := action2.Do(cdp.WithExecutor(fs.ctx, fs.session)); err != nil {
-		return fmt.Errorf("setting window bounds: %w", err)
+	if fs.windowID != nil {
```
inancgumus, line 1200: Instead of using a pointer, why don't we just use `hasUIWindow` parameter here?

inancgumus (commented):
Glad we're fixing this. I have a question.

common/frame_session.go
ankur22: Not sure what you mean. As in, save the flag as a field of the `FrameSession` type and then use it here instead of windowID being a pointer?

inancgumus: Ah, sorry, I wasn't clear. Yep, instead of using a pointer (which can be race-condition-prone and make an allocation for each frame), adding a `hasUIWindow` field is less dangerous and also communicates better that the current frame has no window UI. Also, we can use it for other purposes in the future if needed. Besides, `windowID` will be zero when it's absent since the Chromium code uses a `nullptr`, which evaluates to zero.

ankur22: I wanted to test what happens when `windowID` was `0`, just incase we could do something a little more clever and to get a better understanding of `nullptr`, but it turns out that chromium must interpret it as non null (or the error is incorrect from chromium) as it returns `Browser window not found`. In conclusion I think your suggestion still stands (`hasUIWindow` as a field on `FrameSession`) as the safest path forward 🙂

ankur22: Updated in ed17d77f74c91ce7905ad3ba326cf01b8bac573c

inancgumus: Thanks for giving this a try 🙇 :)

inancgumus (approved):
Nice! 🎉

PR #1418, opened 2024-09-09
Fix a few integration tests that failed often

ankur22:
## What?

Fixes a few of the integration tests that were failing often locally (not related to CI flakey tests, but it might help there too).

## Why?

Better DX.

## Checklist

### DIFF
--- a/k6ext/k6test/vu.go
+++ b/k6ext/k6test/vu.go
@@ -139,7 +139,9 @@ func (v *VU) RunOnEventLoop(tb testing.TB, js string, args ...any) (sobek.Value,
 func (v *VU) RunAsync(tb testing.TB, js string, args ...any) (sobek.Value, error) {
 	tb.Helper()
 
-	return v.RunOnEventLoop(tb, fmt.Sprintf("(async function() { "+js+" })();", args...))
+	jsWithArgs := fmt.Sprintf(js, args...)
+
+	return v.RunOnEventLoop(tb, "(async function() { %s })();", jsWithArgs)
 }
 
 // RunPromise runs the given JavaScript code on the VU's event loop and returns
--- a/tests/browser_test.go
+++ b/tests/browser_test.go
@@ -99,8 +99,19 @@ func TestTmpDirCleanup(t *testing.T) {
 
 	b.Close()
 
-	matches, err = filepath.Glob(tmpDirPath + "/xk6-browser-data-*")
-	assert.NoError(t, err)
+	// We need to wait for something (k6 browser, chromium or the os) to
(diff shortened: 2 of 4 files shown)


tests/browser_test.go
```diff
@@ -99,8 +99,19 @@ func TestTmpDirCleanup(t *testing.T) {
 
 	b.Close()
 
-	matches, err = filepath.Glob(tmpDirPath + "/xk6-browser-data-*")
-	assert.NoError(t, err)
+	// We need to wait for something (k6 browser, chromium or the os) to
+	// actually complete the removal of the directory. It's a race condition.
+	// To try to mitigate the issue, we're adding a retry which waits half a
+	// second if the dir still exits.
```
inancgumus, line 105: Nice comment. I'd also add `TODO` note here, saying, fix this race condition - or might even make sense to create an issue for it.

inancgumus (approved):
Thanks!

k6ext/k6test/vu.go
```diff
@@ -139,7 +139,10 @@ func (v *VU) RunOnEventLoop(tb testing.TB, js string, args ...any) (sobek.Value,
 func (v *VU) RunAsync(tb testing.TB, js string, args ...any) (sobek.Value, error) {
 	tb.Helper()
 
-	return v.RunOnEventLoop(tb, fmt.Sprintf("(async function() { "+js+" })();", args...))
```
ankur22, line 142: Hoping to fix https://github.com/grafana/xk6-browser/actions/runs/10773846755/job/29877212064 with this change 🤞

inancgumus: The number of args don't match to the number of values, I guess.

ankur22: Never seen this issue before 🤷 Seems to only occur with the tip.

inancgumus: Related (IDK :))?

* https://github.com/argoproj/argo-cd/issues/19772
* https://cs.opensource.google/go/x/tools/+/refs/tags/v0.24.0:go/analysis/passes/printf/printf.go;l=502-534
* https://code.waarp.fr/apps/gateway/gateway/-/blob/master/.golangci.yml#L841-844

ankur22: Right, i get it, it's a new change in tip, that's why it's failing on the build step 👍

ankur22: I still don't understand why the original change wasn't building, but the new change seems to have solved it.

inancgumus (approved):

Issue #1410, opened 2024-08-30
VU context should control the iteration


ankur22:
### Brief summary

Currently the browser module relies on the the [events](https://github.com/grafana/xk6-browser/blob/7cc9ea748a213d104a744e689543a8614c8c3034/browser/registry.go#L225-L231) that it is subscribed to control its lifecycle. The reasons we have this is to allow the browser module to control the number of chromium instances that are used per test run.

An unexpected consequence of working with the event system is that we were unable to easily work with the vu context to allow k6 to abort tests early. If a vu context was used to control the chromium subprocess/connection then there would be a [race condition](https://github.com/grafana/xk6-browser/pull/910/commits/4464d33337ac1ff9f65144449f0b53d88f9b233a) between chromium unexpectedly shutting down when the iteration ends and the current API calls being made via CDP.

The work around for this is to use a [background context that is created when the module is created](https://github.com/grafana/xk6-browser/blob/4350722189cf533a31228ff0c032b88f14ae394b/browser/module.go#L107). This means the module relies on the events system to correctly shut down iterations.

When this doesn't work is when a [gracefulStop](https://github.com/grafana/k6/blob/7b62bc9f7f3f801cba5cf6dfe3da75894f597db2/cmd/run.go#L349) occurs (SIGTERM for example) and the iteration's vu context is closed. At this point k6 expects the current API call to cancel itself since the context is closed and allow it to clean itself up and gracefully shutdown.

Since the browser module isn't working with the vu context, it fails to shut it self down promptly, which can result in various other timeouts elsewhere. We should look at using the vu context to control the API calls, but rely on the events to control the chromium browser lifecycle -- `IterEnd` and `exit` events are called once the iteration's vu context is closed and the browser module hands over the main thread back to k6.

### xk6-browser version

v0.53.0

### OS

NA

### Chrome version

NA

### Steps to reproduce the problem

Run the following script, which will block the iteration and then send a `SIGTERM` signal (`kill -15 {pid}`) to the k6 process—it won't quit the process (until a default of 30 seconds or a user-specified timeout elapses):

```js
import { browser } from 'k6/browser';

export const options = {
  scenarios: {
    ui: {
      executor: 'shared-iterations',
      options: {
        browser: {
            type: 'chromium',
        },
      },
    },
  },
}

export default async function() {
  const context = await browser.newContext();
  const page = await context.newPage();

  try {
    await page.goto('https://test.k6.io/', { waitUntil: 'networkidle' });
[...]

PR #1420, opened 2024-09-10
Fix browser not stop during aborts (e.g. SIGTERM)

ankur22:
## What?

This change fixes an issue whereby the browser module could block a test abort if the browser module was in the middle of waiting on something e.g. working with `waitFor*` apis.

The fix is to use the VU context for the iteration, but keep to the background context for running the chromium subprocess and connections. Now when the test is aborted, the vu context is closed by k6, which in turn will cancel all running browser API calls, and finally allow k6 to shutdown the test.

## Why?

It was blocking the test from aborting. The reason was due to the use of the background context for the iteration as well as for running the browser subprocess and connection. k6 signals a test abort by cancelling the VU context, but since the browser module wasn't relying on that, the browser module would first complete the exiting API call/iteration before realising that the test was aborted.

We need to work with both the VU context to control the iteration lifecycle, and the background context so that we allow the k6 event system to control the chromium subprocess/connection lifecycle. They need to be separate to prevent race conditions when working with the chromium subprocess/connection.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1410

### DIFF
--- a/browser/registry_test.go
+++ b/browser/registry_test.go
@@ -204,9 +204,8 @@ func TestBrowserRegistry(t *testing.T) {
 		t.Parallel()
 
 		var (
-			ctx             = context.Background()
 			vu              = k6test.NewVU(t)
-			browserRegistry = newBrowserRegistry(ctx, vu, remoteRegistry, &pidRegistry{}, nil)
+			browserRegistry = newBrowserRegistry(vu, remoteRegistry, &pidRegistry{}, nil)
 		)
 
 		vu.ActivateVU()
@@ -246,9 +245,8 @@ func TestBrowserRegistry(t *testing.T) {
 		t.Parallel()
 
 		var (
-			ctx             = context.Background()
 			vu              = k6test.NewVU(t)
-			browserRegistry = newBrowserRegistry(ctx, vu, remoteRegistry, &pidRegistry{}, nil)
+			browserRegistry = newBrowserRegistry(vu, remoteRegistry, &pidRegistry{}, nil)
 		)
 
 		vu.ActivateVU()
@@ -281,9 +279,8 @@ func TestBrowserRegistry(t *testing.T) {
 		t.Parallel()
 
 		var (
-			ctx             = context.Background()
 			vu              = k6test.NewVU(t)
-			browserRegistry = newBrowserRegistry(ctx, vu, remoteRegistry, &pidRegistry{}, nil)
+			browserRegistry = newBrowserRegistry(vu, remoteRegistry, &pidRegistry{}, nil)
 		)
 
 		vu.ActivateVU()
@@ -305,7 +302,7 @@ func TestBrowserRegistry(t *testing.T) {
 		vu := k6test.NewVU(t)
 		var cancel context.CancelFunc
 		vu.CtxField, cancel = context.WithCancel(vu.CtxField)
-		browserRegistry := newBrowserRegistry(context.Background(), vu, remoteRegistry, &pidRegistry{}, nil)
+		browserRegistry := newBrowserRegistry(vu, remoteRegistry, &pidRegistry{}, nil)
 
 		vu.ActivateVU()
 
--- a/browser/registry_test.go
+++ b/browser/registry_test.go
@@ -219,9 +219,7 @@ func TestBrowserRegistry(t *testing.T) {
 		assert.Equal(t, 3, browserRegistry.getBrowserCount())
 
 		// Verify iteration traces are started
-		browserRegistry.tr.mu.Lock()
-		assert.Equal(t, 3, len(browserRegistry.tr.m))
-		browserRegistry.tr.mu.Unlock()
+		assert.Equal(t, 3, browserRegistry.tr.getIterationTracesCount())
 
 		// Send IterEnd events
 		vu.EndIteration(t, k6test.WithIteration(0))
@@ -232,9 +230,7 @@ func TestBrowserRegistry(t *testing.T) {
 		assert.Equal(t, 0, browserRegistry.getBrowserCount())
 
 		// Verify iteration traces have been ended
-		browserRegistry.mu.RLock()
-		assert.Equal(t, 0, len(browserRegistry.m))
-		browserRegistry.mu.RUnlock()
+		assert.Equal(t, 0, browserRegistry.tr.getIterationTracesCount())
 	})
 
 	t.Run("close_browsers_on_exit_event", func(t *testing.T) {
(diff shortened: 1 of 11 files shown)


browser/registry_test.go
```diff
@@ -296,6 +296,48 @@ func TestBrowserRegistry(t *testing.T) {
 
 		assert.True(t, browserRegistry.stopped.Load())
 	})
+
+	// This test ensures that the chromium browser's lifecycle is not controlled
+	// by the vu context.
+	t.Run("dont_close_browser_on_vu_context_close", func(t *testing.T) {
```
ankur22, line 302: This test will work with main too, but it's good to have still.

What is missing is an actual test to see what happens when SIGTERM is sent. Do you think it's a good idea to create a test using bash to:

1. Run a blocking test
2. run `kill -15 {PID}`
3. Ensure the test aborts soonish (within 10secs)

Is there a way to do this within the go code?

common/browser.go
```diff
@@ -86,13 +88,13 @@ type browserVersion struct {
 
 // NewBrowser creates a new browser, connects to it, then returns it.
 func NewBrowser(
```
ankur22, line 90: I believe that the `Browser` shouldn't need both the `backgroundCtx` and `k6Ctx`, it should only work with one, `k6Ctx`. It's responsibility shouldn't be to interact with the chromium subprocess/connection through `browserProc` and `backgroundCtx`. I need to open a refactor issue for this.

ankur22: Refactored background context out, so this isn't totally true.

browser/registry_test.go
inancgumus: Yes, we can do this with Go as we've similarly done before here: https://github.com/grafana/xk6-browser/blob/cc94235cae6259435d234b371264055510526308/tests/browser_test.go#L286-L324

browser/registry_test.go
```diff
@@ -296,6 +293,48 @@ func TestBrowserRegistry(t *testing.T) {
...
+		vu.ActivateVU()
+
+		// Send a few IterStart events
+		vu.StartIteration(t, k6test.WithIteration(0))
+
+		// Verify browsers are initialized
+		browserRegistry.mu.RLock()
+		assert.Equal(t, 1, len(browserRegistry.m))
+		browserRegistry.mu.RUnlock()
+
```
inancgumus, line 316: I'd move this to a method in `browserRegistry` as the test is coupled to the registry's internals and becomes unnecessarily lengthy because of mutex management (which is supposed to be the registry's job).

chromium/browser_type.go
```diff
@@ -148,13 +158,23 @@ func (b *BrowserType) link(
 
 // Launch allocates a new Chrome browser process and returns a new Browser value,
 // which can be used for controlling the Chrome browser.
-func (b *BrowserType) Launch(ctx context.Context) (_ *common.Browser, browserProcessID int, _ error) {
-	ctx, browserOpts, logger, err := b.init(ctx, false)
+//
+//   - backgroundCtx is used when launching an instance of chromium. This context
+//     is controlled by the k6 event system (e.g. iterStart, iterEnd, etc) and so
+//     the chromium sub process's lifecycle is also controlled by the k6 event system.
```
inancgumus, line 164: Is this comment still relevant?

common/browser.go
```diff
@@ -28,8 +26,11 @@ const (
 
 // Browser stores a Browser context.
 type Browser struct {
-	ctx      context.Context
-	cancelFn context.CancelFunc
+	initContext  context.Context
+	initCancelFn context.CancelFunc
```
inancgumus, line 30: Can you add a comment to clarify what "init" means in this context? 🙇

common/browser.go
```diff
@@ -191,33 +186,39 @@ func (b *Browser) getPages() []*Page {
 }
 
 func (b *Browser) initEvents() error { //nolint:cyclop
-	var cancelCtx context.Context
-	cancelCtx, b.evCancelFn = context.WithCancel(b.ctx)
 	chHandler := make(chan Event)
 
-	b.conn.on(cancelCtx, []string{
+	// Using backgroundCtx here. Using vuCtx would close the connection/subprocess
```
inancgumus, line 191: `backgroundCtx` is also here.

common/browser.go
```diff
@@ -191,33 +186,39 @@ func (b *Browser) getPages() []*Page {
...
 
 	go func() {
 		defer func() {
-			b.logger.Debugf("Browser:initEvents:defer", "ctx err: %v", cancelCtx.Err())
 			b.browserProc.didLoseConnection()
-			if b.cancelFn != nil {
-				b.cancelFn()
+			// Closing the vuCtx incase it hasn't already been closed. Very likely
+			// already closed since the vuCtx is controlled by the k6 iteration,
+			// whereas the backgroundCtx is controlled by the k6 event system. k6
```
inancgumus, line 207: The `backgroundCtx` is also here.

inancgumus (approved):
Looks promosing!

common/browser.go
ankur22: Thanks for spotting this, updated in b52ef8a21fc0c009429b265b8b9c2b37c7a11cf5

common/browser.go
ankur22: Sure, added in 41ad3e0844a83865b13ac1009cefa118b20ce775

browser/registry_test.go
ankur22: So create a method like `getBrowserCount`? It would only be used by the test. I've added it in f3719419bc805b165c9d488d43ebec8c5eb2ebaf, and also done it for iteration traces too in ac2df75082aab2e644be878e8c4131f5787bea9f.

browser/registry_test.go
ankur22: Ah, so i need to perform a SIGTERM on the k6 process, not on chromium.

browser/registry_test.go
inancgumus: The only user for the test is fine. I'd rename `getIterationTracesCount` to `iterationTracesCount` as "get" is frowned upon in idiomatic Go (or leave it as it is, no strong opinion :).

inancgumus (approved):

browser/registry_test.go
ankur22: No problem, changed in 89bcc73bf88660cdd1d1de62de76bcf1d093046c :)

inancgumus (approved):

Issue #1437, opened 2024-09-23
element_handle.newAction function can wait indefinitely when context is cancelled


ankur22:
### Brief summary

We've found another issue where the browser module seems to hang indefinitely. This time it is due to a goroutine not reading from the channel once the context is closed.

1. [helper func calls newPointerAction](https://github.com/grafana/xk6-browser/blob/29f6ef3049fe1ce603fcd90f9dbfe1827ee5dbb9/common/helpers.go#L103)
2. newPointerAction is then called. On result it tries to [write to the resultCh](https://github.com/grafana/xk6-browser/blob/4350722189cf533a31228ff0c032b88f14ae394b/common/element_handle.go#L1651).
3. In the meantime the context has been closed and the goroutine that is supposed to read from the `resultCh` has [shutdown](https://github.com/grafana/xk6-browser/blob/29f6ef3049fe1ce603fcd90f9dbfe1827ee5dbb9/common/helpers.go#L106).

This could be a reason for the browser module not shutting down, although i would have expected the goroutine to be cleaned up by the go runtime 🤔 

stack trace of issue:

```bash
goroutine 0 gp=0x2ef51e0 m=0 mp=0x2ef6420 [idle]:
runtime.futex(0x2ef6560, 0x80, 0x0, 0x0, 0x0, 0x0)
    runtime/sys_linux_amd64.s:557 +0x21 fp=0x7ffc7c4c4b50 sp=0x7ffc7c4c4b48 pc=0x47ee21
runtime.futexsleep(0x7ffc7c4c4bc8?, 0x40f630?, 0x7ffc7c4c4bf8?)
    runtime/os_linux.go:69 +0x30 fp=0x7ffc7c4c4ba0 sp=0x7ffc7c4c4b50 pc=0x437dd0
runtime.notesleep(0x2ef6560)
    runtime/lock_futex.go:170 +0x87 fp=0x7ffc7c4c4bd8 sp=0x7ffc7c4c4ba0 pc=0x40f7c7
runtime.mPark(...)
    runtime/proc.go:1866
runtime.stopm()
    runtime/proc.go:2885 +0x8c fp=0x7ffc7c4c4c08 sp=0x7ffc7c4c4bd8 pc=0x4432ac
runtime.findRunnable()
    runtime/proc.go:3622 +0xd5c fp=0x7ffc7c4c4d80 sp=0x7ffc7c4c4c08 pc=0x444d1c
runtime.schedule()
[...]

PR #1438, opened 2024-09-23
Fix write channel to exit when ctx is done

ankur22:
## What?

Ensuring that the goroutine can exit early if the context is closed.

## Why?

There's a chance that the goroutine [reading](https://github.com/grafana/xk6-browser/blob/29f6ef3049fe1ce603fcd90f9dbfe1827ee5dbb9/common/helpers.go#L105-L113) off the `resultCh` has or is exiting due to the context being closed. We don't want [this goroutine](https://github.com/grafana/xk6-browser/blob/4350722189cf533a31228ff0c032b88f14ae394b/common/element_handle.go#L1648-L1652) to be hanging around waiting since nothing will read off the `resultCh` or `errCh`.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1437

### DIFF
--- a/common/element_handle.go
+++ b/common/element_handle.go
@@ -1558,9 +1558,15 @@ func (h *ElementHandle) newAction(
 
 	return func(apiCtx context.Context, resultCh chan any, errCh chan error) {
 		if res, err := actionFn(apiCtx); err != nil {
-			errCh <- err
+			select {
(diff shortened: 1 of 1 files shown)


inancgumus (approved):
👍

Issue #1440, opened 2024-09-24
Browser hangs when waiting for the browserProc to close


ankur22:
### Brief summary

Here's the stack trace:

```bash
Logger format: RAW
SIGQUIT: quit
PC=0x47ee21 m=0 sigcode=0

goroutine 0 gp=0x2ef51e0 m=0 mp=0x2ef6420 [idle]:
runtime.futex(0x2ef6560, 0x80, 0x0, 0x0, 0x0, 0x0)
    runtime/sys_linux_amd64.s:557 +0x21 fp=0x7fffbfcd7a80 sp=0x7fffbfcd7a78 pc=0x47ee21
runtime.futexsleep(0x7fffbfcd7af8?, 0x40f630?, 0x100000000?)
    runtime/os_linux.go:69 +0x30 fp=0x7fffbfcd7ad0 sp=0x7fffbfcd7a80 pc=0x437dd0
runtime.notesleep(0x2ef6560)
    runtime/lock_futex.go:170 +0x87 fp=0x7fffbfcd7b08 sp=0x7fffbfcd7ad0 pc=0x40f7c7
runtime.mPark(...)
    runtime/proc.go:1866
runtime.stopm()
    runtime/proc.go:2885 +0x8c fp=0x7fffbfcd7b38 sp=0x7fffbfcd7b08 pc=0x4432ac
runtime.findRunnable()
    runtime/proc.go:3622 +0xd5c fp=0x7fffbfcd7cb0 sp=0x7fffbfcd7b38 pc=0x444d1c
runtime.schedule()
    runtime/proc.go:3995 +0xb1 fp=0x7fffbfcd7ce8 sp=0x7fffbfcd7cb0 pc=0x445df1
runtime.park_m(0xc000404540)
    runtime/proc.go:4102 +0x1eb fp=0x7fffbfcd7d40 sp=0x7fffbfcd7ce8 pc=0x44620b
runtime.mcall()
    runtime/asm_amd64.s:459 +0x4e fp=0x7fffbfcd7d58 sp=0x7fffbfcd7d40 pc=0x47b16e

goroutine 1 gp=0xc0000061c0 m=nil [chan receive, 5 minutes]:
runtime.gopark(0x600008?, 0x7f575940ec08?, 0x40?, 0x0?, 0x0?)
    runtime/proc.go:424 +0xce fp=0xc00096e988 sp=0xc00096e968 pc=0x474bce
[...]

PR #1442, opened 2024-09-24
Fix hanging close browser cdp req

ankur22:
## What?

We're putting a timeout on the CDP close request.

## Why?

This is so that the connection is closed and we allow the iteration to end. Without this, the browser module can be left hanging indefinitely on the cdp send (well, waiting for the response from chromium) when actually chromium has already exited.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1440


common/browser.go
```diff
@@ -538,9 +538,14 @@ func (b *Browser) Close() {
...
 	if !b.browserOpts.isRemoteBrowser {
 		var closeErr *websocket.CloseError
-		// Using a internal context here since vu context will very likely be
-		// closed.
-		err := cdpbrowser.Close().Do(cdp.WithExecutor(b.browserCtx, b.conn))
+		// Using the internal context with a timeout of 10 seconds here since
+		// 1. vu context will very likely be closed;
+		// 2. there's a chance that the process has died but the connection still
+		//    thinks it's open.
+		toCtx, toCancelCtx := context.WithTimeout(b.browserCtx, time.Second*10)
```
inancgumus, line 545: What is the reason it's exactly 10 seconds?

chromium/browser_type.go
```diff
@@ -144,9 +144,10 @@ func (b *BrowserType) connect(
 }
 
 func (b *BrowserType) link(
+	ctx context.Context,
 	wsURL string, logger *log.Logger,
 ) (*common.BrowserProcess, error) {
-	bProcCtx, bProcCtxCancel := context.WithCancel(context.Background())
+	bProcCtx, bProcCtxCancel := context.WithCancel(ctx)
```
inancgumus, line 150: It's good to get `Context` from outside for maintainability and testing. I'm trying to understand the whole fix, and I wonder if this change is related to it. I looked at [this](https://github.com/grafana/xk6-browser/blob/529532ec9371b0f283bca1b377c5481e7ad24f7c/browser/module.go#L106), and we pass `context.Background`.

inancgumus (approved):
👍 Thanks! Before merging, should we wait for the k6 side of the fix?

common/browser.go
ankur22: Feels like a good timeout. I was going to do 5 seconds, but 10 seemed safer in case the browser process was still alive but the k6 process was under heavy load. Any suggestions on what might be better?

chromium/browser_type.go
ankur22: Initially i was wondering whether this could have caused the issue and reverted the changes from commits [1](https://github.com/grafana/xk6-browser/pull/1420/commits/e439161ccd7e32d1c00407970535c1362cf4d10d) and [2](https://github.com/grafana/xk6-browser/pull/1420/commits/c9734fe1a7f9350f9fdebb320d1c81fc14ac0dad). This didn't make a difference, but i left this in since it was better to pass the context in than to create a `context.Background`.

ankur22:
> should we wait for the k6 side of the fix?

Yep, good idea. I can then update k6 in this PR with the fix.

.github/workflows/lint.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.21.x
+          go-version: 1.23.x
```
ankur22, line 22: These changes are needed due to updating to the latest version of k6, which works with dependencies that require at least v1.22 of go.

chromium/browser_type.go
inancgumus: It's OK. Just asked if it's related to the fix. It was not :)

common/browser.go
inancgumus: I worry that we would need to change this whenever a larger timeout is needed. A larger timeout can cause prematurely aborted errors from the engine team. Two hopefully possible solutions:
- Implementing our own `Context` implementation that automatically cancels when the process is dead
- Making this an environment variable

.github/workflows/lint.yml
inancgumus: I'm fine with this, but should this part go into a precursory PR?

ankur22: Yeah, i realised that after doing it. Next time!

common/browser.go
ankur22: Yep, that makes sense. Actually, maybe we can check something in the `browserProc` for clues that something has gone wrong 🤔 

I'll create another PR for this though.

inancgumus: Sure, I've created an issue for it from the comment: https://github.com/grafana/xk6-browser/pull/1442

inancgumus (approved):
Reapproving :)

PR #1448, opened 2024-09-27
Update sync examples with `check` to async `check`

inancgumus:
## What?

Updates sync examples with `check` to async `check`.

Some of the examples don't require an async check, but it's better to use async `check` in them for consistency.

## Why?

The sync k6 `check` is cumbersome when working with async functions.

## Checklist

## Related PR(s)/Issue(s)

grafana/k6-docs#1725

### DIFF
--- a/examples/cookies.js
+++ b/examples/cookies.js
@@ -1,5 +1,5 @@
-import { check } from 'k6';
 import { browser } from 'k6/x/browser/async';
+import { check } from 'https://jslib.k6.io/k6-utils/1.5.0/index.js';
 
 export const options = {
   scenarios: {
@@ -23,8 +23,8 @@ export default async function () {
 
   try {
     // get cookies from the browser context
-    check((await context.cookies()).length, {
-        'initial number of cookies should be zero': n => n === 0,
+    await check(await context.cookies(), {
+        'initial number of cookies should be zero': c => c.length === 0,
     });
 
     // add some cookies to the browser context
@@ -63,10 +63,10 @@ export default async function () {
       }
     ]);
     let cookies = await context.cookies();
-    check(cookies.length, {
+    await check(cookies.length, {
       'number of cookies should be 2': n => n === 2,
     });
-    check(cookies[0], {
+    await check(cookies[0], {
       'cookie 1 name should be testcookie': c => c.name === 'testcookie',
       'cookie 1 value should be 1': c => c.value === '1',
       'cookie 1 should be session cookie': c => c.expires === -1,
@@ -76,7 +76,7 @@ export default async function () {
       'cookie 1 should be httpOnly': c => c.httpOnly === true,
       'cookie 1 should be secure': c => c.secure === true,
     });
-    check(cookies[1], {
+    await check(cookies[1], {
       'cookie 2 name should be testcookie2': c => c.name === 'testcookie2',
       'cookie 2 value should be 2': c => c.value === '2',
     });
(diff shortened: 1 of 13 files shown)


ankur22 (approved):
Thanks for working on this change.

My personal preference would be to avoid `then` chaining as i think it looks complicated to a new user who is new to JS.

I'd also only try to retrieve data to compare within a `check` rather than do any changes to the DOM or elsewhere in the website/system. I like to think that a `check` is almost like an assert, and generally we only compare two things in an assert, rather than perform state changes to the system within the assert.

Since these are both style issue though, happy to approve and get them merged in 👍

examples/dispatch.js
```diff
@@ -27,9 +27,10 @@ export default async function() {
     const contacts = page.locator('a[href="/contacts.php"]');
     await contacts.dispatchEvent("click");
 
-    const h3 = page.locator("h3");
-    const ok = await h3.textContent() == "Contact us";
-    check(ok, { "header": ok });
+    await check(page.locator('h3'), {
+      'header': async lo => lo.textContent()
```
allansson, line 31: `async` is not needed because there's no `await`.

examples/cookies.js
```diff
@@ -103,23 +103,22 @@ export default async function () {
...
       'the second filtered cookie value should be 44': c => c.value === '44',
     });
 
     // clear cookies
     await context.clearCookies();
-    cookies = await context.cookies();
-    check(cookies.length, {
-      'number of cookies should be zero': n => n === 0,
+    await check(await context.cookies(), {
+      'number of cookies should be zero': c => c.length === 0,
```
allansson, line 121: Seeing this, I realized that we could make this a bit nicer by allow the check value to be a promise. 

```ts
await check(context.cookies(), {
  // Branches can be synchronous, because they are executed after the value has resolved.
  'number of cookies should be zero': c => c.length === 0,
})
```

I have a [PR](https://github.com/grafana/k6-jslib-utils/pull/14) up.

allansson (approved):
Thanks for doing this. 🙏 

I left a few comments.

examples/cookies.js
inancgumus: This is awesome! Thanks :) We can update the examples when the new lib version is released.

examples/dispatch.js
inancgumus: So, it doesn't matter whether `textContent` returns a promise, right?

examples/cookies.js
```diff
@@ -63,10 +63,10 @@ export default async function () {
       }
     ]);
     let cookies = await context.cookies();
-    check(cookies.length, {
+    await check(cookies.length, {
```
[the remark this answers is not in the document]
inancgumus: I want to keep it consistent with other `check`s. Also, it's easy to add a Promise in the branches and forget to add `await` to `check` (at least for me, it was like so :)).

examples/dispatch.js
```diff
@@ -27,9 +27,10 @@ export default async function() {
...
+        .then(text => text == 'Contact us')
+    });
```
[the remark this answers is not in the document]
allansson: I can agree with doing mutations inside the check-functions.

But I wouldn't consider using `then` an expert-level feature. 

I actually prefer it in some situations, such as when you're just doing a series of transformations on the value. Sort of like you would with an array:

```ts
const names = myArray
  .filter(item => item.isActive)
  .map(item => item.name)

const userName = await fetch("https://fake.api/users/1")
  .then(response => response.json())
  .then(response => response.userName)
```

But maybe that's because I'm an old goat who remember when `Promise`s promised to save us from callback hell. 😅

examples/dispatch.js
allansson: Nope, doesn't matter.

As long as it's a promise it will be awaited by the `check` function. Non-promise values will be used as-is.

The `async` keyword is only needed if you're using `await`. Without an `await`, all it will do is make sure that the return value is wrapped in a `Promise` (basically `Promise.resolve(returnValue)`). Since we support non-promise values, having `async`

inancgumus: Good to know! Thanks for explaining 🙇

examples/elementstate.js
```diff
@@ -33,34 +33,37 @@ export default async function() {
...
-  let isCheckedUnchecked = !await page.$(".checked").then((e) => e.isChecked());
-
-  check(page, {
-    isUncheckedChecked: isUncheckedChecked,
-    isCheckedUnchecked: isCheckedUnchecked,
+  await check(page, {
+    'is unchecked checked':
+      async p => p.$(".unchecked")
+        .then(e => e.setChecked(true))
+        .then(() => p.$(".unchecked"))
```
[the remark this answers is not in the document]
inancgumus: > Although having an async check allows us to do this, i don't think we should be doing state changes to the DOM from within the check.

I've tried that approach, but it looked more complicated. It creates a lot of one-usage variables/constants. I'm not against making state changes in a `check` and then checking them if it looks simpler. Especially in a corner case, like `setChecked`. Otherwise, I'd be against it, too.

examples/dispatch.js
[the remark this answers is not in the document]
inancgumus: I got rid of `then`s to make it simpler. WDYT now @ankur22?

ankur22 (approved):
Thanks! LGTM 🚀

Issue #1451, opened 2024-10-01
Browser hangs for an unknown reason


ankur22:
### Brief summary

```bash
Logger format: RAW
SIGQUIT: quit
PC=0x47ee21 m=0 sigcode=0

goroutine 0 gp=0x2ef61e0 m=0 mp=0x2ef7420 [idle]:
runtime.futex(0x2ef7560, 0x80, 0x0, 0x0, 0x0, 0x0)
	runtime/sys_linux_amd64.s:557 +0x21 fp=0x7fffd0bb3cc0 sp=0x7fffd0bb3cb8 pc=0x47ee21
runtime.futexsleep(0x7fffd0bb3d38?, 0x40f630?, 0x40f630?)
	runtime/os_linux.go:69 +0x30 fp=0x7fffd0bb3d10 sp=0x7fffd0bb3cc0 pc=0x437dd0
runtime.notesleep(0x2ef7560)
	runtime/lock_futex.go:170 +0x87 fp=0x7fffd0bb3d48 sp=0x7fffd0bb3d10 pc=0x40f7c7
runtime.mPark(...)
	runtime/proc.go:1866
runtime.stopm()
	runtime/proc.go:2885 +0x8c fp=0x7fffd0bb3d78 sp=0x7fffd0bb3d48 pc=0x4432ac
runtime.findRunnable()
	runtime/proc.go:3622 +0xd5c fp=0x7fffd0bb3ef0 sp=0x7fffd0bb3d78 pc=0x444d1c
runtime.schedule()
	runtime/proc.go:3995 +0xb1 fp=0x7fffd0bb3f28 sp=0x7fffd0bb3ef0 pc=0x445df1
runtime.park_m(0xc000006e00)
	runtime/proc.go:4102 +0x1eb fp=0x7fffd0bb3f80 sp=0x7fffd0bb3f28 pc=0x44620b
runtime.mcall()
	runtime/asm_amd64.s:459 +0x4e fp=0x7fffd0bb3f98 sp=0x7fffd0bb3f80 pc=0x47b16e

goroutine 1 gp=0xc0000061c0 m=nil [chan receive, 9 minutes]:
runtime.gopark(0x482e08?, 0x7f8afe1f3ca0?, 0x40?, 0x0?, 0x0?)
	runtime/proc.go:424 +0xce fp=0xc0004f4988 sp=0xc0004f4968 pc=0x474bce
[...]

Issue #1450, opened 2024-10-01
Browser hangs when page.close not called when using page.on


ankur22:
### Brief summary

When a test script works with `page.on`, and the iteration is ended without calling `page.close`, the k6 process hangs/deadlocks. This was a known issue and the resolution was to ensure users used `page.close`. However this doesn't work when the test iteration is ended due an abort signal such as SIGINT. In this case even if there is a `page.close` in the `finally` block in the test script, it won't be called.

### xk6-browser version

v1.5.0+

### OS

NA

### Chrome version

NA

### Docker version and image (if applicable)

NA

### Steps to reproduce the problem

Run the following script and send a SIGINT (ctrl+c) before the test ends and while waiting for the `page.goto` to complete, the test will not complete.

```js
import { browser } from 'k6/x/browser';

export const options = {
  scenarios: {
    ui: {
      executor: 'shared-iterations',
      options: {
        browser: {
            type: 'chromium',
        },
      },
    },
  },
}

export default async function() {
  const page = await browser.newPage();

  try {
    await page.on('console', async (msg) => {
      console.log(msg);
    });

    await page.goto('http://2022wfh.ddns.net/test', {waitUntil:'networkidle'});
  } finally {
    await page.close();
  }
}
```

### Expected behaviour

Test it interrupted and end.

### Actual behaviour

Deadlocks.

```[tasklist]
- [ ] https://github.com/grafana/xk6-browser/pull/1452
- [x] Add to release notes: https://github.com/grafana/k6/commit/705866817c8b8fffebb31c147be98b2dfc43ee21
```

PR #1452, opened 2024-10-01
Fix taskqueue not closing when the iteration is aborted

ankur22:
## What?

This ensures that the `tasqueue` is closed if it hasn't already been closed. We can now rely on the vu context when the iteration ends to clean up any remaining `taskqueue`s. This means that if the user forgets to use `page.close` then we will be able to close the `taskqueue`.

## Why?

We need to closed the `taskqueue` to avoid the k6 process from hanging indefinitely.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1451
Updates: https://github.com/grafana/xk6-browser/issues/1450


inancgumus (approved):
🎉 Thanks for all these fixes!

Issue #371, opened 2022-06-03
URL Grouping/Aggregation


tom-miseur:
It is often useful to aggregate endpoint URLs that contain dynamic values. This is critical in the k6 Cloud due to the limits we have in place to prevent tests from emitting `too-many-metrics`/`too-many-urls`.

The [URL Grouping](https://k6.io/docs/using-k6/http-requests#url-grouping) documentation provides a solution for k6 scripts using the `http` module, but because `xk6-browser` operates at the browser-level, there is no opportunity for the user to apply the `name` tag to requests that require it.

The situation is compounded by the fact that `xk6-browser` gains visibility of *all* HTTP requests incurred by the browser, including 3rd party hosts that would not normally be interacted with at all using HTTP k6 scripts.

### Allowlist/blocklist hosts in xk6-browser

A cursory browse through Playwright docs suggests there is no convenient way of preventing/allowing requests to certain hosts, e.g. through specifying regular expressions. There is, however, a request interception mechanism involving [Page.route](https://playwright.dev/docs/api/class-page#page-route) or  [BrowserContext.route](https://playwright.dev/docs/api/class-browsercontext#browser-context-route) that could be used to abort requests that don't fit the criteria.

Pros:
- would appear to support regex allow/block-listing which should be fairly easy for users to apply
- doesn't actually send requests to undesired hosts, so no need to wait for [#1321](https://github.com/grafana/k6/issues/1321) and no need to worry about errors from 3rd party hosts (e.g. errors caused by rate limiting)

Cons:
- users need to encounter the problem before then figuring out how to resolve it

### Allowlist/blocklist hosts after-the-fact

This means xk6-browser still sends requests to the additional hosts, but that traffic can be filtered out of results.

Pros:
- the user wouldn't need to run the test again to have filtering applied

Cons:
- requests are sent to 3rd parties who may have rate limiting/bot protections in place that cause errors
- k6 OSS would need some mechanism to ignore metrics from certain hosts ([#1321](https://github.com/grafana/k6/issues/1321))
- k6 Cloud would need to be able to filter out hosts (unless [#1321](https://github.com/grafana/k6/issues/1321) would result in k6 Cloud not receiving the metrics at all which is quite likely)

### Aggregation Rules

This would involve the user specifying URL grouping regular expressions (likely in `options`) ahead of time. Before any metric is generated, we check if the URL matches any of the patterns and apply the transformation as necessary.

Example:

```javascript
export const options = {
  aggregations: [
[...]

Issue #1227, opened 2024-02-29
Generalize `page.on` events


ankur22:
### What

Generalize `page.on` events architecture.

### Why

So that [new events](https://github.com/grafana/xk6-browser/issues/1460) can easily be added to the existing code.

```[tasklist]
### Tasks
- [ ] #1473
- [ ] #1474
- [ ] #1477
- [ ] #1478
- [ ] #1475
- [ ] #1479
- [x] Update k6 release notes and tasks
```

inancgumus:
I've changed this issue's title to focus on the generalization of `page.on` as the total work is lengthy. Please revert if you think otherwise.

ankur22:
During the [PR](https://github.com/grafana/xk6-browser/pull/1456) to add `page.on('metric')` helpful comments were added which identified areas that could be refactored to work with generics:

- https://github.com/grafana/xk6-browser/pull/1456#discussion_r1786277948
- https://github.com/grafana/xk6-browser/pull/1456#discussion_r1786278776
- https://github.com/grafana/xk6-browser/pull/1456#discussion_r1786282485
- https://github.com/grafana/xk6-browser/pull/1456#discussion_r1786285905
- https://github.com/grafana/xk6-browser/pull/1456#discussion_r1786291372
- https://github.com/grafana/xk6-browser/pull/1456#discussion_r1786332552
- Another idea instead of generics: https://github.com/grafana/xk6-browser/pull/1456#issuecomment-2392410774

PR #1456, opened 2024-10-03
Add page.on('metric')

ankur22:
## What?

This adds the functionality to group url tags by intercepting all metrics that the **browser module** emits, compare and match with the supplied regex(s), and then set the `name` and `url` tags with the new name supplied by the user in the script.

## Why?

A lot of websites will perform requests that contain dynamic parts in the request url, which could be ids that are generated on every new session, account ids, or for cache busting purposes.

This will help in two ways:
1. Avoids high cardinality issues where tests work with websites with a lot of these dynamic urls.
2. It becomes trivial to correlate urls which are for the same page but contain dynamic parts in the url (ids/cache busting/etc).

## How

The user first needs to register an event handler:

```mermaid
sequenceDiagram
  box main goroutine
  participant page.on('metric')
  participant p.eventHandlers
  end
  
  page.on('metric') ->> p.eventHandlers: add metric handler
```

This is a high level overview of what happens when a [http request is started and a metric is about to be emitted](https://github.com/grafana/xk6-browser/blob/d6a878e7f6f9b454bb7089996238ffd15b883ae5/common/network_manager.go#L183):

```mermaid
sequenceDiagram
    box CDP handler goroutine
    participant emitRequestMetrics
    participant handleURLTag
    participant page.urlTagName
    end
    box taskqueue goroutine
    participant handler
    participant MetricEvent.Tag
    end
    box sobek context
    participant _k6BrowserURLGroupingTest
    end

    emitRequestMetrics ->> handleURLTag: url, tags
    handleURLTag ->> page.urlTagName: url
    page.urlTagName ->> page.urlTagName: create MetricEvent{url}
    loop each p.eventHandler
    page.urlTagName ->> handler: MetricEvent{url}
    Note right of handler: Regex supplied by user in handler in URLOverrides{}
    handler ->> MetricEvent.Tag: URLTagPatterns{}
    loop each URLTagPatterns
    MetricEvent.Tag ->> _k6BrowserURLGroupingTest: URLTagPattern.urlRegEx, url
    Note right of _k6BrowserURLGroupingTest: if match found, return true, else false
    _k6BrowserURLGroupingTest ->> MetricEvent.Tag: return true|false
    end
    Note left of MetricEvent.Tag: assign user defined name to MetricEvent.name if match found.
    MetricEvent.Tag ->> handler: return
    handler ->> page.urlTagName: return
    end
    Note left of page.urlTagName: read MetricEvent.name when it's not nil and set urlMatched.
    page.urlTagName ->> handleURLTag: return newTagName, urlMatched
    Note left of handleURLTag: set tags.url and tags.name as newTagName if urlMatched is true
    handleURLTag ->> emitRequestMetrics: return tags
```

The same happens when we're about to emit [response](https://github.com/grafana/xk6-browser/blob/d6a878e7f6f9b454bb7089996238ffd15b883ae5/common/network_manager.go#L205) and [web vital](https://github.com/grafana/xk6-browser/blob/7274e17b078ed1274d0075aa7bc65671b0ba91b3/common/frame_session.go#L326) metrics.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/371

### DIFF
--- a/browser/metric_event_mapping.go
+++ b/browser/metric_event_mapping.go
@@ -25,7 +25,7 @@ func mapMetricEvent(vu moduleVU, cm *common.MetricEvent) (mapping, error) {
 	}
 
 	return mapping{
-		"Tag": func(urls common.URLOverrides) error {
+		"Tag": func(urls common.URLTagPatterns) error {
 			callback := func(pattern, url string) (bool, error) {
 				js := fmt.Sprintf(`_k6BrowserURLGroupingTest(%s, '%s')`, pattern, url)
 
--- a/browser/metric_event_mapping.go
+++ b/browser/metric_event_mapping.go
@@ -13,7 +13,7 @@ func mapMetricEvent(vu moduleVU, cm *common.MetricEvent) (mapping, error) {
 	// We're setting up the function in the Sobek context that will be reused
 	// for this VU.
 	_, err := rt.RunString(`
-	function _k6BrowserURLGroupingTest(pattern, url) {
+	function _k6BrowserCheckRegEx(pattern, url) {
 		let r = pattern;
 		if (typeof pattern === 'string') {
 			r = new RegExp(pattern);
@@ -27,7 +27,7 @@ func mapMetricEvent(vu moduleVU, cm *common.MetricEvent) (mapping, error) {
 	return mapping{
 		"Tag": func(urls common.URLTagPatterns) error {
 			callback := func(pattern, url string) (bool, error) {
-				js := fmt.Sprintf(`_k6BrowserURLGroupingTest(%s, '%s')`, pattern, url)
+				js := fmt.Sprintf(`_k6BrowserCheckRegEx(%s, '%s')`, pattern, url)
 
 				val, err := rt.RunString(js)
 				if err != nil {
--- a/browser/metric_event_mapping.go
+++ b/browser/metric_event_mapping.go
@@ -29,12 +29,12 @@ func mapMetricEvent(vu moduleVU, cm *common.MetricEvent) (mapping, error) {
 			callback := func(pattern, url string) (bool, error) {
 				js := fmt.Sprintf(`_k6BrowserCheckRegEx(%s, '%s')`, pattern, url)
 
-				val, err := rt.RunString(js)
+				matched, err := rt.RunString(js)
 				if err != nil {
 					return false, fmt.Errorf("matching url with regex: %w", err)
 				}
 
-				return val.ToBoolean(), nil
+				return matched.ToBoolean(), nil
 			}
 
 			return cm.Tag(callback, urls)
--- a/browser/metric_event_mapping.go
+++ b/browser/metric_event_mapping.go
@@ -25,7 +25,7 @@ func mapMetricEvent(vu moduleVU, cm *common.MetricEvent) (mapping, error) {
 	}
 
 	return mapping{
-		"Tag": func(urls common.URLGroups) error {
+		"Tag": func(urls common.URLOverrides) error {
 			callback := func(pattern, url string) (bool, error) {
 				js := fmt.Sprintf(`_k6BrowserURLGroupingTest(%s, '%s')`, pattern, url)
 
--- a/browser/metric_event_mapping.go
+++ b/browser/metric_event_mapping.go
@@ -6,8 +6,8 @@ import (
 	"github.com/grafana/xk6-browser/common"
 )
 
-// mapMetric to the JS module.
-func mapMetric(vu moduleVU, cm *common.ExportedMetric) (mapping, error) {
+// mapMetricEvent to the JS module.
+func mapMetricEvent(vu moduleVU, cm *common.MetricEvent) (mapping, error) {
 	rt := vu.VU.Runtime()
 
 	// We're setting up the function in the Sobek context that will be reused
@@ -21,7 +21,7 @@ func mapMetric(vu moduleVU, cm *common.ExportedMetric) (mapping, error) {
 		return r.test(url);
 	}`)
 	if err != nil {
-		return nil, fmt.Errorf("evaluating url grouping: %w", err)
+		return nil, fmt.Errorf("evaluating regex function: %w", err)
 	}
 
 	return mapping{
@@ -31,7 +31,7 @@ func mapMetric(vu moduleVU, cm *common.ExportedMetric) (mapping, error) {
 
 				val, err := rt.RunString(js)
 				if err != nil {
-					return false, fmt.Errorf("evaluating metric tag url grouping: %w", err)
+					return false, fmt.Errorf("matching url with regex: %w", err)
 				}
 
 				return val.ToBoolean(), nil
--- a/browser/metric_mapping.go
+++ b/browser/metric_mapping.go
@@ -1,11 +1,29 @@
 package browser
 
 import (
+	"fmt"
+
 	"github.com/grafana/xk6-browser/common"
 )
 
 // mapMetric to the JS module.
 func mapMetric(vu moduleVU, cm *common.ExportedMetric) (mapping, error) {
+	rt := vu.Runtime()
+
+	// We're setting up the function in the Sobek context that will be reused
+	// for this VU.
+	_, err := rt.RunString(`
+	function _k6BrowserURLGroupingTest(pattern, url) {
+		let r = pattern;
+		if (typeof pattern === 'string') {
+			r = new RegExp(pattern);
+		}
+		return r.test(url);
+	}`)
+	if err != nil {
+		return nil, fmt.Errorf("evaluating url grouping: %w", err)
+	}
+
 	return mapping{
 		"groupURLTag": func(urls common.URLGroups) error {
 			return cm.GroupURLTag(urls)
--- a/browser/metric_mapping.go
+++ b/browser/metric_mapping.go
@@ -1,0 +1,14 @@
+package browser
+
+import (
+	"github.com/grafana/xk6-browser/common"
+)
+
+// mapMetric to the JS module.
+func mapMetric(vu moduleVU, cm *common.ExportedMetric) (mapping, error) {
+	return mapping{
+		"groupURLTag": func(urls common.URLGroups) error {
+			return cm.GroupURLTag(urls)
+		},
+	}, nil
+}
--- a/browser/metric_mapping.go
+++ b/browser/metric_mapping.go
@@ -8,7 +8,7 @@ import (
 
 // mapMetric to the JS module.
 func mapMetric(vu moduleVU, cm *common.ExportedMetric) (mapping, error) {
-	rt := vu.Runtime()
+	rt := vu.VU.Runtime()
 
 	// We're setting up the function in the Sobek context that will be reused
 	// for this VU.
--- a/browser/metric_mapping.go
+++ b/browser/metric_mapping.go
@@ -26,7 +26,18 @@ func mapMetric(vu moduleVU, cm *common.ExportedMetric) (mapping, error) {
 
 	return mapping{
 		"groupURLTag": func(urls common.URLGroups) error {
-			return cm.GroupURLTag(urls)
+			callback := func(pattern, url string) (bool, error) {
+				js := fmt.Sprintf(`_k6BrowserURLGroupingTest(%s, '%s')`, pattern, url)
+
+				val, err := rt.RunString(js)
+				if err != nil {
+					return false, fmt.Errorf("evaluating metric tag url grouping: %w", err)
+				}
+
+				return val.ToBoolean(), nil
+			}
+
+			return cm.GroupURLTag(callback, urls)
 		},
 	}, nil
 }
--- a/browser/metric_mapping.go
+++ b/browser/metric_mapping.go
@@ -25,7 +25,7 @@ func mapMetric(vu moduleVU, cm *common.ExportedMetric) (mapping, error) {
 	}
 
 	return mapping{
-		"groupURLTag": func(urls common.URLGroups) error {
+		"Tag": func(urls common.URLGroups) error {
 			callback := func(pattern, url string) (bool, error) {
 				js := fmt.Sprintf(`_k6BrowserURLGroupingTest(%s, '%s')`, pattern, url)
 
@@ -37,7 +37,7 @@ func mapMetric(vu moduleVU, cm *common.ExportedMetric) (mapping, error) {
 				return val.ToBoolean(), nil
 			}
 
-			return cm.GroupURLTag(callback, urls)
+			return cm.Tag(callback, urls)
 		},
 	}, nil
 }
--- a/browser/page_mapping.go
+++ b/browser/page_mapping.go
@@ -228,6 +228,12 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 						return nil
 					})
 				}
+			case common.EventPageMetricCalled:
+				runInTaskQueue = func(a any) {
+					tq.Queue(func() error {
+						return nil
+					})
+				}
 			default:
 				return fmt.Errorf("unknown page event: %q", event)
 			}
--- a/browser/page_mapping.go
+++ b/browser/page_mapping.go
@@ -230,17 +230,18 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 				}
 			case common.EventPageMetricCalled:
 				runInTaskQueue = func(a any) {
+					// The function on the taskqueue runs in its own goroutine
+					// so we need to use a channel to wait for it to complete
+					// since we're waiting for updates from the handler which
+					// will be written to the ExportedMetric.
+					c := make(chan bool)
 					tq.Queue(func() error {
+						defer close(c)
+
 						m, ok := a.(*common.ExportedMetric)
 						if !ok {
 							return errors.New("incorrect metric message")
 						}
-						// page.on('metric') needs to be synchronized so that
-						// the name can be returned back to the caller of the
-						// handler. Once the handler has complete we call
-						// Completed so that the caller knows that the handler
-						// has complete and to work with the new metric.
-						defer m.Completed()
 
 						mapping, err := mapMetric(vu, m)
 						if err != nil {
@@ -252,6 +253,7 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 						}
 						return nil
 					})
+					<-c
 				}
 			default:
 				return fmt.Errorf("unknown page event: %q", event)
--- a/browser/page_mapping.go
+++ b/browser/page_mapping.go
@@ -229,11 +229,23 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 					})
 				}
 			case common.EventPageMetricCalled:
+				mapMsgAndHandleEvent := func(m *common.ExportedMetric) error {
+					mapping, err := mapMetric(vu, m)
+					if err != nil {
+						return err
+					}
+					_, err = handler(sobek.Undefined(), vu.Runtime().ToValue(mapping))
+					return err
+				}
 				runInTaskQueue = func(a any) {
 					tq.Queue(func() error {
-						_, ok := a.(*common.ExportedMetric)
+						m, ok := a.(*common.ExportedMetric)
 						if !ok {
 							return errors.New("incorrect metric message")
+						}
+
+						if err := mapMsgAndHandleEvent(m); err != nil {
+							return fmt.Errorf("executing page.on handler: %w", err)
 						}
 						return nil
 					})
--- a/browser/page_mapping.go
+++ b/browser/page_mapping.go
@@ -238,12 +238,12 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 					tq.Queue(func() error {
 						defer close(c)
 
-						m, ok := a.(*common.ExportedMetric)
+						m, ok := a.(*common.MetricEvent)
 						if !ok {
 							return errors.New("incorrect metric message")
 						}
 
-						mapping, err := mapMetric(vu, m)
+						mapping, err := mapMetricEvent(vu, m)
 						if err != nil {
 							return fmt.Errorf("mapping the metric: %w", err)
 						}
--- a/browser/page_mapping.go
+++ b/browser/page_mapping.go
@@ -228,6 +228,8 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 						return nil
 					})
 				}
+			default:
+				return fmt.Errorf("unknown page event: %q", event)
 			}
 
 			return p.On(event, runInTaskQueue) //nolint:wrapcheck
--- a/browser/page_mapping.go
+++ b/browser/page_mapping.go
@@ -212,7 +212,7 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 			case common.EventPageConsoleAPICalled:
 				mapMsgAndHandleEvent := func(m *common.ConsoleMessage) error {
 					mapping := mapConsoleMessage(vu, m)
-					_, err := handler(sobek.Undefined(), vu.Runtime().ToValue(mapping))
+					_, err := handler(sobek.Undefined(), vu.VU.Runtime().ToValue(mapping))
 					return err
 				}
 				runInTaskQueue = func(a any) {
@@ -234,7 +234,7 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 					if err != nil {
 						return err
 					}
-					_, err = handler(sobek.Undefined(), vu.Runtime().ToValue(mapping))
+					_, err = handler(sobek.Undefined(), vu.VU.Runtime().ToValue(mapping))
 					return err
 				}
 				runInTaskQueue = func(a any) {
--- a/browser/page_mapping.go
+++ b/browser/page_mapping.go
@@ -231,6 +231,10 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 			case common.EventPageMetricCalled:
 				runInTaskQueue = func(a any) {
 					tq.Queue(func() error {
+						_, ok := a.(*common.ExportedMetric)
+						if !ok {
+							return errors.New("incorrect metric message")
+						}
 						return nil
 					})
 				}
--- a/browser/page_mapping.go
+++ b/browser/page_mapping.go
@@ -2,6 +2,7 @@ package browser
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
@@ -211,8 +212,13 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 				_, err := handler(sobek.Undefined(), vu.Runtime().ToValue(mapping))
 				return err
 			}
-			runInTaskQueue := func(m *common.ConsoleMessage) {
+			runInTaskQueue := func(a any) {
 				tq.Queue(func() error {
+					m, ok := a.(*common.ConsoleMessage)
+					if !ok {
+						return errors.New("incorrect message")
+					}
+
 					if err := mapMsgAndHandleEvent(m); err != nil {
 						return fmt.Errorf("executing page.on handler: %w", err)
 					}
--- a/browser/page_mapping.go
+++ b/browser/page_mapping.go
@@ -243,6 +243,12 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 						if !ok {
 							return errors.New("incorrect metric message")
 						}
+						// page.on('metric') needs to be synchronized so that
+						// the name can be returned back to the caller of the
+						// handler. Once the handler has complete we call
+						// Completed so that the caller knows that the handler
+						// has complete and to work with the new metric.
+						defer m.Completed()
 
 						if err := mapMsgAndHandleEvent(m); err != nil {
 							return fmt.Errorf("executing page.on handler: %w", err)
--- a/browser/page_mapping.go
+++ b/browser/page_mapping.go
@@ -207,23 +207,27 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 		"on": func(event string, handler sobek.Callable) error {
 			tq := vu.taskQueueRegistry.get(vu.Context(), p.TargetID())
 
-			mapMsgAndHandleEvent := func(m *common.ConsoleMessage) error {
-				mapping := mapConsoleMessage(vu, m)
-				_, err := handler(sobek.Undefined(), vu.Runtime().ToValue(mapping))
-				return err
-			}
-			runInTaskQueue := func(a any) {
-				tq.Queue(func() error {
-					m, ok := a.(*common.ConsoleMessage)
-					if !ok {
-						return errors.New("incorrect message")
-					}
+			var runInTaskQueue func(any)
+			switch event {
+			case common.EventPageConsoleAPICalled:
+				mapMsgAndHandleEvent := func(m *common.ConsoleMessage) error {
+					mapping := mapConsoleMessage(vu, m)
+					_, err := handler(sobek.Undefined(), vu.Runtime().ToValue(mapping))
+					return err
+				}
+				runInTaskQueue = func(a any) {
+					tq.Queue(func() error {
+						m, ok := a.(*common.ConsoleMessage)
+						if !ok {
+							return errors.New("incorrect console message")
+						}
 
-					if err := mapMsgAndHandleEvent(m); err != nil {
-						return fmt.Errorf("executing page.on handler: %w", err)
-					}
-					return nil
-				})
+						if err := mapMsgAndHandleEvent(m); err != nil {
+							return fmt.Errorf("executing page.on handler: %w", err)
+						}
+						return nil
+					})
+				}
 			}
 
 			return p.On(event, runInTaskQueue) //nolint:wrapcheck
--- a/browser/page_mapping.go
+++ b/browser/page_mapping.go
@@ -229,14 +229,6 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 					})
 				}
 			case common.EventPageMetricCalled:
-				mapMsgAndHandleEvent := func(m *common.ExportedMetric) error {
-					mapping, err := mapMetric(vu, m)
-					if err != nil {
-						return err
-					}
-					_, err = handler(sobek.Undefined(), vu.VU.Runtime().ToValue(mapping))
-					return err
-				}
 				runInTaskQueue = func(a any) {
 					tq.Queue(func() error {
 						m, ok := a.(*common.ExportedMetric)
@@ -250,8 +242,13 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 						// has complete and to work with the new metric.
 						defer m.Completed()
 
-						if err := mapMsgAndHandleEvent(m); err != nil {
-							return fmt.Errorf("executing page.on handler: %w", err)
+						mapping, err := mapMetric(vu, m)
+						if err != nil {
+							return fmt.Errorf("mapping the metric: %w", err)
+						}
+
+						if _, err = handler(sobek.Undefined(), vu.VU.Runtime().ToValue(mapping)); err != nil {
+							return fmt.Errorf("executing page.on('metric') handler: %w", err)
 						}
 						return nil
 					})
--- a/browser/sync_page_mapping.go
+++ b/browser/sync_page_mapping.go
@@ -1,6 +1,7 @@
 package browser
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/grafana/sobek"
@@ -121,8 +122,13 @@ func syncMapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 				_, err := handler(sobek.Undefined(), vu.Runtime().ToValue(mapping))
 				return err
 			}
-			runInTaskQueue := func(m *common.ConsoleMessage) {
+			runInTaskQueue := func(a any) {
 				tq.Queue(func() error {
+					m, ok := a.(*common.ConsoleMessage)
+					if !ok {
+						return errors.New("incorrect message")
+					}
+
 					if err := mapMsgAndHandleEvent(m); err != nil {
 						return fmt.Errorf("executing page.on handler: %w", err)
 					}
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -355,13 +355,7 @@ func (fs *FrameSession) parseAndEmitWebVitalMetric(object string) error {
 	state := fs.vu.State()
 	tags := state.Tags.GetCurrentValues().Tags
 	if state.Options.SystemTags.Has(k6metrics.TagURL) {
-		if name, ok := fs.page.urlGroupingName(fs.ctx, wv.URL); ok {
-			tags = tags.With("url", name)
-			tags = tags.With("name", name)
-		} else {
-			tags = tags.With("url", wv.URL)
-			tags = tags.With("name", wv.URL)
-		}
+		tags = handleURLTag(fs.ctx, fs.page, wv.URL, tags)
 	}
 
 	tags = tags.With("rating", wv.Rating)
--- a/common/frame_session.go
+++ b/common/frame_session.go
@@ -355,7 +355,7 @@ func (fs *FrameSession) parseAndEmitWebVitalMetric(object string) error {
 	state := fs.vu.State()
 	tags := state.Tags.GetCurrentValues().Tags
 	if state.Options.SystemTags.Has(k6metrics.TagURL) {
-		if name, ok := fs.page.URLGroupingName(fs.ctx, wv.URL); ok {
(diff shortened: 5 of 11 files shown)


examples/pageon-metric.js
```diff
@@ -0,0 +1,33 @@
...
+export default async function() {
+  const page = await browser.newPage();
+
+  page.on('metric', (msg) => {
+    msg.groupURLTag({
+      groups: [
+        {url: /^https:\/\/test\.k6\.io\/\?q=[0-9a-z]+$/, name:'test'},
+      ]
+    });
+  });
```
ankur22, line 25: I'm not sure of the naming of the API (method, fields, etc). Any suggestions?

browser/page_mapping.go
```diff
@@ -206,18 +207,57 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
-				_, err := handler(sobek.Undefined(), vu.Runtime().ToValue(mapping))
-				return err
-			}
-			runInTaskQueue := func(m *common.ConsoleMessage) {
-				tq.Queue(func() error {
-					if err := mapMsgAndHandleEvent(m); err != nil {
-						return fmt.Errorf("executing page.on handler: %w", err)
+			var runInTaskQueue func(any)
+			switch event {
+			case common.EventPageConsoleAPICalled:
```
ankur22, line 212: When we come to generalise `page.on` to allow for other events, this is where I think we can do something. All the other changes are very specific to each event.

inancgumus, line 212: Did you consider using a type switch and assert on the `func(any)` type (instead of asserting in each branch in the current `switch`)?—it'll be slightly more type-safe and added to the language for this type of situation.

browser/page_mapping.go
```diff
@@ -206,18 +207,57 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
-					return nil
-				})
+					_, err = handler(sobek.Undefined(), vu.VU.Runtime().ToValue(mapping))
+					return err
+				}
+				runInTaskQueue = func(a any) {
+					tq.Queue(func() error {
+						m, ok := a.(*common.ExportedMetric)
+						if !ok {
+							return errors.New("incorrect metric message")
```
inancgumus, line 244: This and the other error message from a type assertion can also include the received type, so that we can better debug it.

browser/page_mapping.go
```diff
@@ -206,18 +207,57 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
+						}
+						// page.on('metric') needs to be synchronized so that
+						// the name can be returned back to the caller of the
+						// handler. Once the handler has complete we call
+						// Completed so that the caller knows that the handler
+						// has complete and to work with the new metric.
```
inancgumus, line 250: Can you help me understand what this comment means? For example, I don't understand what synchronized and "the name is returning to the caller of the handler" mean.

browser/page_mapping.go
```diff
@@ -206,18 +207,57 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
+						defer m.Completed()
+
+						if err := mapMsgAndHandleEvent(m); err != nil {
+							return fmt.Errorf("executing page.on handler: %w", err)
+						}
+						return nil
+					})
+				}
+			default:
+				return fmt.Errorf("unknown page event: %q", event)
```
inancgumus, line 260: It's nice that we point out the unknown event type so that we can better debug 👍

browser/page_mapping.go
```diff
@@ -206,18 +207,57 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
+						}
+						// page.on('metric') needs to be synchronized so that
+						// the name can be returned back to the caller of the
+						// handler. Once the handler has complete we call
+						// Completed so that the caller knows that the handler
+						// has complete and to work with the new metric.
+						defer m.Completed()
+
+						if err := mapMsgAndHandleEvent(m); err != nil {
+							return fmt.Errorf("executing page.on handler: %w", err)
```
inancgumus, line 254: Is it executing or mapping? WDYT?

```suggestion
							return fmt.Errorf("mapping page.on handler: %w", err)
```

browser/sync_page_mapping.go
```diff
@@ -121,8 +122,13 @@ func syncMapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 				_, err := handler(sobek.Undefined(), vu.Runtime().ToValue(mapping))
 				return err
 			}
-			runInTaskQueue := func(m *common.ConsoleMessage) {
+			runInTaskQueue := func(a any) {
```
inancgumus, line 125: It's nice that we generalize it this way 👍 But, also consider if we can use Go generics here.

common/frame_session.go
```diff
@@ -355,7 +355,13 @@ func (fs *FrameSession) parseAndEmitWebVitalMetric(object string) error {
...
 	tags := state.Tags.GetCurrentValues().Tags
 	if state.Options.SystemTags.Has(k6metrics.TagURL) {
-		tags = tags.With("url", wv.URL)
+		if name, ok := fs.page.urlGroupingName(fs.ctx, wv.URL); ok {
+			tags = tags.With("url", name)
+			tags = tags.With("name", name)
+		} else {
+			tags = tags.With("url", wv.URL)
+			tags = tags.With("name", wv.URL)
+		}
```
inancgumus, line 364: Should we put this part into a function because we duplicate it thrice?

common/page.go
```diff
@@ -364,6 +363,127 @@ func (p *Page) initEvents() {
 	}()
 }
 
+func (p *Page) urlGroupingName(ctx context.Context, urlTag string) (string, bool) {
```
inancgumus, line 366: As far as I can see, a new reader (e.g., me) doesn't know what this method does or what its purpose is in general. I haven't seen it explained anywhere. I'd add a comment here to explain its purpose and how it works concisely.

common/page.go
```diff
@@ -364,6 +363,127 @@ func (p *Page) initEvents() {
...
+		return "", false
+	}
+	p.eventHandlersMu.RUnlock()
+
+	var name string
+	var nameChanged bool
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	p.eventHandlersMu.RLock()
```
inancgumus, line 382: Should the same lock protect the entire method instead of taking it twice here? There can be a race condition between checking the event handlers above and then looping over them below. We might not need this explanation if we go that route (I might be wrong):

```
// A handler can register another handler from within itself. This is
// the reason to unlock the mutex before calling the handler.
```

I'd use another lock for a handler if this approach I suggest is not possible.

common/page.go
```diff
@@ -364,6 +363,127 @@ func (p *Page) initEvents() {
...
+func (p *Page) urlGroupingName(ctx context.Context, urlTag string) (string, bool) {
+	p.eventHandlersMu.RLock()
+
+	// If there are no handlers for EventConsoleAPICalled.
+	if _, ok := p.eventHandlers[EventPageMetricCalled]; !ok {
+		p.eventHandlersMu.RUnlock()
+
+		return "", false
+	}
+	p.eventHandlersMu.RUnlock()
```
inancgumus, line 375: Did you consider using a `defer` to ensure that we unlock the lock?

common/page.go
```diff
@@ -364,6 +363,127 @@ func (p *Page) initEvents() {
...
+
+		nameCh := make(chan string)
+		em := &ExportedMetric{
+			ctx:    ctx,
+			urlTag: urlTag,
+			nameCh: nameCh,
+		}
+
+		// handler runs on another goroutine.
+		h(em)
```
inancgumus, line 397: Where does it run in another goroutine? It seems like a blocking call here.

common/page.go
```diff
@@ -364,6 +363,127 @@ func (p *Page) initEvents() {
...
+	defer cancel()
+
+	p.eventHandlersMu.RLock()
+
+	for _, h := range p.eventHandlers[EventPageMetricCalled] {
+		// A handler can register another handler from within itself. This is
+		// the reason to unlock the mutex before calling the handler.
+		p.eventHandlersMu.RUnlock()
+
+		nameCh := make(chan string)
```
inancgumus, line 389: I'd remove `Ch` from `nameCh` as its type is already a `chan`. What about `names`?

common/page.go
```diff
@@ -364,6 +363,127 @@ func (p *Page) initEvents() {
...
+		// We want to wait for the name to come in from the handler. When the
+		// nameCh is closed by the handler the name is returned back upstream.
+		ok := true
+		for ok {
+			select {
+			case n, o := <-nameCh:
+				ok = o
+				if ok {
+					name = n
+					nameChanged = true
```
inancgumus, line 408: What does `nameChanged` do in this context?

common/page.go
```diff
@@ -364,6 +363,127 @@ func (p *Page) initEvents() {
...
+	urlTag string
+
+	// nameCh is used to return the url tag metric name when a match is found.
+	// It is also used to help sync the call to the handler and the caller.
+	// We need to wait for the handler to complete before proceeding to return
+	// the name of the url grouping.
+	nameCh chan<- string
+}
+
+// URLGroups will contain all the url groupings.
```
inancgumus, line 440:
```suggestion
// URLGroups will contain all the URL groupings.
```

common/page.go
```diff
@@ -364,6 +363,127 @@ func (p *Page) initEvents() {
...
+	ctx context.Context
+
+	// The url value from the metric's url tag. It will be used to match
+	// against the url grouping regexs.
+	urlTag string
+
+	// nameCh is used to return the url tag metric name when a match is found.
+	// It is also used to help sync the call to the handler and the caller.
+	// We need to wait for the handler to complete before proceeding to return
+	// the name of the url grouping.
```
inancgumus, line 436:
```suggestion
	// the name of the URL grouping.
```

common/page.go
```diff
@@ -364,6 +363,127 @@ func (p *Page) initEvents() {
...
+	return name, nameChanged
+}
+
+// ExportedMetric is the type that is exported to JS. It is currently only used to
+// match on the urlTag and return a name when a match is found.
+type ExportedMetric struct {
+	ctx context.Context
+
+	// The url value from the metric's url tag. It will be used to match
+	// against the url grouping regexs.
```
inancgumus, line 430:
```suggestion
	// The URL value from the metric's url tag. It will be used to match
	// against the URL grouping regexs.
```

common/page.go
```diff
@@ -364,6 +363,127 @@ func (p *Page) initEvents() {
...
+	urlTag string
+
+	// nameCh is used to return the url tag metric name when a match is found.
+	// It is also used to help sync the call to the handler and the caller.
+	// We need to wait for the handler to complete before proceeding to return
+	// the name of the url grouping.
+	nameCh chan<- string
```
inancgumus, line 437: It's nice that we specify that this channel is only for send 👍

common/page.go
```diff
@@ -364,6 +363,127 @@ func (p *Page) initEvents() {
...
+}
+
+// URLGroups will contain all the url groupings.
+type URLGroups struct {
+	Groups []URLGroup
+}
```
inancgumus, line 443: I'd remove this type as we can easily use `[]URLGroup`. It doesn't have any methods on it. `[]URLGroup` is as descriptive as `URLGroups`.

common/page.go
```diff
@@ -364,6 +363,127 @@ func (p *Page) initEvents() {
...
+
+// URLGroup contains the single url regex and the name to give to the metric
+// if a match is found.
+type URLGroup struct {
+	// This is a regex.
+	URL string
```
inancgumus, line 449: Should we name it `RegEx` then (like `URLGroup.RegEx` is very descriptive to me)?

common/page.go
```diff
@@ -364,6 +363,127 @@ func (p *Page) initEvents() {
...
+
+// Completed will allow the caller of the handler to know that the handler
+// has completed and therefore to continue the flow.
+func (e *ExportedMetric) Completed() {
+	close(e.nameCh)
+}
+
+// GroupURLTag will find a match given the URLGroups and the URL from the metric
+// tag and send the name via a channel back to the caller of the handler.
+func (e *ExportedMetric) GroupURLTag(callBack func(pattern, url string) (bool, error), groups URLGroups) error {
```
inancgumus, line 462: I'd make this function value type signature a more descriptive type. It isn't easy to read now.

```
callBack func(pattern, url string) (bool, error)
```

common/page.go
```diff
@@ -364,6 +363,127 @@ func (p *Page) initEvents() {
...
+	for _, g := range groups.Groups {
+		if g.Name == "" {
```
inancgumus, line 464: Do we need `strings.Trim` here as users can specify `"   "` as a name?

common/page.go
```diff
@@ -990,18 +1110,21 @@ func (p *Page) NavigationTimeout() time.Duration {
...
-func (p *Page) On(event string, handler func(*ConsoleMessage)) error {
-	if event != eventPageConsoleAPICalled {
-		return fmt.Errorf("unknown page event: %q, must be %q", event, eventPageConsoleAPICalled)
+func (p *Page) On(event string, handler func(any)) error {
+	switch event {
+	case EventPageConsoleAPICalled:
+	case EventPageMetricCalled:
+	default:
+		return fmt.Errorf("unknown page event: %q", event)
 	}
```
inancgumus, line 1119: Generics can also prevent checking the event type. As far as I can tell, we did this check twice, which can be error-prone when we extend the events.

examples/pageon-metric.js
inancgumus: I'd suggest:

```
  page.on('metric', m => {
    m.tagURLs({
      urls: [
        {url: /^https:\/\/test\.k6\.io\/\?q=[0-9a-z]+$/, name:'test'},
      ]
    });
  });
```

Or, even:

```
  page.on('metric', m => {
    m.tagURLs([
      { url: /^https:\/\/test\.k6\.io\/\?q=[0-9a-z]+$/, name:'test' },
    ]);
  });
```

tests/page_test.go
```diff
@@ -1873,3 +1882,174 @@ func TestPageGetAttributeEmpty(t *testing.T) {
...
+			</body>
+		</html>`)
+		require.NoError(t, err)
+	})
+	tb.withHandler("/ping", func(w http.ResponseWriter, r *http.Request) {
+		_, err := fmt.Fprintf(w, `pong`)
+		require.NoError(t, err)
+	})
+
+	ignoreURLs := map[string]interface{}{
```
inancgumus, line 1913:
```suggestion
	ignoreURLs := map[string]any{
```

tests/page_test.go
```diff
@@ -1873,3 +1882,174 @@ func TestPageGetAttributeEmpty(t *testing.T) {
...
+
+						// The name and url tags should have the same value.
+						assert.Equal(t, u, n)
+
+						// If the url is in the ignoreURLs map then this will
+						// not have been matched on by the regex, so continue.
+						if _, ok := ignoreURLs[u]; ok {
+							foundMu.Lock()
+							foundUnamended++
+							foundMu.Unlock()
```
inancgumus, line 2005: I'd use the [atomic package](https://pkg.go.dev/sync/atomic) here.

inancgumus (commented):
Nicely tackled with this complex issue 🙇! TBH, I should also admit that I'm not sure I understand this PR. So, I have some comments to discuss. None of them are strong opinions. Some suggestions are to make this change more descriptive, extendable, and maintainable.

inancgumus:
Idea: If the generics solution doesn't turn out OK, we can also invent a new type, `PageOnEvent`, for `Page.On` events (`eventHandlers   map[string][]func(PageOnEvent)`) to get rid of fragile runtime type assertions. `PageOnEvent struct { ConsoleMessage *ConsoleMessage; ExportedMetric *ExportedMetric }`. Each event can use only the field it is interested in, removing the need for type assertions.

browser/page_mapping.go
```diff
@@ -206,18 +207,57 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
+			switch event {
+			case common.EventPageConsoleAPICalled:
+				mapMsgAndHandleEvent := func(m *common.ConsoleMessage) error {
+					mapping := mapConsoleMessage(vu, m)
+					_, err := handler(sobek.Undefined(), vu.VU.Runtime().ToValue(mapping))
+					return err
+				}
+				runInTaskQueue = func(a any) {
+					tq.Queue(func() error {
+						m, ok := a.(*common.ConsoleMessage)
```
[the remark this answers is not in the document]
ankur22: I've linked this comment to a [comment](https://github.com/grafana/xk6-browser/issues/1227#issuecomment-2393348036) in the generalize `page.on` events [issue](https://github.com/grafana/xk6-browser/issues/1227).

We can tackle the generalization after merging in this PR 👍 

I'll resolve this comment so that I can work on the other feedback.

browser/page_mapping.go
ankur22: I've linked this comment to a [comment](https://github.com/grafana/xk6-browser/issues/1227#issuecomment-2393348036) in the generalize `page.on` events [issue](https://github.com/grafana/xk6-browser/issues/1227).

We can tackle the generalization after merging in this PR 👍 

I'll resolve this comment so that I can work on the other feedback.

browser/page_mapping.go
ankur22: I've linked this comment to a [comment](https://github.com/grafana/xk6-browser/issues/1227#issuecomment-2393348036) in the generalize `page.on` events [issue](https://github.com/grafana/xk6-browser/issues/1227).

We can tackle the generalization after merging in this PR 👍 

I'll resolve this comment so that I can work on the other feedback.

browser/sync_page_mapping.go
ankur22: I've linked this comment to a [comment](https://github.com/grafana/xk6-browser/issues/1227#issuecomment-2393348036) in the generalize `page.on` events [issue](https://github.com/grafana/xk6-browser/issues/1227).

We can tackle the generalization after merging in this PR 👍 

I'll resolve this comment so that I can work on the other feedback.

common/page.go
ankur22: I've linked this comment to a [comment](https://github.com/grafana/xk6-browser/issues/1227#issuecomment-2393348036) in the generalize `page.on` events [issue](https://github.com/grafana/xk6-browser/issues/1227).

We can tackle the generalization after merging in this PR 👍 

I'll resolve this comment so that I can work on the other feedback.

ankur22:
> Idea: If the generics solution doesn't turn out OK, we can also invent a new type, `PageOnEvent`, for `Page.On` events (`eventHandlers map[string][]func(PageOnEvent)`) to get rid of fragile runtime type assertions. `PageOnEvent struct { ConsoleMessage *ConsoleMessage; ExportedMetric *ExportedMetric }`. Each event can use only the field they are interested in, removing the need for type assertions.

Nice alternative!

I've linked this comment to a [comment](https://github.com/grafana/xk6-browser/issues/1227#issuecomment-2393348036) in the generalize `page.on` events [issue](https://github.com/grafana/xk6-browser/issues/1227).

We can tackle the generalization after merging in this PR 👍

tests/page_test.go
ankur22: Sounds good, updated in a3fefb17d024ec9038340cee0540cc02b98a0183

common/page.go
ankur22: Good spot. Fixed in 14f9fd9302c7638948f64ecc1c5cbfb0d30abe77

common/page.go
ankur22: I prefer `url` since in the future we may add functionality to work with `glob` or even exact matches.

common/frame_session.go
ankur22: Nice idea. Done in 70c9de8a640b77035f016a2f6e5a1f9b1aa2eacf

browser/metric_mapping.go
```diff
@@ -0,0 +1,43 @@
...
+	// We're setting up the function in the Sobek context that will be reused
+	// for this VU.
+	_, err := rt.RunString(`
+	function _k6BrowserURLGroupingTest(pattern, url) {
+		let r = pattern;
+		if (typeof pattern === 'string') {
+			r = new RegExp(pattern);
+		}
+		return r.test(url);
+	}`)
```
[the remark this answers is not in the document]
ankur22: Unfortunately there are differences between the Golang and JS (ECMA) regex engines (as there are with [all regex engines](https://en.wikipedia.org/wiki/Comparison_of_regular_expression_engines)). Since the user will be writing the regex in the JS test script, I believe it is better to stay consistent and use the JS (Sobek's) regex engine instead of Go's implementation to avoid confusion.

browser/page_mapping.go
ankur22: Right, `mapMsgAndHandleEvent` does the mapping and calling of the handler. I don't quite understand why `mapMsgAndHandleEvent` is its own function. I've removed `mapMsgAndHandleEvent` and now the mapping and call to the handler are done from here, with better errors. Done in fd8aa9778ca160ede9d775bc60c0799fabb49e97

common/page.go
ankur22: So the handler is called, and within the handle there is a call to [add it to the taskqueue](https://github.com/grafana/xk6-browser/blob/fd8aa9778ca160ede9d775bc60c0799fabb49e97/browser/page_mapping.go#L233-L254). When something is ran on the task queue it is ran on a separate goroutine to what added the function to the task queue.

I was looking at the [waitForEvent](https://github.com/grafana/xk6-browser/blob/8fce16a75360a2317dfe2ff4c8ac4e5a2c292b5b/browser/browser_context_mapping.go#L113-L132) implementation and i'm going to attempt to work with that.

ankur22: I think i've made it a little easier now in ef24fb2782d1956292e2fa989fbf5e5b4e89f181. It doesn't rely on a channel to return the name updates, instead a new name filed is updated in ExportedMetric which can be read by the caller in `urlGroupingName`.

The call to `h(em)` is now also synchronous since we're synchronising the use of the taskqueue in the mapping layer, like with [waitForEvent](https://github.com/grafana/xk6-browser/blob/8fce16a75360a2317dfe2ff4c8ac4e5a2c292b5b/browser/browser_context_mapping.go#L113-L132).

common/page.go
ankur22: I've removed `nameCh` in ef24fb2782d1956292e2fa989fbf5e5b4e89f181. If we need it back, i'll ensure to use `names`.

common/page.go
ankur22: Sure, added in 34f3eaab5dbec984cff25f1f626924831a699a52

common/page.go
ankur22: The reason the release the lock is called before calling the handler is because the handler can itself add another handler. This allows for that. If we didn't release the lock and a suer tried to add another handler from another handler, then it would deadlock. I've added a test to showcase that behaviour: 814f73c577ccb8e73358c873afdbf2372851ab72

ankur22:
> Nicely tackled with this complex issue 🙇! TBH, I should also admit that I'm not sure I understand this PR's code fully 😄 (my bad) So, I have some comments to discuss. None of them are strong opinions. Some suggestions are to hopefully make this change more descriptive, extendable, and maintainable.

Thanks for the review 🙏 I agree that this is a bit confusing. I will create a diagram to make things a bit clearer.

examples/pageon-metric.js
ankur22: +1 for `urls` instead of `groups` 👍

ankur22: What about `groupTags` instead of `groupURLTag`?:

  page.on('metric', (msg) => {
    msg.groupTags({
      urls: [
        {url: /^https:\/\/test\.k6\.io\/\?q=[0-9a-z]+$/, name:'test'},
      ]
    });
  });

We're not duplicating the word url as much, and we're allowing for it to be extended in case we need to group another tag in the future.

common/page.go
ankur22: Sure, added it in f53dfa5d96ddba47bd86de98135dd87ae34c3eba

common/page.go
ankur22: I think i need to wait for what we decide on in https://github.com/grafana/xk6-browser/pull/1456/files#r1786131678 before actioning this.

examples/pageon-metric.js
inancgumus: Are we grouping tags, or are we telling k6 to tag metrics that match the criteria(s) like URLs? So, our goal here is not to "group tags"; instead, "tagging the metrics" (not even URLs in general) so that k6 can automatically group them. That's why I suggested `tagURLs`. Now I think calling it `tag` might be better after seeing your reasoning 👍 

Something like:

```javascript
  page.on('metric', metric => {
    // give a tag to the metric that matches the URL
    metric.tag({
      urls: [
        {url: /^https:\/\/test\.k6\.io\/\?q=[0-9a-z]+$/, name:'test'},
      ]
    });
  });
```

_PS: Making it extendable is good, but it also complicates the API :)_

browser/metric_mapping.go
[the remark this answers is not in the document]
inancgumus: 👍

browser/page_mapping.go
inancgumus: 👍

common/frame_session.go
inancgumus: 👍

common/page.go
inancgumus: Thanks!

common/page.go
inancgumus: Why do we allow handlers to add handlers? Do you think it's valid use case?

common/page.go
inancgumus: Perfect!

common/page.go
inancgumus: 👍

common/page.go
inancgumus: Clicking on that link didn't take me anywhere :( I would like to know which decision we're waiting for.

common/page.go
inancgumus: Fine, but we can always change it to `URL` in the future when we need it?

common/page.go
ankur22: I'm not sure of what the design philosophy is for this. I've tested with PW and they allow it:

```js
// @ts-check
const { test, chromium } = require('@playwright/test');

test.describe('lifecycle', () => {
  test('lifecycle', async () => {
    const browser = await chromium.launch();
  
    page.on('response', (response) => {
      console.log(`Received response: ${response.url()}`);
  
      page.on('request', (request) => {
        console.log(`Captured request: ${request.url()}`);
      });
    });
  
    await page.goto('https://test.k6.io');
    
    await browser.close();
  });
});
```

common/page.go
ankur22: Keeps track of the whether the name was changed, and returns true if it has or false. Makes it simpler upstream to then work with the name or not based on whether it is true or false.

common/page.go
ankur22: That would be a breaking change.

I think `url` is a good choice, just as k6 have done [here](https://grafana.com/docs/k6/latest/using-k6/http-requests/#url-grouping).

common/page.go
ankur22: `nameCh` has been removed. If the channel is reintroduced it'll be named `names`.

examples/pageon-metric.js
ankur22: > JavaScript is not like Go. We can detect whether users passing an Object or an array without making any breaking changes. Making the API extendable is good, but it also complicates it. So, I'd vote for keeping the API simpler rather than complicating it for unforeseen future needs.

That's true, but I think that's messy and not a great DX. It complicates other things too, such as the implementation in go, type defs and documentation. I'd still prefer to keep the object rather than an array.

> By definition, tagging means grouping in k6 terms.

Yeah I think that makes sense 👍 

> Saying "group" is also a bit confusing as k6 has a group concept.

I don't quite align with you here, but since i agree with what you mean by tagging, i'm happy to drop `group`.

> Are we grouping tags, or are we telling k6 to tag metrics that match the criteria(s), like URLs?

Good bit of analysis on what we're trying to do here, and I agree.

I like the latest suggestion 👍 

```js
  page.on('metric', metric => {
    // give a tag to the metrics that matches the URL
    metric.tag({
      urls: [
        {url: /^https:\/\/test\.k6\.io\/\?q=[0-9a-z]+$/, name:'test'},
      ]
    });
  });
```

> // getting a `msg` on a `metric` sounds a bit weird?

It's technically an event, so i think msg is fine, we're not passing the `metric` itself to the event handler. Happy with metric though as it better represents where the action will take place and easier to reason with 👍

browser/page_mapping.go
ankur22: I don't think this is relevant now due to the change in 8f21d3ae70de26741628e2309087bd359940abcd. If we revert back to the previous approach then I'll ensure to make this clearer.

common/page.go
ankur22: The comment thread which starts with `I'm not sure of the naming of the API (method, fields, etc). Any suggestions?`

examples/pageon-metric.js
ankur22: Changed to the latest suggestion in 597ddd7e65d2edf694206edd2431639ed27df5d4

ankur22: Changes `msg` to `metric` in 72d59070025cb4a645d9db117a4a9d532aca755b

ankur22:
@inancgumus i created the diagram and placed it in the description of the issue. Let me know if it helps or not 🙂

examples/pageon-metric.js
inancgumus: Thanks :)

common/page.go
inancgumus: 👍 To streamline this and avoid mistakes, I believe we can add a helper function that checks if an event is registered because we have similar logic in [urlTagName](https://github.com/grafana/xk6-browser/blob/31050eb39c01897d346070cc4669dfdf84499183/common/page.go#L375-L382) and [onConsoleAPICalled](https://github.com/grafana/xk6-browser/blob/31050eb39c01897d346070cc4669dfdf84499183/common/page.go#L1484-L1490). By the way, in `urlTagName`, we might have forgotten to read-unlock the lock before the first `if`.

common/page.go
inancgumus: Ah, got it. We export it.

common/page.go
inancgumus: I guess `name` here means a "tag name." And, `nameChanged` means that the current URL matches to a user-provided URL pattern. So, we change the tag name. By upstream, you mean the metric system. Right? How can we make this flow/logic clearer in the code? It might be obvious to you since you have worked on it—but not others like me 😅

common/page.go
ankur22: Sure, I feel like that could be something done in the generalize issue. I've added the link to the comment 👍

common/page.go
ankur22: Where's the confusion? Is it the names of the variables?

> By upstream, you mean the metric system. Right?
 👍

common/page.go
```diff
@@ -364,6 +363,105 @@ func (p *Page) initEvents() {
...
+func (e *MetricEvent) Tag(callBack regexCallback, overrides URLOverrides) error {
+	for _, o := range overrides.URLs {
+		name := strings.TrimSpace(o.Name)
+		if name == "" {
+			return fmt.Errorf("name %q is invalid", o.Name)
+		}
+
+		// callback is a function that will perform the regex test in the Sobek
+		// runtime.
+		val, err := callBack(o.URL, e.urlTag)
```
inancgumus: `val` or `matched`? What does a value mean in this context? I'd make it `matched` if it's about matching a URL to a user-provided pattern. And, `e.tagName = userProvidedTagName` would make it clearer instead of `e.name = &name`. Using a pointer is also confusing. Maybe add a field like `isUserURLTagNameExist bool` instead?

browser/metric_event_mapping.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+		return r.test(url);
+	}`)
+	if err != nil {
+		return nil, fmt.Errorf("evaluating regex function: %w", err)
+	}
+
+	return mapping{
+		"Tag": func(urls common.URLOverrides) error {
+			callback := func(pattern, url string) (bool, error) {
```
inancgumus, line 29: I'd call this `matchRegex` or something more descriptive to be clearer and easier to understand.

browser/page_mapping.go
```diff
@@ -206,18 +207,56 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
-				return err
-			}
-			runInTaskQueue := func(m *common.ConsoleMessage) {
-				tq.Queue(func() error {
-					if err := mapMsgAndHandleEvent(m); err != nil {
-						return fmt.Errorf("executing page.on handler: %w", err)
-					}
-					return nil
-				})
+			var runInTaskQueue func(any)
```
inancgumus, line 210: Is it possible to group similar code under each event? They almost do the same thing with seemingly tiny differences.

common/page.go
ankur22: Sure, I've changed it in 1f2b350eb5aecfe2527cd9ab7880e2f1254d0df6

common/network_manager.go
```diff
@@ -57,6 +57,10 @@ func (c *Credentials) Parse(ctx context.Context, credentials sobek.Value) error
 	return nil
 }
 
+type metricInterceptor interface {
+	urlTagName(ctx context.Context, urlTag string) (string, bool)
+}
```
inancgumus, line 62: It's fine this way. But why don't we use a simple function value instead of declaring an interface—if this is only about testing? I wouldn't use an interface only for testing as they are mostly for extending the functionality for unforeseen cases (think io.Writer).

common/network_manager.go
```diff
@@ -89,6 +94,7 @@ type NetworkManager struct {
 // NewNetworkManager creates a new network manager.
 func NewNetworkManager(
 	ctx context.Context, customMetrics *k6ext.CustomMetrics, s session, fm *FrameManager, parent *NetworkManager,
+	mi metricInterceptor,
```
inancgumus, line 97: I'd reformat and lay out these inputs over separate lines as they got long.

common/page.go
ankur22: I've updated the variable names: 760a38042cc4c3554cc53f4c002e8343614982f4

common/network_manager_test.go
```diff
@@ -209,6 +209,12 @@ func TestOnRequestPausedBlockedIPs(t *testing.T) {
 	}
 }
 
+type MetricInterceptorMock struct{}
+
+func (m *MetricInterceptorMock) urlTagName(ctx context.Context, urlTag string) (string, bool) {
+	return "", false
+}
```
inancgumus, line 216: It is not the best win, but this would be a one-liner if we had used a function value instead of an interface as well.

tests/page_test.go
```diff
@@ -1873,3 +1882,187 @@ func TestPageGetAttributeEmpty(t *testing.T) {
...
+					});
+				});
+			});`,
+			want: "ping-4",
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
```
inancgumus, line 1999: Nice tests, but do we forget to call `t.Parallel`?

common/page.go
```diff
@@ -364,6 +363,105 @@ func (p *Page) initEvents() {
...
+// the supplied name is returned back upstream to the caller of urlTagName.
+func (p *Page) urlTagName(ctx context.Context, urlTag string) (string, bool) {
+	p.eventHandlersMu.RLock()
+
+	// If there are no handlers for EventConsoleAPICalled.
+	if _, ok := p.eventHandlers[EventPageMetricCalled]; !ok {
+		p.eventHandlersMu.RUnlock()
+
+		return "", false
+	}
```
inancgumus, line 382: We might put this in a different function so that the lock handling would be simpler. We could also call that function from `Page.onConsoleAPICalled`. This way we won't need to provide `RUnlock` in the loop as a side benefit.

inancgumus (approved):
Awesome work 🎉 I'm approving this because it's functionally correct 👍 I've also added some suggestions for you to consider before merging it in.

common/page.go
```diff
@@ -364,6 +363,105 @@ func (p *Page) initEvents() {
...
+	urlTag string
+
+	// When a match is found this name field should be updated.
+	name *string
+}
+
+// URLOverrides will contain all the URL groupings.
+type URLOverrides struct {
+	URLs []URLOverride `js:"urls"`
+}
```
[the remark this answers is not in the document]
ankur22: This was changed in 45b6cd532616541d5e6eb34d91b1b8a2b53e5482

browser/metric_event_mapping.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// mapMetricEvent to the JS module.
+func mapMetricEvent(vu moduleVU, cm *common.MetricEvent) (mapping, error) {
+	rt := vu.VU.Runtime()
+
+	// We're setting up the function in the Sobek context that will be reused
+	// for this VU.
+	_, err := rt.RunString(`
+	function _k6BrowserURLGroupingTest(pattern, url) {
```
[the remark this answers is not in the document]
ankur22: Changed in 044cc031695cfe8db1e2c8fb2c3e6225e90e2cc9

common/page.go
inancgumus: The confusion arises because of the naming and also the nature of the change. It has many indirections. So, it's not easy to wrap my head around. Naming adds more confusion to the party 😄

common/page.go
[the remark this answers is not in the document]
inancgumus: Sorry, I might have re-added some of my comments because I lost them. Something happened on my computer while reviewing.

common/page.go
```diff
@@ -364,6 +363,105 @@ func (p *Page) initEvents() {
...
+	URL string
+	// The name to send back to the caller of the handler.
+	Name string
+}
+
+type regexCallback func(pattern, url string) (bool, error)
+
+// Tag will find the first match given the URLOverrides and the URL from
+// the metric tag and update the name field.
+func (e *MetricEvent) Tag(callBack regexCallback, overrides URLOverrides) error {
```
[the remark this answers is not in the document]
ankur22: Done in 8fd880a9c1141a6a467cd1fd0dd0782b6f99deae

common/page.go
ankur22: Updated val to matched in e6d3c398589090d56c89349478c7a1fc98c57504

ankur22: Renamed name to userProvidedTagName 826b52711ea6ca3e9522372f3b6cf873bc2e970d

examples/pageon-metric.js
```diff
@@ -0,0 +1,33 @@
...
+  },
+}
+
+export default async function() {
+  const page = await browser.newPage();
+
+  page.on('metric', (metric) => {
+    metric.Tag({
+      urls: [
+        {urlRegEx: /^https:\/\/test\.k6\.io\/\?q=[0-9a-z]+$/, tagName:'test'},
```
inancgumus, line 22: This was simpler like:

```js
{url: /^https:\/\/test\.k6\.io\/\?q=[0-9a-z]+$/, name:'test'},
```

My suggestion was about the Go code, not the JavaScript API. I'm sorry I didn't make it clearer back then 😢 The updated Go code is fine, but the previous JS API was more streamlined. Is it possible to revert it to the old API using `json:...` tags while keeping the updated Go code intact? Skip this suggestion if you prefer the new JS API.

inancgumus (approved):
One last note.

common/page.go
ankur22: Added bool check in a42e4e709031a79471fbc373da09563c3e84264b

browser/metric_event_mapping.go
ankur22: Done in 8fd880a9c1141a6a467cd1fd0dd0782b6f99deae

browser/page_mapping.go
ankur22: Yep, you've already commented on this earlier i believe. It's in the generalize issue 👍

common/network_manager.go
ankur22: Right, that didn't occur to me earlier. I think i'll leave it as an interface though as it's not critical.

common/network_manager.go
ankur22: This is a nit pick but done in 4d9fa0c5c49fabf128adf3721e64f2988b11cef5

common/network_manager_test.go
ankur22: I think i'll leave it as it for now, thanks for the info though.

common/network_manager.go
inancgumus: Yep, it's a small detail. Nothing important. Feel free to skip some of my suggestions. I've already approved the PR. Just noting what I see.

common/network_manager.go
inancgumus: Sure.

tests/page_test.go
ankur22: Good spot, done in bcc15bc7cd5b6e10835688d650a6a5613a662b01

common/page.go
ankur22: Yep, you've mentioned this earlier. It's in the generalize issue 👍

examples/pageon-metric.js
ankur22: Right, ok, done in a6f9e3758130b139c9a872b6748b6d9b675f6a4e

inancgumus (approved):

PR #1469, opened 2024-10-08
Add workaround to retry `waitFor` until it is success or retry count is 0

ankur22:
## What?

We're allowing retries on `waitFor` when the following errors occur when `waitFor` is waiting:

- "Inspected target navigated or closed"
- "Cannot find context with specified id"
- "Execution context was destroyed"

To prevent a stack overflow, it will retry for a maximum of 20 times before failing. In my tests i've seen the retry count drop from 20 to 12 at most. I feel 20 is a safe number that shouldn't be reached in normal SSO login flows. I want to avoid using a larger retry count to force us to fix this permanently asap.

## Why?

This PR fixes an issue where `waitFor` fails to wait while the underlying page it is waiting on is navigating. A common scenario where we might see this is during a login/logout flow using SSO. The initial page will be redirected multiple times until we get to the desired page/state.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1472

### DIFF
--- a/common/frame.go
+++ b/common/frame.go
@@ -564,13 +564,18 @@ func (f *Frame) waitForSelector(selector string, opts *FrameWaitForSelectorOptio
 	return handle, nil
 }
 
-func (f *Frame) waitFor(selector string, opts *FrameWaitForSelectorOptions) error {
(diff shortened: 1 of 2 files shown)


inancgumus (approved):
🚀

PR #1475, opened 2024-10-09
Streamline the addition and running of new `page.on` events

inancgumus:
## What?

Generalize `page.on` events.

## Why?

To streamline the addition and running of additional `page.on` events.

## Checklist

## Related PR(s)/Issue(s)

Updates: #1227

### DIFF
--- a/browser/page_mapping.go
+++ b/browser/page_mapping.go
@@ -449,17 +449,11 @@ func mapPageOn(vu moduleVU, p *common.Page) func(common.PageOnEventName, sobek.C
 				<-done
 			}
 		}
-		_ = queueHandler
 
 		onEventPageConsoleAPICalled := func(event common.PageOnEvent) {
-			tq.Queue(func() error {
-				mapping := mapConsoleMessage(vu, event)
-				_, err := handleEvent(sobek.Undefined(), rt.ToValue(mapping))
-				if err != nil {
-					return fmt.Errorf("executing page.on handler: %w", err)
-				}
-				return nil
-			})
+			mapp = mapConsoleMessage
+			wait = false
+			queueHandler(event)
 		}
 
 		onEventPageMetricCalled := func(event common.PageOnEvent) {
--- a/browser/page_mapping.go
+++ b/browser/page_mapping.go
@@ -434,6 +434,9 @@ func mapPageOn(vu moduleVU, p *common.Page) func(common.PageOnEventName, sobek.C
 		},
 		common.EventPageMetricCalled: {
 			mapp: mapMetricEvent,
+			prep: func() error {
+				return prepK6BrowserRegExChecker(rt)
+			},
 			wait: true,
 		},
 	}
--- a/browser/page_mapping.go
+++ b/browser/page_mapping.go
@@ -419,14 +419,14 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 }
 
 // mapPageOn maps the requested page.on event to the Sobek runtime.
-// It generalizes the handling of page.on events on a taskqueue.
-func mapPageOn(vu moduleVU, p *common.Page) func(common.PageOnEventName, sobek.Callable) error { //nolint:funlen
+// It generalizes the handling of page.on events.
+func mapPageOn(vu moduleVU, p *common.Page) func(common.PageOnEventName, sobek.Callable) error {
 	rt := vu.Runtime()
 
 	pageOnEvents := map[common.PageOnEventName]struct {
 		mapp func(vu moduleVU, event common.PageOnEvent) mapping
-		prep func() error
-		wait bool // should we wait for the handler to complete?
+		init func() error // If set, runs before the event handler.
+		wait bool         // Whether to wait for the handler to complete.
 	}{
 		common.EventPageConsoleAPICalled: {
 			mapp: mapConsoleMessage,
@@ -434,7 +434,7 @@ func mapPageOn(vu moduleVU, p *common.Page) func(common.PageOnEventName, sobek.C
 		},
 		common.EventPageMetricCalled: {
 			mapp: mapMetricEvent,
-			prep: prepK6BrowserRegExChecker(rt),
+			init: prepK6BrowserRegExChecker(rt),
 			wait: true,
 		},
 	}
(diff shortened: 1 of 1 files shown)


browser/page_mapping.go
```diff
@@ -423,70 +423,87 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 func mapPageOn(vu moduleVU, p *common.Page) func(common.PageOnEventName, sobek.Callable) error { //nolint:funlen
 	rt := vu.Runtime()
 
+	pageOnEvents := map[common.PageOnEventName]struct {
+		mapp func(vu moduleVU, event common.PageOnEvent) mapping
+		prep func() error
```
ankur22, line 428: Might be worth adding a comment here. Maybe something like:

```suggestion
		prep func() error // If set, this will run before the handler itself.
```

ankur22, line 428: I saw that `initiating` is used in the error if the prep function fails, and actually I quite like that over prepare. Maybe this can be:

```suggestion
		init func() error
```

browser/page_mapping.go
```diff
@@ -423,70 +423,87 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
-			tq.Queue(func() error {
-				mapping := mapConsoleMessage(vu, event)
-				_, err := handleEvent(sobek.Undefined(), rt.ToValue(mapping))
-				if err != nil {
-					return fmt.Errorf("executing page.on handler: %w", err)
-				}
-				return nil
-			})
+		// Prepare the environment for the page.on event handler if necessary.
+		if pageOnEvent.prep != nil {
```
ankur22, line 449: WDYT?

```suggestion
		// Initialize the environment for the page.on event handler if necessary.
		if pageOnEvent.init != nil {
```

browser/page_mapping.go
```diff
@@ -423,70 +423,87 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
+		tq := vu.taskQueueRegistry.get(vu.Context(), p.TargetID())
+		queueHandler := func(event common.PageOnEvent) {
 			done := make(chan struct{})
+
 			tq.Queue(func() error {
 				defer close(done)
 
-				mapping := mapMetricEvent(vu, event)
-				_, err := handleEvent(sobek.Undefined(), rt.ToValue(mapping))
+				mapping := pageOnEvent.mapp(vu, event)
```
ankur22, line 464: I know this isn't due to your change, but I'm wondering if we can perform the mapping outside the taskqueue? 🤔

browser/page_mapping.go
```diff
@@ -423,70 +423,87 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
-			// to comply with what users expect when using the `tag` method.
-			_, err := rt.RunString(`
-				function _k6BrowserCheckRegEx(pattern, url) {
-					let r = pattern;
-					if (typeof pattern === 'string') {
-						r = new RegExp(pattern);
-					}
-					return r.test(url);
+		// Run the the event handler in the task queue to ensure that
+		// the handler is executed in the event loop.
```
ankur22, line 483: I think this comment should be placed above `tq.Queue(func() error {`.

browser/page_mapping.go
```diff
@@ -423,70 +423,87 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
+		_, err := rt.RunString(`
+			function _k6BrowserCheckRegEx(pattern, url) {
+				let r = pattern;
+				if (typeof pattern === 'string') {
+					r = new RegExp(pattern);
 				}
-			`)
-			if err != nil {
-				return fmt.Errorf("evaluating regex function: %w", err)
+				return r.test(url);
```
ankur22, line 499: I just remembered that this isn't correct. Happy to change this myself in another PR, but we only accept JS regex types not strings. So this should just be:

```suggestion
				return pattern.test(url);
```

I think in the future we will need to accept glob patterns and maybe even exact string matches.

I don't think we will be able to accept a string pattern that is used in `new RegExp` due to working with Sobek.

If a user uses `'/^https:\/\/example\.com\/checkout\/[0-9a-z]+$/'` Sobek removes the delimieters `/^https://example.com/checkout/[0-9a-z]+$/`, which makes it impossible to work with in `new RegExp`.

ankur22 (approved):
Great work on this! I like the solution a lot.

I've added suggestions, which are mainly nit picks tbh.

browser/page_mapping.go
```diff
@@ -423,70 +423,87 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
...
 
-		onEventPageMetricCalled := func(event common.PageOnEvent) {
-			// The function on the taskqueue runs in its own goroutine
-			// so we need to use a channel to wait for it to complete
-			// since we're waiting for updates from the handler which
-			// will be written to the ExportedMetric.
+		// Queue the event handler in the task queue.
+		// Wait for the handler to complete if necessary.
+		tq := vu.taskQueueRegistry.get(vu.Context(), p.TargetID())
+		queueHandler := func(event common.PageOnEvent) {
```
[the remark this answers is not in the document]
inancgumus: Here, I mean "queue the handler", not a queue handler :)

browser/page_mapping.go
```diff
@@ -423,70 +423,87 @@ func mapPage(vu moduleVU, p *common.Page) mapping { //nolint:gocognit,cyclop
 func mapPageOn(vu moduleVU, p *common.Page) func(common.PageOnEventName, sobek.Callable) error { //nolint:funlen
 	rt := vu.Runtime()
 
+	pageOnEvents := map[common.PageOnEventName]struct {
+		mapp func(vu moduleVU, event common.PageOnEvent) mapping
+		prep func() error
+		wait bool // should we wait for the handler to complete?
```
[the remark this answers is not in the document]
inancgumus: I like to write comments as questions so that the reader can think about it while reading. But OK :)

browser/page_mapping.go
inancgumus: Nice idea 👍

browser/page_mapping.go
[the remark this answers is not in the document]
inancgumus: Done anyways 👍

Issue #1480, opened 2024-10-11
Memory usage high


ankur22:
#### io.ReadAll

We are currently using the Gorilla Websocket dependency to communicate with Chromium over the websocket using CDP. The particular issue is when [reading off the websocket](https://github.com/grafana/xk6-browser/blob/6d3fb2c948cc0b56bbca6b148d9101aa0d10e6e0/common/connection.go#L323).

When a test is navigating to a media rich website, all the images are returned to k6. They are read using `io.ReadAll` in the websocket library. ~~This isn't a memory efficient way of reading bytes from the websocket. We should take a look at alternate ways of reading off the websocket, such as streaming the bytes.~~

The memory was being held by the read [queue](https://github.com/grafana/xk6-browser/blob/main/common/event_emitter.go#L172) when the element wasn't overwritten with a new Event before aliasing the slice.

#### Request/Response

Request and responses are being stored in memory when the browser module intercepts them. ~~When the loading completes of a request it is [deleted](https://github.com/grafana/xk6-browser/blob/427cbbf323102a2bf74a9f4b84d9cbf23693fe72/common/network_manager.go#L181) (there are other times the request is deleted). It is only deleted from `reqIDToRequest`, but the request and response hold references to each other so they never get picked up by the GC. What makes things difficult is if the test script is holding on to a reference to the either the request or response, then we shouldn't delete them from memory otherwise we end up with NPDs if the user tries to access them from the test script.~~

This issue is related to the above issue with the read queue.

#### Connection and other leaks

We need to `nil` the `conn` field in the `Browser` so that it is correctly picked up by GC.

### xk6-browser version

v1.8.5

### OS

NA

### Chrome version

NA

### Steps to reproduce the problem

1. Find a website with lots of images and use PPROF to track the memory usage.
2. If you increase the VU count, the memory usage will keep increasing.
3. There will be a point when an OOM will occur when increasing the VU count.

### Expected behaviour

The memory usage shouldn't be as high as it currently is.

### Actual behaviour

Memory usage is too high.

```[tasklist]
- [ ] https://github.com/grafana/xk6-browser/pull/1488
- [ ] https://github.com/grafana/xk6-browser/pull/1493
- [x] Add to release notes: https://github.com/grafana/k6/commit/705866817c8b8fffebb31c147be98b2dfc43ee21
```

PR #1482, opened 2024-10-11
Fix recvloop so it doens't use as much memory

ankur22:
## What?

Avoiding working with `io.ReadAll` when reading off the websocket connection.

## Why?

`recvloop` was using `ws.ReadMessage`, which was in turn working with `io.ReadAll`. It turns out for the current use case, this isn't a good way to read the message off the web socket connection. It works better to buffer the read, store it in a buffer that can increase with the test, which is then cloned before being unmarshaled. We have more control over the memory by working in this way.

When working with a simple test against a test website the memory consumption dropped from 120MB to 20MB. This memory usage drop will depend on the website under test. Websites that use a lot of high res images should see a drop in memory usage.

There is still one more thing to do try (in a new PR) to improve the memory/CPU here, which is to try a different JSON unmarshaling tool.

## Checklist

## Related PR(s)/Issue(s)

### DIFF
--- a/common/connection.go
+++ b/common/connection.go
@@ -1,13 +1,11 @@
 package common
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
-	"io"
 	"net/http"
 	"sync"
 	"sync/atomic"
@@ -321,13 +319,7 @@ func (c *Connection) findTargetIDForLog(id target.SessionID) target.ID {
 }
 
 func (c *Connection) recvLoop() {
-	// Setting the wsBuffer as the same size as bufio.NewReader. We should look at
-	// optimizing it depending on the average website, or allow it to be
-	// configurable, but it depends on how much this affects the user.
-	wsBuffer := make([]byte, 4096)
-	// This buffer grows as the test progresses.
-	var recvBuffer []byte
-	var bufReader bufio.Reader
+	var buffer bytes.Buffer
 
 	c.logger.Debugf("Connection:recvLoop", "wsURL:%q", c.wsURL)
 	for {
@@ -337,28 +329,17 @@ func (c *Connection) recvLoop() {
 			return
 		}
 
-		// Buffered reads from the websocket connection.
-		bufReader.Reset(reader)
-		for {
-			n, err := bufReader.Read(wsBuffer)
-			if err == io.EOF {
-				break
-			}
-			if err != nil {
-				c.handleIOError(err)
-				return
-			}
-			recvBuffer = append(recvBuffer, bytes.Clone(wsBuffer[:n])...)
+		if _, err := buffer.ReadFrom(reader); err != nil {
+			c.handleIOError(err)
+			return
 		}
 
 		// Clone the bytes from the recvBuffer to buf before unmarshaling
-		// the message. Without this test the unmarshaling will fail as
-		// it reuses the buf as the underlying bytes buffer.
-		buf := make([]byte, len(recvBuffer))
-		copy(buf, recvBuffer)
+		// the message. Without this, the unmarshaling will fail as it
+		// reuses the buf as the underlying bytes buffer.
+		buf := bytes.Clone(buffer.Bytes())
 
-		// Reset recvBuffer so it can be reused.
-		recvBuffer = recvBuffer[:0]
+		buffer.Reset()
 
 		c.logger.Tracef("cdp:recv", "<- %s", buf)
 
--- a/common/connection.go
+++ b/common/connection.go
@@ -1,10 +1,13 @@
 package common
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


inancgumus (commented):
Correctly using `io.Reader` directly is risky and highly likely to create unforeseen issues. Since our goal is to reuse the same buffer instead of allocating a new one, we can use `bytes.Buffer` to let it handle the `io.Reader`. It reuses the same internal bytes buffer. Go has `bytes.Buffer` for these kind of situations:

```go
var buffer bytes.Buffer

// ...

if _, err := buffer.ReadFrom(reader); err != nil {
    // ...
}

// Clone so that we can share the underlying bytes
buf := bytes.Clone(buffer.Bytes())

// Reuse the underlying byte slice
buffer.Reset()
```

----

💡 It also makes sense to add a benchmark to see how much memory usage we reduce and keep it at that level from now on.

ankur22:
Thanks for the suggestion @inancgumus. I agree that this is less maintainable and more error prone. I did test with `bytes.Buffer` (i implemented it in the exact same way as you've shown). In my tests i found `bytes.Buffer` used more memory than the current implementation. I need to rerun the tests. I like the idea of using benchmark to make it clearer which is the better approach 👍

I'll do three tests:
1. Without the fix;
2. With `bytes.buffer`;
3. With the current approach.

common/network_manager.go
```diff
@@ -179,9 +179,27 @@ func parseTTL(ttlS string) (time.Duration, error) {
...
-	defer m.reqsMu.Unlock()
+	req = m.reqIDToRequest[reqID]
 	delete(m.reqIDToRequest, reqID)
+	m.reqsMu.Unlock()
+
+	// We need to nil the request in response otherwise they hold onto each
+	// others reference preventing the GC from cleaning the memory up.
+	req.responseMu.Lock()
+	if req.response != nil {
+		req.response.request = nil
```
ankur22, line 192: This was the main issue causing the memory leak.

ankur22: Also leaks in `browser.conn` which requires it to be `nil`ed when the iteration ends, and other such objects needs to be `nil`ed when the iteration ends.

ankur22: I have moved the commits to explore the other memory leaks into a new branch and [PR](https://github.com/grafana/xk6-browser/pull/1484).

common/connection.go
```diff
@@ -321,13 +319,7 @@ func (c *Connection) findTargetIDForLog(id target.SessionID) target.ID {
...
 
 func (c *Connection) recvLoop() {
-	// Setting the wsBuffer as the same size as bufio.NewReader. We should look at
-	// optimizing it depending on the average website, or allow it to be
-	// configurable, but it depends on how much this affects the user.
-	wsBuffer := make([]byte, 4096)
-	// This buffer grows as the test progresses.
-	var recvBuffer []byte
-	var bufReader bufio.Reader
+	var buffer bytes.Buffer
```
ankur22, line 322: > It's nice that we find a solution for this tedious issue 🎉
> 
> I have suggestions.
> 
> This PR adds this:
> 
> https://github.com/grafana/xk6-browser/blob/66179668cb8bc691b1150931dd2f56161e12df36/common/connection.go#L324-L362
> 
> I believe keeping this already tricky and central part of the code simpler will serve us better in the long run.
> 
> Also, correctly using `io.Reader` directly is risky and highly likely to create unforeseen issues. Since our goal is to reuse the same buffer instead of allocating a new one, we can use `bytes.Buffer` to let it handle the `io.Reader`. It reuses the same internal bytes buffer. Go has `bytes.Buffer` for these kind of situations:
> 
> ```go
> var buffer bytes.Buffer
> 
> // loop:
> if _, err := buffer.ReadFrom(reader); err != nil {
>     // ...
> }
> 
> // Clone so that we can share the underlying bytes
> buf := bytes.Clone(buffer.Bytes())
> 
> // Reuse the underlying byte slice
> buffer.Reset()
> 
> // Use buf here
> ...
> ```

I ran this suggested change through my test and it all checks off Perfectly 👍 

> If your profiling shows that `bufio.Reader` is helpful, you can wrap `reader` with it: `buffer.ReadFrom(bufio.NewReader(reader))`.

This didn't make a difference, so I've left it out.

> 💡 It also makes sense to add a `Benchmark` to see how much memory usage we reduce and keep it at that level.

I've added a task in the original [issue](https://github.com/grafana/xk6-browser/issues/1480) tracking all the memory related issues. I can't create a benchmark test yet as they aren't behaving as we would expect since the memory keeps going up due to the other memory leaks (investigating in https://github.com/grafana/xk6-browser/pull/1484). Once memory leaking/usage has come under a bit more control then I think it would be best to complete the benchmarks.

inancgumus: > I ran this suggested change through my test and it all checks off Perfectly 👍 https://github.com/grafana/xk6-browser/commit/a631b38457c5ff4497ae5292cc3e7735ee230c8e

Nice 🎉 

> I've added a task in the original https://github.com/grafana/xk6-browser/issues/1480 tracking all the memory related issues. I can't create a benchmark test yet as they aren't behaving as we would expect since the memory keeps going up due to the other memory leaks (investigating in https://github.com/grafana/xk6-browser/pull/1484). Once memory leaking/usage has come under a bit more control then I think it would be best to complete the benchmarks.

This is fine, but benchmarks can also help with pprofing.

ankur22:
Closing this in favour of the actual solution to the problem https://github.com/grafana/xk6-browser/pull/1488

PR #1484, opened 2024-10-14
Investigate other memory leaks

ankur22:
## What?

This PR is investigating one thing which is `nil`ling references to break connections so that the GC will collect unused memory/objects.

## Why?

To prevent and reduce memory usage/leaks.

## Checklist


ankur22:
Closing this in favour of the actual solution to the problem https://github.com/grafana/xk6-browser/pull/1488

Issue #1487, opened 2024-10-16
Add the ability to filter by request type in `page.on('metric')`


ankur22:
### Feature Description

We have `page.on('metric')` which allows us to group metrics when a URL match is found. The issue arises when the website under test makes different types of requests using the same URL, for example a GET to api/cart, and a POST to api/cart. In this case it's not going to be useful to group the same request together as the behaviour of them will be different.

What would be useful is to be able to filter/group also by the request type.

### Suggested Solution (optional)

A possible solution is to extend the existing API so that it can also take a `type` field. If the `type` field isn't present, it will assume that all requests are to be grouped regardless of the type. If a `type` field is present it will take that into account:

```js
  page.on('metric', (metric) => {
    metric.tag({
      urls: [
        {url: /^https:\/\/test\.k6\.io\/\?q=[0-9a-z]+$/, type:'GET', name:'test'},
      ]
    });
  });
```

### Already existing or connected issues / PRs (optional)

https://github.com/grafana/xk6-browser/issues/371
https://github.com/grafana/xk6-browser/issues/1434

```[tasklist]
- [ ] https://github.com/grafana/xk6-browser/pull/1489
- [ ] https://github.com/grafana/xk6-browser/pull/1490
- [x] Add to release notes: https://github.com/grafana/k6/commit/da3a11112ce4cfb2f7bc9544ae9199a40df7be53
- [ ] https://github.com/grafana/k6-DefinitelyTyped/pull/73
- [ ] https://github.com/grafana/k6-docs/pull/1807
```

inancgumus:
The suggested solution is mixing the tag name and properties. It was fine when there was only `url` and `name`. I believe the API might be less confusing if it would group these `url` by a `name` instead of having `name` as another property. Maybe something like:

```javascript
page.on('metric', metric => {
  metric.tag({
    name: 'tests',
    urls: [
      {url: /^https:\/\/test\.k6\.io\/\?q=[0-9a-z]+$/, type:'GET'},
    ]
  }, {
    name: 'other',
    urls: [
      {url: /^https:\/\/test\.k6\.io\/\?q=[0-9a-z]+$/, type:'POST'},
      {url: /^https:\/\/foo\.com/, type:'GET'},
    ]
  });
});
```

ankur22:
@inancgumus I can see where you are coming from, I like this approach 👍

PR #1489, opened 2024-10-17
Move the name field on the `page.on('metric') API

ankur22:
## What?

This change does two things:
1. Moves the `name` field to the level above.
2. Rename the `urls` field to `matches`.

So from:

```js
  page.on('metric', (metric) => {
    metric.tag({
      urls: [
        {url: /^https:\/\/test\.k6\.io\/\?q=[0-9a-z]+$/, name:'test'},
      ]
    });
  });
```

to:

```js
  page.on('metric', (metric) => {
    metric.tag({
      name:'test',
      matches: [
        {url: /^https:\/\/test\.k6\.io\/\?q=[0-9a-z]+$/},
      ]
    });
  });
```

## Why?

It was unclear what `name` was doing when it was part of the "`match`" object and it made it difficult to add other fields to the object to match against.

`urls` doesn't lend itself to extending with more fields to match against. It also doesn't represent what it is doing, whereas `matches` is a helpful verb that makes it clearer to the user/reader of the API what its impact is.

By moving the `name` field one level up, and renaming `urls` to `matches`, we can now read this API:

```js
    metric.tag({
      name:'test',
      matches: [
        {url: /^https:\/\/test\.k6\.io\/\?q=[0-9a-z]+$/},
      ]
    });
```

as:

1. tag metric;
2. with this name;
3. that matches.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1487

### DIFF
--- a/browser/mapping_test.go
+++ b/browser/mapping_test.go
@@ -378,7 +378,7 @@ type consoleMessageAPI interface {
 
 // metricEventAPI is the interface of a metric event.
 type metricEventAPI interface {
-	Tag(matchesRegex common.K6BrowserCheckRegEx, patterns common.URLTagPatterns) error
+	Tag(matchesRegex common.K6BrowserCheckRegEx, patterns common.TagMatches) error
 }
 
 // frameAPI is the interface of a CDP target frame.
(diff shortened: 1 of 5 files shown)


inancgumus (approved):
LGTM 🎉

PR #1490, opened 2024-10-17
Add method field on Match object for `page.on('metric')`

ankur22:
## What?

This change adds a new field to the `Match` object in `page.on('metric')`. It allows the user to add the request method to it so that the tagging can be fine tuned to group requests with the same url and method. `method` is an optional field, and if it's not defined then the grouping will be done for all urls that match regardless of the request method.

```js
page.on('metric', (metric) => {
  metric.tag({
    name:'ping-1',
    matches: [
      {url: /^http:\/\/127\.0\.0\.1\:[0-9]+\/ping\?h=[0-9a-z]+$/, method: 'GET'},
    ]
  });
});
```

## Why?

The use case for this is when requests are made against the same URL but with different request methods. A simple example is a `PUT` to `/cart/123` to add an item to the cart, and a `GET` to `/cart/123` to retrieve the cart.

## Checklist

## Related PR(s)/Issue(s)

Updates: https://github.com/grafana/xk6-browser/issues/1450

### DIFF
--- a/common/page.go
+++ b/common/page.go
@@ -436,6 +436,10 @@ func (e *MetricEvent) Tag(matchesRegex K6BrowserCheckRegEx, matches TagMatches) 
 			default:
 				return fmt.Errorf("method %q is invalid", m.Method)
 			}
+
+			if method != e.method {
+				continue
+			}
 		}
 
 		// matchesRegex is a function that will perform the regex test in the Sobek
--- a/common/page.go
+++ b/common/page.go
@@ -447,7 +447,7 @@ func (e *MetricEvent) Tag(matchesRegex K6BrowserCheckRegEx, matches TagMatches) 
 // `page.on('metric')`. The user will need to use `Tag` to supply the
 // url regexes and the matching is done from within there. If a match is found,
 // the supplied name is returned back upstream to the caller of urlTagName.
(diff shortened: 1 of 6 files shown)


common/page.go
```diff
@@ -421,6 +426,22 @@ func (e *MetricEvent) Tag(matchesRegex K6BrowserCheckRegEx, matches TagMatches)
...
+		// Validate the request method type if it has been assigned in a Match.
+		method := strings.TrimSpace(m.Method)
+		if method != "" {
+			method = strings.ToUpper(method)
+			switch method {
+			case http.MethodGet, http.MethodPost, http.MethodPut, http.MethodDelete, http.MethodPatch,
+				http.MethodHead, http.MethodOptions, http.MethodConnect, http.MethodTrace:
+			default:
+				return fmt.Errorf("method %q is invalid", m.Method)
+			}
```
inancgumus, line 438: I'd move this to a helper function in helpers.go. I believe it's a bit out of place here and obscures the `Tag` method's actual logic flow.

inancgumus (approved):
LGTM 👍

common/page.go
ankur22: Thanks for the suggestion. I was wondering if this needed to be extracted out into a separate validation function. I think for now though i'm going to leave it as is.

PR #1495, opened 2024-10-18
Add error return value from `page.on` handlers

ankur22:
## What?

This adds an error return value from the `page.on` handlers.

## Why?

It's being more defensive to ensure that when an error occurs (due to the iteration ending) we exit early and don't try to process any other handlers, and to ensure that we don't use incorrect url name values when a `page,.on(metric)` handler returns early.

## Checklist

## Related PR(s)/Issue(s)

Relates to: https://github.com/grafana/xk6-browser/pull/1493#discussion_r1806534456


inancgumus (approved):
Nice, but we need to take these debugs and errors under control to reduce verbosity.

ankur22:
> Nice, but we need to take these debugs and errors under control to reduce verbosity (my comment is not specific to this PR :)).

Yeah, definitely agree.

PR #1510, opened 2024-11-04
Desobekify `BrowserContext` option parsing

inancgumus:
## What?

This is an experimental PR that is open for discussion.

It moves the Sobek-dependent option parsing part out of the business logic. Once we finish doing the same with the rest of the option parsing, we can entirely remove the Sobek dependency from the business logic.

## Why?

We currently parse Sobek values into our business logic types, but the parsing dependency is in an unrelated package (`common`). To enhance maintainability, we move the parsing process and keep related components together (in the mapping layer).

Additionally, we could implement a `Validate` method for the business types (in the `common` package) to check their validity separately from the Sobek transformation. We can do this in a later PR when needed.

## Checklist

## Related PR(s)/Issue(s)

Updates #1064


ankur22 (dismissed):
This makes sense to me. Moving the parsing into the `browser` package (mapping layer) means that it is closer to where it is used, removing it from a package where it isn't used. This should make things clearer for new devs too.

Issue #1270, opened 2024-04-18
De-sobekify `BrowserContext` methods


inancgumus:
See grafana/xk6-browser#1064.

```[tasklist]
### Tasks
- [ ] grafana/xk6-browser#1504
- [ ] grafana/xk6-browser#1507
- [ ] grafana/xk6-browser#1510
- [ ] grafana/xk6-browser#1511
- [ ] https://github.com/grafana/xk6-browser/pull/1512
- [ ] https://github.com/grafana/xk6-browser/pull/1513
- [ ] https://github.com/grafana/xk6-browser/pull/1518
- [ ] grafana/xk6-browser#1519
- [ ] https://github.com/grafana/xk6-browser/pull/1520
- [ ] https://github.com/grafana/xk6-browser/pull/1521
- [ ] https://github.com/grafana/xk6-browser/pull/1526
- [ ] https://github.com/grafana/xk6-browser/pull/1527
- [ ] Add release notes
```

PR #1518, opened 2024-11-06
Desobekify `Credentials`

inancgumus:
## What?

- Desobekify `Credentials`.
- Turn `Credentials` into a value type.

## Why?

See "Desobekifying Sobek transformation" at https://github.com/grafana/xk6-browser/issues/1064.

## Checklist

## Related PR(s)/Issue(s)

- #1270


common/network_manager.go
```diff
@@ -34,6 +35,15 @@ type Credentials struct {
 	Password string `js:"password"`
 }
 
+// IsEmpty returns true if the credentials are empty.
+func (c Credentials) IsEmpty() bool {
+	c = Credentials{
+		Username: strings.TrimSpace(c.Username),
+		Password: strings.TrimSpace(c.Password),
+	}
+	return c == (Credentials{})
```
ankur22: Preferably the `TrimSpace` would have been done when the object was created rather than when `IsEmpty` was called. It leaves us with a scenario where it might be possible to accidentally work with credentials that have fields with spaces.

Could we not have a validator/cleanup method that is called to remove the spaces after parsing/exporting these from the option?

ankur22 (dismissed):
Functionally LGTM 🚀

common/network_manager.go
inancgumus: `IsEmpty` ensures that the `Credentials` value is truly empty (with none of its fields set), so I've tried to preserve the previous behavior. Still, it makes sense to add validator methods to the options, including this one 👍

inancgumus: I've added an issue here: https://github.com/grafana/xk6-browser/issues/1530

PR #1526, opened 2024-11-06
Refactor `parseBrowserContextOptions` to Sobek

inancgumus:
## What?

Refactor `parseBrowserContextOptions` to use direct Sobek transformation.

## Why?

After the refactorings we made in #1270, it's now possible to let Sobek perform this transformation alone.

## Checklist

- [x] I have added tests for my changes (locally tested it—see the script)

## Script

I also used the script to test it using my old friend `Printf` in `Browser.NewContext` etc..

```javascript
  const context = await browser.newContext({
    acceptDownloads: true,
    downloadsPath: "/tmp",
    bypassCSP: true,
    colorScheme: "dark",
    deviceScaleFactor: 1,
    extraHTTPHeaders: {
        "X-Header": "value",
    },
    geolocation: { latitude: 51.509865, longitude: -0.118092 },
    hasTouch: true,
    httpCredentials: { username: "admin", password: "password" },
    ignoreHTTPSErrors: true,
    isMobile: true,
    javaScriptEnabled: true,
    locale: "fr-FR",
    offline: true,
    permissions: ["camera", "microphone"],
    reducedMotion: true,
    screen: { width: 800, height: 600 },
    timezoneID: "Europe/Paris",
    userAgent: "my agent",
    viewport: { width: 800, height: 600 },
  });
}
```

## Related PR(s)/Issue(s)

- #1270

### DIFF
--- a/browser/browser_context_options_test.go
+++ b/browser/browser_context_options_test.go
@@ -12,7 +12,7 @@ import (
 func TestBrowserContextOptionsPermissions(t *testing.T) {
 	vu := k6test.NewVU(t)
 
-	opts, err := parseBrowserContextOptions(vu.Context(), vu.ToSobekValue((struct {
+	opts, err := parseBrowserContextOptions(vu.Runtime(), vu.ToSobekValue((struct {
 		Permissions []any `js:"permissions"`
 	}{
 		Permissions: []any{"camera", "microphone"},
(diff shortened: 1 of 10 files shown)


browser/browser_context_options_test.go
```diff
@@ -40,3 +41,24 @@ func TestBrowserContextSetGeolocation(t *testing.T) {
 	assert.Equal(t, 2.0, opts.Geolocation.Longitude)
 	assert.Equal(t, 3.0, opts.Geolocation.Accuracy)
 }
+
+func TestBrowserContextDefaultOptions(t *testing.T) {
```
ankur22, line 45: I think i would feel more comfortable with this change with a test that ensures that the parsing/export/merge does indeed work with all the possible fields.

ankur22 (approved):
It seems like a good idea, but I think we should add another test to ensure that all the fields are parsed/exported/merged as we expect.

browser/browser_context_options_test.go
inancgumus: That's a double-edged sword. We'd be testing Sobek's export behavior. On the other hand, it makes sense because we might set incorrect `js` tag values in the struct. I'll send another PR soon.

PR #1542, opened 2024-11-18
Fix frame panic

ankur22:
## What?

The change fixes a panic by returning nil instead of erroring.

## Why?

If the frame is nil at this point, then the cause of this is likely due to chrome not sending a frameAttached event ahead of time. This isn't a bug in chrome, and seems to be intended behavior. Instead of worrying about the nil frame and causing the test to fail when the frame is nil, we can instead return early. The frame will be initialized when getFrameTree CDP request is made, which will call onFrameAttached and onFrameNavigated.

## Checklist

## Related PR(s)/Issue(s)

https://github.com/grafana/xk6-browser/issues/1500

### DIFF
--- a/common/frame_manager.go
+++ b/common/frame_manager.go
@@ -246,7 +246,19 @@ func (m *FrameManager) frameNavigated(frameID cdp.FrameID, parentFrameID cdp.Fra
 	frame := m.frames[frameID]
 
 	if !isMainFrame && frame == nil {
-		return errors.New("we either navigate top level or have old version of the navigated frame")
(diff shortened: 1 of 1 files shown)


inancgumus (approved):
This might make sense as PW also returns similarly.

ankur22:
> This might make sense as PW also returns similarly.

I think it's one of those ones where we need to see how it reacts in the wild (on remote runs). I'll start on a deployment soon.

olegbespalov (approved):

Issue #1543, opened 2024-11-19
Evaluate and evaluateHandle cause NPD on nil pageFunc


ankur22:
### Brief summary

The cause of this NPD is due to the browser module not correctly checking the arguments before trying to work with them. With `JSHandle.evaluate` the `pageFunc` argument is required, but the [browser module doesn't correctly check](https://github.com/grafana/xk6-browser/blob/4350722189cf533a31228ff0c032b88f14ae394b/browser/js_handle_mapping.go#L27) whether it is `null`/`undefined` before working with it.

There are many other examples within the browser module where it doesn't first validate the argument before working with it (page, frame, elementHandle).

This issue can be replicated with:

```js
import { browser } from 'k6/browser';

export const options = {
  scenarios: {
    ui: {
      executor: 'shared-iterations',
      options: {
        browser: {
            type: 'chromium',
        },
      },
    },
  },
}

export default async function() {
  const page = await browser.newPage();

  try {
    await page.goto('https://test.k6.io/my_messages.php');

    const jsHandle = await page.waitForSelector('input[name="login"]');
    // This is incorrect, the pageFunction arg is required on evaluate and this is
    // why we get a NPD since we don't check it is nil first.
    await jsHandle.evaluate();
  } finally {
    await page.close();
  }
}
```

There's also an issue with `waitForSelector` -- it returns an `ElementHandle`, but it is mapped to a `JSHandle` 🤔 

```bash
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x1606eea]

goroutine 1536 [running]:
github.com/grafana/xk6-browser/browser.mapJSHandle.func3.1()
github.com/grafana/xk6-browser@v1.9.1/browser/js_handle_mapping.go:27 +0x16a
github.com/grafana/xk6-browser/k6ext.promise.func1()
github.com/grafana/xk6-browser@v1.9.1/k6ext/promise.go:24 +0x2c
created by github.com/grafana/xk6-browser/k6ext.promise in goroutine 53
github.com/grafana/xk6-browser@v1.9.1/k6ext/promise.go:23 +0x9a
```

test run: 3526237

### xk6-browser version

v1.9.1

```[tasklist]
### Tasks
- [ ] #1544
- [x] Update release notes
```

ankur22:
[...]

PR #1544, opened 2024-11-19
Fix `mapJSHandle` invalid memory address or nil pointer dereference

inancgumus:
## What?

Add missing function arguments check to `JSHandle` mapping.

## Why?

#1543 

## Checklist

## Related PR(s)/Issue(s)

Updates #1543

### DIFF
--- a/tests/page_test.go
+++ b/tests/page_test.go
@@ -264,6 +264,11 @@ func TestPageEvaluateMappingError(t *testing.T) {
 			script:  "(6)",
 			wantErr: "Given expression does not evaluate to a function",
 		},
+		{
+			name:    "void",
+			script:  "",
+			wantErr: "evaluate requires a page function",
+		},
 	}
 
 	for _, tt := range tests {
--- a/browser/frame_mapping.go
+++ b/browser/frame_mapping.go
@@ -59,19 +59,25 @@ func mapFrame(vu moduleVU, f *common.Frame) mapping { //nolint:gocognit,cyclop
 				return nil, f.DispatchEvent(selector, typ, exportArg(eventInit), popts) //nolint:wrapcheck
 			}), nil
 		},
-		"evaluate": func(pageFunction sobek.Value, gargs ...sobek.Value) *sobek.Promise {
+		"evaluate": func(pageFunc sobek.Value, gargs ...sobek.Value) (*sobek.Promise, error) {
+			if sobekEmptyString(pageFunc) {
+				return nil, fmt.Errorf("evaluate requires a page function")
+			}
 			return k6ext.Promise(vu.Context(), func() (any, error) {
-				return f.Evaluate(pageFunction.String(), exportArgs(gargs)...) //nolint:wrapcheck
-			})
+				return f.Evaluate(pageFunc.String(), exportArgs(gargs)...)
+			}), nil
 		},
-		"evaluateHandle": func(pageFunction sobek.Value, gargs ...sobek.Value) *sobek.Promise {
+		"evaluateHandle": func(pageFunc sobek.Value, gargs ...sobek.Value) (*sobek.Promise, error) {
+			if sobekEmptyString(pageFunc) {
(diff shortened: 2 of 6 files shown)


ankur22 (commented):
Thanks for the fix! I like where it is going.

Two points:

1. Can we look to fix the other `evaluate`/`evaluateHandle` methods on `page` and `frame`? In a new PR if you prefer, but happy to have them as separate commits in this PR 🙂 
2. This does indeed fix the issues when `evaluate()` is used without pageFunc. It however breaks one thing which is that you can no longer pass in a string:

```js
// As a string that is callable
console.log(await jsHandle.evaluate("node => node.innerText"));
// As a function
console.log(await jsHandle.evaluate(node => node.innerText));
```

3. Do we need tests for at least one of these methods (maybe just on `page`?).

inancgumus:
@ankur22, sure thing!

> 1. Can we look to fix the other `evaluate`/`evaluateHandle` methods on `page` and `frame`? In a new PR if you prefer, but happy to have them as separate commits in this PR 🙂 

Sure :) Can you repurpose issue #1543? Maybe remove [this explanation](https://github.com/grafana/xk6-browser/issues/1543#issuecomment-2486143338) from the issue description so that we can understand what the issue entails. Currently, it looks like it's about `mapJSHandle`.

> 2. This does indeed fix the issues when `evaluate()` is used without pageFunc. It however breaks one thing which is that you can no longer pass in a string:

Are you sure that it doesn't work with strings? I've tried this, and it worked:

```js
const n = await handle.evaluate('() => 5 * 42');
```

But, this is a nice catch because now we don't accept non-functions.

> 3. Do we need tests for at least one of these methods (maybe just on `page`?).

Would adding a test to `examples/` be fine?

ankur22:
> Sure :) Can you repurpose issue https://github.com/grafana/xk6-browser/issues/1543? Maybe move https://github.com/grafana/xk6-browser/issues/1543#issuecomment-2486143338 to the issue description so that we can understand what the issue entails. Currently, it looks like it's about mapJSHandle.

Sure, done.

> Are you sure that it doesn't work with strings?

Yeah, this script below no longer works with this branch:

```js
import { browser } from 'k6/browser';

export const options = {
  scenarios: {
    ui: {
      options: {
        browser: {
            type: 'chromium',
        },
      },
    },
  },
}

  try {
    await page.goto('https://test.k6.io/my_messages.php');

    const jsHandle = await page.waitForSelector('h2');

    // As a string that is callable
    console.log(await jsHandle.evaluate("node => node.innerText"));
    
    // As a function
    console.log(await jsHandle.evaluate(node => node.innerText));
  } finally {
    await page.close();
  }
}
```

> Would adding a test to examples/ be fine?

I think we should have table tests (integration tests) in the `tests` package, unless there's a reason it's not possible to create them? A test in the examples folder would be a good addition too.

tests/page_test.go
```diff
@@ -264,6 +264,11 @@ func TestPageEvaluateMappingError(t *testing.T) {
 			script:  "(6)",
 			wantErr: "Given expression does not evaluate to a function",
 		},
+		{
+			name:    "void",
+			script:  "",
+			wantErr: "evaluate requires a page function",
+		},
```
ankur22, line 271: Perfect 👍 I was mainly looking for this.

Would be good to get a test to make sure that this works as i demonstrated in the script:

    // As a string that is callable
    console.log(await jsHandle.evaluate("node => node.innerText"));
    // As a function
    console.log(await jsHandle.evaluate(node => node.innerText));

inancgumus: Don't we already have tests for that?

https://github.com/grafana/xk6-browser/blob/d19534e927bfe19133205d1b16110973f3f86ca4/tests/page_test.go#L228-L239

I've also added a test for `SobekEmptyString`. Lastly, I ran your script, and it works.

ankur22 (approved):
LGTM 🚀

Issue #1292, opened 2024-05-02
Migrating to k6-core's linter


inancgumus:
### What?

* k6-core and k6-browser linter configurations do not match.
* Use k6-core's version directly or use it as the base and extend it with k6-browser's additional rules.

### Why?

* To rely on a shared and reviewed configuration across teams.
* To prevent security issues like `setInputFiles`.

### How?

k6-core prefers to use the same rules used in the linter configuration to sync across the projects (or extensions in general). k6-core has created [a reusable workflow](https://github.com/grafana/k6-ci) that they've been using for other minor extensions, as well.  A concrete example is related to the `os` package [which is denied](https://github.com/grafana/k6/blob/master/.golangci.yml#L64) in the k6-core configuration.

They suggest k6-browser to use the entire workflow [here](https://github.com/grafana/xk6-grpc/blob/main/.github/workflows/all.yml#L12-L14) for example. Or, only [the linter configuration part](https://github.com/grafana/k6-ci/blob/883239d9b74eba603d530375685c55ec82621478/.github/workflows/all.yml#L45-L46).  They believe a better case would be the ability to merge two golangci configurations, but they don't find it as a solution at the moment.

### Tasks

```[tasklist]
### Tasks
- [ ] Work with the same linter + dependency job -- see [comment](https://github.com/grafana/xk6-browser/issues/1292#issuecomment-2092665993)
- [ ] Work with the same test jobs -- see [comment](https://github.com/grafana/xk6-browser/issues/1292#issuecomment-2092708331)
- [ ] Fix older lint issues
- [ ] Migrate fully to work with https://github.com/grafana/k6-ci
```

inancgumus:
cc: @codebien

ankur22:
@codebien @inancgumus I've taken a look at this for the browser project. I think this is a great idea. Unfortunately the browser codebase deviates quite a bit from the k6 codebase, and therefore the k6 linter rules point out many issues in the browser code base:

<details>
</details>

ankur22:
@codebien @inancgumus apart from the lint, there are also the following jobs:

- `dependencies`
- `test-go-versions`
- `test-build`

`test-go-versions` isn't going to work with the browser codebase since we need to be able to add some env vars when running the test like we've done [here](https://github.com/grafana/xk6-browser/blob/main/.github/workflows/test.yml#L42). The possible solutions for this are:

[...]

PR #1557, opened 2024-12-10
Lint fixes

mstoykov:
## What?

Fixes lint errors or ignroes them

## Why?

Works towards #1292 so it can be merged into k6 

## Checklist

### DIFF
--- a/browser/registry_test.go
+++ b/browser/registry_test.go
@@ -289,7 +289,7 @@ func TestBrowserRegistry(t *testing.T) {
 
 		vu := k6test.NewVU(t)
 		var cancel context.CancelFunc
-		vu.CtxField, cancel = context.WithCancel(vu.CtxField)
+		vu.CtxField, cancel = context.WithCancel(vu.CtxField) //nolint:fatcontext
 		browserRegistry := newBrowserRegistry(context.Background(), vu, remoteRegistry, &pidRegistry{}, nil)
 
 		vu.ActivateVU()
@@ -320,6 +320,7 @@ func TestBrowserRegistry(t *testing.T) {
 }
 
 func TestParseTracesMetadata(t *testing.T) {
+	t.Parallel()
 	testCases := []struct {
 		name        string
 		env         map[string]string
@@ -388,6 +389,7 @@ func TestParseTracesMetadata(t *testing.T) {
 	for _, tc := range testCases {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
 			lookup := func(key string) (string, bool) {
 				v, ok := tc.env[key]
 				return v, ok
--- a/common/connection.go
+++ b/common/connection.go
@@ -157,7 +157,12 @@ func NewConnection(
 
 	ctx, cancelCtx := context.WithCancel(ctx)
 
-	conn, _, connErr := wsd.DialContext(ctx, wsURL, header)
+	conn, response, connErr := wsd.DialContext(ctx, wsURL, header)
+	if response != nil {
+		defer func() {
+			_ = response.Body.Close()
+		}()
+	}
 	if connErr != nil {
 		cancelCtx()
 		return nil, connErr
@@ -251,18 +256,21 @@ func (c *Connection) closeAllSessions() {
 }
 
 func (c *Connection) createSession(info *target.Info) (*Session, error) {
-	c.logger.Debugf("Connection:createSession", "tid:%v bctxid:%v type:%s", info.TargetID, info.BrowserContextID, info.Type)
+	c.logger.Debugf("Connection:createSession", "tid:%v bctxid:%v type:%s",
+		info.TargetID, info.BrowserContextID, info.Type)
 
 	var sessionID target.SessionID
 	var err error
 	action := target.AttachToTarget(info.TargetID).WithFlatten(true)
 	if sessionID, err = action.Do(cdp.WithExecutor(c.ctx, c)); err != nil {
-		c.logger.Debugf("Connection:createSession", "tid:%v bctxid:%v type:%s err:%v", info.TargetID, info.BrowserContextID, info.Type, err)
+		c.logger.Debugf("Connection:createSession", "tid:%v bctxid:%v type:%s err:%v",
+			info.TargetID, info.BrowserContextID, info.Type, err)
 		return nil, err
 	}
 	sess := c.getSession(sessionID)
 	if sess == nil {
-		c.logger.Warnf("Connection:createSession", "tid:%v bctxid:%v type:%s sid:%v, session is nil", info.TargetID, info.BrowserContextID, info.Type, sessionID)
+		c.logger.Warnf("Connection:createSession", "tid:%v bctxid:%v type:%s sid:%v, session is nil",
+			info.TargetID, info.BrowserContextID, info.Type, sessionID)
 	}
 	return sess, nil
 }
@@ -317,6 +325,7 @@ func (c *Connection) findTargetIDForLog(id target.SessionID) target.ID {
 	return s.targetID
 }
 
+//nolint:funlen,gocognit,cyclop
 func (c *Connection) recvLoop() {
 	c.logger.Debugf("Connection:recvLoop", "wsURL:%q", c.wsURL)
 	for {
@@ -343,13 +352,14 @@ func (c *Connection) recvLoop() {
 
 		// Handle attachment and detachment from targets,
 		// creating and deleting sessions as necessary.
+		//nolint:nestif
 		if msg.Method == cdproto.EventTargetAttachedToTarget {
 			ev, err := cdproto.UnmarshalMessage(&msg)
 			if err != nil {
 				c.logger.Errorf("cdp", "%s", err)
 				continue
 			}
-			eva := ev.(*target.EventAttachedToTarget)
+			eva := ev.(*target.EventAttachedToTarget) //nolint:forcetypeassert
 			sid, tid := eva.SessionID, eva.TargetInfo.TargetID
 
 			if c.onTargetAttachedToTarget != nil {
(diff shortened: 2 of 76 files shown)


common/connection.go
```diff
@@ -157,7 +157,12 @@ func NewConnection(
 
 	ctx, cancelCtx := context.WithCancel(ctx)
 
-	conn, _, connErr := wsd.DialContext(ctx, wsURL, header)
+	conn, response, connErr := wsd.DialContext(ctx, wsURL, header)
+	if response != nil {
+		defer func() {
+			_ = response.Body.Close()
```
olegbespalov, line 163: `wsd.DialContext` docs stays: 

```
The response body may not contain the entire response and does not need to be closed by the application.
```

So not sure if we need this :shrug:

common/execution_context.go
```diff
@@ -307,9 +309,7 @@ func (e *ExecutionContext) Eval(apiCtx context.Context, js string, args ...any)
 		returnByValue: true,
 	}
 	evalArgs := make([]any, 0, len(args))
-	for _, a := range args {
-		evalArgs = append(evalArgs, a)
-	}
+	evalArgs = append(evalArgs, args...)
```
olegbespalov, line 312: Do we need here (and same applies for the file) this `evalArgs` at all? :thinking:

common/remote_object.go
```diff
@@ -78,28 +77,27 @@ func parseRemoteObjectPreview(op *cdpruntime.ObjectPreview) (map[string]any, err
...
 ) (any, error) {
-	switch t {
-	case cdpruntime.TypeAccessor:
+	switch t { //nolint:exhaustive
+	case runtime.TypeAccessor:
 		return "accessor", nil
-	case cdpruntime.TypeBigint:
-		n, err := strconv.ParseInt(strings.Replace(val, "n", "", -1), 10, 64)
+	case runtime.TypeBigint:
+		n, err := strconv.ParseInt(strings.ReplaceAll(val, "n", ""), 10, 64)
```
olegbespalov, line 87: Not related to the PR, but isn't this an error? I mean that we're using `strconv.ParseInt` for the parsing of `runtime.TypeBigint`? :thinking:

k6ext/k6test/vu.go
```diff
@@ -13,12 +13,10 @@ import (
 	"github.com/grafana/xk6-browser/k6ext"
 
 	"go.k6.io/k6/event"
-	k6event "go.k6.io/k6/event"
 	k6common "go.k6.io/k6/js/common"
-	k6eventloop "go.k6.io/k6/js/eventloop"
+	eventloop "go.k6.io/k6/js/eventloop"
```
olegbespalov, line 17: no need for aliasing

storage/file_persister.go
```diff
@@ -25,11 +25,11 @@ func (l *LocalFilePersister) Persist(_ context.Context, path string, data io.Rea
 	cp := filepath.Clean(path)
 
 	dir := filepath.Dir(cp)
-	if err = os.MkdirAll(dir, 0o755); err != nil {
+	if err = os.MkdirAll(dir, 0o755); err != nil { //nolint:forbidigo,gosec
 		return fmt.Errorf("creating a local directory %q: %w", dir, err)
 	}
 
-	f, err := os.OpenFile(cp, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600)
+	f, err := os.OpenFile(cp, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600) //nolint:forbidigo
```
olegbespalov, line 32: Not related to the PR, but this probably should be refactored, it shouldn't use directly os, right?

storage/storage_test.go
```diff
@@ -13,14 +13,14 @@ import (
...
 	t.Parallel()
 
-	tmpDir := os.TempDir()
+	tmpDir := os.TempDir() //nolint:forbidigo
 
 	t.Run("dir_provided", func(t *testing.T) {
 		t.Parallel()
 
-		dir, err := os.MkdirTemp("", "*")
+		dir, err := os.MkdirTemp("", "*") //nolint:forbidigo
```
olegbespalov, line 21: We probably can just replace these lines with `dir := t.TempDir()`

common/connection.go
mstoykov: great but then we need to put a nolint line, this way at least nobody needs to go read what DialContext does and whether for this http.response it needs or does not need to be closed :shrug:

common/execution_context.go
mstoykov: I have no idea - fixing nearly 400 lint errors and some point means that I stopped trying to figure out if a given piece of code is needed or I can remove it as goign up and down the stack is way too much work for each call place.

storage/file_persister.go
mstoykov: Yes - I have a list of things that shoudl be fixed, but rewritign browser as part of getting linter issues fixed turned out to be way too much work.

k6ext/k6test/vu.go
mstoykov: ah ... god catch

common/remote_object.go
mstoykov: This is parsin the BigInt from its string representation, why not directly from the type ... :shrug: 

Arguably one of  the places where I am more questioning why we are doing this at all than anythign else. 

In this case we do first remove `n` which arguably will only be the *last* symbol in the string representation. 

Just changing more than how the code actually does a thing means more chances for error and given the size of the required changes, it already is very likely some corner case isn't handled.

olegbespalov (approved):

olegbespalov:
There are also some linter reports, but overall LGTM :+1:

mstoykov:
> There are also some linter reports, but overall LGTM 👍

Differences in 1.62.2 and 1.60.2 :scream: 

I have been updating k6 golangci-lint upwards so we can just fix a bunch of this at the same time ...

common/remote_object.go
ankur22: That bigInt is coming back from chrome itself in a json payload as a CDP event over the websocket, not from Sobek's runtime, so i think this is a valid way of parsing it from its CDP string representation into one that is needed in k6 browser, and eventually in the test script. The case for this is when we get a piece of data from chrome (e.g. a console message, request, response, etc, containing a JS type of some kind) and the user wants to read that value into the test script.

I may have missed something though?

mstoykov: If it comes over CDP as JSON - there is no json representation for bigint in the specification, so I guess this needs this speicifc code. So yeah this is probably is the correct thing to do

olegbespalov: The confusing part is that how we try to restore the `bigInt`, like instead of `math/bigInt` from GO SDK we treat `bigInt` as an integer.

Consider example:

```golang
package main

import (
	"fmt"
	"math/big"
	"strconv"
	"strings"
)

func main() {
	// A large number string
	bigIntString := "123456789012345678901234567890"

	bigInt := new(big.Int)
	_, success := bigInt.SetString(bigIntString, 10) // Base 10

	if !success {
		fmt.Println("Failed to parse big integer")
		return
	}
	fmt.Println("Parsed big.Int:", bigInt)

	n, err := strconv.ParseInt(strings.ReplaceAll(bigIntString, "n", ""), 10, 64)
	if err != nil {
		fmt.Println("Failed to parse int64 " + err.Error())
		return
	}

	fmt.Println("Parsed N:", n)
}
```

mstoykov: I guess you are correct for this, but please open an issue for it as it changes the current behaviour.

browser/registry_test.go
```diff
@@ -320,6 +320,7 @@ func TestBrowserRegistry(t *testing.T) {
 }
 
 func TestParseTracesMetadata(t *testing.T) {
```
ankur22, line 322: I think this needs to be ran sequentially 🤔 Let me double check.

common/remote_object.go
olegbespalov: >I guess you are correct for this, but please open an issue for it as it changes the current behaviour.

What I'm trying to check with @ankur22 (because of lack of my context of browser) if it's the case and we should open the issue

common/execution_context.go
ankur22: Does look like we might not need it. I'll need to run some tests though. I think for this PR, it's probably best not to incase there's a good reason for it.

common/remote_object.go
olegbespalov: Extracted https://github.com/grafana/xk6-browser/issues/1558, if not relevant let's close it

common/frame_session.go
```diff
@@ -1029,7 +1023,11 @@ func (fs *FrameSession) attachIFrameToTarget(ti *target.Info, sid target.Session
 		return nil
 	}
 	// Remove all children of the previously attached frame.
-	fs.manager.removeChildFramesRecursively(fr)
+	err := fs.manager.removeChildFramesRecursively(fr)
+	if err != nil {
+		return fmt.Errorf("removing childs from frame with ID %v : %w",
```
ankur22, line 1028: Instead of returning an error, can we log an error and continue? That way we came the same behaviour as before.

common/remote_object.go
ankur22: Yeah, that's a good point. Worth having the issue open to explore the reason why it's not represented as a go's bigInt.

storage/file_persister.go
ankur22: This should only occur when running tests locally, when running in the cloud the env is set so that the remote_persister is used.. When we wrote this there was no way to write arbitrary files using the k6 implemented FS API (i think).

ankur22 (approved):
I found it tiring to follow, i can't imagine how tedious it was to work on 😅 Thank you!

olegbespalov (approved):
After a resolving the conflict, I'm fine with merging this as it's an anything else fixing separately. It's too big and easily can get new conflicts.

common/frame_session.go
mstoykov: done

olegbespalov (approved):

ankur22 (approved):

